PR #53, opened 2014-11-17
Store GovDelivery credentials as environment variables.

kalleth:
When we want to make email alert API work in development, we need to provide a "real" username and password. We don't, however, want to run the risk of committing these details.

If we store them as environment variables, we'll never commit them.

### DIFF
--- a/config/gov_delivery.yml.erb
+++ b/config/gov_delivery.yml.erb
@@ -1,6 +1,6 @@
 development:
-  username: govdelivery@example.com
-  password: nottherealpassword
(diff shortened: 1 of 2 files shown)


kalleth:
Hi,

Look at this great and exetremely interesting stuff, it's admired by so many people, just take a look <http://ntedikati.andycarr.us/aeqgjo>

My Best, Stephen Best

kalleth:
Hi friend! 

I wanted to tell you how  much I miss you,  here  are some fresh photos from me and  my family, please take a look <http://spephuntepu.normelaine.com/aedaxpmp>


All best, Tom Russell

PR #67, opened 2015-01-15
Rescue Zero Subscriber Errors on bulletin send

james:
Bug: If a single user creates a topic, and then either fails to
complete the subscription process, or later unsubscribes, we will
continue to send emails to that topic. While that topic has zero
subscriptions, govdelivery will return an error message, which will
cause sidekiq to continually retry the bulletin send until success.
This means that the first user to then subscribe to that topic, will
soon receive all emails that we were trying to send during the time the
topic had zero subscribers.

This commit captures that specific error. It has to check both error
code and message, as that error code is used for other errors by
govdelivery.

### DIFF
--- a/app/services/gov_delivery/client.rb
+++ b/app/services/gov_delivery/client.rb
@@ -2,6 +2,7 @@ module GovDelivery
   class Client
     class UnknownError < StandardError; end
     class TopicAlreadyExistsError < UnknownError; end
+    class ZeroSubscriberError < UnknownError; end
 
     def initialize(options = {})
       @options = options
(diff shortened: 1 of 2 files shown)


james:
I'm currently not doing anything when we receive this error message. I think it will just silently be marked as successful. Is this OK, and if not, any proposed solution?

bradleywright:
@james I think that outcome is okay - in our existing GovDelivery go-between we also ignore these errors as they don't tell us anything interesting.

rboulton:
Agree with @bradleywright.  I'll merge.

PR #68, opened 2015-01-22
Govdelivery logging

james:
Adds sidekiq logging, and seperate logfile for what we send to govdelivery.


evilstreak:
One comment, good to merge other than that :+1:

james:
@evilstreak I seem to have fixed that up so good your comment disappeared forever

PR #76, opened 2015-06-17
Upgrade Rails to 4.1.11

alext:
To pick up fix for CVE-2015-3225


Gemfile.lock
```diff
@@ -134,19 +133,15 @@ GEM
     sidekiq-logging-json (0.0.14)
       sidekiq (~> 3)
     slop (3.6.0)
-    sprockets (2.12.3)
-      hike (~> 1.2)
-      multi_json (~> 1.0)
+    sprockets (3.2.0)
```
bradleywright: This is a big upgrade. Anything to worry about?

bradleywright: Although this is just an API app, so maybe we don't care...

alext: Yeah, this app has no assets, so I can't see that this will affect anything.

PR #79, opened 2015-09-30
Use a Services module to clean up external service location

mobaig:
This approach provides easier stubbing, a simpler implementation, and
a more obvious file location for the register of external services. It
is currently being applied to other GDS apps as well.


rboulton:
:+1:

PR #80, opened 2015-09-30
Add content IDs to Subscriber Lists

mobaig:
NOTE: review this by clicking on each commit, the entire diff is a bit noisy.

As part of the move towards identifying taggable content via unique content_ids,
and consequently handling email subscriptions and alerts via the same,
this commit does the following:

- Adds content_id to the SubscriberList model.
- Adds a factory method to build SubscriberLists, moving this logic out
  of the controller.

Trello: https://trello.com/c/ZD65QUBN/326-email-alerts-based-on-links-hash

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -18,8 +18,7 @@ class SubscriberListsController < ApplicationController
   end
 
   def create
-    list = create_or_fetch_subscriber_list
-
+    list = build_subscriber_list
     if list.save
       respond_to do |format|
         format.json { render json: list.to_json, status: 201 }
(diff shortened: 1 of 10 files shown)


bradwright:
Not sure about this - does it make moving the Whitehall matrix-like topics to this API impossible? They're more like stored queries, not pieces of content. cc @elliotcm

rboulton:
Ah, this isn't quite how I'd envisioned it: I'd expected to maintain the tags structure, but reference the tags by content_id instead of path.  The easiest way to migrate to this might be to add a new `hstore` field similar to the existing tags field, but using content_ids. Since these tags will always be found in the "links" field, I'd suggest calling that new field "links".

mobaig:
I've clearly missed something here. Will rework and push some changes.

elliotcm:
:+1: to @rboulton's comment

mobaig:
Re-opening after I've made some changes.

elliotcm:
Looks like this was superseded by https://github.com/alphagov/email-alert-api/pulls/82 so I'm deleting this branch.

PR #82, opened 2015-10-01
Add links field to SubscriberList

mobaig:
As part of the move towards identifying taggable content via unique content_ids,
and consequently handling email subscriptions and alerts via the same,
this commit does the following:

- Adds a links hstore to the SubscriberList model. This is intended to
  contain tags which are identified by content_id rather than path.
- Adds a factory method to build SubscriberLists, moving this logic out of the controller.

Trello: https://trello.com/c/guDFlWXp

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -1,8 +1,6 @@
 require "gov_delivery/client"
 
 class SubscriberListsController < ApplicationController
-  before_filter :validate_request, only: :create
-
   def show
     subscriber_list = SubscriberList.where_tags_equal(params[:tags]).first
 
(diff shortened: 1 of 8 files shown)


mobaig:
Rework of #80 .

@bradwright @rboulton @tijmenb - ready for another review when you are.

app/controllers/subscriber_lists_controller.rb
```diff
@@ -18,37 +16,28 @@ def show
...
-
-    response = gov_delivery.create_topic(params[:title])
-
-    SubscriberList.new(
-      title: params[:title],
-      gov_delivery_id: response.to_param,
-      tags: params[:tags]
+  def build_subscriber_list
+    gov_delivery_response = Services.gov_delivery.create_topic(params[:title])
+    SubscriberList.build_from(
```
tijmenb: Does the `build_from` factory method add anything now? We can now use `SubscriberList.new(subscriber_list_params.merge(gov_delivery_id: gov_delivery_response.to_param))`.

db/migrate/20151001102405_add_links_to_subscriber_list.rb
```diff
@@ -0,0 +1,7 @@
+class AddLinksToSubscriberList < ActiveRecord::Migration
+  def change
+    add_column :subscriber_lists, :links, :hstore, null: false, default: {}
+
+    add_index :subscriber_lists, :links, using: :gin
```
tijmenb: Can you add a comment explaining `gin`?

tijmenb:
Looks good to me, but I'd like @rboulton to merge so make sure we're doing the right thing here.

rboulton:
Looks good to me, too.  I'll not merge right now, because @tijmenb's nitpickky comments might be something you want to address, but what's going on here looks right to me now.  Thanks!

PR #83, opened 2015-10-06
Support querying of SubscriberLists with both tags and links

mobaig:
Modifies SubscriberList retrieval and creation of notifications to work with both links and tags, required as part of the migration to using only links.

Includes extraction of the (reasonably complex) querying logic into its own class. Could probably be factored much better than it is, but hopefully it provides a starting point for making this code easier to understand.

See each commit for more detail.

Trello: https://trello.com/c/x9pN83lI/365-email-alert-api-supports-links-to-topics-being-passed-as-content-ids-as-well-as-by-the-current-slug-method


app/controllers/notifications_controller.rb
```diff
@@ -1,6 +1,7 @@
 class NotificationsController < ApplicationController
   def create
-    NotificationWorker.perform_async(notification_params.to_json)
+    query_field = notification_params[:links].present? ? :links : :tags
```
rboulton: The links field could be present but not yet have been updated to contain topic links.  For example, there might be an entry in links for "available_translations", but no entry in topics (but some entries for topics under `tags`).  I think this needs to actually check if there are any matches in the links field, and add those to any matches based on the tags field.

KushalP:
Apart from what @rboulton brought up above, I think the rest of the code in here looks good. Once that's taken care of let me know; it would be good to see how this plugs together in the dev VM.

mobaig:
@rboulton have added a couple of commits which address your concerns.

mobaig:
@KushalP ready for another look whenever you are.

KushalP:
For the commits in 2d9e03 and ff61c5 do we have tests for them? It would be good to track any regressions in these instances.

mobaig:
@KushalP I feel like notifications are adequately tested at both integration and unit level, but on second look a basic `link`-oriented test in `spec/requests/send_notification_spec.rb` might be a good idea. Will add.

mobaig:
@KushalP @rboulton updated.

FYI, as noted on the Trello ticket I expect this PR to resolve [this bug](https://errbit.preview.alphagov.co.uk/apps/54257afd0da1154003001a0a/problems/56153beb65786306aee20500) in errbit.

PR #86, opened 2015-10-14
Add tasks to support conversion of topic tags into topic links

mobaig:
As part of the move towards using content IDs to uniquely identify
content across gov.uk, SubscriberLists need to have their existing tags
converted from slugs to content IDs.

This commit adds rake tasks to:
- Query the Content Store with a slug and write the returned content ID
  to the corresponding SubscriberList links hash.
- Delete SubscriberLists that don't return a content ID from the Content
  Store.

Trello: https://trello.com/c/ZD65QUBN/326-email-alerts-based-on-links-hash

### DIFF
--- a/lib/tasks/links_migration/link_migrator.rb
+++ b/lib/tasks/links_migration/link_migrator.rb
@@ -1,41 +1,45 @@
-class LinkMigrator
-  def populate_topic_links
-    relevant_subscriber_lists.each do |sl|
-      base_path = base_path_from(sl)
-      content_item = Services.content_store.content_item(namespaced(base_path))
-      if content_item && content_item.content_id.present?
-        puts "Updating links in SubscriberList #{sl.id}"
-        sl.update(links: {topics: [content_item.content_id]})
+module Tasks
+  module LinksMigration
+    class LinkMigrator
+      def populate_topic_links
+        relevant_subscriber_lists.each do |sl|
+          base_path = base_path_from(sl)
+          content_item = Services.content_store.content_item(namespaced(base_path))
+          if content_item && content_item.content_id.present?
+            puts "Updating links in SubscriberList #{sl.id}"
+            sl.update(links: {topics: [content_item.content_id]})
+          end
+        end
+      end
+
+      def destroy_non_matching_subscriber_lists
+        relevant_subscriber_lists.each do |sl|
+          base_path = base_path_from(sl)
+          content_item = Services.content_store.content_item(namespaced(base_path))
+
+          if content_item.blank? || content_item.content_id.blank?
+            puts "Destroying Subscriber List id: #{sl.id}"
+            sl.destroy!
+          end
+        end
+      end
+
+    private
+
+      def base_path_from(subscriber_list)
+        # Only one base path is ever specified in topic tags
+        subscriber_list.tags[:topics].first
+      end
+
+      def relevant_subscriber_lists
+        @relevant_subscriber_lists ||= SubscriberListQuery.new(query_field: :tags)
+          .subscriber_lists_with_key(:topics)
+      end
+
+      def namespaced(base_path)
+        # Because the base paths in email-alert-api don't include '/topic'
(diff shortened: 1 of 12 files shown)


mobaig:
Could do with review from @elliotcm and @rboulton in particular.

lib/tasks/links_migration/link_migrator.rb
```diff
@@ -0,0 +1,45 @@
...
+        # Only one base path is ever specified in topic tags
+        subscriber_list.tags[:topics].first
+      end
+
+      def relevant_subscriber_lists
+        @relevant_subscriber_lists ||= SubscriberListQuery.new(query_field: :tags)
+          .subscriber_lists_with_key(:topics)
+      end
+
+      def namespaced(base_path)
```
elliotcm: Technically this is returning the base path.  The argument to this method is the slug.

lib/tasks/links_migration/link_migrator.rb
```diff
@@ -0,0 +1,45 @@
+module Tasks
+  module LinksMigration
+    class LinkMigrator
+      def populate_topic_links
+        relevant_subscriber_lists.each do |sl|
+          base_path = base_path_from(sl)
+          content_item = Services.content_store.content_item(namespaced(base_path))
+          if content_item && content_item.content_id.present?
```
elliotcm: Under what circumstances would we not have a content ID here, and is skipping them the right thing to do?

lib/tasks/links_migration/link_migrator.rb
```diff
@@ -0,0 +1,45 @@
...
+            puts "Updating links in SubscriberList #{sl.id}"
+            sl.update(links: {topics: [content_item.content_id]})
+          end
+        end
+      end
+
+      def destroy_non_matching_subscriber_lists
+        relevant_subscriber_lists.each do |sl|
```
elliotcm: Using low-meaning variables is discouraged in [the styleguide](https://github.com/alphagov/styleguides/blob/master/ruby.md) when not on a single line.  Perhaps `list` here?

elliotcm:
Generally looks good.  Having we checked that there's no subscribers to these phantom subscriber lists before we delete them?

lib/tasks/links_migration/link_migrator.rb
mobaig: :+1: will update...

lib/tasks/links_migration/link_migrator.rb
mobaig: Redirects.

Good shout though. Both methods were originally one thing, but in the current form of this class it probably makes sense to fail loudly if this is the case - the idea being the cleanup task below should be run first. Will update...

lib/tasks/links_migration/link_migrator.rb
mobaig: Updating...

mobaig:
@elliotcm have reworked the approach here a bit. Key changes are detailed in the latest commit - would appreciate another look when you're able.

lib/tasks/links_migration/link_migrator.rb
```diff
@@ -0,0 +1,83 @@
...
+          end
+
+          if content_item.content_id.blank?
+            raise DodgyBasePathError, <<-ERROR.strip_heredoc
+              No content ID found for #{base_path_from(list)},
+              run `bundle exec rake links_migration:report_non_matching`
+              and fix these cases before continuing migration.
+            ERROR
+          end
+
```
jackscotti: I thought content-store was checking for the validity of UUIDs, but it's not.
https://github.com/alphagov/content-store/blob/master/app/models/content_item.rb#L30

Do you think is worth checking for it here?

jackscotti: I was wrong. It's doing it here: https://github.com/alphagov/content-store/blob/master/app/models/content_item.rb#L216

PR #89, opened 2015-10-26
Add temporary fix to correctly handle NotificationWorker jobs

mobaig:
This adds a temporary bandaid which correctly handles queued jobs that
may have been enqueued with or without encoding of the params hash at
the point of calling NotificationWorker.perform.

Relevant history can be observed in the October history of these two views:
https://github.com/alphagov/email-alert-api/commits/master/app/controllers/notifications_controller.rb
https://github.com/alphagov/email-alert-api/commits/master/app/workers/notification_worker.rb

Trello: https://trello.com/c/AuskAQ9S/390-fix-nomethoderror-in-email-alert-api-notifications

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -1,8 +1,6 @@
 class NotificationsController < ApplicationController
   def create
-    # Ensure the Sidekiq worker receives a simple data type as its argument -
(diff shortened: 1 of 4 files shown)


tijmenb:
LGTM - maybe some links to previous PR's in this description would be handy for future devs.

PR #92, opened 2015-11-03
Rename policy SubscriberList links to 'parent'

mobaig:
The policy-publisher and govuk-content-schemas have been updated (https://github.com/alphagov/policy-publisher/pull/111) such that email signup content items now carry a link to their parent content item. The key used in the link is 'parent'. The implication of this is that email-alert-api will start receiving links keyed with 'parent' for policy email subscriptions and notifications. This commit supports that transition.

Trello: https://trello.com/c/EzjSpzQl/402-update-email-alert-frontend-to-start-passing-content-ids-as-well-as-slugs


mobaig:
cc:  @rboulton @alext @tijmenb

PR #95, opened 2015-11-13
Improve `/healthcheck` endpoint

joshmyers:
As part of improving our Sidekiq monitoring, we
now graph all stats in Graphite. We are also 
moving to a 12factor style of application. 

Applications should have a notion of what is 
‘normal’ and can therefore set their thresholds 
rather than having these set by Puppet

Depends on: https://github.gds/gds/puppet/pull/3671

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -7,14 +7,18 @@ class HealthcheckController < ActionController::Base
         render json: {
           checks: {
             queue_size: {
+              queues: sidekiq_queues,
               status: queue_size_status
             },
             queue_age:  {
+              queues: queue_latencies,
               status: queue_age_status
-            }
+            },
+            queue_retry_size: {
+              status: queue_retry_size_status
+            },
           },
-          status: 'ok' #FIXME: probably need to pin this on DB connectivity and
-                       #Redis connectivity
+          status: status
         }
       }
     end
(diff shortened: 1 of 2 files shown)


tijmenb:
This looks okay. https://github.gds/gds/puppet/pull/3671 should be merged/deployed first right?

tijmenb:
@joshmyers Sure :) 

I would pull everything out into a class, so you can test it without having to go through the controller. 

You could pass in references to `Sidekiq::Stats.new` and a list of queues from the controller, so you don't have to stub `queue_latencies` and `sidekiq_queues` (which is an example of [stubbing the system under test](https://robots.thoughtbot.com/don-t-stub-the-system-under-test)).

We probably want `ENV.fetch('SIDEKIQ_RETRY_SIZE_CRITICAL').to_i`, which will raise an exception if the env variable isn't defined (`ENV['SIDEKIQ_RETRY_SIZE_CRITICAL'].to_i` will return `0` currently).

joshmyers:
@mattbostock I've addressed @tijmenb comment re the `ENV.fetch` but haven't re written tests so they don't have to go through the controller. That's probably our of scope for this PR, as I just added to what was already there.

joshmyers:
Builds are failing due to ENV variables not being set. I could set some sane defaults?

tijmenb:
Some of them are set here: https://github.com/alphagov/email-alert-api/blob/improve_healthcheck/spec/controllers/healthcheck_controller_spec.rb#L3-L7. You could add the missing too?

PR #99, opened 2016-02-05
Add document_type to subscriber lists and notifications

danielroseman:
Allow an optional document_type field to be set on subscriber lists, and supplied in notifications.

If present on subscriber lists that are already matched on links or tags, only notifications with a matching document_type will be sent. Notifications will always be sent on matching subscriber lists without a value for document_type, whether or not that value is provided in the notification.

```
subscriber list document_type | content item document_type | send email?
------------------------------+----------------------------+------------
travel-advice                 | travel-advice              | yes
travel-advice                 | publication                | no
travel-advice                 | [missing]                  | no
[missing]                     | publication                | yes
[missing]                     | [missing]                  | yes
```

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -46,5 +46,6 @@ private
     params.slice(:title)
       .merge(tags: params.fetch(:tags, {}))
       .merge(links: params.fetch(:links, {}))
+      .merge(document_type: params.fetch(:document_type, ""))
   end
 end
--- a/app/workers/notification_worker.rb
+++ b/app/workers/notification_worker.rb
@@ -24,7 +24,7 @@ class NotificationWorker
 
     @tags_hash  = Hash(notification_params[:tags])
     @links_hash = Hash(notification_params[:links])
-    lists = (lists_matched_on_links + lists_matched_on_tags).uniq { |l| l.id }
+    @document_type = notification_params[:document_type]
     delivery_ids = lists.map(&:gov_delivery_id)
 
     if lists.any?
@@ -63,5 +63,17 @@ private
   def lists_matched_on_links
     @lists_matched_on_links ||= SubscriberListQuery.new(query_field: :links)
       .where_all_links_match_at_least_one_value_in(@links_hash)
+  end
+
+  def filter_by_document_type(matching_lists)
+    matching_lists.select { |l| l.document_type.blank? || l.document_type == @document_type }
+  end
+
+  def matching_lists
+    (lists_matched_on_links + lists_matched_on_tags).uniq { |l| l.id }
+  end
+
+  def lists
+    @lists ||= filter_by_document_type(matching_lists)
   end
 end
--- a/spec/workers/notification_worker_spec.rb
+++ b/spec/workers/notification_worker_spec.rb
@@ -71,6 +71,50 @@ RSpec.describe NotificationWorker do
       end
     end
 
+    context "filtering on document_type" do
+      context "when the optional document_type is set in the subscriber list" do
+        before do
+          create(:subscriber_list, document_type: 'travel-advice', tags: {topics: ['foo/bar']})
+          notification_params.merge!(tags: {topics: ['foo/bar']})
+        end
+
+        it "sends a bulletin when document_type matches" do
+          notification_params[:document_type] = 'travel-advice'
+          make_it_perform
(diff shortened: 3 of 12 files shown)


rboulton:
:+1: for intent and API, haven't looked at implementation.

app/workers/notification_worker.rb
```diff
@@ -24,7 +24,7 @@ def perform(notification_params)
 
     @tags_hash  = Hash(notification_params[:tags])
     @links_hash = Hash(notification_params[:links])
-    lists = (lists_matched_on_links + lists_matched_on_tags).uniq { |l| l.id }
+    @document_type = notification_params[:document_type]
     delivery_ids = lists.map(&:gov_delivery_id)
 
     if lists.any?
```
tuzz: `lists` is now a method so we're calling this twice. It might be worth memoizing this as operations like `#uniq` can be quite expensive – usually O(n log n).

spec/workers/notification_worker_spec.rb
```diff
@@ -71,6 +71,50 @@ def make_it_perform
       end
     end
 
+    context "filtering on document_type" do
+      context "when present in subscriber list" do
```
tuzz: Might be more clearly written as: "when the optional document_type is set on the subscriber list"

Or similar?

spec/workers/notification_worker_spec.rb
```diff
@@ -71,6 +71,50 @@ def make_it_perform
...
+          expect(@gov_delivery).not_to have_received(:send_bulletin)
+        end
+
+        it "does not send a bulletin when content item has no document_type" do
+          make_it_perform
+          expect(@gov_delivery).not_to have_received(:send_bulletin)
+        end
+      end
+
+      context "when missing from subscriber list" do
```
tuzz: Might be more clearly written as: "when the optional document_type is absent from the subscriber list"

Or similar?

spec/workers/notification_worker_spec.rb
```diff
@@ -71,6 +71,50 @@ def make_it_perform
...
+          expect(@gov_delivery).to have_received(:send_bulletin)
+        end
+
+        it "send a bulletin when content item has no document_type" do
+          make_it_perform
+          expect(@gov_delivery).to have_received(:send_bulletin)
+        end
+      end
+    end
+
```
tuzz: Is it worth testing the case where the document_type is set to an empty string on the subscriber list?

app/workers/notification_worker.rb
```diff
@@ -64,4 +64,13 @@ def lists_matched_on_links
...
   end
+
+  def filter_by_document_type(matching_lists)
+    matching_lists.select { |l| l.document_type.blank? || l.document_type == @document_type }
+  end
+
+  def lists
+    matching_lists = (lists_matched_on_links + lists_matched_on_tags).uniq { |l| l.id }
+    filter_by_document_type(matching_lists)
+  end
```
tuzz: Is this kind of filtering what the `SubscriberListQuery` is for, or is that for something else?

app/controllers/subscriber_lists_controller.rb
```diff
@@ -43,7 +43,7 @@ def find_subscriber_list
   end
 
   def subscriber_list_params
-    params.slice(:title)
+    params.slice(:title, :document_type)
       .merge(tags: params.fetch(:tags, {}))
       .merge(links: params.fetch(:links, {}))
   end
```
tuzz: Is there any workflow for updating subscriber lists?

tuzz:
A few minor comments. Generally looks good.

spec/workers/notification_worker_spec.rb
danielroseman: I've updated the subscriberlist model and controller so that document_type is always an empty string if not supplied, so these cases are now the same.

There's no support for modifying subscriber lists; the API only supports the create action.

app/workers/notification_worker.rb
danielroseman: I did wonder about that, but the existing interface is built around a querying on a specific field, so it didn't really seem appropriate to do this there.

app/controllers/subscriber_lists_controller.rb
danielroseman: No it seems that the API only supports get and create.

bradwright:
@tuzz is this mergable now?

PR #100, opened 2016-02-10
Add Travel Advice Subscriber Lists

bradwright:
Add a migration which provides a mapping from GovDelivery Topic IDs to country slugs.

https://trello.com/c/zdbzUecF/511-create-email-alert-api-subscriber-lists-for-all-travel-advice-topics


bradwright:
Note: the all countries list doesn't currently work as the way we match on `document_format` doesn't have the right boolean logic.

bradwright:
The fix for the all countries list is going to be hairier than I thought - rescoping this PR to just include the individual countries (which works fine).

tuzz:
:+1: looks good

db/migrate/20160209143102_add_foreign_travel_advice_subscriber_lists.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class AddForeignTravelAdviceSubscriberLists < ActiveRecord::Migration
+  def change
+    csv_path = File.join(File.dirname(__FILE__), "20160209143102_add_foreign_travel_advice_subscriber_lists.csv")
+
+    CSV.foreach(csv_path) do |gov_delivery_id, country_slug|
+      SubscriberList.create!(
+        gov_delivery_id: gov_delivery_id,
+        links: {
+          countries: [country_slug]
```
elliotcm: There's a problem here, not immediately but it'll bite us later.

The `links` field is intended to house content IDs rather than slugs.  Slugs are still supported in the (deprecated) `tags` field but that's a legacy thing intended to match against a `tags` field within the `details` hash.

The original ticket did say slugs but was a little vague about the details of these fields.  We should really be using content IDs now.

This will still work for the specific use case because the travel advice publisher controls the format of the strings sent direct to the API and nothing else needs to trigger it, but it will break once we switch it over to the email alert service via the message bus since the contents of the links hash won't match.

bradwright: I was aware of that trade off - but because we haven't minted countries as a real thing yet (eg travel advice vs world locations etc.) I decided to park that entire conversation.

elliotcm: :+1:

tuzz: I may be missing some context? Couldn't we use the content_ids from here:
https://github.com/alphagov/travel-advice-publisher/blob/master/lib/data/countries.yml

bradwright: Going to address this to make it future proof,

PR #102, opened 2016-02-23
Remove a bandaid that no longer applies...

tuzz:
There was a bandaid added which meant that items
on the queue would work irrespective of whether
they were hashes or hashes encoded as strings. All
of these items should have now been processed by
the queue.


mobaig:
LGTM!

PR #103, opened 2016-02-24
Add document_type only query to email-alert-api

tuzz:
https://trello.com/c/gPjk5P2A/611-add-document-type-only-query-to-email-alert-api

### DIFF
--- a/spec/models/subscriber_list_spec.rb
+++ b/spec/models/subscriber_list_spec.rb
@@ -29,6 +29,46 @@ RSpec.describe SubscriberList, type: :model do
     end
   end
 
+  describe "validations" do
+    subject { FactoryGirl.build(:subscriber_list) }
+
+    it "is valid for the default factory" do
+      expect(subject).to be_valid
+    end
+
+    it "is invalid without either a document_type, tags or links" do
+      subject.document_type = ""
+      subject.tags = {}
+      subject.links = {}
+
+      expect(subject).to be_invalid
+    end
+
+    it "is valid with a document type but no tags or links" do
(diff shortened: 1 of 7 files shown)


spec/workers/notification_worker_spec.rb
```diff
@@ -102,6 +102,24 @@ def make_it_perform
       end
     end
 
+    context "when the subscriber list has no tags or links" do
+      before do
+        create(:subscriber_list, document_type: 'travel-advice', tags: {}, links: {})
+      end
+
+      it "sends a bulletin when document_type matches" do
```
bradwright: Could we add a test for when the bulletin has some tags as well?

bradwright: Also the case where it should trigger more than 1 list (eg "all countries" + "travel advice for spain")

tuzz:
Addressed @bradwright – please could you take another look.

spec/models/subscriber_list_spec.rb
```diff
@@ -29,6 +29,46 @@
     end
   end
 
+  describe "validations" do
+    subject { FactoryGirl.build(:subscriber_list) }
```
tijmenb: Would the tests be more independently readable if `subject` were changed to `let(:subscriber_list)`?

tuzz: Can do. I'm somewhat indifferent. `subject` is supposed to be the main object that you're testing (system under test).

bradwright:
Merging on the basis that this is feature complete and tested, and I want to turn some things off.

PR #108, opened 2016-04-05
Add health check for GovDelivery

tijmenb:
During maintenance this afternoon we realised that if our GovDelivery API credentials would stop working, we would perhaps not notice instantly. This PR proposes a canary for this situation.

It adds a check to the `/healthcheck` endpoint that will fail if GovDelivery returns anything other than `200` for a routine request. This will be picked up by [Icinga](https://github.com/alphagov/govuk-puppet/blob/359752e58e26f6dfc799e7915b0fb23924a52f19/modules/icinga/files/usr/lib/nagios/plugins/check_json_healthcheck).

Notes:

- I've used the `categories.xml` because it's unlikely to change, doesn't rely on state and is reasonably performant, since we don't use categories. But it's not ideal probably.
- This will raise an Nagios alert that _something_ is wrong with GovDelivery, but not really what.

Thoughts, @elliotcm @danielroseman @alphagov/team-finding-things ?

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -17,6 +17,9 @@ class HealthcheckController < ActionController::Base
             queue_retry_size: {
               status: queue_retry_size_status
             },
+            govdelivery: {
+              status: govdelivery_status
+            },
           },
           status: status
         }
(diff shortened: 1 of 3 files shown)


mattbostock:
:+1:

PR #110, opened 2016-04-15
Add govuk-request-id data atrribute to html email alert body

steventux:
https://trello.com/c/tFRwaUPB/26-ensure-request-id-is-set-in-message-body-by-email-alert-api

As part of ongoing monitoring for email alerts and other request tracing work we'd like to be able to trace an email alert from origin to delivery using the existing request id header created by the load balancer and used in other parts of the stack.
This change only adds a data attribute to an html body and no-ops with blank or text bodies.

See also work by @elliotcm for how this links to travel advice, publishing-api etc...
https://github.com/alphagov/govuk-puppet/pull/4320
https://github.com/alphagov/travel-advice-publisher/pull/139
https://github.com/alphagov/publishing-api/pull/269

paired with Pablo on this.

/cc @gpeng

### DIFF
--- a/lib/govuk_request_id.rb
+++ b/lib/govuk_request_id.rb
@@ -1,0 +1,23 @@
+class GovukRequestId
+  class << self
+    def insert(body)
+      return body unless body.present?
+
+      if body =~ /^</
+        body += govuk_request_id_html_comment
+      end
+
+      body
+    end
+
+  private
+
+    def govuk_request_id_html_comment
+      "\n<!-- govuk_request_id: #{govuk_request_id} -->\n"
+    end
+
+    def govuk_request_id
+      GdsApi::GovukHeaders.headers[:govuk_request_id]
+    end
+  end
+end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -10,8 +10,13 @@ class NotificationsController < ApplicationController
 private
 
   def notification_params
-    params.slice(:subject, :body, :from_address_id, :urgent, :header, :footer, :document_type)
+    params.slice(:subject, :from_address_id, :urgent, :header, :footer, :document_type)
       .merge(tags: params.fetch(:tags, {}))
       .merge(links: params.fetch(:links, {}))
+      .merge(body: notification_body)
+  end
+
+  def notification_body
+    GovukRequestId.insert(params[:body])
   end
 end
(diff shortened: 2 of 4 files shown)


lib/govuk_request_id.rb
```diff
@@ -0,0 +1,13 @@
+class GovukRequestId
+  def self.insert(body)
+    return body unless body.present?
+
+    doc = Nokogiri::HTML::DocumentFragment.parse(body)
```
gpeng: Maybe we'd be better adding the ID within an HTML comment which would mean we could avoid Nokogiri and any potential fragility that might introduce. We could treat body just as a string then and append it to the end.

pmanrubia: Good call @gpeng 👍 
Not sure if this applies here, but there are servers that remove comments before serving the HTML so we might be having issues here?
I am not strongly attached to Nokogiri and the data attribute, but it feels to me like a bit more robust solution thinking in the long term?

gpeng: We'd have to check that the comments got through the rest of the stack. There's also the issue of plain text emails. I don't know if we send them and if we do I don't know whether it would necessarily be a problem to append an identifier.

steventux: https://github.com/alphagov/govdelivery-email-templates/blob/master/README.md doesn't make mention of any plain text emails, though this doc isn't necessarily up-to-date or complete.

steventux: According to [govdelivery docs](http://developer.govdelivery.com/api/comm_cloud_v1/Default.htm#API/Comm%20Cloud%20V1/API_CommCloudV1_Bulletins_CreateandSendBulletin.htm) anything in the CDATA tags is untouched so if we were to use html comments instead then they should make it into the email body

pmanrubia: Fair enough. I was thinking on Govdelivery, or any automatic tool/sw within the validation loop, deciding to remove HTML comments 😃

steventux:
Thanks @pmanrubia @gpeng for review, I've simplified to an html comment. Not sure if we want to, or need to, detect html or not. Still need to decide if/how the id is presented in plain text emails.

gpeng:
Plain text will be a problem unless we either accept that the id will be visible in the message or we can work some way out that gives us access to the email headers. Something to ask in the meeting on Monday...

steventux:
@gpeng [we attempt to detect html in the body here](https://github.com/alphagov/email-alert-api/pull/110/files#diff-c77faca1663fea5b5ef6e2e6de2527a8R6) as the basis for adding the ID.
Having chatted with GovDelivery we are also confident that the plain text part they send is the text stripped from the html sent so this should also ensure the ID never gets sent in plain text mails.

gpeng:
:+1:

PR #111, opened 2016-04-20
Add govuk request id to email body...

steventux:
https://trello.com/c/tFRwaUPB/26-ensure-request-id-is-set-in-message-body-by-email-alert-api

Adds the govuk request id from GdsApi::Headers populated in the adapter
middleware into the email body as an html div with the request id as a data attribute.
No-op for non html email body content.

This is an amendment to https://github.com/alphagov/email-alert-api/pull/110/commits/b2ecfeb23f5471303c44c3519a56a06c49459ffb as html comments no not appear in the received emails.

### DIFF
--- a/lib/govuk_request_id.rb
+++ b/lib/govuk_request_id.rb
@@ -4,7 +4,7 @@ class GovukRequestId
       return body unless body.present?
 
       if body =~ /^</
-        body += govuk_request_id_html_comment
+        body += govuk_request_id_html
       end
 
       body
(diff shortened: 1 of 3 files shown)


elliotcm:
Should we use something like a `span` instead, just to be on the same side regarding rendered layout?

If you do (and you don't have to) then might as well rename the `govuk_request_id_html_comment` method since it's no longer an HTML comment.

steventux:
@elliotcm I've amended the method name to `govuk_request_id_html` - thanks for spotting this. :+1:

I asked the same question of block level elements altering the layout, the received wisdom was that `div` elements should have a 0x0px default. We'll do a few staging tests and see, the risk is some spurious style rules applied by an untested email client.

elliotcm:
@steventux Yeah there's the risk of odd clients but also if any whitespace makes its way in it will suddenly not be quite as invisible.  I'm on the fence, if you want to put it through as-is I'm fine merging this.

steventux:
@elliotcm in the interests of not accidentally introducing formatting issues for the purposes of monitoring I've `span`d it :+1:

PR #115, opened 2016-04-28
Document how to manually retry failed notifications

steventux:
/cc @alexmuller


alexmuller:
Looks like commands that work to me, thanks @steventux.

(cc @deanwilson)

PR #116, opened 2016-07-15
Upgrade Ruby to 2.1.4

alexmuller:
This is one of the last 2 apps running Ruby 2.1.2 on our machines.

I deliberately have only bumped this by the smallest amount I have to in order to get off 2.1.2. If you'd like to go further, feel free.

I haven't tested this change locally. The test failures don't seem to be related to this change.


elliotcm:
@alexmuller The issue on CI is that whatever DB user the test is running as doesn't have superuser rights to add the hstore extension.  This is added by puppet in prod but (I'm guessing) we can't do that here because it's frequently dropped and recreated.

I vaguely remember us fixing this last time by sshing into some CI box or other and adding it by hand, but I can't find creds to try that.  Could you help me out?

Once that's fixed I'll take on this upgrade and deploy, we can probably bump it to 2.3.0

elliotcm:
@alexmuller I'm looking at an alternative fix for this, closing for now.

elliotcm:
The alternative fix got stalled, it's in a new backlog and will be sorted at some point.

In the meantime we have another ticket to upgrade to 2.3.x so I'll do that in another PR.

PR #117, opened 2016-07-18
Add JSON columns to SubscriberList.

elliotcm:
This preps us for moving away from hstore which is
a postgres extension which makes puppet/CI more
complicated.

We're now on postgres 9.3+ which means we have JSON
columns available by default.


elliotcm:
Tests are broken on CI because of hstore faff which this PR is the first step towards fixing.  They're passing locally.

deanwilson:
This code looks fine. Merge away.

PR #118, opened 2016-07-18
Patch out hstore

elliotcm:
This is temporary until we can remove HSTORE, the
setup of the extension will now be done in CI itself.


deanwilson:
Merge away.

PR #119, opened 2016-07-21
Fix migration class name.

elliotcm:
Rails likes to make assumptions.


deanwilson:
Code change looks fine. Merge away

PR #126, opened 2016-08-16
Fix duplicate subscriptions.

elliotcm:
For *some reason* we have duplicate entries in the
subscriber list table for the same govdelivery
topic.

This removes all where the gov_delivery_id is null
which were specialist document subscriptions and
have been checked with Chris Patuzzo.

For the rest it removes all but the most recent.
The tags, links, and title all match for these, so
there should be no effect on the end user.

Finally, both Ruby and DB uniqueness constraints
are added to the subscriber list model.

### DIFF
--- a/app/models/subscriber_list.rb
+++ b/app/models/subscriber_list.rb
@@ -9,6 +9,7 @@ class SubscriberList < ActiveRecord::Base
   validate :tag_values_are_valid
   validate :link_values_are_valid
   validate :either_document_type_tags_or_links_present
+  validates :gov_delivery_id, uniqueness: true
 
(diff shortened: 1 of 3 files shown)


elliotcm:
@tijmenb looks like this goes against some work your team did: https://github.com/alphagov/email-alert-api/commit/26a5eccd7c8bed786e41cb6c76ef20c364f8ff74

tuzz:
👍  from me

elliotcm:
I'm going to close this pending discussion with @tijmenb and work around it in the sync script.

elliotcm:
Workaround: https://github.com/alphagov/email-alert-api/pull/127

PR #129, opened 2016-09-05
Query against JSON fields.

elliotcm:
PR https://github.com/alphagov/email-alert-api/pull/124
started reading from the temporary JSON fields,
but our query objects did not.

This switches the query over so we can replace the
underlying HStore columns.

### DIFF
--- a/app/queries/subscriber_list_query.rb
+++ b/app/queries/subscriber_list_query.rb
@@ -1,6 +1,9 @@
 class SubscriberListQuery
   def initialize(query_field: :tags)
-    @query_field = query_field
+    raise ArgumentError.new("query_field must be `:tags`, `:links`, or `:neither`") unless %i{tags links neither}.include?(query_field)
+
+    # Query against temporary JSON fields
+    @query_field = "#{query_field}_json".to_sym
   end
 
   # Find all lists in which all the links present have at least one match in the
@@ -42,9 +45,9 @@ class SubscriberListQuery
   end
 
   def subscriber_lists_with_key(key)
-    # This uses the `?` hstore operator, which returns true only if the hstore
-    # contains the specified key.
-    SubscriberList.where("#{@query_field} ? :key", key: key)
+    # This uses the `->` JSON operator to select lists only if the field's keys
+    # contain the specified key.
+    SubscriberList.where("(#{@query_field} -> :key) IS NOT NULL", key: key)
   end
 
 private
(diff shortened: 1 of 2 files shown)


app/queries/subscriber_list_query.rb
```diff
@@ -42,9 +45,11 @@ def find_exact_match_with(query_hash, document_type)
   end
 
   def subscriber_lists_with_key(key)
-    # This uses the `?` hstore operator, which returns true only if the hstore
+    # This uses the `ANY` array operator, which returns true only if the field's keys
     # contains the specified key.
-    SubscriberList.where("#{@query_field} ? :key", key: key)
+    SubscriberList.where(":key = ANY(ARRAY(SELECT json_object_keys(#{@query_field})))",
+      key: key,
+    )
```
MatMoore: I suspect I might be misunderstanding the query but can this be simplified to something like this?
`SubscriberList.where("#{@query_field} -> :key IS NOT NULL", key: key)`

app/queries/subscriber_list_query.rb
```diff
@@ -71,12 +76,14 @@ def subscriber_lists_with_keys_overlapping(query_hash)
...
     # contains all the specified keys.
-    SubscriberList.where("#{@query_field} ?& Array[:keys]", keys: query_hash.keys)
+    SubscriberList.where("ARRAY(SELECT json_object_keys(#{@query_field})) = Array[:keys]",
+      keys: query_hash.keys,
+    )
   end
 
   def subscriber_lists_without_tags_or_links
-    SubscriberList.where(tags: "", links: "")
+    SubscriberList.where("tags_json::text = '{}'::text AND links_json::text = '{}'::text")
```
MatMoore: I don't know if it will make any real difference, but you should also be able to write
`tags_json = '{}'::json AND links_json = '{}'::json` to skip some of that casting.

MatMoore:
I think the queries could be simplified slightly (see inline comments) but looks good to me. Feel free to merge 👍

app/queries/subscriber_list_query.rb
elliotcm: Looks like it needs to be `SubscriberList.where("(#{@query_field} -> :key) IS NOT NULL", key: key)` but yeah that should work.

app/queries/subscriber_list_query.rb
elliotcm: `PG::UndefinedFunction: ERROR:  operator does not exist: json = json`

:o

elliotcm: There's no JSON equality in 9.3, I think there is with JSONB in 9.4.

MatMoore: Ah! Fair enough.

elliotcm:
Updated where possible.

PR #131, opened 2016-09-15
Read from new json fields

elliotcm:
Next step once deployed is to remove the temporary *_json equivalents.


tijmenb:
I'm going to close this, as Elliot has left. I'll put it on a todo list for his successor.

elliotcm:
;_;

PR #132, opened 2016-09-29
Fix integration / staging sync task

36degrees:
See individual commits for details.

### DIFF
--- a/lib/tasks/data_hygiene.rake
+++ b/lib/tasks/data_hygiene.rake
@@ -7,10 +7,10 @@ task duplicate_record_for_tag: :environment do
 
   if from_topic_tag.blank?
     $stderr.puts "A from_topic_tag must be supplied"
-    exit 1
+    abort
   elsif to_topic_tag.blank?
     $stderr.puts "A to_topic_tag must be supplied"
-    exit 1
+    abort
   end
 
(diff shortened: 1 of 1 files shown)


thehenster:
It'd be nice if we didn't hardcode "UKGOVUK" in there but I can't think of an alternative. Even if we detected the presence of "DUP" we'd still need to substitute the string.. Probably the hardcoded value is a lesser evil than the combination of "DUP" being a reliable thing and the more complicated substitution rule.

LGTM.

thehenster:
It looks like there are two different templates for the sync job for integration and staging.

https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/data_sync_complete_integration.yaml.erb
https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/templates/jobs/data_sync_complete_staging.yaml.erb

It only needs to run in integration right? There is the option of guarding against running in other environments too.

36degrees:
It'll already only have any effect in integration because on staging the account code will match production. I could modify the jenkins jobs to pass another environment variable or something but it seems like overkill? Also, this is more generic in that it'd still work if we e.g. added another environment or changed staging to use a different account id.

PR #136, opened 2017-01-19
[DO NOT MERGE] Change handling of email alert titles

rubenarakelyan:
This commit splits the `title`, `short_name` and `description` attributes so that they can be sent separately to GovDelivery.

Since `title` and `short_name` have a 255 character limit enforced by GovDelivery, this commit adds the ability to pass a different, shorter string for these fields whilst keeping all the data in the `description` field.

https://github.com/alphagov/finder-frontend/pull/284 makes changes to `finder-frontend` so that a separate `short_name` is sent to `email-alert-api`.

Trello: https://trello.com/c/KTHRa4a0/422-fix-email-alert-api-errors-when-the-email-alert-short-name-is-longer-than-255-characters

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -29,7 +29,11 @@ class SubscriberListsController < ApplicationController
 
 private
   def build_subscriber_list
-    gov_delivery_response = Services.gov_delivery.create_topic(params[:title])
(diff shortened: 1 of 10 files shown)


rubenarakelyan:
I'm closing this because it's becoming too complicated and the email team are looking at changing this all anyway. I'll put in a smaller change to get the CMA ones working.

PR #139, opened 2017-02-08
Remove temporary *_json fields

dwhenry:
These `links_json` and tags_json` fields where added 
for the `hstore` to `json` data type migration which 
has now been completed, so we can now remove this fields.

As an additional task we need to manually remove the 
`hstore` extension from the postgres database instances.


jennyd:
Puppet PR to remove hstore: https://github.com/alphagov/govuk-puppet/pull/5458

PR #142, opened 2017-02-17
Speed up the data sync process

dwhenry:
# Tests to follow

This checks the records on GovDelivery against
what is in the database. Any matches on GovDelievryID
and Title are left with any changed being deleted or
created as appropriate.

The delete code runs in threads, this is a temporary
measure as it may be required to purge GovUkDelivery
data (approx 16000 records) and should be removed once
all data has been moved across.

https://trello.com/c/0DFyGhHj/114-rethink-integration-email-alert-api-rake-task-to-sync-topics


lib/tasks/data_hygiene.rake
```diff
@@ -51,42 +51,65 @@ task sync_govdelivery_topic_mappings: :environment do
 
   puts "Fetching topics.."
   topics = Services.gov_delivery.fetch_topics["topics"]
+  topics = topics.map { |topic| [topic["name"], topic["code"]] }
+  subscriber_lists = SubscriberList.where.not(gov_delivery_id: nil)
+                                   .pluck(:title, :gov_delivery_id)
+                                   .map { |title, gov_delivery_id| [(title || gov_delivery_id).strip, gov_delivery_id] }
```
jennyd: For the subscriber lists without titles, GovDelivery gives them names like "MISSING TITLE UKGOVUKDUP_391", so the first time this runs all of these will be considered to have changed. It's a good thing to give them predictable titles ourselves so I think that's fine. We should also investigate at some point why we have lists without titles in the database, but that's a different question...

lib/tasks/data_hygiene.rake
```diff
@@ -51,42 +51,65 @@ task sync_govdelivery_topic_mappings: :environment do
...
+      end
     end
+    threads.each(&:join)
 
     # GovDelivery delete topics asynchronously and/or are only eventually
     # consistent on deletes.  We have to wait until all the deletes take effect
     # or we'll get conflicts when trying to recreate them.
-    attempts = 0
-    while Array(Services.gov_delivery.fetch_topics["topics"]).count > 0
+    puts 'Wainting for 30 seconds for topics to be asynchronously deleted'
```
jennyd: typo - "Wainting"

lib/tasks/data_hygiene.rake
```diff
@@ -51,42 +51,65 @@ task sync_govdelivery_topic_mappings: :environment do
...
   else
-    puts "Deleting all remote topics.."
-    topics.each do |topic|
-      puts "-- Deleting #{topic["name"]} (#{topic["code"]})"
-      Services.gov_delivery.delete_topic(topic["code"])
+    puts "Deleting remote topics.."
+    threads = to_be_deleted.each_slice(DataHygiene::BATCH_SIZE).map do |batch|
+      Thread.new do
+        batch.each do |name, code|
+          puts "-- Deleting #{name} (#{code})"
```
jennyd: Does `puts` work reliably now in threads? This suggests it used not to: http://stackoverflow.com/questions/5044299/whats-the-easiest-way-to-print-output-from-parallel-operations-in-ruby-without

config/gov_delivery.yml
```diff
@@ -1,7 +1,7 @@
 development:
   username: govdelivery@example.com
   password: nottherealpassword
-  account_code: UKGOVUK
```
jennyd: Did you mean to include this change?

lib/tasks/data_hygiene.rake
```diff
@@ -51,42 +51,65 @@ task sync_govdelivery_topic_mappings: :environment do
...
-      end
+      print '.'
     end
   end
 
-  puts "Creating remote topics to match the #{SubscriberList.count} local topics.."
-  created = {}
-  SubscriberList.find_each do |list|
-    next if created.has_key?(list.gov_delivery_id)
+  3.times.each do |i|
```
jennyd: We should check whether there are any topics to be created before going into this loop.

lib/tasks/data_hygiene.rake
```diff
@@ -51,42 +51,65 @@ task sync_govdelivery_topic_mappings: :environment do
...
+end
 
-    created[list.gov_delivery_id] = true
+module DataHygiene
+  # will hopefully limit delete runtime to 1 per sec * 300 / 60 =~ 5 minutes
+  # a maximum thread count of 16000 / 300 =~ 54
+  BATCH_SIZE = 300
+
+  def self.create_topics(list)
+    list.each do |gov_delivery_id, title|
```
jennyd: Isn't this the wrong way round? We build these lists with `title, gov_delivery_id` above.

lib/tasks/data_hygiene.rake
```diff
@@ -51,42 +51,65 @@ task sync_govdelivery_topic_mappings: :environment do
...
+      title = title || "MISSING TITLE #{gov_delivery_id}"
```
jennyd: This isn't compatible with the way we check for matching titles above - I don't mind which style of title we use but we need to be consistent :)

config/gov_delivery.yml
dwhenry: Let talk about this as I feel it should be UKGOVUK for development, but that means we need to support different values in staging/production so might need to add a new confirm variable somewhere.

spec/lib/data_hygiene/data_sync_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    create(:subscriber_list, title: 'Topic A', gov_delivery_id: 'TA')
+    create(:subscriber_list, title: 'Topic B', gov_delivery_id: 'TB')
+    create(:subscriber_list, title: 'Topic C', gov_delivery_id: 'TC')
+
+    expect(gov_delivery_client).to receive(:create_topic).with('Topic C', 'TC')
+
+    subject.run
+  end
+
+  it "will detele and recreate topcis that don't match on title and gov_devlivery_id" do
```
jennyd: A few typos in this line

spec/lib/data_hygiene/data_sync_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:logger) { double(:logger, info: true, warn: true) }
+  let(:topics) do
+    [{ 'name' => 'Topic A', 'code' => 'TA' },{ 'name' => 'Topic B', 'code' => 'TB' }]
+  end
+  let(:gov_delivery_client) { double(:gov_delivery_client, fetch_topics: { 'topics' => topics }) }
+  subject { described_class.new(logger, 0) }
+
+  before do
+    allow(ENV).to receive(:[]).and_return('')
+    allow(ENV).to receive(:[]).with('ALLOW_GOVDELIVERY_SYNC').and_return('allow')
```
jennyd: It would be good to test the `validate!` method as well, to be sure that we can't run this in production.

lib/data_hygiene/data_sync.rb
```diff
@@ -0,0 +1,152 @@
...
+      logger.info "Retrieving data from servers.."
+
+      # determine which topics/subscriber_lists already exist and which have been
+      # modified/created since the last data sync
+      matching = topics & subscriber_lists
+      to_be_deleted = topics - subscriber_lists
+      to_be_created = subscriber_lists - topics
+
+      # filter out any duplicate ID with at least one match from the list to be delete/created
+      # - this avoids errors during creation and avoid unnecessary deletions and creations
```
jennyd: I think this is to handle rows with duplicate codes but different titles, because one of those may exist in GovDelivery already so we don't want to try to create/delete another topic with the same id - is that right? Could you expand this comment? It took me a while to work out why we'd need to do this.

dwhenry: copied comment into codebase

lib/data_hygiene/data_sync.rb
```diff
@@ -0,0 +1,152 @@
...
+        )
+
+        logger.info "Done."
+      end
+    end
+
+    def subscriber_lists
+      @subscriber_lists ||= SubscriberList
+        .where.not(gov_delivery_id: nil)
+        .distinct
```
dwhenry: the distinct applies to the fields that are plucked.. :)

spec/lib/data_hygiene/data_sync_spec.rb
dwhenry: add tests for when code should NOT run.

lib/data_hygiene/data_sync.rb
```diff
@@ -0,0 +1,152 @@
...
+          end
+        end
+        threads.each(&:join)
+
+        # GovDelivery delete topics asynchronously and are only eventually
+        # consistent on deletes. We wait here to give the deletes a chance to
+        # process, and then perform a retry if the create fails
+        logger.info 'Waiting for 30 seconds for topics to be asynchronously deleted'
+
+        @delete_wait.times do
```
dwhenry: As discussed this is a hand-wavy thing as we know GovDelivery is eventual consistent, but have no real feel for timing.

spec/lib/data_hygiene/data_sync_spec.rb
dwhenry: Fixed and added an additional test for when records with topic ID and different titles.

lib/data_hygiene/data_sync.rb
jennyd: good point :)

lib/data_hygiene/data_sync.rb
jennyd: Ok, let's stick with this and keep an eye on it to see if timing is a problem in either Staging or Integration.

PR #141, opened 2017-02-14
Delete SubscriberList's without subscribers on GovDelivery

dwhenry:
For each SubscriberList query GovDelivery for the 
count of subscribers and filter out those that have 
a count of 0 so they can be deleted. Any subscriber 
lists that fail to successfully get data from GovDelivery
are logged for future investigation.

PR #144, opened 2017-02-21
Default footer

dwhenry:
This is pending #141 being merged

Send a default footer via the API. This footer can be overwritten if the source app passed in a footer (special publisher)

This PR also fixes a bug related to squiggly heredocs (<<~) and ruby 2.3.0

https://trello.com/c/hRFWQxHc/77-make-footers-consistent

### DIFF
--- a/app/services/gov_delivery/request_builder.rb
+++ b/app/services/gov_delivery/request_builder.rb
@@ -35,10 +35,14 @@ module GovDelivery
             xml.cdata options[:header]
           } if options[:header]
           xml.footer {
-            xml.cdata options[:footer]
-          } if options[:footer]
+            xml.cdata options[:footer] || default_footer
+          }
         }
       }.to_xml
+    end
+
+    def self.default_footer
+      @default_footer ||= File.read(File.expand_path(File.join(__dir__, 'templates', 'default_footer.html')))
     end
   end
 end
--- a/app/services/gov_delivery/request_builder.rb
+++ b/app/services/gov_delivery/request_builder.rb
@@ -35,7 +35,7 @@ module GovDelivery
             xml.cdata options[:header]
           } if options[:header]
           xml.footer {
-            xml.cdata options[:footer] || default_footer
+            xml.cdata default_footer
           }
         }
       }.to_xml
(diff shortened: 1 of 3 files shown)


app/services/gov_delivery/request_builder.rb
```diff
@@ -35,10 +35,14 @@ def self.send_bulletin_xml(topic_ids, subject, body, options = {})
...
-            xml.cdata options[:footer]
-          } if options[:footer]
+            xml.cdata options[:footer] || default_footer
+          }
         }
       }.to_xml
     end
+
+    def self.default_footer
+      @default_footer ||= File.read(File.expand_path(File.join(__dir__, 'templates', 'default_footer.html')))
```
MatMoore: Was wondering if it's cleaner to use rails views for this, rather than adding a templates dir. I'm fine with either approach though.

spec/fixtures/bulletin_post.xml
```diff
@@ -0,0 +1,40 @@
+<bulletin>
```
MatMoore: These don't feel like fixtures, since they describe the outcome of a test, rather than setting it up.

I think it would read more clearly if you built the expectation in the test (as before) and just substituted in the footer from `RequestBuilder.default_footer`.

like

```ruby
expect(req.body).to be_equivalent_to(
  %{
    <bulletin>
    <!-- some xml -->
    <footer><![CDATA[#{GovDelivery::RequestBuilder.default_footer}
]]></footer>
    </bulletin>
  }
```

app/services/gov_delivery/request_builder.rb
dwhenry: I think you are correct a rails view might be nicer, however I left it as is for the moment as this keeps it in line with the GovUkDelivery (Python app) that has the same file. Ideally, this should be moved once we have killed GovUkDelivery.

spec/fixtures/bulletin_post.xml
dwhenry: Done. :)

app/services/gov_delivery/request_builder.rb
MatMoore: Cool, feel free to merge 👍

PR #145, opened 2017-02-23
Log email criteria

dwhenry:
Start doing some basic logging for each notification that is received, as well as creating an endpoint for external (govuk_delivery) apps to send logs in.

This will allow us to track how notifications are matched to GovDelivery topics in both apps.
https://trello.com/c/b9ZonGXn/121-logging


db/migrate/20170221141514_create_notification_log.rb
```diff
@@ -0,0 +1,19 @@
+class CreateNotificationLog < ActiveRecord::Migration
+  def change
+    create_table :notification_logs do |t|
```
tijmenb: Perhaps we can set `null: false` for most/all of these fields so we won't need a lot of validation on the requests from `govuk_delivery`.

app/models/notification_log_reporting.rb
```diff
@@ -0,0 +1,45 @@
...
+  # Generate a list of duplicated gov request ID coming through from a single app
+  # this would be a sign we are sending emails twice.
+  def duplicates
+    @scope
+      .group(:app, :gov_request_id)
+      .having('COUNT(1) > 1')
+      .pluck(:app, :gov_request_id, 'ARRAY_AGG(gov_delivery_ids)')
+  end
+
+  # Report when a notification is sent by GovUkDelivery and not by EmailAlertApi.
```
MatMoore: Assuming `gov_delivery_ids` is the topics, isn't this the number of topics that aren't notifying via EmailAlertApi, rather than the number of notifications itself?

app/controllers/notification_logs_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+private
+
+  def notification_log_params
+    params.permit(
+      :gov_request_id,
+      :content_id,
+      :public_updated_at,
+      :document_type,
+      :app,
```
MatMoore: Is there a better name than just app for this? Like `email_notification_app` or something?

dwhenry: Have changed this to emailing_app

db/migrate/20170221141514_create_notification_log.rb
dwhenry: @tijmenb @MatMoore I was avoiding adding validations on this model as:

* I don't want the process to fail if this model fails to save. A logging failure should not trigger an email to be resent (i.e. don't raise an error).
* I would prefer to save all the records (including invalid ones) so that I can understand what missing/invalid data is occurring and identify issues in the process. Once we have a better view of the data I would potentially look to add some validation at that stage.

tijmenb: Make a lot of sense!

app/models/notification_log_reporting.rb
dwhenry: Yes, that is correct. I have added a comment to try and clarify my intention, let me know if it is  still unclear.

spec/models/notification_log_reporting_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        create(govuk_request_id: govuk_request_id, emailing_app: 'gov_uk_delivery')
+
+        expect(subject.missing).to eq({})
+      end
+    end
+  end
+
+  context '#different' do
+    context 'when notification for GovUkDelivery and not for EmailAlertApi' do
+      it 'does not detects any differences' do
```
MatMoore: s/detects/detect

spec/models/notification_log_reporting_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        govuk_request_id = SecureRandom.uuid
+        create(govuk_request_id: govuk_request_id, emailing_app: 'gov_uk_delivery')
+
+        expect(subject.different).to eq({})
+      end
+    end
+
+    context 'when notification for GovUkDelivery and EmailAlertApi which is sent to different gov_delivery_ids lists' do
+      it 'detects notificatiosn with different gov delivery ids' do
```
MatMoore: typo in "notifications"

spec/models/notification_log_reporting_spec.rb
dwhenry: fixed

spec/models/notification_log_reporting_spec.rb
dwhenry: fixed

PR #146, opened 2017-02-27
Also delete topics which 404 as well as ones with 0 subscribers

jennyd:
Deleting these cleans up our data and also may fix some errors for users. If we have a subscriber list in our database which doesn't exist in GovDelivery, we'll redirect users to GovDelivery to sign up but that will fail because they don't recognise the topic id. We don't have any of these right now in this app, but since GovDelivery data can be modified in their UI it seems useful to support this here as well as in govuk-delivery (where we have about 2000 topics in this state).

This changes the error-raising behaviour of the client but this shouldn't cause any problems elsewhere in the application because `TopicNotFound` is a subclass of `UnknownError` which should already be handled.

It's not very nice to have to check for a string zero, but now that we're using `parse_topic_response` for the extra error handling we'd need to change the [behaviour of `ResponseParser`](https://github.com/alphagov/email-alert-api/blob/f5558d1a09cd74528740883c9374ec2075fe611d/app/services/gov_delivery/response_parser.rb#L30-L32) to fix this, and that's outside the scope of this PR (but worth looking at separately).

https://trello.com/c/Ruf2l6R9/120-also-delete-topics-from-our-databases-if-they-don-t-exist-in-govdelivery

### DIFF
--- a/lib/data_hygiene/delete_unneeded_topics.rb
+++ b/lib/data_hygiene/delete_unneeded_topics.rb
@@ -1,7 +1,7 @@
 require 'thread'
 
 module DataHygiene
-  class DeleteTopicsWithoutSubscribers
+  class DeleteUnneededTopics
     attr_reader :client
     delegate :puts, to: :@output
     delegate :gets, to: :@input
--- a/lib/data_hygiene/delete_unneeded_topics.rb
+++ b/lib/data_hygiene/delete_unneeded_topics.rb
@@ -26,15 +26,19 @@ module DataHygiene
         with_gov_delivery_error.each { |sl| puts "#{sl.gov_delivery_id} - #{sl.title}"}
       end
 
-      puts ''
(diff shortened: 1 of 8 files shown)


lib/data_hygiene/delete_unneeded_topics.rb
```diff
@@ -38,43 +48,58 @@ def subscriber_lists
...
+        .select { |c, s| c == :unknown_error }
         .map(&:last)
     end
 
   private
 
     def subscriber_count(gov_delivery_id)
       topic = client.fetch_topic(gov_delivery_id)
-      topic && topic['subscribers_count']
+      topic.subscribers_count
```
MatMoore: Could you convert to integers at this point? Otherwise, would be useful to have a comment somewhere that explains this value is intended to be a string for now.

jennyd: It depends how much we want to trust GovDelivery to only return a number in this field - `'abc'.to_i` also returns `0` which might mean that we end up deleting some topics which returned some funky data instead of actually having 0 subscribers. I've added a comment about it in a separate commit.

MatMoore: Ah... ruby. 😞

PR #149, opened 2017-03-16
Fix create subscriber lists testing

dwhenry:
Clean up the tests, inline helper that is only called
once, remove random un-necessary references to the
helper and add a test to check we are actually doing
the thing we need.

https://trello.com/c/R9GjkEJ6/118-1-write-tests-linking-201-response-from-email-alert-api-to-actually-creating-a-topic-in-the-database-and-govdelivery


mobaig:
👍

jennyd:
This is definitely an improvement, but we're still not testing that the newly created topic is stored in our database.

PR #150, opened 2017-03-17
Stop GD-12004 errors being retried by sidekiq

dwhenry:
GD-12004 - No subscribers for topics, should not be
automatically retried by the sidekiq worker.

This change ensures that we are still notified via
Airbrake/Errbit when an error occurs but swallows 
the error so that it doesn't trigger a retry.


jennyd:
I'm a bit confused now about why this is necessary, since [`send_bulletin` rescues `ZeroSubscriberError`](https://github.com/alphagov/email-alert-api/blob/f7a456e6fe1db7195ec6d8baf11ba8ab7ffdca9a/app/services/gov_delivery/client.rb#L64) - I can't see how we can get into this situation...

PR #151, opened 2017-03-17
Refactor main query method to be easier to parse

dwhenry:
Use better naming and expand implementation to help 
parse the method and understand what it is doing

* Use postgres `<@` method instead of `&&` for 
pre-filtering [https://www.postgresql.org/docs/9.3/static/functions-array.html]
* Better method and variable naming


spec/queries/top_query_spec.rb
```diff
@@ -0,0 +1,24 @@
+# require 'rails_helper'
```
mobaig: This looks a bit wrong.

app/queries/find_exact_match.rb
```diff
@@ -0,0 +1,39 @@
+class FindExactMatch
+  def initialize(query_field: :tags)
+    raise ArgumentError.new("query_field must be `:tags` or `:links`") unless %i{tags links}.include?(query_field)
+
+    @query_field = query_field.to_sym
+  end
+
+
+  def call(query_hash, document_type)
```
mobaig: I'd make query_hash a keyword arg for greater clarity at call sites. But not document type, that's probably fine as it is.

app/controllers/subscriber_lists_controller.rb
```diff
@@ -37,13 +37,13 @@ def build_subscriber_list
   end
 
   def find_subscriber_list
```
mobaig: With the shorter method names you've introduced I'd be even more inclined to golf this entire method down to 3 lines, for brevity.

app/queries/find_exact_match.rb
```diff
@@ -0,0 +1,39 @@
...
+    return [] unless query_hash.present?
+
+    subscriber_lists = subscriber_lists_with_all_matching_keys(query_hash)
```
mobaig: Appreciate this is an existing method, but `subscriber_lists_where_all_keys_match(query_hash)` reads better.

spec/queries/find_exact_match_with_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
```
mobaig: spec name

app/queries/has_topic_value.rb
```diff
@@ -0,0 +1,19 @@
+class HasTopicValue
```
mobaig: Rename suggestion: `FindByTopic`, with method signature `call(topic:)`.

app/queries/with_one_matching_value_for_each_key.rb
```diff
@@ -0,0 +1,45 @@
+class WithOneMatchingValueForEachKey
```
mobaig: It's late, and I might be misunderstanding this, but `WithAtLeastOneMatchingValue` is more accurate, I think.

app/queries/with_one_matching_value_for_each_key.rb
```diff
@@ -0,0 +1,45 @@
...
+  def initialize(query_field: :tags, scope: SubscriberList)
+    raise ArgumentError.new("query_field must be `:tags` or `:links`") unless %i{tags links}.include?(query_field)
+
+    @query_field = query_field.to_sym
+    @scope = scope
+  end
+
+  # Find all lists in which at least one match in the supplied list.
```
mobaig: `in which at least one value in query_field matches a value in the supplied hash`

app/queries/with_one_matching_value_for_each_key.rb
```diff
@@ -0,0 +1,45 @@
...
+        (
+          content_item_tags_or_links[key] & subscriber_list_tags_or_links[key]
+        ).any?
+      end
+    end
+  end
+
+private
+
+  # Return all SubscriberLists which have a subscet of keys from those requested.
```
mobaig: typo: subscet

spec/queries/top_query_spec.rb
dwhenry: deleted

app/queries/find_exact_match.rb
dwhenry: changed to `subscriber_lists_where_all_keys_present` to keep the right intent

app/controllers/subscriber_lists_controller.rb
dwhenry: done

spec/queries/find_exact_match_with_spec.rb
dwhenry: done

app/queries/has_topic_value.rb
dwhenry: done

app/queries/with_one_matching_value_for_each_key.rb
dwhenry: I have changed this to `WithMatchingValuesForAllKeys` its a bit wordy but it's also not a super easy concept

app/queries/with_one_matching_value_for_each_key.rb
dwhenry: fixed

app/queries/find_exact_match.rb
dwhenry: I understand your reasoning here, however I would prefer to leave this as is for the moment as we will need to add `email_document_supertype` and `government_document_supertype` to the argument list. I would prefer to update the interface when we do that PR. (which is what I am working on today)

app/queries/with_one_matching_value_for_each_key.rb
dwhenry: Made this into an epic.. not sure it is better or not..

dwhenry:
```
  # Filter by subscriber lists where the keys from the supplied hash (content_item_tags_or_links)
  # are a superset of the keys in corresponding subscriber list field (tags or links) where
  # the array of values for each key (on the subscriber list) have at least one matching value
  # with the array of values from the supplied hash for the matching key.
  # ** This means that not keys from the the supplied hash are required to be matched.
```

app/queries/with_matching_values_for_all_keys.rb
```diff
@@ -1,13 +1,16 @@
-class WithOneMatchingValueForEachKey
+class WithMatchingValuesForAllKeys
```
mobaig: It occurs to me that this class name is probably trying too hard to explain the implementation rather than what this object is for. What if you simply called it something like `MatchedForNotification`, leaving the explanation of the (not exactly simple) query in the comment?

app/queries/with_matching_values_for_all_keys.rb
```diff
@@ -1,13 +1,16 @@
...
     @scope = scope
   end
 
-  # Find all lists in which at least one match in the supplied list.
-  # Note - does not require that all the supplied hash is matched.
+  # Filter by subscriber lists where the keys from the supplied hash (content_item_tags_or_links)
+  # are a superset of the keys in corresponding subscriber list field (tags or links) where
+  # the array of values for each key (on the subscriber list) have at least one matching value
+  # with the array of values from the supplied hash for the matching key.
+  # ** This means that not keys from the the supplied hash are required to be matched.
```
mobaig: Agree that fleshing this out is better. Slightly terser and with a fixed typo:

```
Filter out subscriber lists:
- where the keys from the supplied hash are a superset of the keys in the specified query_field AND
- the key values in the subscriber list have at least one corresponding match in the supplied hash.
Note that this means that not all keys from the the supplied hash are required to be matched.
```

app/queries/with_matching_values_for_all_keys.rb
dwhenry: Good shout

app/queries/with_matching_values_for_all_keys.rb
dwhenry: done

spec/queries/find_exact_match_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      :subscriber_list,
+      tags: {
+        topics: ["oil-and-gas/licensing"],
+        organisations: ["environment-agency", "hm-revenue-customs"]
+      },
+      document_type: "policy",
+    )
+  end
+
+  it "requires all tag types in the document to be present in the list" do
```
dwhenry: This was copied from the existing tests. It will need to be refactored with the supertype changes so will try and make it clearer when I do that.

PR #153, opened 2017-03-17
Optional gov_delivery_id when creating a subscriber_list

dwhenry:
If a gov_delivery_id is passed in when creating a 
subscriber list, then then subscriber list is marked
as being disabled by default.

https://trello.com/c/PKxw58Hu/126-1-backfill-existing-topics-into-email-alert-api-db-disabled


jennyd:
I haven't tried running this, but it looks fine 👍

PR #154, opened 2017-03-22
Add supertype fields

dwhenry:
Add supertype handling

This enables the following functionality:

* create subscriber list with supertype criteria
* subscriber list filtering now honours supertype
criteria
* added testing around edge cases

In addition this change fixes a bug where searching
for a subscriber list with links or tags that failed
to find a match would fall back to searching by
document type (where links and tags are blank).

https://trello.com/c/5yVMib2f/124-2-add-support-for-whitehall-topics-in-new-email-stack


app/queries/find_exact_query.rb
```diff
@@ -0,0 +1,32 @@
...
+    @tags = tags.symbolize_keys
+    @links = links.symbolize_keys
+    @document_type = document_type
+    @email_document_supertype = email_document_supertype
+    @government_document_supertype = government_document_supertype
+  end
+
+  def exact_match
+    if @links.any? || @tags.any?
+      find_exact(:links, @links) || find_exact(:tags, @tags)
```
jennyd: I think this would be a bit clearer if we split `@tags` into an `elsif` - as it is, if there are tags but no links we'll call `find_exact` for links unnecessarily.

app/workers/notification_worker.rb
```diff
@@ -6,12 +6,13 @@ def perform(notification_params)
...
 
     query = SubscriberListQuery.new(
-      tags_hash,
-      links_hash,
-      document_type,
+      tags: tags_hash,
+      links: links_hash,
+      document_type: notification_params[:document_type],
+      email_document_supertype: notification_params[:email_document_supertype],
+      government_document_supertype: notification_params[:government_document_supertype],
```
jennyd: The notifications controller isn't passing the supertype params to the worker yet - is that intentional?

spec/models/subscriber_list_spec.rb
```diff
@@ -36,36 +36,28 @@
       expect(subject).to be_valid
     end
 
-    it "is invalid without either a document_type, tags or links" do
-      subject.document_type = ""
-      subject.tags = {}
-      subject.links = {}
+    it "is valid when tags are an array" do
```
dwhenry: done

spec/queries/subscriber_list_query_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'rails_helper'
+
+RSpec.describe SubscriberListQuery do
+  subject do
+    described_class.new(
+      tags: { polices: ['eggs'] },
```
dwhenry: corrected

spec/requests/create_subscriber_list_spec.rb
```diff
@@ -134,14 +136,21 @@
         expect(response.status).to eq(201)
       end
     end
+  end
+
+  context "when creating a subscriber list with 'email' and 'government' document supertypes" do
+    it "returns a 201" do
+      create_subscriber_list(
+        email_document_supertype: "publications",
+        government_document_supertype: "health",
```
jennyd: Could we use an actual gov supertype here eg `news-stories`, to avoid confusion?

app/workers/notification_worker.rb
dwhenry: nope - this is added with a test

spec/requests/create_subscriber_list_spec.rb
dwhenry: done everywhere...

app/queries/find_exact_query.rb
dwhenry: done

spec/models/subscriber_list_spec.rb
```diff
@@ -36,36 +36,28 @@
       expect(subject).to be_valid
     end
 
-    it "is invalid without either a document_type, tags or links" do
```
dwhenry: done the test..

spec/queries/find_exact_query_spec.rb
```diff
@@ -10,14 +10,14 @@
     end
 
     it "matched when subscriber list has the same links and matching document_type" do
-      query = build_query(links: { policies: ['aa-11'] }, document_type: 'document_type')
-      subscriber_list = create_subscriber_list(links: { policies: ['aa-11'] }, document_type: 'document_type')
+      query = build_query(links: { policies: ['aa-11'] }, document_type: 'travel_advise')
```
jennyd: s/travel_advise/travel_advice

spec/queries/subscriber_list_query_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+        government_document_supertype: 'other',
+      )
+    end
+  end
+
+  context 'when matching has tags fields' do
+    it_behaves_like "#links matching", { tags: { policies: ['eggs'] }, links: {} }
+
+    it "excluded when non-matching tags" do
+      subscriber_list = create_subscriber_list(tags: { policies: ['apples'] }, document_type: 'document_type')
```
jennyd: In this test and the next one, shouldn't `document_type` match the query one (or be empty) for this to be testing what it claims to?

dwhenry: it doesn't effect the test to have it but you are correct it deviates from the description so I have removed it.

jennyd: Actually we don't have any tests yet for this query for the cases where the list has no document/super type - only for ones where it matches or is a different type, rather than being `''`.

dwhenry: I can add some.. maybe.. give me a second..

dwhenry: we do have tests.. it just not explicitly, anywhere the fields are not set they have `''` in the database. I will make this more obvious in helper methods.

jennyd: Ah yes... it is pretty confusing, but less so than not using shared examples 👍

jennyd:
Yay \o/ 🎉 ➕ 🎹 🏑

PR #155, opened 2017-03-23
Support content items where the tags hash values are not arrays

dwhenry:
This is required to support `specialist publisher` which send through
the `format` criteria in the `tags` hash as a string value insteade of
an array of strings.

We have revert this to the previous format rathger than fixing the
bug at source as we are not able to get 100% confidence that this
is the only source of data in this format.

### DIFF
--- a/spec/queries/matched_for_notification_spec.rb
+++ b/spec/queries/matched_for_notification_spec.rb
@@ -69,6 +69,26 @@ RSpec.describe MatchedForNotification do
       end
     end
 
+    context "Specialist publisher edge case" do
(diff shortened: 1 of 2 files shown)


spec/queries/matched_for_notification_spec.rb
```diff
@@ -69,6 +69,26 @@ def execute_query(field:, query_hash:)
       end
     end
 
+    context "Specialist publisher edge case" do
+      let!(:subscriber_list) { create(:subscriber_list, tags: { format: ["employment_tribunal_decision"] }) }
+
+      it "finds the list when the criteria values is a string that is present in the subscriber list valeus for the field" do
```
dwhenry: Fixed.

PR #156, opened 2017-03-23
Use `gov_delivery_id` when finding exact match

dwhenry:
When we create a subscriber list via the api it first 
does a lookup to see if a subscriber list with the
same criteria already exists.

With the migration of whitehall data it is possible 
that multiple subscriber lists with different 
gov_delivery_id will exists. As this is the case
it is important that when we are filtering we get 
the correct record. 

This is more important during the create process as 
otherwise you would find a subscriber list with same
criteria, and would skip the creation of a record with
the desired gov_delivery_id resulting in users not
being notified.

example:

```
'https://www.gov.uk/government/feed?topics%5B%5D=uk-economy' has topic_id UKGOVUK_6888
'https://www.gov.uk/government/topics/uk-economy.atom' has topic_id UKGOVUK_8609
```

https://trello.com/c/5yVMib2f/124-2-add-support-for-whitehall-topics-in-new-email-stack

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -2,7 +2,7 @@ require "gov_delivery/client"
 
 class SubscriberListsController < ApplicationController
   def show
-    subscriber_list = find_subscriber_list
+    subscriber_list = FindExactQuery.new(find_exact_query_params).exact_match
     if subscriber_list
       respond_to do |format|
         format.json { render json: subscriber_list.to_json }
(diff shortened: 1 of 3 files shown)


jennyd:
I'm not sure why we'd have lists with duplicated criteria but different govdelivery ids, can we talk about it tomorrow?

dwhenry:
@jennyd it is caused by changes to the URLs that have been used over time:

```
"https://www.gov.uk/government/feed?topics%5B%5D=uk-economy",2013-10-23T11:51:51Z,"UKGOVUK_6888"
"https://www.gov.uk/government/topics/uk-economy.atom",2014-02-18T10:24:03Z,"UKGOVUK_8609"
```

jennyd:
Ah, ok - could you add that example to the commit message?

dwhenry:
@jennyd done

dwhenry:
@jennyd updated comment

PR #158, opened 2017-03-31
Add subscriber list ID to the title on gov delivery

dwhenry:
This is required to avoid name collisions which result in errors
as the name is required to be unique on gov delivery.

### DIFF
--- a/app/services/subscriber_list_creator.rb
+++ b/app/services/subscriber_list_creator.rb
@@ -1,0 +1,26 @@
+class SubscriberListCreator
+  attr_reader :record
+
+  def initialize(params)
+    @params = params
+  end
+
+  def save
+    @record = SubscriberList.new(@params)
+    return false unless @record.save
+
+    # if the topic_id is passed in use the provided title.
+    return true if @record.gov_delivery_id.present?
+
+    title_append_string = " (#{@record.id})"
+    @record.title = "#{@record.title[0..(254 - title_append_string.size)]}#{title_append_string}"
+    gov_delivery_id = build_on_gov_delivery(@record)
+    @record.gov_delivery_id = gov_delivery_id
+    @record.save
+  end
+
+  def build_on_gov_delivery(record)
+    gov_delivery_response = Services.gov_delivery.create_topic(record.title)
+    gov_delivery_response.to_param
+  end
+end
--- a/spec/services/subscriber_list_creator_spec.rb
+++ b/spec/services/subscriber_list_creator_spec.rb
@@ -1,0 +1,95 @@
+require "rails_helper"
+
+RSpec.describe SubscriberListCreator do
+  context "when gov delivery ID is not passed in" do
+    let(:params) do
+      {
+        title: "This is a short title",
+        enabled: true,
+        tags: {},
+        links: {},
+        migrated_from_gov_uk_delivery: false,
+        document_type: 'news-article',
+        email_document_supertype: 'publications',
+        government_document_supertype: 'news-articles',
+      }
+    end
+
+    before do
+      allow(Services.gov_delivery).to receive(:create_topic).and_return('TOPIC_123')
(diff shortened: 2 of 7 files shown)


app/services/subscriber_list_creator.rb
```diff
@@ -0,0 +1,28 @@
...
+    @params = params
+  end
+
+  def save
+    ActiveRecord::Base.transaction do
+      @record = SubscriberList.new(@params)
+      return false unless @record.save
+
+      # if the topic_id is passed in use the provided title.
+      return true if @record.gov_delivery_id
```
mobaig: maybe `@record.present?` would be a touch more robust?

spec/services/subscriber_list_creator_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    it "creates the topic on gov_delivery" do
+      creator = described_class.new(params)
+      creator.save
+      expect(Services.gov_delivery).to have_received(:create_topic)
+        .with("This is a short title (#{creator.record.id})")
+    end
+
+    context "and the title is longer than 255 characters" do
+      it "trims the title" do
+        title = 260.times.map { ('A'..'Z').to_a.sample }.join
```
mobaig: I think this could do with being simpler. It doesn't have to be a randomly generated string. A static value (eg `'A' * 260) would do the job and make the expectation below a touch easier to read too.

app/services/subscriber_list_creator.rb
dwhenry: done

spec/services/subscriber_list_creator_spec.rb
dwhenry: done

app/services/subscriber_list_creator.rb
```diff
@@ -0,0 +1,28 @@
+class SubscriberListCreator
+  attr_reader :record
+
+  def initialize(params)
+    @params = params
+  end
+
+  def save
+    ActiveRecord::Base.transaction do
```
dwhenry: removed

dwhenry:
@jennyd related to the above comment about name/short_name is it only short_name with the uniq constraint? IF it is both fields then added the ID at the end is still the best idea. :(

dwhenry:
Just checked in integration and both name and short_name need to be unique. As such adding the ID to the end of the string is the best solution we have at this stage.

jennyd:
[Policy publisher already appends " policy" to the title to avoid duplicates](https://github.com/alphagov/policy-publisher/commit/dc7e64286d5803cff6e9bb0c4058967091906e6e) - it looks like that was added after some policy subscriptions were already created, and those ones didn't have their titles updated at the same time, so they're most likely to be responsible for any duplicates with taxons. I think for topics we only support subscriptions on subtopics, and they use [combined titles like "Topic title: Subtopic title"](https://github.com/alphagov/collections/blob/1af678b9c3ebf20ed8e6e24e52a9a3809d690566/app/models/topic.rb#L34) so they would be very unlikely to duplicate taxon titles.

I think that appending the row id to the title to avoid occasional duplicate errors is the wrong tradeoff - it makes the title more confusing to users who subscribe to any new list. We should update the names and titles of the existing policy lists to the newer format - that will clean up some tech debt as well as helping resolve this problem.

The possibility of duplicates is a side-effect of only creating the topics on GovDelivery when someone first subscribes, which is sensible for the Whitehall ones where many combinations of filters are possible but not so much for the subscriptions to known topics/policies/taxons/countries etc. We should consider changing that model so that we know in advance if creating any topics fails in GovDelivery, before users discover for us.

dwhenry:
After discussion with Rob Rankin and @jennyd this PR is being closed for the time being. Rob will be documenting why this decision was made and looking at alternative solutions.

PR #159, opened 2017-04-03
Add a rake task to fetch missing subscriber list titles from GovDelivery

jennyd:
We have lots of subscriber lists in our database with missing titles, because they have been migrated to use email-alert-api from govuk-delivery which doesn't store titles for topics. This makes it harder to see from our data what a subscriber list is, and means that in our Staging and Integration GovDelivery accounts most topics have unhelpful ["MISSING TITLE [id]" names](https://github.com/alphagov/email-alert-api/blob/0dd0bae92ac39c66c0fca7f3cf0b99dc53c8e080/lib/data_hygiene/data_sync.rb#L83). GovDelivery requires a name for topics, so we can fetch the data from there in Production and update our database using it.

This task fetches all the topics in our database from GovDelivery, updates the title for lists which don't have one, and logs any discrepancies between the titles we have and the ones in GovDelivery.

I wrote this first to fetch every topic from GovDelivery individually, as we do in [`DeleteUnneededTopics`](https://github.com/alphagov/email-alert-api/blob/0dd0bae92ac39c66c0fca7f3cf0b99dc53c8e080/lib/data_hygiene/delete_unneeded_topics.rb), but then realised that getting the list of all topics would give us all the data we need for this task - `DeleteUnneededTopics` also needs the subscriber counts which aren't included in that response, but we only need the topic code and name here. I've updated this to do that instead, which should make it much faster overall, but have left the first version in an earlier commit in case we start getting timeouts when fetching all topics, in which case we may need to revert back to that method.

### DIFF
--- a/lib/data_hygiene/title_fetcher.rb
+++ b/lib/data_hygiene/title_fetcher.rb
@@ -1,8 +1,8 @@
 module DataHygiene
   class TitleFetcher
-    # Fetch titles for subscriber lists from GovDelivery and update our database
-    # with them if we don't have a title already. Log when we have a title which
-    # doesn't match GovDelivery's.
+    # Fetch all topics from GovDelivery and update the subscriber lists in our
+    # database with the title for each if we don't have a title already. Log
+    # when we have a title which doesn't match GovDelivery's.
 
     attr_reader :client, :logger, :stats
 
@@ -13,11 +13,14 @@ module DataHygiene
     end
 
     def run
+      all_topics = fetch_all_topics
+
       count = subscriber_lists.count
+      logger.info "Updating titles for #{count} subscriber lists"
 
-      logger.info "Fetching and updating titles for #{count} subscriber lists"
-
-      subscriber_lists.each { |subscriber_list| update_title(subscriber_list) }
+      subscriber_lists.each do |subscriber_list|
+        update_title(subscriber_list, all_topics[subscriber_list.gov_delivery_id])
+      end
 
       logger.info ''
       logger.info "Done:"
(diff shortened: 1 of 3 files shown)


lib/data_hygiene/title_fetcher.rb
```diff
@@ -0,0 +1,74 @@
...
+      logger.info "Updating titles for #{count} subscriber lists"
+
+      subscriber_lists.each do |subscriber_list|
+        update_title(subscriber_list, all_topics[subscriber_list.gov_delivery_id])
+      end
+
+      logger.info ''
+      logger.info "Done:"
+      stats.each do |k, v|
+        logger.info "  #{k}: #{v.size}"
```
dwhenry: Any reason for storing the topic_ids if we are only doing a count?

lib/data_hygiene/title_fetcher.rb
```diff
@@ -0,0 +1,74 @@
+module DataHygiene
+  class TitleFetcher
```
dwhenry: This name is slightly misleading, maybe TitleUpdater would be more accurate

spec/lib/data_hygiene/title_fetcher_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+            subject.send(:update_title, subscriber_list, 'Education')
+            expect(subject.stats).to eq({ update_failed: ['ABC_123'] })
+          end
+        end
+
+        it_behaves_like 'updates the title', nil
+        it_behaves_like 'updates the title', ''
+
+        it_behaves_like 'logs when saving fails', nil
+        it_behaves_like 'logs when saving fails', ''
```
dwhenry: I would potentially argue that when we only have 1 test in a shared_example it is easier to follow by just have a duplicated test. :)

lib/data_hygiene/title_fetcher.rb
jennyd: I considered that, but thought it sounded like it would update all of our titles, overwriting them in our database if they're different in GovDelivery, whereas this only sets titles on lists which don't have one and logs everything else. "Fetch" also makes it clearer that it's getting them from somewhere else as opposed to modifying them based on something local. I can change it if you like?

lib/data_hygiene/title_fetcher.rb
jennyd: Hm, I think I was going to use them in some tests to make sure that multiple lists were being counted in the right buckets, but looks like I didn't actually do that. The tests with single lists and on the logging for `#run` cover that enough, so I'll make it just keep count instead.

jennyd: done

spec/lib/data_hygiene/title_fetcher_spec.rb
jennyd: I got rid of one of the cases for that so it's just a single test now.

lib/data_hygiene/title_fetcher.rb
dwhenry: Shrug, naming. :(

PR #160, opened 2017-04-04
When deleting unneeded topics, only consider lists older than a day

jennyd:
Now that we're creating new subscriber lists for taxons and Whitehall things,
we want to avoid accidentally deleting a topic while the first user is trying
to subscribe to it. Only considering lists older than a day for deletion
should do this comfortably while still letting us clear up unneeded data.

We really only need to change this behaviour for lists with 0 subscribers
rather than for the other categories as well, but it was much simpler to
exclude them from the start rather than having to check each one as we're
deleting, or word all the logging differently to make it clear that we're
making this change only for lists with 0 subscribers. I've added tests for the
other cases to document the new behaviour.

Nothing else in the app cares about `created_at` so it's safe to change it in
the default factory.

### DIFF
--- a/lib/data_hygiene/delete_unneeded_topics.rb
+++ b/lib/data_hygiene/delete_unneeded_topics.rb
@@ -115,8 +115,8 @@ module DataHygiene
       end
 
       def call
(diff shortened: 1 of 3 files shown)


jennyd:
[Equivalent govuk-delivery PR](https://github.gds/gds/govuk_delivery/pull/30)

PR #161, opened 2017-04-18
Update docs about the data sync rake task

jennyd:
We changed the behaviour of this task in https://github.com/alphagov/email-alert-api/pull/142 to make it faster, and to enable us to subscribe to topics in Integration and Staging and have those subscriptions persist rather than disappear overnight when all the data was deleted.


dwhenry:
👍

PR #162, opened 2017-04-18
Add documentation about subscriber list matching criteria

jennyd:
This gives an overview of how subscriber lists work. I haven't included many examples because the tests are more comprehensive, but I've tried to explain how the two sides of the app work and how the data matches up. The query code can be complex to read so hopefully this will help people new to the app to understand how it works.

### DIFF
--- a/doc/matching-content-to-subscriber-lists.md
+++ b/doc/matching-content-to-subscriber-lists.md
@@ -1,0 +1,83 @@
+# Matching content to subscriber lists
+
+## Format of matching criteria
+
+A subscriber list uses several fields to store the criteria which define how it
+matches published content:
+
+### JSON fields
+
+- `links`
+- `tags`
+
+Only one of these should be used on a given subscriber list; `tags` is deprecated.
+They can also both be empty `{}` if they aren't needed.
+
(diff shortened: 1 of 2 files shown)


doc/matching-content-to-subscriber-lists.md
```diff
@@ -0,0 +1,83 @@
...
+
+`links` uses content ids and `tags` uses slugs - but the queries used on these
+fields don't care about that difference.
+
+It's crucial that the key names stored in these JSON fields match the key names
+used in the requests to email-alert-api (which are usually the same as the ones
+on the content item itself). It's easy to break existing subscriptions by
+renaming a field in the links hash on published content without updating the
+corresponding subscriber lists, which would mean that subscribers stop getting
+emails for that kind of content.
```
dwhenry: but be worth mentioning `taxon_tree` which is a special case field that is generated in `email_alert_service` and not `publishing_api`

jennyd: I've added a note about `taxon_tree` above, with a link for some context on why we did it that way.

dwhenry:
👍

PR #163, opened 2017-04-19
Fix link formatting in data sync docs

jennyd:
There were line breaks in the middle of these links, preventing them from
rendering properly.


dwhenry:
👍

PR #166, opened 2017-07-20
Keep track of number of subscribers to lists

tijmenb:
This commit adds a rake task that will update the "subscriber_count" of subscriber lists ("topics"). It does this by getting the data from GovDelivery. Separately we'll set up a rake task to periodically run the rake task so that the counts are refreshed regularly.

We're adding the counts because we're considering adding something to the content tagger UI that shows how many people are subscribed to a certain taxon. This can help the information architects determine the relative popularity of taxons.

The actual work happens in a sidekiq worker to make sure we're not tying up the rake task executors on Jenkins, and to make sure it's retried when an error occurs. We'll also get data from statsd on the job performance.

https://trello.com/c/YwfSmNQX

### DIFF
--- a/app/workers/update_subscriber_counts.rb
+++ b/app/workers/update_subscriber_counts.rb
@@ -1,0 +1,14 @@
+class UpdateSubscriberCounts
+  include Sidekiq::Worker
+
+  def perform
+    SubscriberList.find_each do |subscriber_list|
+      begin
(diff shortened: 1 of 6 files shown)


tijmenb:
@rubenarakelyan correct, the stand-alone changes are in https://github.com/alphagov/email-alert-api/pull/166/commits/7ba543c29de969d8c9e9451c8c5c39aa1c065b8d.

PR #168, opened 2017-07-25
Fix a warning

tuzz:
>Passing 'info' command to redis as is; administrative commands cannot be
>effectively namespaced and should be called on the redis connection
>directly; passthrough has been deprecated and will be removed in
>redis-namespace 2.0 (at healthcheck_controller.rb:34:in `block in status')


davidbasalla:
Looks good!

PR #169, opened 2017-07-27
Add a reporting script to help retire govuk-delivery

tuzz:
https://trello.com/c/eKdM4dZr/44-improve-notification-reporting-when-comparing-email-alert-api-with-govuk-delivery-2

Add a reporting script to help retire govuk-delivery

Before we can migrate, we need to ensure
email-alert-api behaves more-or-less the same as
govuk-delivery. This report provides detailed
output that we can use to fix problems with topic
matching between the two systems.

It also provides a ‘global’ view of which topics
are most frequently mismatched so we can focus our
attention on those.

## Example output

screen shot 2017-07-27 at 11 42 10

Possible improvements if we have more time:

- Add support for `--request_id` parameter 
- Incorporate `enabled_gov_delivery_ids` and `disabled_` fields
- Investigate why lots of subscriber lists appear as `<unknown>`
- Output a CSV for further analysis


spec/lib/notification_report_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+
+      context "when multiple notifications were sent from a system" do
+        before do
+          create(:notification_log, emailing_app: "email_alert_api", gov_delivery_ids: %w(a b c))
+
+          create(:notification_log, emailing_app: "gov_uk_delivery", gov_delivery_ids: %w(b c d))
+          create(:notification_log, emailing_app: "gov_uk_delivery", gov_delivery_ids: %w(c d))
+        end
+
+        it "uses the gov_delivery_ids of the last notification log" do
```
davidbasalla: I'm not sure here whether using the gov_delivery_ids of the _last_ notification log captures the actual behaviour of the system correctly? I had a brief look through the data yesterday and saw that for multiple notification logs (for the same govuk_request_id and same emailing_app), there was little or no overlap between the gov_delivery_ids. So by just using the gov_delivery_ids of the last notification log as this spec suggests, we might be missing out on some data...?

davidbasalla:
A quick thought about the output, would it be beneficial to say that the main 'key' is a govuk_request_id? It's not obvious from just looking at the output what that value represents, and anyone using it would have to look through the code to see what it refers to.

spec/lib/notification_report_spec.rb
davidbasalla: Ah, just read through the rest of the spec 😄  It looks like there's alerting in place for when the multiple notifications' gov_delivery_ids don't match (e.g. `#email_alert_api_notifications_have_the_same_topics`), so that would alert us to the fact that something fishy is going on, and we can then investigate. So you can ignore the previous comment!

davidbasalla:
This looks good! 👍 The listed recommendations for possible improvements also sound valuable, in particular I'd be interested in some output around enabled/disabled `gov_delivery_ids` but that's something we can tackle once we get into the analysis part (and adapt the script as needed).

spec/lib/notification_report_spec.rb
tuzz: Yeah – I came to the same conclusion that it's risky to just use the last notification without also checking if the notifications from each system have the same topics. 👍

tuzz:
>A quick thought about the output...

Addressed.

>I'd be interested in some output around enabled/disabled...

Agreed, I started working on this but it turned out to be quite complicated so decided to leave this out for now. We can add this in later, perhaps after we have a better understanding from this work:
https://trello.com/c/CaD3UZFZ/69-make-sure-toggling-enabled-disabled-across-systems-works-quickly-and-reliably-1

PR #171, opened 2017-08-01
Add rake tasks to enable/disable lists

rubenarakelyan:
This PR adds two rake tasks that enable/disable subscriber lists in email-alert-api and govuk-delivery to facilitate the migration of all email lists to email-alert-api.

Trello: https://trello.com/c/CaD3UZFZ/69-make-sure-toggling-enabled-disabled-across-systems-works-quickly-and-reliably-1

### DIFF
--- a/lib/tasks/switch_list_to_email_alert_api.rake
+++ b/lib/tasks/switch_list_to_email_alert_api.rake
@@ -1,0 +1,34 @@
+desc "Switch a subscriber list to email-alert-api from govuk-delivery"
+task switch_list_to_email_alert_api: [:environment] do |_, args|
+  TICK = "\e[32m✓\e[0m"
+  CROSS = "\e[31m✗\e[0m"
+
+  args.extras.each do |gov_delivery_id|
+    # Enable the list in email-alert-api
+    subscriber_list = SubscriberList.where(gov_delivery_id: gov_delivery_id).first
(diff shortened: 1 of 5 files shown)


lib/tasks/switch_list_to_email_alert_api.rake
```diff
@@ -0,0 +1,30 @@
...
+    exit
+  end
+
+  # Disable the list in govuk-delivery
+  response = Services.govuk_delivery.disable_list(gov_delivery_id).to_hash
+
+  if !response['enabled']
+    puts "#{TICK} #{gov_delivery_id} has been disabled in govuk-delivery"
+  else
+    puts "#{CROSS} #{gov_delivery_id} could not be disabled in govuk-delivery"
```
davidbasalla: Would we want to disable the list in email-alert-api in case this happens?

davidbasalla:
Would we be running this rake task for each subscriber list individually? If not, would it be useful to pass it a batch of subscriber lists to process?

lib/tasks/switch_list_to_email_alert_api.rake
rubenarakelyan: Yes, I've added this

rubenarakelyan:
I was assuming we'd be doing them individually, but I've added the ability to specify multiple IDs in a single call.

PR #173, opened 2017-08-01
Add CSV output for notification log reports

davidbasalla:
This allows for exporting the notification logs to CSV. It will export a CSV file to the `/public/data` folder which can then be copied and viewed with any spreadsheet application. It will make it easier to spot patterns when identifying different mismatches, as we will be able to sort by different columns (eg created_at).

The code has been copied and adapted from [LocalLinksManager’s export_links rake task](https://github.com/alphagov/local-links-manager/blob/master/lib/tasks/export/link_exporter.rake)

Related to https://trello.com/c/UVYYmT7t/67-understand-the-differences-between-email-alert-apis-subscriber-lists-and-govuk-deliverys-feedurls-and-decide-what-to-do-about-th

### DIFF
--- a/lib/notification_report.rb
+++ b/lib/notification_report.rb
@@ -1,3 +1,5 @@
+require "csv"
+
 class NotificationReport
   def self.all
     new(NotificationLog.all)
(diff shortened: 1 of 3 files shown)


lib/notification_report.rb
```diff
@@ -1,3 +1,5 @@
+require 'csv'
```
gpeng: This file is already using double quotes. We should be consistent (also on lines 135, 154)

davidbasalla: Thanks, updated! As a side note, it looks like we don't run the lint checker this app... might pick that up at a later point.

PR #174, opened 2017-08-07
Add an endpoint for retrieving topic matches

tuzz:
We want to see which topics are matched against
query parameters so we can cross-reference these
against govuk-delivery.

This provides an endpoint which will be used in
a Whitehall rake task that generates a report.

### DIFF
--- a/app/controllers/topic_matches_controller.rb
+++ b/app/controllers/topic_matches_controller.rb
@@ -1,0 +1,30 @@
+class TopicMatchesController < ApplicationController
+  def show
+    respond_to do |format|
+      format.json { render json: topic_matches }
+    end
+  end
+
+private
+
+  def topic_matches
+    hash = params.permit!.to_h
+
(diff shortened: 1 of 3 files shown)


app/controllers/topic_matches_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+      government_document_supertype: hash[:government_document_supertype],
+    )
+
+    topics = query.lists
+    enabled, disabled = topics.partition(&:enabled?)
+
+    {
+      topics: topics.map(&:gov_delivery_id),
+      enabled: enabled.map(&:gov_delivery_id),
+      disasbled: disabled.map(&:gov_delivery_id),
```
davidbasalla: Small typo: disasbled -> disabled, pops up in other parts of the code in this PR too

tuzz: Thanks, fixed.

PR #178, opened 2017-09-05
Add subscriber list export task

gpeng:
As part of the migration of email topic management from govuk_delivery
to email-alert-api we need to try a subset before switching them all
over. This rake task produces a CSV of all subscriber lists to help us
identify an initial group (or groups).

It can be removed once the migration is complete.

[Trello](https://trello.com/c/4y8S9j1S/85-transition-whitehalls-email-sending-to-use-email-alert-api)

### DIFF
--- a/lib/tasks/subscriber_list_export.rake
+++ b/lib/tasks/subscriber_list_export.rake
@@ -1,0 +1,36 @@
+require 'csv'
+
+namespace :subscriber_list do
+  task export: :environment do
+    links_keys = [
+      :service_manual_topics,
+      :countries,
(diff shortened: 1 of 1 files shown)


lib/tasks/subscriber_list_export.rake
```diff
@@ -0,0 +1,36 @@
+require 'csv'
+
+namespace :subscriber_list do
+  task export: :environment do
+    links_keys = [
```
davidbasalla: Out of curiosity, where did you get this list of keys from?

PR #179, opened 2017-09-08
Add tasks to switch all subscriber lists over at once

tuzz:
https://trello.com/c/4y8S9j1S/85-transition-whitehalls-email-sending-to-use-email-alert-api

### DIFF
--- a/lib/tasks/switch_list_to_email_alert_api.rake
+++ b/lib/tasks/switch_list_to_email_alert_api.rake
@@ -32,3 +32,20 @@ task switch_list_to_email_alert_api: [:environment] do |_, args|
     end
   end
 end
+
+desc "Switch all subscriber lists to email-alert-api from govuk-delivery"
+task switch_all_lists_to_email_alert_api: :environment do
+  SubscriberList.where(migrated_from_gov_uk_delivery: true).find_each do |list|
+    begin
+      Services.govuk_delivery.disable_list(list.gov_delivery_id)
+
+      list.enabled = true
+      list.save!
+
+      print "."
+    rescue => e
+      puts e.message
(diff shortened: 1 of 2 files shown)


lib/tasks/switch_list_to_email_alert_api.rake
```diff
@@ -32,3 +32,20 @@ task switch_list_to_email_alert_api: [:environment] do |_, args|
     end
   end
 end
+
+desc "Switch all subscriber lists to email-alert-api from govuk-delivery"
+task switch_all_lists_to_email_alert_api: :environment do
+  SubscriberList.where(migrated_from_gov_uk_delivery: true).find_each do |list|
+    begin
+      Services.govuk_delivery.disable_list(list.gov_delivery_id)
```
gpeng: Does this always `raise` if it fails or is there a return value we should check before enabling the other one?

tuzz: It always raises. There is a `success` attribute in the bod, we could check, but I've tested on staging and this seems fine.

govuk-delivery responds with a 200, 400 or 404 from the `disable_list` endpoint:
https://github.com/alphagov/govuk-delivery/blob/master/service.py#L324-L330

Additionally, we could get 5xx errors from Nginx.

All these cases raise errors in gds-api-adapters:
https://github.com/alphagov/gds-api-adapters/blob/5513aabfc2de805f0c005eb9b69ea4179aa7e342/lib/gds_api/exceptions.rb#L58

Which means the simple `rescue` should suffice.

I can add an extra check for the `success` attribute if you think, but I can't see how we'd get past that line with a `success=false` without also raising an error.

PR #187, opened 2017-10-17
Send email to `Subscriber`

gpeng:
This PR adds a basic concept of a `Subscriber` and the functionality to send an email to one. It is intended only as an initial implementation to be built upon. Validation on the model and retrying of the service call has intentionally been omitted as out of the scope for now.

[Trello](https://trello.com/c/R53XrWOI/229-email-alert-api-can-send-an-email-to-a-user)

This PR depends upon incoming PRs from [Trello](https://trello.com/c/E0FTbAS9/228-email-alert-api-can-send-an-email-via-notify) and [Trello](https://trello.com/c/4eOYL9WC/227-expose-notify-credentials-as-environment-variables)

### DIFF
--- a/app/commands/deliver_to_subscriber.rb
+++ b/app/commands/deliver_to_subscriber.rb
@@ -1,0 +1,22 @@
+class DeliverToSubscriber
+  attr_reader :subscriber
+
+  def initialize(subscriber:)
+    @subscriber = subscriber
+    raise ArgumentError, "subscriber cannot be nil" if subscriber.nil?
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    email_sender.call(address: subscriber.address)
+  end
+
+private
+
+  def email_sender
+    Services.email_sender
+  end
+end
--- a/app/models/subscriber.rb
+++ b/app/models/subscriber.rb
@@ -1,0 +1,3 @@
+class Subscriber < ActiveRecord::Base
+
+end
--- a/db/migrate/20171016143522_create_subscribers.rb
+++ b/db/migrate/20171016143522_create_subscribers.rb
@@ -1,0 +1,8 @@
+class CreateSubscribers < ActiveRecord::Migration[5.1]
+  def change
+    create_table :subscribers do |t|
+      t.string :address
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170922091721) do
+ActiveRecord::Schema.define(version: 20171016143522) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -43,6 +43,12 @@ ActiveRecord::Schema.define(version: 20170922091721) do
     t.string "email_document_supertype", default: "", null: false
     t.string "government_document_supertype", default: "", null: false
     t.integer "subscriber_count"
+  end
+
+  create_table "subscribers", force: :cascade do |t|
+    t.string "address"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
   end
 
 end
--- a/lib/tasks/deliver.rake
+++ b/lib/tasks/deliver.rake
@@ -1,0 +1,6 @@
+namespace :deliver do
+  task :to_subscriber, [:id] => :environment do |_t, args|
+    subscriber = Subscriber.find(args[:id])
+    DeliverToSubscriber.call(subscriber: subscriber)
+  end
+end
--- a/spec/commands/deliver_to_subscriber_spec.rb
+++ b/spec/commands/deliver_to_subscriber_spec.rb
@@ -1,0 +1,32 @@
+require "rails_helper"
+
+RSpec.describe DeliverToSubscriber do
+  describe ".call" do
+    let(:email_sender){ double }
+    before do
+      allow(Services).to receive(:email_sender).and_return(
+        email_sender
+      )
+    end
+
+    it "calls email_sender with address" do
+      subscriber = double(address: "test@test.com")
+
+      expect(email_sender).to receive(:call)
+        .with(address: "test@test.com")
+
+      DeliverToSubscriber.call(
+        subscriber: subscriber,
+      )
+    end
+
+    it "requires subscriber" do
+      expect{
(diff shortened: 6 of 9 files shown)


app/commands/deliver_to_subscriber.rb
```diff
@@ -0,0 +1,18 @@
+class DeliverToSubscriber
+  attr_reader :subscriber, :email_sender
+
+  def initialize(options)
```
thomasleese: Might be nicer to do `def initialize(subscriber:, email_sender:)` so it's clear what arguments are available.

lib/tasks/deliver.rake
```diff
@@ -0,0 +1,6 @@
+namespace :deliver do
+  task :to_subscriber, [:id] => :environment do |_t, args|
+    subscriber = Subscriber.find(args[:id])
+    DeliverToSubscriber.call(subscriber, Services.email_sender)
```
thomasleese: I think this will need to be given as keyword arguments, `call(subscriber: subscriber, email_sender: Services.email_sender)`

app/commands/deliver_to_subscriber.rb
gpeng: I did have it with keyword arguments (in fact I've been back and forth on it a couple of times, I've forgotten how to do new things!) but I was thinking that as the interaction with everything is currently so fluid using a hash might make it more flexible and require less changes. I also had it with `email` and `subscriber` as 'kwargs' and then `options` after that as a hash but I had an architectural crisis of confidence with that :) I'll change it now I've had a second opinion.

gpeng: Any thoughts on passing `Services.email_sender` as an argument vs just using that in the code. It's dependency injection but we already abstract it away by wrapping it in the `Services` module and we don't 'really' get any benefit in injecting it. What do you think?

bevanloon: personally don't think you need to pass it in given it's going to be wrapped in the `Services`.

thomasleese: Yeah, I would agree, it's probably not worth needing to pass it in.

app/commands/deliver_to_subscriber.rb
```diff
@@ -0,0 +1,22 @@
+class DeliverToSubscriber
+  attr_reader :subscriber
+
+  def initialize(subscriber:)
+    @subscriber = subscriber
+    raise ArgumentError, "subscriber is required" if subscriber.nil?
```
thomasleese: Very picky... but I would probably change this to “subscriber cannot be nil”.

app/commands/deliver_to_subscriber.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def self.call(subscriber:)
+    DeliverToSubscriber.new(subscriber: subscriber).call
+  end
```
tuzz: I quite like writing these convenience methods like:

```ruby
def self.call(*args)
  new(*args).call
end
```

It means you don't have to update the method signature if you change `#initialize` or if you change the class name.

db/migrate/20171016143522_create_subscribers.rb
```diff
@@ -0,0 +1,8 @@
+class CreateSubscribers < ActiveRecord::Migration[5.1]
+  def change
+    create_table :subscribers do |t|
+      t.string :address
```
tuzz: Would it make sense to make this `null: false` and add a unique index to this column?

lib/tasks/deliver.rake
```diff
@@ -0,0 +1,6 @@
+namespace :deliver do
+  task :to_subscriber, [:id] => :environment do |_t, args|
+    subscriber = Subscriber.find(args[:id])
+    DeliverToSubscriber.call(subscriber: subscriber)
```
tuzz: 👍

spec/commands/deliver_to_subscriber_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    it "requires subscriber" do
+      expect{
+        DeliverToSubscriber.call(subscriber: nil)
+      }.to raise_error(
+        ArgumentError,
+        "subscriber is required"
+      )
+    end
```
tuzz: I think rubocop will complain about a multi-line `{ ... }` block. Saying that, I don't think we have rubocop yet in this project. Maybe we should add it?

app/models/subscriber.rb
```diff
@@ -0,0 +1,3 @@
+class Subscriber < ActiveRecord::Base
+
+end
```
tuzz: Would it be worth adding validations or shall we do that later?

e.g. unique address (case insensitive), matches an email regex(?)

db/schema.rb
```diff
@@ -45,4 +45,10 @@
     t.integer "subscriber_count"
   end
 
+  create_table "subscribers", force: :cascade do |t|
+    t.string "address"
```
kevindew: Might it be more explicit to call this email or email_address?

app/models/subscriber.rb
gpeng: I intentionally left that off as it wasn't required to complete the ticket and we're not accepting user input yet. Seemed like that would be something to think about then or when we're looking at the data we need to migrate.

db/migrate/20171016143522_create_subscribers.rb
gpeng: Yes, probably but again them needing to be unique hasn't been 'established' yet (although it's obviously very unlikely we won't want them to be). Trying not to get to far ahead of 'minimal'.

db/schema.rb
gpeng: I had it as `email_address` but it seemed redundant at this stage as we're in the email-alert-api and it seemed sufficiently implicit. Happy to change it if you disagree.

kevindew: ah yes true 👍

app/commands/deliver_to_subscriber.rb
gpeng: Added in a75098d

app/models/subscriber.rb
gpeng: [Trello](https://trello.com/c/BhKdFEXN/170-subscriber-validation)

db/migrate/20171016143522_create_subscribers.rb
gpeng: [Trello](https://trello.com/c/BhKdFEXN/170-subscriber-validation)

PR #188, opened 2017-10-18
Send an email via notify

bevanloon:
Basic functionality to send an email via GOV.UK Notify.

Define a minimal interface for sending individual emails by wrapping the [Ruby client for notifications API](https://github.com/alphagov/notifications-ruby-client) in a class and exposing it as `Services.email_sender`.

Relies on an env var for the notify api key as specified in this [Trello](https://trello.com/c/4eOYL9WC/227-expose-notify-credentials-as-environment-variables).

Related [Trello](https://trello.com/c/E0FTbAS9/228-email-alert-api-can-send-an-email-via-notify)

Prerequisite for [PR 187](https://github.com/alphagov/email-alert-api/pull/187) and it's associated [Trello](https://trello.com/c/R53XrWOI/229-email-alert-api-can-send-an-email-to-a-user)


config/notify.yml
```diff
@@ -0,0 +1,2 @@
+api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] || "TEST_KEY"%>
+template_id: <%= ENV['GOV_UK_NOTIFY_TEMPLATE_ID'] || "80c6e214-8ed3-4437-bf57-79005ec74719" %>
```
bevanloon: I think `template_id` is a required field when we call notify - I've hard coded the fallback for now, but I think we should probably add a default in puppet? thoughts @rubenarakelyan

thomasleese: Can we stick with either `GOVUK` or `GOV_UK` for environment variable names?

app/services/email_sender/notify_sender.rb
```diff
@@ -0,0 +1,37 @@
+require "notifications/client"
+
+module EmailSender
+  class NotifySender
+    def call(options)
```
thomasleese: Can we use `def call(address:, template_id:)` here instead of having to `unpack` and check that we were given a hash?

app/services/email_sender/notify_sender.rb
```diff
@@ -0,0 +1,37 @@
...
+      raise ArgumentError, "address is required" if @address.nil?
+      send
+    end
+
+  private
+    attr_accessor :address, :template_id
+
+    def unpack(options)
+      raise ArgumentError, "call expects a hash" unless options.is_a?(Hash)
+      @address = options[:address]
```
thomasleese: I'm not sure I like the idea of storing the current state (`address` and `template_id`) of the sender as instance variables instead of just using local variables and passing arguments. It doesn't sit well with me, and it presumably means this class isn't thread safe (I don't know if this matters or not in Rails, but it seems unnecessary to have that restriction on such a simple class).

app/services/email_sender/notify_sender.rb
```diff
@@ -0,0 +1,37 @@
...
+      @template_id = options[:template_id] || notify_config.fetch(:template_id)
```
thomasleese: I would maybe move `notify_config.fetch(:template_id)` into its own method called `default_template_id`, for clarity.

config/notify.yml
rubenarakelyan: `GOVUK` is the one that's consistently used in other apps

rubenarakelyan: Would it not be better to fail if these env vars don't exist?

thomasleese: Perhaps it would be nicer to have this file with a layout similar to [gov_delivery.yml](https://github.com/alphagov/email-alert-api/blob/master/config/gov_delivery.yml) with the environments.

app/services/email_sender/notify_sender.rb
bevanloon: so the reason I've left this without keyword args for the moment is because it is a minimal interface. We will most probably want to add more than just address and template....things like our reference to pass through to notify, the subject, perhaps a user's name for personalisation. It felt premature to start future proofing that right now given we are creating the MVP.

Those may or may not be good reasons for the decision though?

thomasleese: Hmm, yeah, I can see where you're coming from.

I would argue though, that we're going to need to modify the code anyway to use any new fields from the `options` hash that it doesn't give us any advantages here.

config/notify.yml
bevanloon: Can I play devil's advocate for a minute and ask if we actually have different data we want for these variables in test, dev and prod yet? 

If we don't then would it be a good idea to wait until we have separate distinct values per env before adding in the code to do this? We will never simplify it again if we add it in now and find we don't need it.

bevanloon: fixed in 1cf52acf8c2c181b83eff5668df537ff1be6d5ce

kevindew:
Sorry wading in so this might have already been discussed.

I'm just thinking about how this will work in a local environment and what you might need. I'm thinking this is probably in the scope of this piece of work but I guess it could be a later thing.

I'm imagining locally you probably want 3 scenarios:

1. No-op action maybe similar to [Publishing API queue_publisher.rb](https://github.com/alphagov/publishing-api/blob/e05a0b33eded5c66e869e3d246606f630eef2adb/lib/queue_publisher.rb) - which I guess would be the default for someone using the app since they didn't want have env vars set up for notify by default
2. Send to only own email address (so overwrite sender) - which I guess most devs who've set up a notify account want to do
3. Actually send emails to people - which is probably quite dangerous for local testing, so might be something that we want someone to actually change code for.

What do you think?

config/notify.yml
rubenarakelyan: API key - yes defo, so we don't accidentally send real-life emails in non-production environments

Template ID - can use the same one since Notify doesn't have the notion of different environments with different data

bevanloon:
@kevindew that sounds fairly sensible to me. 
I'm wondering whether it's worth capturing those on a ticket so we've sort of documented the decision for the behaviour somewhere.

Do you think that no-op / all of the functionality is the responsibility of the email sender? or the thing that calls the email sender? Since this class sends an email to an address passed in, I'm just wondering how long the chain of passing in whatever override mechanism we decide on would get and whether to just cut it off at the head. Perhaps there's still value in exercising the code paths though? I'm not sure.

How do you picture number 3 working?

rubenarakelyan:
A developer can't get an API key for sending real emails unless it is approved by the Notify team, so I don't think we should worry about that scenario happening by mistake.

app/services/email_sender/notify_sender.rb
rubenarakelyan: Could the list of arguments get quite large when we start to add new features? That would be a good argument for using `options` and unpacking it; it's also then easier to have a mix of required and optional params in any order.

thomasleese: Required and optional params can be achieved with `def func(a:, b: "optional")` and the arguments can be written in any order when calling the method.

I don't like using `options` because it makes it much harder to immediately see what arguments a method takes without writing documentation for the method, plus it means Ruby gives us a load of `ArgumentError`s for free.

kevindew:
> @kevindew that sounds fairly sensible to me.
I'm wondering whether it's worth capturing those on a ticket so we've sort of documented the decision for the behaviour somewhere.

Yeah sure, I'll see where I can fit in.

> Do you think that no-op / all of the functionality is the responsibility of the email sender? or the thing that calls the email sender? Since this class sends an email to an address passed in, I'm just wondering how long the chain of passing in whatever override mechanism we decide on would get and whether to just cut it off at the head. Perhaps there's still value in exercising the code paths though? I'm not sure.

Sure. I'm not sure where the exact responsibility of that would come in, but I imagine you're right that it doesn't belong in EmailSender. I guess we'd have some form of EmailService which decides whether Emails go to the sender or not

> How do you picture number 3 working?

I wonder if we could do it with two env vars: SEND_EMAILS and OVERRIDE_EMAIL_RECIPIENT. I'm not sure if there are already precedents set in this regard though with Action Mailer that could be worth following.

>  A developer can't get an API key for sending real emails unless it is approved by the Notify team, so I don't think we should worry about that scenario happening by mistake.

Ah awesome! that'll be good to avoid disasters.

app/services/email_sender/notify_sender.rb
bevanloon: In the first instance, the list of arguments could be as large as the list of arguments to notify https://github.com/alphagov/notifications-ruby-client#arguments:

From their example:
```
  email_address: email_address,
  template_id: template_id,
  personalisation: {
    name: "name",
    year: "2016"
  },
  reference: "your_reference_string"
  email_reply_to_id: email_reply_to_id
```

Personalisation is the bit that can get really large since that's where we'll pass in any template based variables.

thomasleese:
```ruby
def call(email_address:, template_id:, personalisation: {}, reference: nil, email_reply_to_id: nil)
```

Personally that seems fine to me, we can always break it down into lines if we want to. The extra documentation we get from doing it like that seems like a big win to me.

Happy to concede on this point though if I'm the only advocate for this!

bevanloon: I'm tinkering with turning the function definition into

```
def call(address:, **keyword_args)
```

that should give us the required `ArgumentError` for free on `address` as @thomasleese points out and allows us to pass in any other number of keyword arguments. If we decide to explicitly require another argument in the future, we can introduce the requisite argument in any consumers first without breaking anything and then pull it out specifically in the function definition.

thomasleese: @bevanloon That seems like a good compromise! 👍

app/services/email_sender/notify_sender.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+  private
+
+    def send(address:, **keyword_args)
+      client.send_email(
+        email_address: address,
+        template_id: template_id(keyword_args)
+      )
+    end
```
tuzz: I wonder if we should call this something other than `#send` to avoid clashing with [`Object#send`](https://ruby-doc.org/core-2.4.2/Object.html#method-i-send)? Not that it really matters.

config/notify.yml
```diff
@@ -0,0 +1,11 @@
...
+  api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] %>
+  template_id: <%= ENV['GOVUK_NOTIFY_TEMPLATE_ID'] || "a5583ad3-adab-4c76-b85d-f88bd4a3dc69" %>
+
+test:
+  api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] %>
+  template_id: <%= ENV['GOVUK_NOTIFY_TEMPLATE_ID'] || "a5583ad3-adab-4c76-b85d-f88bd4a3dc69" %>
+
+production:
+  api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] %>
+  template_id: <%= ENV['GOVUK_NOTIFY_TEMPLATE_ID'] %>
```
kevindew: I know you've already had a discussion about this so apologies for bringing it back.

If we're controlling these by env vars I don't think we get anything by these being separate ones for different environments - changing env vars changes them all. 

So I'd imagine we'd only want to separate these out if we hard code some attributes for particular environments (which I generally don't think is great since it's not very 12 factor but sometimes pragmatic)

/cc @thomasleese

app/services/email_sender/notify_sender.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module EmailSender
+  class NotifySender
+    def call(address:, **keyword_args)
+      send(address: address, **keyword_args)
+    end
+
+  private
+
+    def send(address:, **keyword_args)
```
tuzz: What are these `keyword_args` for? I can't see they're used anywhere?

config/notify.yml
thomasleese: I imagine we might want to hardcode these for the test environment, and perhaps not allow them to be configured with environment variables.

Otherwise, I have no opinion on this, I'm happy with keeping the environments or leaving them out.

rubenarakelyan: We may need to put something for the development and test environments directly into the file like some other apps.

app/services/email_sender/notify_sender.rb
bevanloon: good point thanks @tuzz fixed that in 49de7db462496913b6feced6d7f84f5c64e81a4d

spec/services/email_sender/notify_sender_spec.rb
```diff
@@ -0,0 +1,48 @@
+require "rails_helper"
+require "notifications/client"
+
+RSpec.describe EmailSender::NotifySender do
+  describe "#call" do
+    it "requires an address" do
```
gpeng: Do we need to test this now it's a keyword argument? We wouldn't test that a method with a 'normal' argument didn't work if you didn't supply any.

app/services/email_sender/notify_sender.rb
bevanloon: so at the moment only used for overriding the default `template_id` from the config - so 
https://github.com/alphagov/email-alert-api/pull/188/files#diff-0cc6c4f93994a3add62321b9d512a6e4R14
which ends up https://github.com/alphagov/email-alert-api/pull/188/files#diff-0cc6c4f93994a3add62321b9d512a6e4R30

Am I using them correctly? (I've not used the double splatted keyword args feature before).

The intention is, however, to be able to pass in anything we might need when we come to extend this so things like personalisation variables, reply to addresses, any reference ids we might want. There's a small chat about it https://github.com/alphagov/email-alert-api/pull/188#pullrequestreview-70129238

spec/services/email_sender/notify_sender_spec.rb
bevanloon: quite right - fixed f6c9a48650302f23b73623ad9318b3165e63cdfb

PR #192, opened 2017-10-19
Validate the uniqueness of email address

tuzz:
We have a unique index on the column, but the
error messages are friendlier if we add the
validation in as well.

### DIFF
--- a/spec/models/subscriber_spec.rb
+++ b/spec/models/subscriber_spec.rb
@@ -41,5 +41,12 @@ RSpec.describe Subscriber, type: :model do
 
       expect(subject).to be_invalid
     end
+
+    it "is invalid if an email address is already taken" do
+      FactoryGirl.create(:subscriber, address: "foo@bar.com")
+
+      subject.address = "foo@bar.com"
+      expect(subject).to be_invalid
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/models/subscriber_spec.rb
```diff
@@ -41,5 +41,19 @@
...
+
+      subject.address = "foo@bar.com"
+      expect(subject).to be_invalid
+    end
+
+    it "is case-insensitive when validating if an address is taken" do
+      FactoryGirl.create(:subscriber, address: "FOO@bar.com")
+
+      subject.address = "foo@BAR.COM"
+      expect(subject).to be_invalid
```
thomasleese: Technically, these are distinct email addresses. The part before the @ is case sensitive, the part after the @ is not.

thomasleese: (massive edge case, and most servers tend to be case insensitive on the part before @ too, so maybe it doesn't matter)

tuzz: Ahh – I didn't know that. I think we shouldn't restrict it in that case. It's simpler without, anyway.

thomasleese: I suppose we could write our own validation which checks the part after the @ being case sensitive and the part before being case insensitive, but that feels a bit extreme.

PR #195, opened 2017-10-20
Add subscription model

thomasleese:
This adds a subscription model which is used to link together the subscriber and the subscriber list classes. 

[Trello Card](https://trello.com/c/JmOA1348/256-subscription-model)

### DIFF
--- a/app/models/subscriber.rb
+++ b/app/models/subscriber.rb
@@ -2,4 +2,7 @@ class Subscriber < ActiveRecord::Base
   validates :address, presence: true
   validates :address, format: { with: /@/, message: "is not an email address" }
   validates :address, uniqueness: true
+
+  has_many :subscriptions, dependent: :destroy
+  has_many :subscriber_lists, through: :subscriptions
 end
--- a/app/models/subscriber_list.rb
+++ b/app/models/subscriber_list.rb
@@ -9,6 +9,9 @@ class SubscriberList < ApplicationRecord
   validate :tag_values_are_valid
   validate :link_values_are_valid
 
+  has_many :subscriptions
+  has_many :subscribers, through: :subscriptions
+
   def self.build_from(params:, gov_delivery_id:)
     new(
       title: params[:title],
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -1,0 +1,3 @@
+class Subscription < ApplicationRecord
+  
+end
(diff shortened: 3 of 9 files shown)


app/models/subscriber.rb
```diff
@@ -2,4 +2,6 @@ class Subscriber < ActiveRecord::Base
   validates :address, presence: true
   validates :address, format: { with: /@/, message: "is not an email address" }
   validates :address, uniqueness: true
+
+  has_many :subscribers, dependent: :destroy
```
gpeng: Should this be `subscriptions`?

app/models/subscriber_list.rb
```diff
@@ -9,6 +9,8 @@ class SubscriberList < ApplicationRecord
   validate :tag_values_are_valid
   validate :link_values_are_valid
 
+  has_many :subscribers, dependent: :destroy
```
gpeng: And `subscriptions` here too?

app/models/subscriber.rb
thomasleese: Yep!

thomasleese:
@gpeng Regarding the `dependent: :destroy` I figured it might be useful between subscriptions and subscribers to easily delete subscriptions, but I left it out on the subscriber lists because I imagine we might want to be absolutely sure about only deleting subscriber lists when they have no subscribers.

thomasleese:
It didn't occur to me to include both the `has_many` and the `has_many :through` associations in each model, which makes a lot of sense, so I've added that now.

app/models/subscriber.rb
```diff
@@ -2,4 +2,7 @@ class Subscriber < ActiveRecord::Base
   validates :address, presence: true
   validates :address, format: { with: /@/, message: "is not an email address" }
   validates :address, uniqueness: true
+
+  has_many :subscriptions, dependent: :destroy
+  has_many :subscription_lists, through: :subscriptions
```
gpeng: `subscriber_lists` here

thomasleese: Clearly struggling with words this morning...

thomasleese:
Thanks @gpeng, good point. I've added some tests now that should cover everything.

gpeng:
Actually... Should we also add the uniqueness constraint as a validation in the model so we get `ActiveRecord` errors before the database ones kick in?

thomasleese:
@gpeng Good idea, have added one if you could take a look.

PR #196, opened 2017-10-20
Create notification model

thomasleese:
This model will be used to store the input we get from the `email-alert-service`.

This is the first part of [this Trello Card](https://trello.com/c/J7TCm10I/254-persist-notification-in-email-alert-api), the next part will be to actually create instances of this model when we receive a notification, which depends on [this Trello Card](https://trello.com/c/10k5XPez/264-add-title-and-changenote-to-email-alert-service).

This probably warrants some discussion on the fields we currently have in the model, since this is based on what we currently have in `NotificationLog` and some of the fields may not be necessary. It may also be necessary for us to store the priority in this model.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -1,0 +1,5 @@
+require_relative "extensions/symbolize_json"
+
+class Notification < ApplicationRecord
+  include SymbolizeJSON
+end
--- a/db/migrate/20171020091213_create_notifications.rb
+++ b/db/migrate/20171020091213_create_notifications.rb
@@ -1,0 +1,20 @@
+class CreateNotifications < ActiveRecord::Migration[5.1]
+  def change
+    create_table :notifications do |t|
+      t.uuid :content_id, null: false
+      t.text :title, null: false
+      t.text :base_path, null: false
+      t.text :change_note, null: false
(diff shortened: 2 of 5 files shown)


app/models/notification.rb
```diff
@@ -0,0 +1,21 @@
...
+class Notification < ApplicationRecord
+  include SymbolizeJSON
+
+  def subscriber_lists
+    subscriber_list_query.lists
+  end
+
+private
+
+  def subscriber_list_query
```
gpeng: I'm not sure this should be the concern of the `Notification` object. Obvs, we haven't really settled on the architecture of the matching but I would imagine it to happen outside of here.

db/migrate/20171020091213_create_notifications.rb
```diff
@@ -0,0 +1,19 @@
+class CreateNotifications < ActiveRecord::Migration[5.1]
+  def change
+    create_table :notifications do |t|
```
gpeng: We also need `base_path`. Sorry, forgot to put it on the ticket.

thomasleese: Ah yep, added that!

PR #199, opened 2017-10-24
Support sending instances of the email model

thomasleese:
Depends on the new `Email` model in [Trello Card](https://trello.com/c/DgNo1WHQ/267-create-an-email-on-receipt-of-a-notification).

[Trello Card](https://trello.com/c/2vIrpbFX/255-email-delivery-queue)

### DIFF
--- a/app/services/email_sender/notify.rb
+++ b/app/services/email_sender/notify.rb
@@ -2,33 +2,33 @@ require "notifications/client"
 
 module EmailSender
   class Notify
-    def call(address:, **keyword_args)
-      send_to_notify(address: address, **keyword_args)
+    def call(address:, subject:, body:)
+      client.send_email(
+        email_address: address,
+        template_id: template_id,
+        personalisation: {
+          subject: subject,
+          body: body,
+        },
+      )
     end
 
   private
 
-    def send_to_notify(address:, **keyword_args)
-      client.send_email(
-        email_address: address,
-        template_id: template_id(keyword_args)
-      )
-    end
-
     def client
-      @client ||= Notifications::Client.new(notify_api_key)
+      @client ||= Notifications::Client.new(api_key)
     end
 
-    def notify_config
+    def config
       EmailAlertAPI.config.notify
     end
 
-    def notify_api_key
-      notify_config.fetch(:api_key)
+    def api_key
+      config.fetch(:api_key)
     end
 
-    def template_id(template_id: nil, **_)
-      template_id || notify_config.fetch(:template_id)
+    def template_id
+      config.fetch(:template_id)
     end
   end
 end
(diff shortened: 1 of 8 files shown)


app/services/email_sender/notify.rb
```diff
@@ -2,16 +2,17 @@
 
 module EmailSender
   class Notify
-    def call(address:, **keyword_args)
-      send_to_notify(address: address, **keyword_args)
+    def call(address:, personalisation: {}, **keyword_args)
```
gpeng: Do we need to pass this as a `Hash` here? having `address:, subject:, body:` is much more email domain-esque although I suppose that couples us a bit to the idea that that's what we're sending. If we are going to pass a `Hash` I don't understand the name `personalisation`. That sounds specific to the recipient rather than the email.

thomasleese: I went back and forth on this... I agree the name `personalisation` is a bit rubbish, but it's what GOV.UK Notify calls it. I came to the conclusion that if we have `address:, subject:, body:` as arguments then the `template_id` because redundant because this class will only work with our defined template, but perhaps that's fine?

I actually prefer `address:, subject:, body:` so if we decide that's the way to go, then I would be up for that.

tuzz: To add more confusion to the mix...

I was thinking we'd just pass an `email` object which would have an association to the subscriber and it'd contain `subject` and `body` as attributes.

thomasleese: That's how I designed the `DeliveryToSubscriber` class, I like the distinction that and `EmailSender` gives us at the moment.

thomasleese: Thinking about it, I actually prefer `address:, subject:, body:` because it means we can completely abstract away that we're using GOV.UK Notify, which means it'll be easy enough to introduce a new, e.g. `EmailSender::Mailgun`, sender class if we needed to.

gpeng: I think we should definitely not use Notify terminology for exactly that reason. On the passing an `Email` instance. I've played a bit with that in the stuff I'm doing. I think if we pass the values where possible it reduces coupling. When I started passing objects I ended up on the slippery slope of having three `FactoryGirl.create` calls going on for each spec.

app/services/email_sender/notify.rb
```diff
@@ -2,33 +2,33 @@
...
-
-  private
-
-    def send_to_notify(address:, **keyword_args)
+    def call(address:, subject:, body:)
       client.send_email(
         email_address: address,
-        template_id: template_id(keyword_args)
+        template_id: template_id,
+        personalisation: {
```
gpeng: This is much better just going on in here. We're in `Notify` land in this class which is defo where the translation from our world should happen.

PR #200, opened 2017-10-24
Notification handling

gpeng:
This PR adds a second path for handling the POST request to `notifications#create`. This is implemented with a `NoficationHandler` which currently just creates a `Notification` object and an `Email` object.

We're effectively adding a v2 of the endpoint but haven't taken the decision to actually version the API so there are a few bits to this PR that result from this. The error handling currently catches everything and sends to Sentry without returning to the caller. This is the first good reason for considering a v2 endpoint that we've encountered as once we start to implement the response it will be difficult to do. For now I've stubbed each path out in each path's request specs, and the new stuff in the controller spec. 

[Trello](https://trello.com/c/DgNo1WHQ/267-create-an-email-on-receipt-of-a-notification)

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -1,0 +1,26 @@
+class Email < ApplicationRecord
+  belongs_to :notification
+  validates :subject, :body, :notification, presence: true
+
+  def self.create_from_params!(params)
+    build_from_params(params).tap(&:save!)
+  end
+
+  def self.build_from_params(params)
+    new.tap do |instance|
+      instance.notification_id = params[:notification_id]
+      instance.subject = params[:title]
+      instance.body = <<~BODY
+        description: #{params[:description]}
+        change_note: #{params[:change_note]}
+        base_path: #{params[:base_path]}
+        updated: #{instance.format_date(params[:public_updated_at])}
+      BODY
+    end
+  end
+
+  def format_date(date)
+    return unless date
+    date.strftime("%H:%M %-d %B %Y")
+  end
+end
--- a/app/services/notification_handler.rb
+++ b/app/services/notification_handler.rb
@@ -1,0 +1,10 @@
+require 'rails_helper'
+class NotificationHandler
+  def self.call(params)
+    notification = Notification.create(params)
+
+    email = Email.create(some_params)
+
+    (Matcher/Deliverer).perform_async(notification.id, email.id)
+  end
+end
--- a/app/services/notification_handler.rb
+++ b/app/services/notification_handler.rb
@@ -9,13 +9,17 @@ class NotificationHandler
   end
 
   def call
-    notification = Notification.create!(
-      notification_params
-    )
+    begin
+      notification = Notification.create!(
+        notification_params
+      )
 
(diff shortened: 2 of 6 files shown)


app/models/email.rb
```diff
@@ -0,0 +1,26 @@
+class Email < ApplicationRecord
+  belongs_to :notification
+  validates :subject, :body, :notification, presence: true
+
+  def self.create_from_params!(params)
+    build_from_params(params).tap(&:save!)
```
thomasleese: Do we need the `tap` here? Can we not just do `...(params).save!`?

app/models/email.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def self.build_from_params(params)
+    new.tap do |instance|
```
thomasleese: Not sure on the benefit of using `tap` here.

app/services/notification_handler.rb
```diff
@@ -0,0 +1,53 @@
...
+  def initialize(params:)
+    @params = params
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    begin
```
thomasleese: Do we need to wrap this in a transaction to think about what happens if only the second operation fails?

app/models/email.rb
gpeng: It's so as the method returns the instance. If we just do `.save!` it returns `true` or raises.

thomasleese: Ah yep, makes sense.

app/models/email.rb
gpeng: I just thought it was neater than doing: 

```
instance = new
instance.foo = params[:foo]
instance.bar = params[:bar]
#etc...
instance
```

gpeng: I suppose we could do `new(<the params>)`

thomasleese: Yeah, I was thinking `new()` but I don't mind really, was just wondering if there was another reason.

gpeng: Oh actually we couldn't because we need to do some stuff to them first.

app/services/notification_handler.rb
gpeng: I thought about that but I think if the first one fails we get the error report and if the second one fails then we get an error report but still have the notification that we can use to fix it. I decided we probably didn't want the first db call to not be committed because of the second one failing.

app/models/email.rb
thomasleese: Oh good point, yep.

gpeng: The params aren't actually 'the params' for the model

app/services/notification_handler.rb
thomasleese: I suppose if we're thinking of `Notification` as a log of all notifications then it is important this gets committed to the db. I imagine we'll add similar sort of Icinga alerts to this new system, so perhaps a mismatched number of emails and notifications could be something we proactively check for.

gpeng: Yeah... That would be good

PR #202, opened 2017-10-25
Add DeliverToSubscriberWorker

thomasleese:
Depends on #199.

[Trello Card](https://trello.com/c/2vIrpbFX/255-email-delivery-queue)

### DIFF
--- a/lib/tasks/deliver.rake
+++ b/lib/tasks/deliver.rake
@@ -1,6 +1,7 @@
 namespace :deliver do
   task :to_subscriber, [:id] => :environment do |_t, args|
     subscriber = Subscriber.find(args[:id])
-    DeliverToSubscriber.call(subscriber: subscriber)
+    email = OpenStruct.new(subject: "Test email", body: "This is a test email.")
+    DeliverToSubscriber.call(subscriber: subscriber, email: email)
   end
 end
--- a/app/commands/deliver_to_subscriber.rb
+++ b/app/commands/deliver_to_subscriber.rb
@@ -1,9 +1,12 @@
 class DeliverToSubscriber
   attr_reader :subscriber
+  attr_reader :email
 
-  def initialize(subscriber:)
+  def initialize(subscriber:, email:)
     @subscriber = subscriber
+    @email = email
     raise ArgumentError, "subscriber cannot be nil" if subscriber.nil?
+    raise ArgumentError, "email cannot be nil" if email.nil?
   end
 
   def self.call(*args)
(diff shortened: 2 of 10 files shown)


lib/tasks/deliver.rake
```diff
@@ -1,6 +1,7 @@
 namespace :deliver do
   task :to_subscriber, [:id] => :environment do |_t, args|
     subscriber = Subscriber.find(args[:id])
-    DeliverToSubscriber.call(subscriber: subscriber)
+    email = OpenStruct.new(subject: "Test email", body: "This is a test email.")
```
kevindew: Just had a thought about this. I see that we've used an openstruct here - which I guess is to get around persisting an email to the database and I wonder if we should actually be creating a model. Which, naturally, brings up the quagmire of what is the responsibilities of the Email class - is it to represent sending of an alert, or that the system sent an email? 

I'd probably argue for the latter as:
- it means they don't bypass the list of all emails we send
- we probably want test emails to be part of whatever system we use to determine if they send correctly

What do you think? are emails already coupled to a sending concept or is this still fair game?

thomasleese: Yeah, it took me a while to decide whether to persist a model or not here. I figured that we didn't need to, because I believe the idea for the future is to have another model which represents the actual email which got sent out via GOV.UK Notify and that will include the data on if the send went through correctly.

Perhaps we can have a single `Email` in the database which represents the test email, and then it can be associated with many `SentEmail`s if we decide that an association is the way to go.

I guess these are questions that need answering when we start to look into getting delivery reports, but perhaps not something to be decided right now?

kevindew: > Yeah, it took me a while to decide whether to persist a model or not here. I figured that we didn't need to, because I believe the idea for the future is to have another model which represents the actual email which got sent out via GOV.UK Notify and that will include the data on if the send went through correctly.

Sure, yeah I didn't think there was going to be another model that represents the actual email so sounds like we need to get our ducks in a row on what the domain model responsibilities are.

> I guess these are questions that need answering when we start to look into getting delivery reports, but perhaps not something to be decided right now?

Yeah seems a challenge knowing when best to resolve these. It seems a shame to know we're likely to have to revisit everything we do, but also wouldn't want to grind progress to a halt because of the unknowns.

🤔 hehe

So is a snapshot of your thoughts on this right now that this should be persisted somewhere eventually but we don't have the right construct yet?

thomasleese: > So is a snapshot of your thoughts on this right now that this should be persisted somewhere eventually but we don't have the right construct yet?

Yeah. In my mind, when we decided that it would be one `Notification` maps to one `Email`, there would be a new model created in the future which had the one to many relation between `Email` and the actual email which got sent to each individual subscriber.

kevindew: I see - that's different to the model in my head. So sounds like we need to get some shared understanding.

PR #209, opened 2017-10-31
Add pseudo email sender for use in dev vm

bevanloon:
Developers need to run `email-alert-api` locally with minimal annoyance. We don't want to have every team setting up Notify accounts and api keys just to run the application in the dev vm.

To accomplish this we:

* Add a Pseudo email sender which implements the same interface as the Notify sender (i.e. responds to `.call` taking three arguments: `address`, `subject` and `body`).
* Log emails sent to the Pseudo email sender to a file.
* Expose a config value `email_sender_provider`.
* Default our provider based on environment - in production it will default to Notify, other environments default to Pseudo.
* Allow overriding of the defaults with an env var

[Trello](https://trello.com/c/yg3QHfM8/277-create-service-to-wrap-sending-emails-to-notify)

### DIFF
--- a/app/services/email_sender/pseudo.rb
+++ b/app/services/email_sender/pseudo.rb
@@ -10,8 +10,7 @@ Body: #{body}
   private
 
     def logger
-      @logger ||= Logger.new("#{Rails.root}/log/pseudo_email.log")
+      @logger ||= Logger.new("#{Rails.root}/log/pseudo_email.log", 5, 4194304)
     end
   end
 end
-
--- a/app/services/email_sender/pseudo.rb
+++ b/app/services/email_sender/pseudo.rb
@@ -1,0 +1,17 @@
+module EmailSender
+  class Pseudo
+    def call(address:, subject:, body:)
+      logger.info(%(Sending email to #{address}
+Subject: #{subject}
+Body: #{body}
+))
+    end
+
+  private
+
+    def logger
+      @logger ||= Logger.new("#{Rails.root}/log/pseudo_email.log")
+    end
+  end
+end
+
--- a/config/email_service_provider.yml
+++ b/config/email_service_provider.yml
@@ -1,0 +1,8 @@
+development:
+  provider: <%= ENV["EMAIL_SERVICE_PROVIDER"] || "PSEUDO" %>
+
+test:
+  provider: <%= ENV["EMAIL_SERVICE_PROVIDER"] || "PSEUDO" %>
+
+production:
+  provider: <%= ENV["EMAIL_SERVICE_PROVIDER"] || "NOTIFY" %>
--- a/lib/email_alert_api/config.rb
+++ b/lib/email_alert_api/config.rb
@@ -23,6 +23,10 @@ module EmailAlertAPI
       @notify ||= notify_environment_config.symbolize_keys.freeze
     end
 
+    def email_service_provider
+      @email_service_provider ||= email_service_provider_config.fetch("provider").freeze
+    end
+
   private
 
     def redis_config_path
@@ -43,6 +47,14 @@ module EmailAlertAPI
 
     def notify_environment_config
       YAML.safe_load(ERB.new(File.read(notify_config_path)).result).fetch(@environment)
+    end
+
+    def email_service_provider_config_path
+      File.join(app_root, "config", "email_service_provider.yml")
+    end
+
+    def email_service_provider_config
+      YAML.safe_load(ERB.new(File.read(email_service_provider_config_path)).result).fetch(@environment)
     end
   end
 end
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -13,10 +13,8 @@ module Services
   end
 
   def self.email_sender
-    if Rails.env.production?
-      @email_sender ||= EmailSender::Notify.new
-    else
-      @email_sender ||= EmailSender::Pseudo.new
-    end
+    return @email_sender ||= EmailSender::Notify.new if EmailAlertAPI.config.email_service_provider == "NOTIFY"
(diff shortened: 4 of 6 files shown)


app/services/email_sender/pseudo.rb
```diff
@@ -0,0 +1,13 @@
+module EmailSender
+  class Pseudo
+    def call(address:, subject:, body:)
+      logger.info("Sending email to #{address}\r\nSubject: #{subject}\r\nBody: #{body}\r\n\r\n")
```
thomasleese: Might be nicer to use a multiline string here:

```ruby
%{
Sending email to #{address}
Subject: ...
}
```

config/email_service_provider.yml
```diff
@@ -0,0 +1,8 @@
+development:
+  provider: <%= ENV["GOVUK_EMAIL_ALERT_API_EMAIL_SERVICE_PROVIDER"] || "PSEUDO" %>
```
thomasleese: Do we need the `GOVUK_EMAIL_ALERT_API_` prefix on these? We don't do that for other configuration values.

thomasleese: Also not a fan of the all caps for the values, but that's just my personal preference.

lib/services.rb
```diff
@@ -12,6 +13,8 @@ def self.content_store
   end
 
   def self.email_sender
-    @email_sender ||= EmailSender::Notify.new
+    return @email_sender ||= EmailSender::Notify.new if
```
thomasleese: I think it would be good if we could raise an exception here if we don't get `notify` or `pseudo` since it's most likely a mistake.

lib/services.rb
```diff
@@ -12,6 +13,7 @@ def self.content_store
   end
 
   def self.email_sender
-    @email_sender ||= EmailSender::Notify.new
+    return @email_sender ||= EmailSender::Notify.new if EmailAlertAPI.config.email_service_provider == "NOTIFY"
```
thomasleese: I think it would be good if could raise an exception here if something is not what we expect. We could still accept `nil` as meaning the same as `pseudo`, but I think it would still be good if an exception is raised for something else.

lib/email_alert_api/config.rb
```diff
@@ -23,6 +23,10 @@ def notify
       @notify ||= notify_environment_config.symbolize_keys.freeze
     end
 
+    def email_service_provider
+      @email_service_provider ||= email_service_provider_config.symbolize_keys.fetch(:provider).freeze
```
thomasleese: Is it worth turning this into a symbol?

app/services/email_sender/pseudo.rb
bevanloon: updated that

config/email_service_provider.yml
bevanloon: probably right. I was being overly pedantic with the naming.

lib/email_alert_api/config.rb
bevanloon: maybe not - was originally exposing the whole yaml file here as with the rest of the class, but decided it wasn't worth it yet.

config/email_service_provider.yml
bevanloon: all caps feels to me more like a constant.....which these sort of represent. it doesn't make much difference to me though - if there's general outcry, I'm happy to lowercase them.

lib/email_alert_api/config.rb
thomasleese: Hmm, yeah, that makes sense if we may expose more in the future.

bevanloon: I've changed it anyway - no big deal to change it back if we need to.

lib/services.rb
bevanloon: added an exception to this now @thomasleese - see what you think

config/email_service_provider.yml
```diff
@@ -0,0 +1,8 @@
+development:
+  provider: <%= ENV["EMAIL_SERVICE_PROVIDER"] || "PSEUDO" %>
```
thomasleese: I know this is how other config files are done, but maybe we should be using `ENV.fetch` here.

lib/services.rb
```diff
@@ -12,6 +13,8 @@ def self.content_store
...
+    return @email_sender ||= EmailSender::Pseudo.new if EmailAlertAPI.config.email_service_provider == "PSEUDO" || EmailAlertAPI.config.email_service_provider.nil?
+    raise NameError, "Email service provider does not exist"
```
thomasleese: Wonder if it's worth including the email service provider that was given in the error message?

thomasleese: Also not sure if `NameError` is the right exception to use here as it feels like it has a very specific use case of undefined variables? But maybe it's normal to use this for other things in Ruby.

I would suggest an ordinary `RuntimeError`.

config/email_service_provider.yml
bevanloon: any particular reason? won't that raise an exception if the key is not found?
My intention behind this env var is that it should be used for override only and not be set by default.....

thomasleese: Oh sorry, I meant like `ENV.fetch("EMAIL_SERVICE_PROVIDER", "PSEUDO")` and similar. It's just a personal preference really, I refer that way of doing it.

bevanloon: I think we should keep it consistent with the rest of the app.....either that or change them all.

lib/services.rb
bevanloon: I liked NameError since it's "Raised when a given name is invalid or undefined." which is basically what we have, albeit a class that is literally undefined/non-existent. Anyhoo, bog standard generic raise now in place.

PR #213, opened 2017-11-02
Add email address override

bevanloon:
Broadly speaking this adds the ability to specify an override email address in `email_service.yml`.

At the same time, instead of exposing yet another config variable in `EmailAlertAPI::Config`, refactor to encapsulate the functionality of choosing the underlying provider and overriding the email addresses in an `EmailSender` class as suggested in the [trello](https://trello.com/c/yg3QHfM8/277-create-service-to-wrap-sending-emails-to-notify).

 * Rename `EmailSender` module to `EmailSenderService` to better describe what the namespace represents.
* Rename `email_service_provider.yml` to `email_service.yml` and add in an email address override option for each environment.
* Refactor `config.rb` to expose an `email_service` as a hash and use the keys on that instead of the previous `email_service_provider` method.
* Add a base `EmailSender` class to orchestrate sending emails
  * decides which underlying provider class to use based on config
  * overrides the email address if specified in config

Part of [Trello](https://trello.com/c/yg3QHfM8/277-create-service-to-wrap-sending-emails-to-notify)


app/services/email_sender_service/email_sender.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+  private
+
+    def config
+      EmailAlertAPI.config.email_service
+    end
+
+    def email_override
+      config.fetch(:email_address_override)
```
thomasleese: I don't know if it makes sense to use `.fetch` here, an absence of this key can be taken to mean they don't want an override.

config/email_service.yml
```diff
@@ -1,8 +1,11 @@
 development:
   provider: <%= ENV["EMAIL_SERVICE_PROVIDER"] || "PSEUDO" %>
+  email_address_override:
```
thomasleese: I wonder if these should also be configurable with environment variables?

app/services/email_sender_service/email_sender.rb
bevanloon: changed in 4a23d1cfb76349031d30739fe883ede788b6beb0

config/email_service.yml
bevanloon: changed in 4a23d1cfb76349031d30739fe883ede788b6beb0

app/services/email_sender_service/email_sender.rb
```diff
@@ -0,0 +1,45 @@
...
+        subject: subject,
+        body: body
+      )
+    end
+
+  private
+
+    def config
+      EmailAlertAPI.config.email_service
+    end
```
kevindew: I'd consider switching this to be injected in when we create this class (as that's only done in one place). 

You might even (with a tweak to the naming format) be able to get it down to this:

```
def initialize(config) # or could have keyword args
  @email_override = config[:email_address_override]
  provider = "EmailSenderService::#{config.fetch(:provider).classify}"
  @provider = provider.constantize.new
rescue NameError
  raise "Email service provider #{provider} does not exist"
end
```

app/services/email_sender_service/email_sender.rb
```diff
@@ -0,0 +1,45 @@
+module EmailSenderService
```
kevindew: I'd consider switching this to be a class of EmailSenderService rather than nesting a class of the same name in the module.

app/services/email_sender_service/email_sender.rb
```diff
@@ -0,0 +1,45 @@
...
+  class EmailSender
+    def call(address:, subject:, body:)
+      provider.call(
+        address: email_or_email_override(address),
```
kevindew: Could probably save a method here and have it as `email_override || address`

app/services/email_sender_service/email_sender.rb
```diff
@@ -0,0 +1,45 @@
+module EmailSenderService
+  class EmailSender
+    def call(address:, subject:, body:)
```
kevindew: Might be nice to have a log entry here that explains if it's overriden or not.

E.g. 
```
overriden = " (overriden to #{email_address_override}) " if email_address_overide
logger.info(%(Sending email to #{address}#{overriden} with subject: #{subject}")
```

and then adjust the pseudo one accordingly.

app/services/email_sender_service/email_sender.rb
bevanloon: do away with the namespacing? 😱 

perhaps we should have a general chat about the naming here as well as the underlying Notify and Pseudo classes sitting in the EmailSenderService namespace.

I was wondering if DeliveryService would be a better name at the module/namespace level and under that have EmailService as our orchestrator of Notify/Pseudo/any other provider?

kevindew: You'd still have namespacing just would be:

```
class EmailSenderService
   class Notify; end
   class Pseudo; end
end
```

rather than 

```
module EmailSenderService
   class EmailSender; end
   class Notify; end
   class Pseudo; end
end
```

bevanloon:
closing in favour of https://github.com/alphagov/email-alert-api/pull/215

app/services/email_sender_service/email_sender.rb
bevanloon: so I've moved this to t'other PR (https://github.com/alphagov/email-alert-api/pull/215) - I've chosen not to do the `constantize` dance because I find it obscures the code a bit, mostly when I come back to it a few months later. It becomes hard to grep where classes are being used when tracing code through and debugging and for me, just provides that extra bit of mental overhead that the more specific individual instantiation of classes provides. Happy to chat about it though if anyone thinks `constantize` is better.

app/services/email_sender_service/email_sender.rb
bevanloon: I've added this in here https://github.com/alphagov/email-alert-api/pull/215/files#diff-51c7eba122752bad14917f721d539715R8 for now. I haven't updated the pseudo class yet because I've yet to decide how to implement it nicely. Either I start passing in extra params when calling the Pseudo method (and then have to change the interface to the Notify class as well, or have a `switch` / `if` condition in here) or I just make the Pseudo class a complete noop and implement all the logging in here.....or something else entirely. haven't figured it out yet.

app/services/email_sender_service/email_sender.rb
bevanloon: 👍 so I've re-implemented this in https://github.com/alphagov/email-alert-api/pull/215
with the suggested structure

PR #214, opened 2017-11-03
Diagram that proposes how the system should work

tuzz:
https://trello.com/c/ygeNLLIt/303-next-round-of-email-alert-api-architecture
https://trello.com/c/CAG6H3hE/304-spike-digest-architecture-delivery-approach-and-size-of-the-work

This diagram tries to map out the steps required to send emails in response to changes to content. We should try to reach agreement on this so we can build the system with the same understanding. Please leave feedback, ask questions, raise concerns, etc.

👉  [diagram](https://raw.githubusercontent.com/alphagov/email-alert-api/80e23730fd9af4ebe0c7457b9b55656c0f2e3495/doc/sequence_diagram.png)


bevanloon:
Thinking about the "if failed" case and looking at the Notify docs (https://github.com/alphagov/notifications-ruby-client#get-the-status-of-one-message) for their return codes....in the first instance, we possibly only need to resend from our side if we get one of the 400 message codes back from Notify. I think that represents a catastrophic failure where we have an email that somehow didn't get saved in Notify properly.

bevanloon:
I was also wondering if the EmailBuildWorker actually needs to receive the enqueues or if it could  just run every x minutes and grab data from the EmailEntry table that it needs to send. I suppose then we would need to build in the logic that decided if it was looking for "daily" or "weekly" things as well based on datetime......so I'm not sure if that makes anything better / easier to maintain.

tuzz:
>Thinking about the "if failed" case ...

Yeah, following on from the chat we had with Notify this morning, the`DeliveryMonitor` can probably be quite simple. They retry temporary failures for up to 3 day. They also have a `permanent-failure` state which is probably when we should remove/blacklist that subscriber from the system.

I'm going to leave that part of the diagram for now as I think we can elaborate it when we work on that bit of the system.

>I was also wondering if the EmailBuildWorker actually needs to receive the enqueues ...

We could just run it every x minutes. Either approach would probably be fine.

PR #215, opened 2017-11-06
Email address override

bevanloon:
Broadly speaking this adds the ability to specify an override email address in `email_service.yml`.

At the same time, instead of exposing yet another config variable in `EmailAlertAPI::Config`, refactor to encapsulate the functionality of choosing the underlying provider and overriding the email addresses in an `EmailSender` class as suggested in the [trello](https://trello.com/c/yg3QHfM8/277-create-service-to-wrap-sending-emails-to-notify).

 * Rename `EmailSender` module to `EmailSenderService` to better describe what the namespace represents.
* Rename `email_service_provider.yml` to `email_service.yml` and add in an email address override option for each environment.
* Refactor `config.rb` to expose an `email_service` as a hash and use the keys on that instead of the previous `email_service_provider` method.
* Add a base `EmailSender` class to orchestrate sending emails
  * decides which underlying provider class to use based on config
  * overrides the email address if specified in config

Part of [Trello](https://trello.com/c/yg3QHfM8/277-create-service-to-wrap-sending-emails-to-notify)

Supersedes https://github.com/alphagov/email-alert-api/pull/213


app/services/email_sender/email_sender_service.rb
```diff
@@ -0,0 +1,30 @@
+class EmailSenderService
+  def initialize(config, email_service_provider)
+    @email_address_override = config[:email_address_override]
+    @provider = email_service_provider
+  end
+
+  def call(address:, subject:, body:)
+    log_override(address: address, subject: subject, body: body) if @email_address_override
+
+    @provider.call(
```
thomasleese: I would be tempted to make this and `email_address_override` an `attr_reader` to look a bit nicer.

PR #217, opened 2017-11-07
Add description to deliver rake tasks

bevanloon:
Add a description to `deliver:to_subscriber` and `deliver:to_test_email` rake tasks.
Refactor and extract the `test_email` into its own method.

### DIFF
--- a/lib/tasks/deliver.rake
+++ b/lib/tasks/deliver.rake
@@ -1,15 +1,19 @@
 namespace :deliver do
+  def test_email
+    OpenStruct.new(subject: "Test email", body: "This is a test email.")
+  end
+
   desc "Send a test email to a subscriber by id"
   task :to_subscriber, [:id] => :environment do |_t, args|
     subscriber = Subscriber.find(args[:id])
-    email = OpenStruct.new(subject: "Test email", body: "This is a test email.")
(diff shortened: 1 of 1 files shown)


lib/tasks/deliver.rake
```diff
@@ -1,13 +1,19 @@
 namespace :deliver do
+  def test_email
+    OpenStruct.new(subject: "Test email", body: "This is a test email,")
+  end
+
+  desc "Send a test email to a subscriber by Id"
```
thomasleese: `Id` with a capital `I` there I think looks a bit odd.

lib/tasks/deliver.rake
```diff
@@ -1,13 +1,19 @@
 namespace :deliver do
+  def test_email
+    OpenStruct.new(subject: "Test email", body: "This is a test email,")
```
thomasleese: I think there should be a `.` at the end of the string, rather than a `,`!

bevanloon: quite right....good spot.

lib/tasks/deliver.rake
bevanloon: fixed

PR #220, opened 2017-11-07
Rename models and services following a discussion

tuzz:
https://trello.com/c/ygeNLLIt/303-next-round-of-email-alert-api-architecture

[discussion document](https://docs.google.com/document/d/15SEs81CPPipxMUFWgmN6KUZFpkiJkV0Uv9bFZpk9dG8/edit)


doc/sequence_diagram.puml
```diff
@@ -5,67 +5,67 @@ skinparam dpi 300
...
 controller -> email_alert_service : acknowledge
 
 group sidekiq
-  subscription_fulfilment_worker -> content_change : find by id
-  content_change -> subscription_fulfilment_worker : instance
-  subscription_fulfilment_worker -> subscription_matcher : content_change
+  subscription_content_worker -> content_change : find by id
+  content_change -> subscription_content_worker : instance
+  subscription_content_worker -> subscription_matcher : content_change
   subscription_matcher -> subscription_matcher : ... existing logic
```
bevanloon: Is this the point that we interact with `Subscribable` (`SubscriberList` in the current system), `Subscription` and `Subscriber` models? Do we want to capture this on the sequence diagram?

tuzz: Yes, at this point the `SubscriptionMatcher` matches the `ContentChange` against `Subscribables` then finds all `Subscriptions` associated with these.

We probably don't interact with the `Subscriber` much at this point. We're more interested in their associated `Subscriptions`.

We could add it to the diagram but I decided not to because:
- the diagram's already quite unwieldy and I didn't want to make it bigger
- the matching behaviour's super complicated and I didn't want to unpick it all

I think there's probably a separate task to take a look at that matching logic and simplify it. For example, it's currently using lots of Postgres JSON querying techniques that I think we could get rid of eventually.

PR #221, opened 2017-11-09
Introduce EmailRenderer class

thomasleese:
Currently this has all the same behaviour as what was in the `Email` class, but it prepares us for the future.

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -7,21 +7,12 @@ class Email < ApplicationRecord
   end
 
   def self.build_from_params(params)
-    new.tap do |instance|
-      instance.notification_id = params[:notification_id]
-      instance.subject = params[:title]
-      instance.body = <<~BODY
-        There has been a change to *#{params[:title]}* on #{instance.format_date(params[:public_updated_at])}.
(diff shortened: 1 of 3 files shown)


app/models/email.rb
```diff
@@ -7,21 +7,12 @@ def self.create_from_params!(params)
...
-        **#{params[:change_note]}**
-      BODY
-    end
-  end
-
-  def format_date(date)
-    return unless date
-    date.strftime("%H:%M %-d %B %Y")
+    new(
+      notification_id: params[:notification_id],
```
bevanloon: What's the reasoning behind not getting the notification id from the rendered?

thomasleese: Just that this is not relevant to the renderer, only the email - and in the future email will have an `address` field. The renderer is only concerned with the user facing parts of the email, i.e. the subject and the body.

bevanloon: is it worth passing all the `params` into the renderer in that case? or are we keeping this flexible until we are more confident about what it needs?

thomasleese: Keeping it flexible, I think this can be looked at when the email content work is done.

PR #222, opened 2017-11-09
DeliveryAttempt Model

thomasleese:
Introduces a new model which will hold delivery attempts.

Depends on #223 .

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -1,5 +1,6 @@
 class Email < ApplicationRecord
   belongs_to :notification
+
   validates :address, :subject, :body, :notification, presence: true
 
   def self.create_from_params!(params)
--- a/app/models/delivery_attempt.rb
+++ b/app/models/delivery_attempt.rb
@@ -1,0 +1,24 @@
+class DeliveryAttempt < ApplicationRecord
+  belongs_to :email
+
+  validates :email, :status, :provider, :reference, presence: true
+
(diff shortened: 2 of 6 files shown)


app/models/email.rb
```diff
@@ -1,5 +1,7 @@
 class Email < ApplicationRecord
   belongs_to :notification
+  has_many :delivery_attempts
+
```
tuzz: I'd be tempted to not add this association to `Email` so that visibility is asymmetric:

- Email doesn't know about DeliveryAttempts
- Delivery attempt knows about Email

It'd mean we'd go from: `email.delivery_attempts` to `DeliveryAttempt.where(email: email)`.

I think the advantage of this approach is it reduces the coupling between these models and encourages us to keep `DeliveryAttempt` specific behaviour out of the `Email` model.

This is probably controversial though? What do you think?

thomasleese: Yeah, I actually completely agree with this. I was hesitant to actually include this in the first place...

kevindew:
Just wanted to check in here with the range of status' have you considered the scenarios for when Notify 400s or similar? Where we might need to set the status without hearing anything back from Notify

thomasleese:
@kevindew Yeah, so when we get a `400` we set it to `internal-failure` which suggests retrying will not help (because it's most likely that our api key is misconfigured) and when we get a `429` we set it to `technical-failure` which will eventually trigger a retry.

PR #224, opened 2017-11-09
Write up of ADR for Notify integration

kevindew:


doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+### SubscriptionContent
+
+**Responsibility**: Represent that a content change should or has been emailed to a
+subscription
+
+#### Shall
+
+- Have an association with a ContentChange
+- Have an association with a Subscription
+- Have a nullable association with Email
```
thomasleese: Why nullable?

thomasleese: In fact, wouldn't this be better as a many-to-many relationship between emails?

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+- Store information such as
+  - Recipient
+  - Subject
+  - Body (in Notify Markdown)
+- Delivery status
+
+#### Shall Not
+
+- Have more than one recipient
+- Know anything about the content that led to its generation
```
thomasleese: If this is true, then I don't think the `SubscriptionContent` above should even have an association (nullable or not) with emails.

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Should
+
+- Only exist in the database for a non permanent time period
+- Be associated with a Subscriber for logging purposes
```
thomasleese: I wanted to avoid this as much as possible, is logging just the email address enough?

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+### DeliveryAttempt
+
+**Responsibility**: Storing the information used to communicate with the
+transportation service for email
+
+#### Shall
+
+- Store information such as
+  - Email associated with
+  - Provider (e.g. Notify)
```
thomasleese: Is this definitely required or are we predicting a future where we officially support different providers?

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+- It is not essential that digests return the same content for each user with
+  same subscriptions
+  - E.g. a digest processed for Joe processed 2 mins after Steve may get
+    slightly different results if something was published in that time
+- If you are subscribed to things that produce the same content for two
+  subscriptions at different digest frequencies, you will only receive
+  notifications for a piece of content once (on the higher frequency)
+- Digests take effect from the point of subscription
+  - E.g. if you joined a weekly digest on a Wednesday that was sent on Fridays
+    you’d only get alerts for content after you joined We can send remaining
```
rubenarakelyan: Is this a separate bullet point?

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+## Specifics
+
+The architecture can be visualised through two diagrams:
+
+1. [Domain model](../domain.png) which represents the data models and how they
+   relate
+2. [Sequence diagram](../sequence_diagram.png) which represents all of the
+   components and how they communicate to achieve specified functionality
+
+This document then goes through each of the concepts in the diagram and outline
```
rubenarakelyan: Nit: outlines

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Shall
+
+- Store information such as
+  - List name
+
+#### Should
+
+- Require that the list name is unique
+- Maintain a relationship with corresponding govdelivery list
```
rubenarakelyan: Should this be flagged as a temporary thing?

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+information specific to that list.
+
+#### Shall
+
+- Store information such as
+  - List name
+
+#### Should
+
+- Require that the list name is unique
```
rubenarakelyan: This is a GovDelivery requirement that we can get rid of in our own system

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+- Maintain a relationship with corresponding govdelivery list
+- Store information about what subscription criteria it matches
+  (links, tags, document_supertypes)
+- Have information to source where the Subscribable originated from
+
+#### May
+
+- In future, have an association to an entity that models the subscription
+  criteria information (links, tags, super types)
+- Release the requirement of a list name being unique if we are able to
```
rubenarakelyan: As per comment above

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### May
+
+- Store preferences that affect all of a users subscriptions
+- Have the ability to be disabled if there are problems with a users account
+- Store other personalisation information such as name
+- Be associated with statistics regarding the subscriber
+
+#### To determine
+- What should we allow in terms of removing a subscriber - do we use the
```
rubenarakelyan: An IA question here: do we need to completely purge a user if they delete themselves from the system? (Guessing yes since we no longer have a reason to store their details). I think this will come out of the IA discussions we're having at the moment.

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+  associated data as a log?
+
+### Subscription
+
+**Responsibility**: Store that a Subscriber has an association to a SubscriberList
```
rubenarakelyan: Should `SubscriberList` be `Subscribable`?

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+  won’t result in a success
+
+#### May
+
+- React to the transportation layer returning a 429 status code
+
+
+#### To determine
+
+- What should happen when an email has exhausted retries and failed
```
rubenarakelyan: Temporary failure: give up but keep subscriber
Permanent failure: give up and delete subscriber
Notify failure: retry sending at next opportunity

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Should
+
+- Be able to retry Emails that fail to send
+- Be able to abort trying to send an email if it’s a situation where retrying
+  won’t result in a success
+
+#### To determine
+
+- What should happen to stop a eternal retry loop?
```
rubenarakelyan: In which scenarios could this happen?

doc/arch/adr-001-notify-integration.md
kevindew: yup 👍

doc/arch/adr-001-notify-integration.md
kevindew: Nah - I'll merge them

doc/arch/adr-001-notify-integration.md
bevanloon: Also interesting is what we do with users who have subscribed to some emails and then unsubscribed from everything. Do we auto-remove? Also something to think about in the future if/when we are relying on links from emails essentially being the user's way to access their account; a user unsubscribed from everything no longer gets any emails and may not be aware that their data is still stored on our system....or may be fairly confused about how to get back to delete their account.

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+- Trigger EmailGenerationWorker for emails that have to run immediately
+
+#### Should not
+
+- Be able to have multiple instances of this worker processing the same
+  ContentChange concurrently
+
+#### May
+
+- Be passed in frequency as an argument to match subscriptions of a given
```
bevanloon: s/Be passed in/Accept ?

doc/arch/adr-001-notify-integration.md
kevindew: Yes these are interesting problems.

@rubenarakelyan yup be great if we can worth that out. Is this something already on an agenda for discussions or should we log it somewhere?

@bevanloon I assume that the account still exists even if there are no subscriptions, it helps maintain our logs better and can offer benefits of maintaining users preferences. I don't know though if this factors into IA if we have any responsibility to remove data after a while.

doc/arch/adr-001-notify-integration.md
kevindew: Seems to just be (assuming our system is working correctly) if Notify returns notify-failure.

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+  frequency, as to allow rebuilding of emails that went out in error
```
bevanloon: nit - s/as/in order 
or just remove it. The comma is potentially not needed either.

doc/arch/adr-001-notify-integration.md
```diff
@@ -0,0 +1,424 @@
...
+
+### EmailDeliveryWorker
+
+**Responsibility**: Finding Emails that should be sent and creating
+DeliveryAttempts for them and updating Email status accordingly
+
+#### Shall
+
+- Communicate with a transportation layer (Notify) to send an Email
+- Create a DeliveryAttempt object to associate with the Email
```
thomasleese: I prefer that the email sending service which communicates with the transportation layer creates the `DeliveryAttempt` object, in that way there's less risk of the system getting into an inconsistent state since it's happening very close to where the "delivery attempt" happened.

kevindew:
Should be all comments covered here. I did some tweaking of the [EmailDeliveryWorker](https://github.com/alphagov/email-alert-api/blob/712e6af75a54941901cfe71c018f1cc10f9eb9d7/doc/arch/adr-001-notify-integration.md#emaildeliveryworker) and [DeliveryMonitor](https://github.com/alphagov/email-alert-api/blob/712e6af75a54941901cfe71c018f1cc10f9eb9d7/doc/arch/adr-001-notify-integration.md#deliverymonitor) for handling their error scenarios

doc/arch/adr-001-notify-integration.md
rubenarakelyan: @kevindew Having spoken to Tim, we should just blank out the email address of the user if they unsubscribe from everything. We can still keep the record since other things rely on it but then it no longer has any personal information in it. I've created a card to do this for planning.

PR #225, opened 2017-11-09
Rename DeliverToSubscriber to DeliverEmail 

thomasleese:
Since emails now include the address directly.

### DIFF
--- a/app/workers/deliver_email_worker.rb
+++ b/app/workers/deliver_email_worker.rb
@@ -1,4 +1,4 @@
-class DeliverToSubscriberWorker
+class DeliverEmailWorker
   include Sidekiq::Worker
 
   def self.queue_for_priority(priority)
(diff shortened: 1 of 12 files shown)


app/workers/deliver_email_worker.rb
```diff
@@ -13,10 +13,9 @@ def self.queue_for_priority(priority)
 
   sidekiq_options retry: 3, queue: queue_for_priority(:low)
 
-  def perform(subscriber_id, email_id)
-    subscriber = Subscriber.find(subscriber_id)
+  def perform(email_id)
     email = Email.find(email_id)
```
bevanloon: In https://github.com/alphagov/email-alert-api/pull/225/files?diff=unified#diff-e067234861884b7cc76410c92f6b6377R6 we are raising if `email.nil?`, here, however, using`find` will raise `RecordNotFound`, which is the point we would get `email` being `nil`. Are we swallowing the `RecordNotFound`? If not will we ever pass `nil` to `DeliverEmail`? 

I know we are also calling from the rake tasks but I'm not sure that's too risky to be honest.

thomasleese: I think it's highly unlikely we will ever send a `nil` email, but I think it's still a useful check and might help with debugging any weird problems.

PR #227, opened 2017-11-10
Return reference and errors from Email Sender

thomasleese:
This returns the provider reference on a successful email send call, and raises a common provider-agnostic `ClientError` exception when there is a failure. This should make it easier to build a `DeliveryAttempt` object from the response of the sender.

[Trello Card](https://trello.com/c/oAaQvqQE/320-create-a-delivery-monitor-to-check-status-of-emails-sent-to-notify)


thomasleese:
Now that the system has simplified somewhat since Notify will provide us with web hooks, this is no longer necessary.

PR #228, opened 2017-11-10
Remove created delivery attempt status

thomasleese:
We will not need this status since we can start it in the sending state.

### DIFF
--- a/app/models/delivery_attempt.rb
+++ b/app/models/delivery_attempt.rb
@@ -3,19 +3,15 @@ class DeliveryAttempt < ApplicationRecord
 
   validates :email, :status, :provider, :reference, presence: true
 
-  enum status: %i(sending delivered permanent_failure temporary_failure technical_failure internal_failure)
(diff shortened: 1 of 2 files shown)


app/models/delivery_attempt.rb
```diff
@@ -3,19 +3,15 @@ class DeliveryAttempt < ApplicationRecord
 
   validates :email, :status, :provider, :reference, presence: true
 
-  enum status: %i(created sending delivered permanent_failure temporary_failure technical_failure internal_failure)
+  enum status: %i(sending delivered permanent_failure temporary_failure technical_failure)
   enum provider: %i(psuedo notify)
 
   def failure?
     permanent_failure? || temporary_failure? || technical_failure? || internal_failure?
```
bevanloon: spurious '|| internal_failure?' here?

thomasleese: Oh yep, fixed!

PR #229, opened 2017-11-10
Add an endpoint to receive status updates for delivery attempts

tuzz:
https://trello.com/c/oAaQvqQE/320-create-a-delivery-monitor-to-check-status-of-emails-sent-to-notify

### DIFF
--- a/app/controllers/status_updates_controller.rb
+++ b/app/controllers/status_updates_controller.rb
@@ -2,13 +2,13 @@ class StatusUpdatesController < ApplicationController
   wrap_parameters false
 
   def create
-    StatusUpdateWorker.perform_async(status_update_params)
+    StatusUpdateWorker.perform_async(**status_update_params)
     render plain: "queued for processing", status: :accepted
   end
 
 private
 
   def status_update_params
-    params.permit(:reference, :status).to_h
+    params.permit(:reference, :status).to_h.symbolize_keys
   end
 end
--- a/app/controllers/status_updates_controller.rb
+++ b/app/controllers/status_updates_controller.rb
@@ -1,0 +1,14 @@
+class StatusUpdatesController < ApplicationController
+  wrap_parameters false
+
+  def create
+    StatusUpdateWorker.perform_async(status_update_params)
+    render plain: "queued for processing", status: :accepted
+  end
+
+private
+
+  def status_update_params
+    params.permit(:reference, :status).to_h
+  end
+end
--- a/app/workers/status_update_worker.rb
+++ b/app/workers/status_update_worker.rb
@@ -1,38 +1,19 @@
 class StatusUpdateWorker
   include Sidekiq::Worker
+  include Sidekiq::Symbols
 
-  attr_accessor :reference, :status
+  def perform(reference:, status:)
+    delivery_attempt = DeliveryAttempt.find_by!(reference: reference)
+    delivery_attempt.update!(status: status.underscore)
 
-  def perform(params)
-    params.deep_symbolize_keys!
-
-    self.reference = params.fetch(:reference)
-    self.status = params.fetch(:status).underscore
-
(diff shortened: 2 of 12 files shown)


app/services/unsubscriber.rb
```diff
@@ -0,0 +1,8 @@
+module Unsubscriber
```
thomasleese: For such a simple module, I wonder if we can just make this is a method on `Subscriber`.

db/migrate/20171110163345_make_subscriber_address_nullable.rb
```diff
@@ -0,0 +1,5 @@
+class MakeSubscriberAddressNullable < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :subscribers, :address, true
```
thomasleese: My only concern with allowing `null` values here is it's possible to end up with strings which are either `null` or `""` which effectively mean the same thing in our system. Perhaps the validation will catch any problems here though.

app/workers/status_update_worker.rb
```diff
@@ -0,0 +1,38 @@
+class StatusUpdateWorker
+  include Sidekiq::Worker
+
+  attr_accessor :reference, :status
+
+  def perform(params)
+    params.deep_symbolize_keys!
+
+    self.reference = params.fetch(:reference)
```
thomasleese: Might be nicer to allow the worker to take `reference` and `status` as specific arguments, and then the caller (in which case the controller) can deal with extracting the values from POST input.

db/migrate/20171110163345_make_subscriber_address_nullable.rb
tuzz: Good point. We should only have one version of empty so it's not confusing.

I've added a test to ensure this can't be an empty string so this column can only be `nil`. We could do it the other way round, Rails has good out-of-the-box support for `allow_nil` that we can take advantage of this way round.

Here's the test: https://github.com/alphagov/email-alert-api/pull/229/files#diff-16d4cbe1eafec712dda5d67ccfcca877R38

What do you think?

tuzz:
Thanks @thomasleese. I've made those changes. Please could you take another look?

I've broken out the healthchecking into separate tickets:
https://trello.com/c/sehNMIbL/246-make-healthcheck-test-for-technicalfailure
https://trello.com/c/31XA1mkY/248-integrate-email-alert-api-with-notifys-callback

And I've added an integration ticket for when Notify have done their callback work:
https://trello.com/c/31XA1mkY/248-integrate-email-alert-api-with-notifys-callback

Have I missed anything you can think of? The change turned out to be quite small.

tuzz:
Tested on my local machine by manually creating the models then posting to the endpoint with sidekiq running. I can see the status change and a `permanent_failure` set the subscriber's email address to `NULL` and deleted its subscriptions.

```
$ curl -H "Content-Type: application/json" -X POST -d '{"reference":"ref-123","status":"permanent_failure"}' http://localhost:3088/status_updates
```

app/controllers/status_updates_controller.rb
```diff
@@ -0,0 +1,14 @@
+class StatusUpdatesController < ApplicationController
+  wrap_parameters false
+
+  def create
```
thomasleese: Do we need to think about how we can secure this webhook now or is that for later? Because if the route is public, it does mean anyone can change the status of delivery attempts right now.

tuzz: Yep, we're proposing we add a HTTP header that sets a bearer token.

I held off implementing this as we're yet to confirm with Notify.

(more context: https://docs.google.com/a/digital.cabinet-office.gov.uk/document/d/1g7NlOQT-e02n1nL4DXSNfrvZfFmkQtiAmh5IjNx-10A/edit?usp=sharing)

PR #230, opened 2017-11-10
Fix running `rake db:migrate` against a new database

tuzz:
Database migrations were broken because:
- they didn’t include the Rails version number
- the hstore migrations require root access

We’re not using hstore anymore so I’ve commented
out the creation of these columns. This meant I
had to disable some data migrations that depend on
those columns.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -71,8 +71,8 @@ ActiveRecord::Schema.define(version: 20171109091838) do
   end
 
   create_table "subscriber_lists", id: :serial, force: :cascade do |t|
-    t.string "title"
-    t.string "gov_delivery_id"
+    t.string "title", limit: 255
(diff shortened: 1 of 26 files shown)


db/schema.rb
```diff
@@ -71,8 +71,8 @@
   end
 
   create_table "subscriber_lists", id: :serial, force: :cascade do |t|
-    t.string "title", limit: 255
-    t.string "gov_delivery_id", limit: 255
+    t.string "title"
```
thomasleese: Do you know why these changed?

tuzz: Yep, I did some digging around. I've added a commit to explicitly set these:
https://github.com/alphagov/email-alert-api/pull/230/commits/bca15f51e92167ad687f3c2eb8d0e7ccd23ca0f5

PR #231, opened 2017-11-10
Add the SubscriptionContent model

tuzz:
This adds the model and connects it up to how the system would look, but synchronously. Splitting the `NotificationHandler` up into the appropriate workers feels like something for a different PR.

[Trello Card](https://trello.com/c/bjVnCN7D/328-create-the-subscriptioncontent-model)

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -5,14 +5,6 @@ class Email < ApplicationRecord
     build_from_params(params).tap(&:save!)
   end
 
-  def self.create_from_subscription_content!(params, subscription_content)
-    create_from_params!(
-      params.merge(
-        address: subscription_content.subscription.subscriber.address
-      )
-    )
-  end
-
   def self.build_from_params(params)
     renderer = EmailRenderer.new(params: params)
 
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -5,6 +5,14 @@ class Email < ApplicationRecord
     build_from_params(params).tap(&:save!)
   end
 
+  def self.create_from_subscription_content!(params, subscription_content)
+    create_from_params!(
+      params.merge(
+        address: subscription_content.subscription.subscriber.address,
+      )
+    )
+  end
+
   def self.build_from_params(params)
     renderer = EmailRenderer.new(params: params)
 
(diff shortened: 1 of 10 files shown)


db/migrate/20171110171050_add_subscription_content.rb
```diff
@@ -0,0 +1,10 @@
+class AddSubscriptionContent < ActiveRecord::Migration[5.1]
+  def change
+    create_table :subscription_contents do |t|
+      t.references :subscription, null: false, foreign_key: true
+      t.references :notification, null: false, foreign_key: true
+      t.references :email, null: true, foreign_key: true
```
kevindew: Are we considering at this point archiving emails? as we might not want this to be an actual foreign key for that?

Haven't thought it through but wondering if we want cascades on other columns at all?

tuzz: I don't really mind either way re. foreign keys. We can remove them later if there's a need to archive this data, perhaps?

Which cascades would you want to add? Do you mean in the context of this change or are you thinking more generally about all the models in the system so far?

kevindew: Yeah if they're situations that aren't too problematic to swap up out later fine to defer until we know more.

I was just referring to this change here. Without any cascades any attempts to remove notification or subscription will fail. Is that cool?

thomasleese: I think that's probably fine for now since we have no code to delete stuff, and it's a useful check to prevent accidental deletion in the console. We can make a decision on cascades when we need to start looking at how we can archive/delete stuff easily I think.

kevindew: Sure, sounds fair

app/models/email.rb
```diff
@@ -5,6 +5,14 @@ def self.create_from_params!(params)
     build_from_params(params).tap(&:save!)
   end
 
+  def self.create_from_subscription_content!(params, subscription_content)
+    create_from_params!(
+      params.merge(
+        address: subscription_content.subscription.subscriber.address
```
gpeng: This couples things really deeply. If all we need is the address would it be better to just add that to the params and use `create_from_params`?

thomasleese: Yep, agreed. That should be fixed now!

PR #232, opened 2017-11-14
Rename Notification Model to ContentChange

1pretz1:
ran migrations to update the email reference associated with
notifications and update tests to now call upon ContentChange

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -1,7 +1,5 @@
 class Email < ApplicationRecord
-  belongs_to :content_change
-
-  validates :address, :subject, :body, :content_change, presence: true
+  validates :address, :subject, :body, presence: true
 
   def self.create_from_params!(params)
     build_from_params(params).tap(&:save!)
@@ -11,7 +9,6 @@ class Email < ApplicationRecord
     renderer = EmailRenderer.new(params: params)
 
     new(
-      content_change_id: params[:content_change_id],
       address: params[:address],
       subject: renderer.subject,
       body: renderer.body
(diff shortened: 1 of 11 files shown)


db/migrate/20171113163731_update_emails_reference_from_notifications_to_content_changes.rb
```diff
@@ -0,0 +1,6 @@
+class UpdateEmailsReferenceFromNotificationsToContentChanges < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :emails, :notification_id
+    add_reference(:emails, :content_change, null: false, foreign_key: true)
```
thomasleese: I'm not sure we actually need this new column back! I think we decided to not have a relation between emails and content changes.

thomasleese: If we decided to keep this in, we'll run into a problem where all the fields will be `NULL` when the column is created, and that will fail because of the `null: false`. I wonder if we can remove the foreign key, rename the column and then bring it back.

app/models/email.rb
```diff
@@ -11,7 +9,6 @@ def self.build_from_params(params)
     renderer = EmailRenderer.new(params: params)
 
     new(
-      notification_id: params[:notification_id],
```
thomasleese: Nice! 🎉 

I just thought... we might have to move this tiny change into a separate PR and deploy that first to prevent an old version of the application from trying to save this field - alternatively, since the errors will be sent to Sentry and we don't really care about the new system yet, we can leave it.

PR #233, opened 2017-11-14
Rate limit Notify calls

gpeng:
The Notify service enforces a rate limit of 50 calls per second. To allow us to respect that this PR adds a very simple rate limiting mechanism.

The theory goes that if we have 3 boxes with 2 workers each, then each worker should only be allowed to make 8 requests per second. (This should be configurable via an env var but it is hard coded for now while I get some feedback and prove it works. I'll add that in another PR).

The `RateSleeper` class measures the length of time it takes to retrieve the stuff from the DB and make the Notify request and then either:

returns 0 if that length of time is already greater than the duration we need to limit to

or

returns the length of time in seconds that, when added to the time already taken by the request will result in the worker being throttled to the desired rate.

Some things for discussion:

* it's a bit tricky to integration test. We could stub the `Email.find` and `DeliverEmail.call` calls and `sleep` in one of those but I expect that will be flakey
* Sidekiq's scheduler runs at 15 second intervals by default. I'm not sure what effect this will have, probably none but might be worth considering
* By not storing the 'actual' time we shouldn't be vulnerable to BST shenanigans so I've not tested it

[Trello](https://trello.com/c/vXfGdgkA/340-implement-rate-limiting)


app/workers/deliver_email_worker.rb
```diff
@@ -14,8 +14,19 @@ def self.queue_for_priority(priority)
...
     DeliverEmail.call(email: email)
+
+    duration = (start - Time.now.utc) * 1000.0
+
+    sleep(
+      RateSleeper.sleep_time_in_seconds(
+        execution_time_in_milliseconds: duration,
+        allowed_rate_per_second: 8
+      )
+    )
```
kevindew: You might want to only run sleep if this class returns a non zero value as (from a quick google) `sleep(0)` doesn't seem to be a no-op

app/workers/deliver_email_worker.rb
```diff
@@ -14,8 +14,19 @@ def self.queue_for_priority(priority)
...
+
     email = Email.find(email_id)
     DeliverEmail.call(email: email)
+
+    duration = (start - Time.now.utc) * 1000.0
+
+    sleep(
+      RateSleeper.sleep_time_in_seconds(
+        execution_time_in_milliseconds: duration,
+        allowed_rate_per_second: 8
```
kevindew: Do we need this magic number to be configurable at all?

tuzz:
I had a kooky idea for an interface:

```ruby
module Services
  def self.rate_limiter
    @rate_limiter ||= RateLimiter.new(per_second: ENV["REQUESTS_PER_SECOND"])
  end
end

# then:

def send_to_notify(email)
  Services.rate_limiter.throttle { NotifyClient.deliver(email) }
end
```

Or something like that? It might make sense to have `RateLimiter.instance` or `.singleton` rather than putting it on the `Services` module.

Just an idea. `RateSleeper` also looks good 👍

app/workers/deliver_email_worker.rb
```diff
@@ -14,8 +14,19 @@ def self.queue_for_priority(priority)
...
 
   def perform(email_id)
+    start = Time.now.utc
+
     email = Email.find(email_id)
     DeliverEmail.call(email: email)
+
+    duration = (start - Time.now.utc) * 1000.0
+
+    sleep(
```
thomasleese: Feels to me like we could have another method in the `RateSleeper` class which does the `sleep` for us and does the check @kevindew mentioned on `sleep(0)`.

thomasleese: Maybe even provide something similar to what @tuzz is suggesting and have a helper method which does all the above and accepts a block to execute.

```ruby
RateSleeper.run do
  email = Email.find(email_id)
  DeliverEmail.call(email: email)
end
```

tuzz:
Rather than skimming your code just before I went home, I've actually read your code properly this morning!

Your implementation of rate limiting is a little different to how I envisioned it, but I think it's much simpler in that it doesn't need to persist state in the process of when the last request was made. 🎉 

I'm actually thinking we could simplify it further and just do this in the worker:

```ruby
def perform(email_id)
  sleep(sleep_time)
  # ...
end

def sleep_time
   ENV.fetch("RATE_LIMIT_SLEEP_TIME", 0.125).to_f
end
```

Regarding testing, I think we could write a test for the worker that does this:

```ruby
it "sleeps for a set amount of time so the rate limit isn't exceeded" do
  time_taken = Benchmark.realtime { subject.perform }
  expect(time_take).to be_within(0.01).of(0.125)
end
```

If the test is unreliable, we could stub `sleep_time` to be a larger number and increase the tolerance?

app/workers/deliver_email_worker.rb
```diff
@@ -14,8 +14,19 @@ def self.queue_for_priority(priority)
   sidekiq_options retry: 3, queue: queue_for_priority(:low)
 
   def perform(email_id)
+    start = Time.now.utc
+
     email = Email.find(email_id)
     DeliverEmail.call(email: email)
+
+    duration = (start - Time.now.utc) * 1000.0
```
bevanloon: is this going to return a negative number?
in which case, will https://github.com/alphagov/email-alert-api/pull/233/files#diff-69752d22c8871b0f2ec09d2de4ef14aeR6 ever be true for a positive `minimum_duration_in_milliseconds`?

gpeng: Back to front... Time is not one of my strong points. 🤕 I've got rid of that stuff now.

gpeng:
I've updated based on the comments and discussions. It's tricky to test so the tests are 'a bit vague' :tm: but I think they cover the essence of what we're trying to do.

gpeng:
@tuzz I left the slightly more complicated version in as if we always sleep for 0.125 or whatever we might end up slowing things down quite a bit with a big queue depending on the duration of the Notify request. e.g if the request takes 150ms we'd end up blocking the worker for 0.275 which will add up over a big queue.

app/services/rate_sleeper.rb
```diff
@@ -0,0 +1,34 @@
+class RateSleeper
+  attr_reader :jobs_per_second
```
bevanloon: super pedantic so ignore if you fancy, but we could set the `attr_reader` in `private` since it is only used internally.

gpeng: I expect nothing less from you... Fixed in 37a16a9

spec/services/rate_sleeper_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+          expect(rate_sleeper).not_to receive(:sleep)
+
+          rate_sleeper.run {
+            Kernel.sleep(0.13)
+          }
+        end
+      end
+    end
+
+    context "with jobs per second 20" do
```
bevanloon: Just wondering what this context "tests" given we are already exercising the code's maths (and hurting my head manually calculating it all) in https://github.com/alphagov/email-alert-api/pull/233/files#diff-7adf437dfb818ed2ba40a94bbea07125R5 ? Is it just that we can pass in different `jobs_per_second` and that will affect the resulting sleep times?

bevanloon:
Nice use of `Benchmark` to neatly encapsulate the logic you had before 👍

spec/services/rate_sleeper_spec.rb
gpeng: yep, it's just testing that the configuration parameter actually configures the thing correctly

PR #234, opened 2017-11-14
Make each of the healthchecks influence overall status

tuzz:
https://trello.com/c/sehNMIbL/339-make-healthcheck-test-for-technicalfailure

Previously, the only healthchecks that influenced
the overall status of the `/healtcheck` endpoint
were the database and redis connections.

This change makes it so that the status is set to
the most severe status of the individual checks.
It also substantially refactors the controller and
extracts lots of `Healthcheck` models for each of 
the checks.

Each of these models has a uniform interface
responding to `#name`, `#status` and `#details` and
the top-level Healthcheck also has the interface
and combines everything into one.

---

Responds with this JSON on integration:
```json
{
  "status": "ok",
  "checks": {
    "database": {
      "status": "ok"
    },
    "govdelivery": {
      "status": "ok",
      "ping_status": 200
    },
    "queue_latency": {
      "status": "ok",
      "queues": {
        "default": 0
      }
    },
    "queue_size": {
      "status": "ok",
      "queues": {
        "default": 0
      }
    },
    "redis": {
      "status": "ok"
    },
    "retry_size": {
      "status": "ok",
      "retry_size": 0
    }
  }
}
```

### DIFF
--- a/app/models/healthcheck.rb
+++ b/app/models/healthcheck.rb
@@ -1,0 +1,38 @@
+class Healthcheck
+  def status
+    if statuses.include?(:critical)
+      :critical
+    elsif statuses.include?(:warning)
+      :warning
+    else
+      :ok
+    end
+  end
+
+  def details
+    { checks: checks }
+  end
+
+private
+
+  def checks
+    all.each.with_object({}) do |check, hash|
+      hash[check.name] = check.details.merge(status: check.status)
+    end
+  end
+
+  def statuses
+    @statuses ||= all.map(&:status)
+  end
+
+  def all
+    @all ||= [
+      DatabaseHealthcheck.new,
+      GovdeliveryHealthcheck.new,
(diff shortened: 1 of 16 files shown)


app/models/healthcheck.rb
```diff
@@ -0,0 +1,42 @@
+class Healthcheck
+  def name
+    "overall"
+  end
+
+  def status
+    if statuses.any? { |s| s == :critical }
```
thomasleese: Might be more succinct to do `statuses.include?(:critical)` here and below.

app/models/healthcheck.rb
```diff
@@ -0,0 +1,42 @@
+class Healthcheck
+  def name
```
thomasleese: This should probably return a symbol like the others, but also I'm not convinced we even need this method.

tuzz:
@thomasleese: updated

PR #235, opened 2017-11-15
Add a healthcheck for technical failures of delivery attempts

tuzz:
https://trello.com/c/sehNMIbL/339-make-healthcheck-test-for-technicalfailure

Alerts in Icinga:

screen shot 2017-11-15 at 14 59 17

screen shot 2017-11-15 at 15 00 24

### DIFF
--- a/app/models/delivery_attempt.rb
+++ b/app/models/delivery_attempt.rb
@@ -6,6 +6,11 @@ class DeliveryAttempt < ApplicationRecord
   enum status: %i(sending delivered permanent_failure temporary_failure technical_failure)
   enum provider: %i(pseudo notify)
 
+  def self.latest_per_email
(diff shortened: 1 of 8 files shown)


tuzz:
You raise a good point. We should keep all `DeliveryAttempt`s and that would mean the alert hangs around for a day unless it's ack'ed.

Maybe I should only check the last `DeliveryAttempt` for an email so that if one gets successfully delivered the technical failure is ignored?

thomasleese:
Oh yeah, that sounds like a better check. Something that is more like "if there is an email which has only 'technical_failure' delivery attempts".

tuzz:
@thomasleese: I've added https://github.com/alphagov/email-alert-api/pull/235/commits/e8c6d66ba0dba25ae381e3ce6a2f917b7c3cd7af to address your comment.

PR #237, opened 2017-11-16
Improve email format

thomasleese:
Update the email format to match the design.

[Trello Card](https://trello.com/c/iWTDQ2sn/256-add-content-to-formatted-email)

### DIFF
--- a/app/models/email_renderer.rb
+++ b/app/models/email_renderer.rb
@@ -4,16 +4,17 @@ class EmailRenderer
   end
 
   def subject
-    params[:title]
+    "GOV.UK Update - #{title}"
   end
 
   def body
     <<~BODY
-      There has been a change to *#{params[:title]}* on #{format_date(params[:public_updated_at])}.
+      #{change_note}: #{description}.
 
-      > #{params[:description]}
+      #{url}
+      Updated on #{public_updated_at}
 
-      **#{params[:change_note]}**
+      Unsubscribe from #{title} - #{unsubscribe_url}
     BODY
   end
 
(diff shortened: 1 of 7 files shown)


spec/models/email_spec.rb
```diff
@@ -31,19 +31,11 @@
     }
 
     it "sets subject" do
-      expect(email.subject).to eq("Title")
+      expect(email.subject).to_not be_nil
```
gpeng: Why are we only testing that these are not `nil` now?

thomasleese: I figured since we already have tests for the `EmailRenderer`, it didn't seem useful to check the contents again.

app/models/email_renderer.rb
```diff
@@ -4,25 +4,49 @@ def initialize(params:)
...
+
+  def description
+    params.fetch(:description)
+  end
+
+  def public_updated_at
+    params.fetch(:public_updated_at).strftime("%I:%M %P, %-d %B %Y")
+  end
+
+  def url
```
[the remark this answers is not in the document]
thomasleese: Good point, fixed that now!

spec/models/email_spec.rb
gpeng: Could we stub the methods on `EmailRenderer` and then test that they are set to what the `EmailRenderer` returns instead?

thomasleese: Sure, fixed that now!

app/models/email_renderer.rb
```diff
@@ -4,25 +4,49 @@ def initialize(params:)
...
+    "#{Plek.new.website_root}#{base_path}"
+  end
+
+  def unsubscribe_url
+    "https://www.gov.uk/email/token/unsubscribe"
```
gpeng: Should we use `Plek` here too just to be consistent? I realise this link doesn't actually work anyway at the moment.

thomasleese: Good point, fixed!

PR #240, opened 2017-11-20
Add Email Generation Worker

thomasleese:
Introduces a new Email Generation Worker responsible for taking Subscription Content instances and turning them into emails for the Delivery Worker to send them.

[Trello Card](https://trello.com/c/dZd19B5G/347-add-the-emailgenerationworker)

### DIFF
--- a/app/workers/email_generation_worker.rb
+++ b/app/workers/email_generation_worker.rb
@@ -1,0 +1,31 @@
+class EmailGenerationWorker
+  include Sidekiq::Worker
+  include Sidekiq::Symbols
+
+  def perform(subscription_content_id:, priority:)
+    subscription_content = SubscriptionContent.find(subscription_content_id)
+
+    email = Email.create_from_params!(email_params(subscription_content))
+
+    subscription_content.update!(email: email)
+
+    DeliverEmailWorker.perform_async_with_priority(
+      email.id, priority: priority.to_sym
+    )
+  end
+
+private
+
+  def email_params(subscription_content)
+    content_change = subscription_content.content_change
+
+    {
+      title: content_change.title,
+      change_note: content_change.change_note,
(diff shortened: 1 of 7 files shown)


app/workers/email_generation_worker.rb
```diff
@@ -0,0 +1,35 @@
...
+    subscription_content.update!(email: email)
+
+    DeliverEmailWorker.perform_async_with_priority(
+      email.id, priority: priority.to_sym
+    )
+  end
+
+private
+
+  def subscription_contents
```
gpeng: I don't think this is being called

app/workers/email_generation_worker.rb
```diff
@@ -0,0 +1,35 @@
...
+  def email_params(subscription_content)
+    content_change = subscription_content.content_change
+
+    {
+      title: content_change.title,
+      change_note: content_change.change_note,
+      description: content_change.description,
+      base_path: content_change.base_path,
+      public_updated_at: content_change.public_updated_at,
+      address: subscription_content.subscription.subscriber.address,
```
gpeng: We've got that deep coupling again here. We should have a rethink about how this works as it keeps cropping up.

app/workers/email_generation_worker.rb
thomasleese: Yep, removed this.

app/workers/email_generation_worker.rb
thomasleese: Agreed, not sure if it's something we should discuss in a dev catchup rather than for this PR as there isn't really a way around it.

gpeng: Ok.. Let's leave it in here then and we can discuss some alternatives.

PR #242, opened 2017-11-20
Add sidekiqued subscription content worker

bevanloon:
Create a SubscriptionContentWorker - a Sidekiq worker that takes a content change id and a priority and then matches subscribers to the content change and queues emails for delivery to said subscribers.

Remove the code from NotificationHandler that was doing this previously and instead call the new Sidekiq worker.

[Trello](https://trello.com/c/nKLn1k1E/350-add-the-subscriptioncontentworker)

### DIFF
--- a/app/services/notification_handler.rb
+++ b/app/services/notification_handler.rb
@@ -11,46 +11,13 @@ class NotificationHandler
   def call
     begin
       content_change = ContentChange.create!(content_change_params)
-      deliver_to_subscribers(content_change)
-      deliver_to_courtesy_subscribers
+      SubscriptionContentWorker.perform_async(content_change_id: content_change.id, priority: priority)
     rescue StandardError => ex
       Raven.capture_exception(ex, tags: { version: 2 })
     end
   end
 
 private
-
-  def deliver_to_subscribers(content_change)
-    subscriptions_for(content_change: content_change).find_each do |subscription|
-      subscription_content = SubscriptionContent.create!(
-        content_change: content_change,
-        subscription: subscription,
-      )
-
-      EmailGenerationWorker.perform_async(
-        subscription_content_id: subscription_content.id,
-        priority: priority
-      )
-    end
-  end
-
-  def deliver_to_courtesy_subscribers
-    addresses = [
-      "govuk-email-courtesy-copies@digital.cabinet-office.gov.uk",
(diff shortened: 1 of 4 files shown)


app/services/notification_handler.rb
```diff
@@ -10,54 +10,16 @@ def self.call(*args)
...
-      )
-    end
-  end
-
-  def subscriptions_for(content_change:)
-    SubscriptionMatcher.call(content_change: content_change)
-  end
-
-  def content_change_params
+  def notification_params
```
thomasleese: I think this should be called `content_change_params`.

app/workers/subscription_content_worker.rb
```diff
@@ -0,0 +1,80 @@
...
+        DeliverEmailWorker.perform_async_with_priority(
+          email.id, priority: priority
+        )
+      rescue StandardError => ex
+        Raven.capture_exception(ex, tags: { version: 2 })
+      end
+    end
+  end
+
+  def subscribers_for(content_change:)
```
thomasleese: I think we've lost of the use of `SubscriptionMatcher` for the various methods here.

app/services/notification_handler.rb
bevanloon: 👍 should be fixed now

app/workers/subscription_content_worker.rb
bevanloon: Yep....reinstated

spec/workers/subscription_content_worker_spec.rb
```diff
@@ -0,0 +1,118 @@
+require "rails_helper"
+
+RSpec.describe SubscriptionContentWorker do
```
thomasleese: There is no test which calls the Sidekiq `perform_async` method which I suspect will fail as the worker does not `include Sidekiq::Symbols`.

gpeng: I think `.perform_async` is in `Sidekiq::Worker` isn't it?

gpeng: Oh... It's needed for keyword arguments. 🙊

bevanloon: okedokey - try that

PR #244, opened 2017-11-21
Rename DeliverEmailWorker

gpeng:
The `DeliverEmailWorker` now does a bit more than delivery email so we're renaming it.

This PR creates a new `DeliveryRequestWorker` which is identical to the current `DeliverEmailWorker` at the moment and switches `NotificationHandler` to use the new one.

The old `DeliverEmailWorker` will be removed after this has been deployed.

Part of [Trello](https://trello.com/c/BxtZNx5v/352-add-the-deliveryrequestworker)

### DIFF
--- a/app/workers/delivery_request_worker.rb
+++ b/app/workers/delivery_request_worker.rb
@@ -1,0 +1,27 @@
+class DeliveryRequestWorker
+  include Sidekiq::Worker
+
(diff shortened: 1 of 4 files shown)


gpeng:
@thomasleese I had to make quite a few changes to fix the merge conflicts from your last PR. Could you just give this another once over please?

PR #247, opened 2017-11-22
Add `ready_to_send` status

gpeng:
Adds another value to the `status` enum of `DeliveryAttempt` and makes this the default value. This value should be used when the `DeliveryAttempt` is first created as per [the flow diagram](https://github.com/alphagov/email-alert-api/blob/master/doc/sequence_diagram.png).

This will change the integer value associated with all of the previously created `DeliveryAttempt` records. These will be deleted once this change is deployed.

We could do this without having to delete everything by just adding the new status value to the end of the enum but `ready_to_send` should be the default which should really be the 0 value. So... while we still have the opportunity to just delete everything I've opted to do that now. We'll probably need to delete all of the existing `DeliveryAttempt` records as some point anyway as they are all in `sending` state and that won't get updated.

### DIFF
--- a/app/models/delivery_attempt.rb
+++ b/app/models/delivery_attempt.rb
@@ -3,7 +3,14 @@ class DeliveryAttempt < ApplicationRecord
 
   validates :email, :status, :provider, presence: true
 
-  enum status: %i(sending delivered permanent_failure temporary_failure technical_failure)
+  enum status: %i(
+    ready_to_send
+    sending
+    delivered
+    permanent_failure
+    temporary_failure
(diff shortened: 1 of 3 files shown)


gpeng:
Discussed this offline. This is part of work to update the `DeliveryRequestWorker` in line with [the flow diagram](https://github.com/alphagov/email-alert-api/blob/master/doc/sequence_diagram.png) We've agreed that it may not be necessary as the absence of a `DeliveryRequest` implies that the process failed. I'll close this for now.

PR #248, opened 2017-11-23
Set processed at

gpeng:
This PR adds a call to `Email#mark_processed!` in the `DeliverEmail` command which updates the `Email` record adding a timestamp.

We did consider doing this within a transaction but having thought about it a partially complete set of data and an exception reported would be more useful than just the exception so I've omitted the transaction for now.

We also discussed changing the order or events within the `DeliverEmail` command and adding a `ready_to_send` state to `DeliveryRequest` but after some discussion within the team we decided that that state is implied by the presence of an `Email` object without a corresponding `DeliveryAttempt` so the extra bit of complexity is probably unnecessary.

[Trello](https://trello.com/c/BxtZNx5v/352-add-the-deliveryrequestworker)

### DIFF
--- a/app/commands/deliver_email.rb
+++ b/app/commands/deliver_email.rb
@@ -17,6 +17,8 @@ class DeliverEmail
       body: email.body,
     )
 
+    email.mark_processed!
+
     record_delivery_attempt(
       email: email,
       status: :sending,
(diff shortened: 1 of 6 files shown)


gpeng:
Closing as we're not going to do this now. `processed_at` will be inferred from the presence of a `DeliveryAttempt`.

PR #251, opened 2017-11-23
Add an /unsubscribe endpoint

tuzz:
https://trello.com/c/o5ffFODH/386-add-an-endpoint-to-email-alert-api-to-process-unsubscriptions

### DIFF
--- a/app/controllers/unsubscribe_controller.rb
+++ b/app/controllers/unsubscribe_controller.rb
@@ -1,8 +1,15 @@
 class UnsubscribeController < ApplicationController
   def unsubscribe
-    uuid = params.fetch(:uuid)
+    Unsubscribe.subscription!(subscription)
+  end
 
-    subscription = Subscription.find_by!(uuid: uuid)
-    subscription.destroy
+private
+
+  def subscription
+    Subscription.find_by!(uuid: uuid)
+  end
+
+  def uuid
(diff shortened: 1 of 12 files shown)


app/controllers/unsubscribe_controller.rb
```diff
@@ -0,0 +1,16 @@
+class UnsubscribeController < ApplicationController
+  def unsubscribe
+    Unsubscribe.subscription!(subscription)
+    render plain: "deleted #{uuid}", status: 200
```
thomasleese: Can we return JSON here? Since we're probably doing that in other places.

I'm also wondering if we should return a status of `204`.

app/services/unsubscribe.rb
```diff
@@ -0,0 +1,27 @@
+module Unsubscribe
```
thomasleese: Should we call this `UnsubscribeService`?

tuzz: I don't mind, @gpeng ?

app/controllers/unsubscribe_controller.rb
tuzz: I've updated to a `204` as that seems more appropriate 👍

tuzz:
Going to rename a bunch of things in a separate PR. Keen to get this merged.

PR #252, opened 2017-11-24
Rescue ActiveRecord::NotFound

gpeng:
Rescue by default and return `status: 404`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,4 +2,12 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :null_session
+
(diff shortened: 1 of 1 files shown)


kevindew:
@gpeng Looks good, but I'm just wondering if we don't get this behaviour from rails by default?

bevanloon:
also - this is in https://github.com/alphagov/email-alert-api/pull/251/commits/de791064e406e87a82def62941dae66378ef207d but with a `render plain` is it important to get it in sooner and do we care about `plain` vs `json`?

gpeng:
@bevanloon We've taken it out from there. @kevindew We were trying to get it to return the 404 in request specs too instead of raising but it turns out there's config that does that so I'll close this.

kevindew:
@gpeng Oh very cool about the config option. I've always done the nasty expect raise and that always felt wrong in request specs.

PR #255, opened 2017-11-24
Use hostname in metrics

kevindew:
Trello: https://trello.com/c/u20s96bS/353-set-up-statistics-monitoring-to-inform-product-decisions

[An issue has been identified][statsd-pr] with the way statsd reports
statistics that can lead to incorrect values for high frequency counts.
A work around this issue is to include the hostname in the key which is
sent to statsd, doing this also gives us the advantage that we can
compare the rates of usage for particular hosts.

[statsd-pr]: https://github.com/alphagov/govuk-puppet/pull/6834

### DIFF
--- a/app/services/email_sender/notify.rb
+++ b/app/services/email_sender/notify.rb
@@ -11,10 +11,10 @@ class EmailSenderService
           body: body,
         },
       )
-      EmailAlertAPI.statsd.increment("notify.email_send_request.success")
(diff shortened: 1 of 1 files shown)


kevindew:
@thomasleese we can aggregate them dynamically in graphite/grafana - does require a bit of graphite witchcraft though.

PR #256, opened 2017-11-26
Update Ruby to 2.4.2 and bump Gemfile dependancies

RoryMacDonald:
Ahead of enabling Dependabot on the repository, we want to be using the latest
dependancies, to minimise the onslaught of incremental PRs.


Gemfile
```diff
@@ -30,7 +30,7 @@ group :development do
 end
 
 group :development, :test do
-  gem 'govuk-lint'
+  gem 'govuk-lint', '3.3.0'
```
thomasleese: Should we do `~> 3.3.0` so we get subsequent compatible updates?

kevindew: Assuming we semvar it `'~> 3.3'` would be better

RoryMacDonald: Thanks @thomasleese @kevindew. Both of these will lock the Gemfile to 3.3.1 which contains a load of incompatible lint changes. So suggest this bump gets handled separately

PR #261, opened 2017-11-28
Introduce more statistics

kevindew:
This includes
- Stats for how many content changes are entering the system
- Timing for the time between when an email is created and when we first send it to a third party service
- Timing for the time between a content change being created and the eventual email being sent to a third party service

See commit messages for further details

It was quite tricky in this to work out what the component should be that has the cross cutting concern of Email through to ContentChange - so I went for `EmailTimingStatsService` which acts kind of like a module, it didn't feel too great. So suggestions welcome on improving the modelling of it (without creating too many classes)

### DIFF
--- a/app/queries/content_changes_for_email_query.rb
+++ b/app/queries/content_changes_for_email_query.rb
@@ -1,0 +1,9 @@
+class ContentChangesForEmailQuery
+  def self.call(email)
+    subscription_content = SubscriptionContent
+      .where("subscription_contents.content_change_id = content_changes.id")
+      .where(email: email)
+      .exists
+    ContentChange.where(subscription_content)
+  end
+end
--- a/app/services/deliver_email_service.rb
+++ b/app/services/deliver_email_service.rb
@@ -17,6 +17,8 @@ class DeliverEmailService
       body: email.body,
     )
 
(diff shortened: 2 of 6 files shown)


app/models/content_change.rb
```diff
@@ -3,6 +3,14 @@
 class ContentChange < ApplicationRecord
   include SymbolizeJSON
 
+  scope :for_email, ->(email) do
+    subscription_content = SubscriptionContent
+      .where("subscription_contents.content_change_id = content_changes.id")
+      .where(email: email)
+      .exists
+    where(subscription_content)
+  end
```
tuzz: I'd maybe consider moving this out of `ContentChange` and either into `Email` or the `Service`. I'm not overly keen on giving `ContentChange` knowledge of how everything associates to it. As far as `ContentChange` is concerned it just holds some information about what's changed and the rest of the system builds its structures around it.

But... I don't really mind as it's only used for stats tracking (which is always a bit of an awkward thing). What do you think?

kevindew: Sure, I get you. Maybe it's better if I create a Query object. I had it originally in the Service but felt like something that should be tested in it's own right.

kevindew: @tuzz I've introduced this into a query class now.

PR #265, opened 2017-11-30
Create subscription endpoint

thomasleese:
This creates a new subscriptions controller with an endpoint (`POST /subscriptions`) which can be used to create new subscriptions.

I made a decision to name the variables and fields `subscribable`s to follow our idealised model, although I realise this may be confusing to mix the two, so we may decide to not do this.

Users of the API are expected to know about the subscriber list IDs which means we're still tied to them being created in GovDelivery in this PR, but that feels like something which should be dealt with in a later PR.

[Trello Card](https://trello.com/c/GpyQh4gq/407-add-a-subscription-creation-endpoint-to-email-alert-api)

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -2,4 +2,12 @@ class SubscriptionsController < ActionController::Base
   def create
     render json: {}, status: :created
   end
+
+private
+
+  def subscription_params
+    params.require(:address)
+    params.require(:subscribable_id)
+    params.permit(:address, :subscribable_id)
+  end
 end
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,11 +1,18 @@
 class SubscriptionsController < ActionController::Base
   def create
-    subscription = Subscription.create!(
+    subscription = Subscription.find_or_initialize_by(
       subscriber: subscriber,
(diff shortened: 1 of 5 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    if subscription.new_record?
+      subscription.save!
+      status = :created
+    else
+      status = :ok
+    end
+
+    render json: { id: subscription.id }, status: status
```
gpeng: I wonder should we use the `uuid` field here as we interact with other parts of the API (unsubscribe at the moment) using that. I've approved with this as is as I think it's fine but curious if anybody else has any opinions.

bevanloon: I don't think it matters too much at this point - I'm usually in favour of not leaking incrementing numbers if we don't have to, but that's probably a false concern really since we can't actually use these against any of our endpoints to mutate data anyway (I think). 

I wonder if the returned Id is even needed? For unsubscribe, we just returned a status code with no body didn't we?

thomasleese: I figured the ID might be useful for debugging if we have to manually trigger that endpoint for whatever reason.

bevanloon: yep, probably will come in handy at some point :)

PR #267, opened 2017-12-01
Add a migration to fix subscriber lists with no title

thomasleese:
For each subscriber list with no title, it fetches the topic from GovDelivery and either puts in the title or deletes the subscriber list if it has no subscribers.

In production there are 740 subscriber lists with no title, and out of them about half have no subscribers.

The migration takes about 15 minutes to run.

[Trello Card](https://trello.com/c/GDln9WNI/405-add-titles-to-subscriber-lists-that-dont-have-titles)


thomasleese:
Closed in preference to using: https://github.com/alphagov/email-alert-api/blob/master/lib/tasks/data_hygiene.rake#L17

PR #273, opened 2017-12-04
Add subscribables#show endpoint

gpeng:
We need to be able to retrieve a `SubscriberList` by gov_delivery_id as that is all we currently have in the frontend.

There is an existing `SubscriberListsController#show` but when only the `gov_delivery_id` param is passed it always returns the same `SubscriberList` as the existing action performs a complex lookup query on other params in addition to the `gov_delivery_id`. This could be considered a bug but we will be changing `SubscriberList` to `Subscribable` at some point soon and to avoid changing existing code that is running in production until we're ready to replace this I've added this `Subscribables` controller.

There is also a bit of an issue with using `gov_delivery_id` but the existing functionality in the frontends relies on it so we should probably continue with that naming for now to avoid too much change in the frontend for the MVP stuff.

`SubscriberList` serialises to `subscriber_list: <attributes>` so there is a bit of funkiness to avoid that.

[Trello](https://trello.com/c/YtmHgU3b/408-add-page-to-email-alert-frontend-to-capture-a-users-email-address)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,7 @@ Rails.application.routes.draw do
   resources :notifications, only: %i[create index show]
   resources :status_updates, path: "status-updates", only: %i[create]
   resources :subscriptions, only: %i[create]
+  get "subscribables/:gov_delivery_id", to: "subscribables#show"
 
   get "/healthcheck", to: "healthcheck#check"
 
--- a/app/controllers/subscribables_controller.rb
+++ b/app/controllers/subscribables_controller.rb
@@ -1,0 +1,12 @@
+class SubscribablesController < ApplicationController
+  def show
+    render json: subscribable_json
+  end
+
+private
+
+  def subscribable_json
+    subscriber_list = SubscriberList.find_by(gov_delivery_id: params[:gov_delivery_id])
+    { subscribable: subscriber_list.attributes }
+  end
+end
--- a/spec/requests/subscribables_controller_spec.rb
+++ b/spec/requests/subscribables_controller_spec.rb
@@ -1,0 +1,14 @@
+require "rails_helper"
+
+RSpec.describe SubscribablesController, type: :request do
+  describe "GET /subscribables/<govuk_delivery_id>" do
+    it "returns the subscribable" do
+      subscribable = create(:subscriber_list, gov_delivery_id: "test135")
+      get "/subscribables/test135"
+
+      subscribable_response = JSON.parse(response.body).deep_symbolize_keys[:subscribable]
+
+      expect(subscribable_response[:id]).to eq(subscribable.id)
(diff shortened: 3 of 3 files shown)


config/routes.rb
```diff
@@ -5,6 +5,7 @@
   resources :notifications, only: %i[create index show]
   resources :status_updates, path: "status-updates", only: %i[create]
   resources :subscriptions, only: %i[create]
+  get "subscribables/:gov_delivery_id", to: "subscribables#show"
```
thomasleese: I'm wondering if the `gov_delivery_id` should be a query parameter. Then it makes it easier for us to change this into a more generic search query in the future?

gpeng: I did consider that but decided it was going down the non-standard 'show' path that the `SubscriberListController` already has (the one that I wasted loads of time scratching my head about yesterday that led me to doing this). I think we should stick with the almost restful pass an id (admittedly not 'the' id here) and get the thing. I think the action that we'll eventually add to replace the stuff on `SubscriberListController` would accept parameters and be more like a search type thing. We might have some trouble with versioning but having this separate controller altogether will hopefully help with that. All that said, happy to change if I'm in the minority though.

thomasleese: My main concern with that was when we eventually get rid of GovDelivery IDs, it could cause us problems or make it a lot harder to move away from them as what seems like the standard `GET /subscribables/:id` endpoint isn't really using an ID.

bevanloon: What do we ultimately want the interface to look like? (i.e. once we get rid of GovDelivery IDs, how will we want this to work)? Probably a silly question, but I'm not really picturing it in my head......

gpeng: I still think we'll need some way of getting the subscribable with 'that' id though. We'll call it something else perhaps but I still don't imagine well just use the incrementing integer id. Like with do already with `Subscription#uuid` for unsubscribing. If we use `subscribables?gov_delivery_id=<id>` then you'd expect that to return a collection as it would be the index by convention. We could return a collection with one thing in and deal with that at the other end but that seems wrong too.

gpeng: I suppose it's analogous to getting a content item by slug. The slug isn't the db id but it's the id that we choose to expose.

gpeng: We've (@thomasleese @bevanloon and me) discussed this offline. It's not ideal but as far as we can see we'll probably end up renaming `gov_delivery_id` but continuing to use a non-DB identifier for this functionality. In that eventuality this won't be too difficult to change and the URL pattern will stay the same, only the parameter name will change internally. (It's already referred to as `topic_id` on the frontend anyway so we've got renaming to do there too in any case). So... we're going to press ahead with this for now.

PR #279, opened 2017-12-06
Add an initial pseudo load test

thomasleese:
This will trigger 83,333 delivery request workers and time how long it takes to do so. We can then monitor the sidekiq queue to check how long it takes to finish all the jobs.

[Trello Card](https://trello.com/c/HBHdooWY/425-psuedo-load-test-from-deliveryrequestworker)

### DIFF
--- a/lib/load_tester.rb
+++ b/lib/load_tester.rb
@@ -1,0 +1,28 @@
+class LoadTester
+  def self.test_delivery_request_workers(number)
+    new.test_delivery_request_workers(number)
+  end
+
+  def test_delivery_request_workers(number)
+    email = create_test_email(to: success_address(0))
+
+    number.times do
+      DeliveryRequestWorker.perform_async(email.id)
+    end
+  end
+
+private
+
+  def create_test_email(to:)
+    Email.create!(
+      address: to,
+      subject: "Subject",
+      body: "Body",
+    )
+  end
+
+  def success_address(n)
+    tag = n.to_s.rjust(8, "0")
+    "success+#{tag}@simulator.amazonses.com"
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/load_tester.rb
```diff
@@ -0,0 +1,21 @@
+class LoadTester
+  def trigger_delivery_request_workers(number)
+    email = create_test_email(to: success_address(0))
```
kevindew: should `success_address` be passed number?

thomasleese: Yeah, I might have been thinking too much for the future here, but `success_address` returns a tagged email address with the argument passed in used as the tag.

kevindew: I'm more just questioning whether you meant to hardcode that always return "success+0@simulator.amazonses.com

thomasleese: Oh, yeah, I figured for this test since it was fine since we only need to send to one address to get the load we need. For future tests, we'll need to have more variations.

kevindew: Oh yeah I understand more now, my brain was interpreting this as being within a loop.

thomasleese: I couldn't decide whether to create an email for each delivery request worker, I figured in the end that we're only testing the worker, and that would be testing in the next level up.

lib/tasks/load_tests.rake
```diff
@@ -0,0 +1,13 @@
+require "benchmark"
+require "load_tester"
+
+namespace :load_tests do
+  desc "Run a load test of the DeliveryRequestWorker by triggering 83,333 requests"
+  task :delivery_request_worker, [] => :environment do |_t, _args|
+    results = Benchmark.measure do
+      LoadTester.new.trigger_delivery_request_workers(83333)
```
kevindew: I'd wonder if this would be slightly more idiomatic as a class method rather than doing new then calling it.

Also missed a great chance to use `83_333`

PR #282, opened 2017-12-06
Return 404 if Subscribable not found

gpeng:
As we are retrieving `SubscriberList` using `find_by` it can be `nil` when not found. This commit returns `nil` and 404 in this scenario rather than letting a 500 happen when we try to get the attributes.

Part of [Trello](https://trello.com/c/YtmHgU3b/408-add-page-to-email-alert-frontend-to-capture-a-users-email-address)


gpeng:
Thanks @tuzz I've put a card in to return appropriate JSON errors and we can revisit this.

PR #284, opened 2017-12-06
Add a feature flag to use email-alert-frontend for email collection

tuzz:
https://trello.com/c/bwRGOWZQ/409-add-feature-to-apps-that-serve-email-signup-pages-to-redirect-to-email-alert-frontend-for-email-address-collection

We serve the subscription_url as a field in the
JSON response when clients call
find_or_create_subscriber_list. That means we can
toggle the subscription_url in email-alert-api to
affect all apps.

If we need more granular control, we can switch
over a few document_types at a time.

Here's where we use `subscription_url` in each client app:

- [collections](https://github.com/alphagov/collections/blob/60fcda67508170fe51b8227139fe217bfa47ca65/app/models/email_signup.rb#L16)
- [finder-frontend](https://github.com/alphagov/finder-frontend/blob/b0fd3ce2aa6405dc9ffd50912b832e471cc7cb37/lib/email_alert_signup_api.rb#L14)
- [whitehall](https://github.com/alphagov/whitehall/blob/a53f622a3e8fb5aa296bc532ae0d48a29e614a38/app/models/email_signup.rb#L41)

### DIFF
--- a/app/models/subscriber_list.rb
+++ b/app/models/subscriber_list.rb
@@ -27,7 +27,11 @@ class SubscriberList < ApplicationRecord
   end
 
   def subscription_url
-    PublicUrlService.deprecated_subscription_url(gov_delivery_id: gov_delivery_id)
+    if use_email_alert_frontend_for_email_collection?
+      PublicUrlService.subscription_url(gov_delivery_id: gov_delivery_id)
+    else
+      PublicUrlService.deprecated_subscription_url(gov_delivery_id: gov_delivery_id)
+    end
   end
 
   def to_json
@@ -50,5 +54,10 @@ private
 
   def gov_delivery_config
     EmailAlertAPI.config.gov_delivery
+  end
+
+  # We could make this more sophisticated if needed, e.g. check document_type.
+  def use_email_alert_frontend_for_email_collection?
+    ENV.include?("USE_EMAIL_ALERT_FRONTEND_FOR_EMAIL_COLLECTION")
   end
 end
--- a/app/services/public_url_service.rb
+++ b/app/services/public_url_service.rb
@@ -4,6 +4,17 @@ module PublicUrlService
       "#{website_root}#{base_path}"
     end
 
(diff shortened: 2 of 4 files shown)


app/models/subscriber_list.rb
```diff
@@ -51,4 +55,9 @@ def link_values_are_valid
   def gov_delivery_config
     EmailAlertAPI.config.gov_delivery
   end
+
+  # We could make this more sophisticated if needed, e.g. check document_type.
+  def use_email_alert_frontend_for_email_collection?
+    ENV["USE_EMAIL_ALERT_FRONTEND_FOR_EMAIL_COLLECTION"] == "true"
```
kevindew: I wonder if we just want to check this env var exists? or is there prior art we can apply about boolean env vars

I have a memory but couldn't find anything very quickly that when we have a boolean env var it set in hieradata it comes out as a 1. This may feel slightly weird if we have to set it in hieradata intentionally wrapped as a string.

I could be wrong though :-/

Rest looks good 👍

thomasleese: I agree, I think we should just check for the existence of this environment variable. It means we can leave it as `undef` in Puppet until we want it set.

tuzz: Ok, sounds good to me. Updated. 👍

app/models/subscriber_list.rb
```diff
@@ -51,4 +55,9 @@ def link_values_are_valid
   def gov_delivery_config
     EmailAlertAPI.config.gov_delivery
   end
+
+  # We could make this more sophisticated if needed, e.g. check document_type.
+  def use_email_alert_frontend_for_email_collection?
+    ENV["USE_EMAIL_ALERT_FRONTEND_FOR_EMAIL_COLLECTION"]
```
thomasleese: I would do `ENV.include?` so we always get a boolean here, but I don't think it matters.

tuzz: @thomasleese: updated

PR #290, opened 2017-12-07
Don't raise a 500 server error when subscribing with no email address

tuzz:
https://trello.com/c/amAa5HkV/444-handle-error-cases-for-capturing-a-users-email-address

Previously, the `find_or_create_by` would fail because the address is
invalid but still instantiate a Subscriber object which later causes
errors in the controller. This change fails early and responds with a
422 error when this happens.

This makes use of Rails default error handling as per
config.action_dispatch.rescue_responses

### DIFF
--- a/spec/controllers/subscriptions_controller_spec.rb
+++ b/spec/controllers/subscriptions_controller_spec.rb
@@ -91,4 +91,16 @@ RSpec.describe SubscriptionsController, type: :controller do
       end
     end
   end
+
+  context "with an invalid email address" do
+    let(:subscribable) { create(:subscriber_list) }
+
+    def create_subscription
+      post :create, params: { subscribable_id: subscribable.id, address: "invalid" }, format: :json
+    end
+
+    it "raises an error" do
+      expect { create_subscription }.to raise_error(ActiveRecord::RecordInvalid)
+    end
+  end
 end
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,4 +1,4 @@
-class HealthcheckController < ActionController::Base
+class HealthcheckController < ApplicationController
   def check
     render json: healthcheck.details.merge(status: healthcheck.status)
   end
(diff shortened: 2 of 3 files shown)


spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -91,4 +91,28 @@ def create_subscription
...
+    it "returns 422, unprocessable entity" do
+      create_subscription
+      expect(response.status).to eq(422)
+    end
+
+    it "returns a useful error message" do
+      create_subscription
+
+      body = JSON.parse(response.body)
+      expect(body).to eq("error" => "record invalid")
```
thomasleese: You could use the `data` let variable here to avoid having to do the parsing.

app/controllers/application_controller.rb
```diff
@@ -3,4 +3,10 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :null_session
+
+  rescue_from ActiveRecord::RecordInvalid, with: :unprocessable_entity
+
+  def unprocessable_entity
+    render json: { error: "record invalid" }, status: 422
+  end
```
kevindew: Are you sure you need this? I wonder if rails catches this for us automatically - though even if it does I wonder if the route needs to be told it's a json format.

kevindew: I hope this is useful as it took me forever to look up (as it does every single time I try to look it up): if you check out: config.action_dispatch.rescue_responses on http://guides.rubyonrails.org/configuring.html and this section is interesting too: config.debug_exception_response_format

tuzz: Good spot, Rails does automatically catch `RecordInvalid` and respond 422, according to `rescue_responses` in that ^ documentation. I've removed the handling from this PR.

`debug_exception_response_format` on changes the format in development mode, which probably isn't what we want. Maybe we can just leave the default response alone for now as nothing's actually using it.

PR #291, opened 2017-12-07
Implement email generation worker as per design

thomasleese:
The email generation worker we've got at the moment does not match the design we had of it running as a regular task and handling many subscription contents in one go. By making it work like this design we should improve performance as the query can act on many subscription contents per worker rather than currently by having one worker for one query.

This enables us to achieve the 83,333 in 5 minutes goal for this worker.

### DIFF
--- a/app/workers/email_generation_worker.rb
+++ b/app/workers/email_generation_worker.rb
@@ -1,3 +1,5 @@
+require "sidekiq-scheduler"
+
 class EmailGenerationWorker
   include Sidekiq::Worker
 
--- a/app/workers/email_generation_worker.rb
+++ b/app/workers/email_generation_worker.rb
@@ -1,21 +1,37 @@
 class EmailGenerationWorker
   include Sidekiq::Worker
 
-  def perform(subscription_content_id)
-    subscription_content = SubscriptionContent.find(subscription_content_id)
+  LOCK_NAME = "email_generation_worker".freeze
 
-    email = Email.create_from_params!(email_params(subscription_content))
+  def perform
+    SubscriptionContent.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
+      subscription_contents.find_each do |subscription_content|
+        email = create_email(subscription_content: subscription_content)
+        subscription_content.email = email
+        subscription_content.save
 
-    subscription_content.update!(email: email)
-
-    DeliveryRequestWorker.perform_async_with_priority(
-      email.id, priority: subscription_content.content_change.priority.to_sym,
-    )
+        DeliveryRequestWorker.perform_async_with_priority(
+          email.id, priority: subscription_content.content_change.priority.to_sym
+        )
+      end
+    end
   end
 
 private
 
-  def email_params(subscription_content)
+  def subscription_contents
+    SubscriptionContent
+      .joins(:content_change, subscription: { subscriber: { subscriptions: :subscriber_list } })
+      .includes(:content_change, subscription: { subscriber: { subscriptions: :subscriber_list } })
+      .where.not(subscribers: { address: nil })
+      .where(email: nil)
+  end
+
+  def create_email(subscription_content:)
+    Email.create_from_params!(email_params(subscription_content: subscription_content))
+  end
+
+  def email_params(subscription_content:)
     content_change = subscription_content.content_change
 
     {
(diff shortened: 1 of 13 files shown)


rubenarakelyan:
As discussed, let's change the way this worker works to be on a (short) timer rather than one worker per email - this should hopefully increase performance along with the query changes.

app/workers/email_generation_worker.rb
```diff
@@ -1,24 +1,38 @@
...
+    end
   end
 
 private
 
-  def email_params(subscription_content)
+  def subscription_contents
+    SubscriptionContent
+      .includes(:content_change)
+      .includes(subscription: :subscriber)
```
tuzz: When a user unsubscribes we delete a `subscription`.

That could mean `subscription_content.subscription` and `.subscriber` are `nil`, which bits of code in this worker might assume are present.

Maybe we should add a test for this case and perhaps change this `includes` to `join` so that records are filtered out that no longer have associated `subscriptions`?

app/workers/email_generation_worker.rb
```diff
@@ -1,24 +1,38 @@
...
-    email = Email.create_from_params!(email_params(subscription_content))
+  def perform
+    while subscription_contents.exists?
+      to_queue = []
 
-    subscription_content.update!(email: email)
+      SubscriptionContent.transaction do
+        subscription_contents.lock.limit(100).find_each do |subscription_content|
+          email = create_email(subscription_content: subscription_content)
+          subscription_content.update!(email: email)
```
tuzz: My understanding is we need to handle this race condition:

- Two of these jobs run at the same time
- Each job creates an `email` for a single `subscription_content`
- The last job 'wins' the race and sets the `subscription_content.email_id` field
- The other email is 'orphaned' and in an inconsistent state

I think your approach of pessimistically locking the table works, but maybe there's a way to do this optimistically:

- Add a unique index to the `subscription_content.email_id` field in the database
- Create the `email` and associate the `subscription_content` with it in a transaction
- Rescue from this operation if we get a `UniqueViolation` and return from the job

This way, we assume there aren't two processes creating/updating the database at the same time and if there are, the first process makes the change and the second returns from the job, doing nothing.

However, the overhead of transactionalising each create/update might add some overhead. We should measure this.

app/workers/email_generation_worker.rb
```diff
@@ -1,24 +1,38 @@
...
+          to_queue << [email.id, subscription_content.content_change.priority.to_sym]
+        end
+      end
 
-    DeliveryRequestWorker.perform_async_with_priority(
-      email.id, priority: subscription_content.content_change.priority.to_sym,
-    )
+      to_queue.each do |email_id, priority|
+        DeliveryRequestWorker.perform_async_with_priority(email_id, priority: priority)
+      end
```
tuzz: Why do we enqueue at the end of the job instead of during?

thomasleese: Because we’re in a transaction here, the emails may not have been committed to the database yet so if we were to trigger the workers before that they wouldn’t be able to find the email.

tuzz: Ahh right. That makes sense.

app/workers/email_generation_worker.rb
thomasleese: I like that solution and I think it should work - we may have to put in a `return if subscription_content.email` somewhere to avoid sending out duplicate emails. I'll leave this discussion open to see what other people think for the moment.

app/workers/email_generation_worker.rb
```diff
@@ -1,21 +1,43 @@
...
 class EmailGenerationWorker
   include Sidekiq::Worker
 
-  def perform(subscription_content_id)
-    subscription_content = SubscriptionContent.find(subscription_content_id)
+  sidekiq_options unique: :until_and_while_executing
 
-    email = Email.create_from_params!(email_params(subscription_content))
+  def perform
+    while subscription_contents.exists?
```
bevanloon: If the plan is still to change over to running this on a timer rather than it being triggered by the SubscriptionContentWorker, I wonder whether we need the `while` here - we could just process everything where `email: nil` and not worry too much?

app/workers/email_generation_worker.rb
tuzz: We could set `success=true` after the `#update!` line and exit the job early if this isn't set, maybe?

thomasleese:
Superseded by #303.

PR #293, opened 2017-12-08
Store priority on content change

thomasleese:
For the new email generation worker, we're not going to be passing in the subscription content or the priority, so we need to get the priority from the initial content change.

Depends on #292.

### DIFF
--- a/db/migrate/20171208081924_add_priority_to_change_content.rb
+++ b/db/migrate/20171208081924_add_priority_to_change_content.rb
@@ -1,0 +1,5 @@
+class AddPriorityToChangeContent < ActiveRecord::Migration[5.1]
+  def change
+    add_column :content_changes, :priority, :integer, default: 0
+  end
+end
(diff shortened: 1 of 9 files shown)


db/migrate/20171208081924_add_priority_to_change_content.rb
```diff
@@ -0,0 +1,5 @@
+class AddPriorityToChangeContent < ActiveRecord::Migration[5.1]
+  def change
+    add_column :content_changes, :priority, :integer, default: 0
+  end
+end
```
tuzz: Maybe we should add an index, too? It seems quite likely we'd want to query by `high` priority content changes only.

thomasleese: I agree we're likely to need an index here, but I wanted to avoid making any "bigger" decisions regarding priorities as we haven't fully established how they're going to work yet.

tuzz: Ok, seems reasonable. I think there was talk of doing a larger review of which fields need indices, so we can leave this for now.

PR #297, opened 2017-12-08
[DO NOT MERGE] Add a lock_version field to emails

thomasleese:
This will be used as part of optimistic locking in the new email generation worker.


thomasleese:
Should be part of #291 .

PR #301, opened 2017-12-11
Rework rate limiting

gpeng:
We need to rate limit our API calls to Notify but our current implementation is inefficient and assumes all workers are calling notify which is not necessarily the case.

This PR switches to a Redis backed implementation using the `ratelimit` gem.

I've included some unit tests for the rate limiting 'workings' but nothing that actually tests it from a higher level as it has proved a bit difficult to do thus far. Suggestions gratefully received.

I have tested it manually on integration both with low thresholds and at higher volumes with the production thresholds and all appears to be working well as far as things appear within the sidekiq monitoring app.

[Trello](https://trello.com/c/GdRc6cNa/446-revisit-rate-limiting-within-the-deliveryrequestworker)


tijmenb:
Just to be sure, have Sidekiq plugins like https://github.com/enova/sidekiq-rate-limiter been considered?

gpeng:
@tijmenb Yes. We did look at that. It doesn't support Sidekiq 5 currently I don't think and interferes with other stuff as it uses it's own fetch class. I think it has a per worker rate as well and we need it to be global. On top of all that, it looks pretty abandoned too.

app/workers/delivery_request_worker.rb
```diff
@@ -13,15 +15,70 @@ def self.queue_for_priority(priority)
...
+    rate_limiter.add("delivery_request")
+  end
+
+  def reschedule_job
+    GovukStatsd.increment("delivery_request_worker.rescheduled")
+    DeliveryRequestWorker.perform_in_with_priority(30.seconds, email.id, priority)
+  end
+
+  def rate_limit_threshold
+    per_minute_to_allow_300_per_second = 18000
```
thomasleese: I think we should aim for 360 req per second.

gpeng: Updated in 14d015d

gpeng:
rebased from master...

PR #303, opened 2017-12-12
Email Generation Service

thomasleese:
This replaces the email generation worker with an email generation service which will be run regularly using a rake task rather than a scheduled sidekiq job.

The service works by obtaining a Postgres advisory lock, querying on all the subscription contents and building emails and triggering delivery request workers for each of them and then releasing the lock.


app/services/email_generation_service.rb
```diff
@@ -0,0 +1,56 @@
...
+  def self.call(*args)
+    new.call(*args)
+  end
+
+  def call
+    SubscriptionContent.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
+      subscription_contents.find_in_batches(batch_size: 500) do |group|
+        to_queue = []
+
+        SubscriptionContent.transaction do
```
gpeng: Is this the right scope for this transaction? Should it be inside the `group.each` so as each email creation and subscription update is in its own?

lib/load_tester.rb
```diff
@@ -1,4 +1,8 @@
 class LoadTester
+  def initialize
```
gpeng: Do we need this (or should `enable_logging` be uncommented?)

app/services/email_generation_service.rb
thomasleese: Yeah, I'm using a transaction here for performance rather than consistency. If we process 500 records in one transaction it's faster than processing each one in its own transaction.

lib/load_tester.rb
thomasleese: It's useful to uncomment while running load tests to see where performance could be increased, that's the only reason I've left it there as we have no better way of passing in arguments to the load tester at the moment.

app/services/email_generation_service.rb
gpeng: ok :+1:

lib/load_tester.rb
gpeng: :+1:

thomasleese:
@gpeng I've made a few more optimisations since you approved, would you be able to have another look.

thomasleese:
Since adding the last few optimisations, I'm able to get the rake task to run just under 5 minutes for 83,333 records! 🎉

PR #304, opened 2017-12-12
Improve subscription content and email generation performance

thomasleese:
This improves the performance of the subscription content matcher and the email generation worker.

I was able to get the subscription matcher to run in 50 seconds for 83,333 matches on my dev machine which is not impressively fast, but it's also not the bottle neck in our system so speeding it up won't dramatically improve the situation.

I was able to get the email generation worker to run in 148 seconds for 83,333 subscription contents which means it's also no longer the bottleneck.

### DIFF
--- a/app/services/email_generation_service.rb
+++ b/app/services/email_generation_service.rb
@@ -6,7 +6,7 @@ class EmailGenerationService
   end
 
   def call
-    SubscriptionContent.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
+    ensure_only_running_once do
       subscription_contents.find_in_batches do |group|
         to_queue = []
 
@@ -31,6 +31,12 @@ class EmailGenerationService
   end
 
 private
+
+  def ensure_only_running_once
+    SubscriptionContent.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
+      yield
+    end
+  end
 
   def queue_delivery_request_workers(queue)
     queue.each do |email_id, priority|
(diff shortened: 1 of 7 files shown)


app/services/email_generation_service.rb
```diff
@@ -7,41 +7,39 @@ def self.call(*args)
 
   def call
     SubscriptionContent.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
```
tuzz: Minor suggestion: it might be nice to have a method name to make it clearer what this does, e.g.

```ruby
only_run_this_code_if_no_other_process_is_running_this_code do
  # ...
end
```

(hopefully with a better name!)

What do you think?

db/migrate/20171212130557_add_indexes_to_subscriber_list.rb
```diff
@@ -0,0 +1,9 @@
+class AddIndexesToSubscriberList < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
```
tuzz: Why are we disabling transactionality? If the migration fails that would put us in a weird state.

thomasleese: It's so we can add the indexes concurrently without locking the table. If the migration fails, we can just run it again.

app/services/email_generation_service.rb
thomasleese: `ensure_only_running_once` ?

PR #305, opened 2017-12-12
Nullify subscription_contents.subscription_id when unsubscribing

tuzz:
This issue was discovered while working on:
https://trello.com/c/waCjUj3J/437-add-high-level-integration-tests-to-email-alert-api

(discussed with @thomasleese and @rubenarakelyan)

I decided to break it into a separate PR. Currently there's a unit test for this, but soon there will also be a feature test once the other PR associated with that Trello ticket is merged.

**Explanation:**

We delete subscription records when users
unsubscribe. This means we need to make this field
nullable on the subscription_contents table.

We considered also deleting the
subscription_contents associated with the deleted
subscription, but decided not to because the
subscription_content still points to the
content_change which is useful information.

### DIFF
--- a/app/services/unsubscribe_service.rb
+++ b/app/services/unsubscribe_service.rb
@@ -12,12 +12,20 @@ module UnsubscribeService
 
     def unsubscribe!(subscriber, subscriptions)
       ActiveRecord::Base.transaction do
+        nullify_references_to_subscriptions!(subscriptions)
+
         subscriptions.each(&:destroy)
 
         if no_other_subscriptions?(subscriber, subscriptions)
           subscriber.nullify_address!
         end
       end
+    end
+
+    def nullify_references_to_subscriptions!(subscriptions)
+      SubscriptionContent
+        .where(subscription: subscriptions)
+        .update_all(subscription_id: nil)
     end
 
     def no_other_subscriptions?(subscriber, subscriptions)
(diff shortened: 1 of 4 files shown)


kevindew:
Just a small suggestion that may not be relevant. But could we achieve this with less code with we had a on_delete nil cascade? I guess that depends on how restrictive we want to make it for deleting subscription records

tuzz:
I thought about this, but would prefer to not to. My reasoning is:

In the [domain model](https://github.com/alphagov/email-alert-api/blob/master/doc/domain.png), `SubscriptionContent` has visibility of `Subscription`, in that it knows what a `Subscription` is and is aware of that part of the domain, but `Subscription` does not have visibility of `SubscriptionContent`.

In my mind, a `Subscription` is a standalone concept that can exist irrespective of how we've decided to model `SubscriptionContent` and what that part of the system does. I intentionally left out visibility from `Subscription` to `SubscriptionContent` so we're quite reservist in which bits of the system we give knowledge of to other bits of the system. I see this as a kind of coupling.

If you consider at the system in a holistic way, the concept of `SubscriptionContent` is an 'extension' that provides the features and functionality for the construction of emails from content changes but, in my opinion, it is desirable to be able to extend the system with new concepts without having to re-open existing models/classes (e.g. to add associations). This is loosely described in the [open/closed principle](https://en.wikipedia.org/wiki/Open/closed_principle), though I'm applying some interpretations of Domain Driven Design to this definition.

This isn't so clear-cut in Rails because it mixes its domain and persistence layer, so I'd be happy to discuss further if you have a strong preference for adding the cascade.

tuzz:
Going to merge because I want to rebase on top of this.

PR #307, opened 2017-12-12
Send reference to Notify (+ lots of refactoring)

tuzz:
https://trello.com/c/onBqZonX/447-send-reference-to-notify-when-sending-emails

As I was working on this I decided to do some refactoring.

I'd strongly recommend reviewing this commit by commit.


app/providers/notify_provider.rb
```diff
@@ -0,0 +1,32 @@
...
+  def initialize(config: EmailAlertAPI.config.notify)
+    api_key = config.fetch(:api_key)
+    base_url = config.fetch(:base_url)
+
+    self.client = Notifications::Client.new(api_key, base_url)
+    self.template_id = config.fetch(:template_id)
+  end
+
+  def call(address:, subject:, body:, reference:)
+    client.send_email(
```
thomasleese: Do we need to pass back the return value of this?

app/services/metric_service.rb
```diff
@@ -0,0 +1,58 @@
+class MetricService
```
thomasleese: I would call this `MetricsService`, but don't think it matters much.

app/services/metric_service.rb
```diff
@@ -0,0 +1,58 @@
...
+  class << self
```
thomasleese: Might be good to get https://github.com/alphagov/email-alert-api/blob/master/app/services/email_sender_service.rb#L12 in here too.

app/services/metric_service.rb
tuzz: renamed

app/services/metric_service.rb
tuzz: added

app/providers/notify_provider.rb
tuzz: We don't currently use the return value of this method call.

Do you think we should return something?

thomasleese: Oh, I guess now that we're using `reference` instead of relying on Notify IDs it doesn't matter.

PR #312, opened 2017-12-14
Re-enable the status update healthchecks

rubenarakelyan:
This PR reverts https://github.com/alphagov/email-alert-api/pull/238 to re-enable the healthchecks since we are now getting status update callbacks from Notify. It also ignores the healthchecks in non-production environments where we don't receive status update callbacks from Notify.

Trello: https://trello.com/c/31XA1mkY/449-integrate-email-alert-api-with-notifys-callback

### DIFF
--- a/db/migrate/20171214113834_delete_test_emails_and_delivery_attempts.rb
+++ b/db/migrate/20171214113834_delete_test_emails_and_delivery_attempts.rb
@@ -1,0 +1,7 @@
+class DeleteTestEmailsAndDeliveryAttempts < ActiveRecord::Migration[5.1]
+  def up
+    # Emails and DeliveryAttempts at this point are all test data
+    Email.destroy_all
+    DeliveryAttempt.destroy_all
+  end
+end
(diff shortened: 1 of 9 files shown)


thomasleese:
We will probably have to clear out the existing delivery attempts otherwise this will be flagged as critical.

thomasleese:
Note that we shouldn't deploy this until we disable these health checks on staging and integration as they won't support the status updates.

db/migrate/20171214113834_delete_test_emails_and_delivery_attempts.rb
```diff
@@ -0,0 +1,7 @@
+class DeleteTestEmailsAndDeliveryAttempts < ActiveRecord::Migration[5.1]
+  def up
+    # Emails and DeliveryAttempts at this point are all test data
+    Email.destroy_all
+    DeliveryAttempt.destroy_all
```
thomasleese: Think we might have to change the order of these statements... https://sentry.io/govuk/app-email-alert-api/issues/419031969/

rubenarakelyan: https://github.com/alphagov/email-alert-api/pull/327

PR #320, opened 2017-12-14
Treat requests as JSON

kevindew:
Trello: https://trello.com/c/hR6rkDdu/454-return-json-errors-from-email-alert-api

This sets all our requests to be treated as a JSON format, which means
we get JSON exceptions.

The config.debug_exception_response_format sets the debug exceptions to
be rendered in JSON rather than HTML.

This would all be slightly neater if we used Rails api_only mode however
GDS SSO is currently incompatible.


kevindew:
Closing in favour of https://github.com/alphagov/email-alert-api/pull/328

PR #326, opened 2017-12-15
Fix lint errors

thomasleese:
This is needed to upgrade govuk-lint as per #316. See the commits for individual details.


spec/units/models/unsubscribe_link_spec.rb
```diff
@@ -9,7 +9,7 @@
   describe "#url" do
     it "returns an unsubscribe url for the subscription" do
       expect(unsubscribe_link.url).to eq(
-        "http://www.dev.gov.uk/email/unsubscribe/1234?title=dave%20crocker%20%26%20friends"
+        "http://www.dev.gov.uk/email/unsubscribe/1234?title=dave+crocker+%26+friends"
```
tuzz: I don't remember the details, but I think we need to escape the `+` symbols otherwise the unsubscribe links are rendered correctly in the email. That's why we used `URI.encode(value, /\W/)`. Maybe @gpeng can remember more?

thomasleese: Ah... I did wonder if that was the case. `+`s are a bit of a weird thing, and I couldn't find a suitable replacement for `URI.encode` which supported the second argument.

gpeng: Yeah... I remembered us having trouble with this but I couldn't remember what the problem was. I thought we were just trying to not have `+` urls. Let's deploy it to integration and check these manually before we merge.

thomasleese: Yep, turns out the pluses get turned into `%2B` so the URLs don't work.

thomasleese: Looks like we can use https://stackoverflow.com/a/39209414 instead.

thomasleese: @gpeng @tuzz Have updated the PR to use `ERB::Util.url_encode`.

PR #331, opened 2017-12-15
Rename FactoryGirl to FactoryBot

kevindew:
This will quieten down these deprecation warnings.

P.S. Feel free to merge this in while I'm on Holiday.

### DIFF
--- a/spec/units/services/unsubscribe_service_spec.rb
+++ b/spec/units/services/unsubscribe_service_spec.rb
@@ -1,6 +1,6 @@
 RSpec.describe UnsubscribeService do
   describe ".subscriber!" do
-    let!(:subscriber) { FactoryBot.create(:subscriber, address: "foo@bar.com") }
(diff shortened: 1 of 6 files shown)


spec/units/services/unsubscribe_service_spec.rb
```diff
@@ -1,6 +1,6 @@
 RSpec.describe UnsubscribeService do
   describe ".subscriber!" do
-    let!(:subscriber) { FactoryGirl.create(:subscriber, address: "foo@bar.com") }
+    let!(:subscriber) { FactoryBot.create(:subscriber, address: "foo@bar.com") }
```
thomasleese: We could just use `create` here too.

kevindew: Dang - I missed that one. I'm very sloppy today! Thanks 👍

PR #341, opened 2017-12-19
Fix an autoload problem for symbolize_json

tuzz:
On my development machine, when running the
email-alert-api server, I was seeing the
healthcheck integration test fail.

The failing spec was caused by a non-zero Sidekiq
retry set size. The failing Sidekiq jobs indicated
a problem loading the symbolize_json file.

I’m not sure why this only happens when the server
is running. It could be to do with Spring, but
moving this file to a directory on the
eager_load_path fixes it.


app/models/subscriber_list.rb
```diff
@@ -1,5 +1,4 @@
 require 'json'
```
thomasleese: Not really related to this PR, but do we need this?

PR #342, opened 2017-12-19
Add logging of missing bearer tokens

gpeng:
All apps should now be sending a bearer token with `internal_app` permission. Prior to enforcing this, this commit adds logging which we can deploy for a while just to ensure we haven't missed any clients.

This is a temporary commit that will be superseded by enforcing the authentication.

[Trello](https://trello.com/c/75sVuXhA/455-require-all-apps-to-authenticate-with-email-alert-api)


thomasleese:
Third time lucky!

PR #343, opened 2017-12-19
Add a rake task to import govdelivery subscriptions

tuzz:
https://trello.com/c/CAktH8HT/467-create-tooling-to-import-govdelivery-data-to-email-alert-api

To run the rake task:
`CSV_PATH=foo.csv bundle exec rake import_govdelivery_csv`

This produces output in the form:
```
..F
Successful rows: 2
Failed rows: 1
 - ["Couldn't find SubscriberList", {"TOPIC_ID"=>"67890", "TOPIC_NAME"=>"Second", "TOPIC_CODE"=>"UKGOVUK_222", "TOPIC_VISIBILITY"=>"1", "SUBSCRIBER_ID"=>"1111111111", "DESTINATION"=>"foo@example.com"}]
```

The data needs to be written to an unencrypted CSV
before importing.


tuzz:
Ok, going to merge.

Quite a lot of rows fail to import because of a missing `SubscriberList`. These look to be for lists that no longer exist in `email-alert-api` but are still in govdelivery. I think this is fine.

I'd recommend backing up the CSV we import so we can refer back to it in the future if we need any of the subscriptions for these lists.

PR #345, opened 2017-12-21
Add authorisation and authentication

gpeng:
This PR adds a requirement for `internal_app` permission to all endpoints except status_updates, which requires the aptly entitled `status_updates` permission and healthcheck which doesn't require authentication.

The diffs are a bit messy but there aren't actually that many significant changes in each of the files. In the integration specs I've added a 'with authentication and authorisation' context that logs in with the necessary permissions and then 'without authentication' and 'without authorisation' contexts that just check for a 403.

I've also added login steps to the functional tests but have not tested the negative cases there. Some of those logins are a bit contrived in that they needed `status_updates` and `internal_app` permissions which doesn't happen in reality currently.

We've been logging permissions in production today and all of the following apps are correctly configured for these changes.

* travel advice publisher
* collections
* email-alert-frontend
* specialist-publisher
* whitehall
* finder-frontend

There are three other 'clients' making requests. The notify callback which has the status_updates permission and two things calling `/healthcheck` without a bearer token which is fine.

[Trello](https://trello.com/c/75sVuXhA/455-require-all-apps-to-authenticate-with-email-alert-api)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,35 +1,11 @@
 class ApplicationController < ActionController::API
   include GDS::SSO::ControllerMethods
 
-  before_action :log_missing_credentials
+  before_action :authorise
 
 private
 
-  def log_missing_credentials
-    auth_header = request.env["HTTP_AUTHORIZATION"]
-    if auth_header.present?
-      authenticate_user!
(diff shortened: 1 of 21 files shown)


gpeng:
Thanks @thomasleese I've updated it to do that in 8d38382

PR #358, opened 2018-01-16
Clean Gemfile and bundle update

kevindew:
Saves us a few dependabot PR's.


Gemfile.lock
```diff
@@ -135,15 +135,15 @@ GEM
...
-    multi_json (1.12.2)
+    multi_json (1.13.1)
     multi_xml (0.6.0)
     multipart-post (2.0.0)
     netrc (0.11.0)
-    nio4r (2.1.0)
-    nokogiri (1.7.2)
-      mini_portile2 (~> 2.1.0)
+    nio4r (2.2.0)
+    nokogiri (1.8.1)
```
thomasleese: We can't upgrade Nokogiri because this version has a bug.

https://github.com/alphagov/email-alert-api/pull/340

PR #359, opened 2018-01-16
Associate a user with authenticated actions

kevindew:
Trello: https://trello.com/c/HqLhaWvN/476-for-models-that-are-created-as-a-result-of-a-request-in-email-alert-api-store-the-user-id-so-we-know-which-user-created-things

or auditing purposes this will store the user id of the user that has
performed an action which has altered the state of the application.

Since we have authentication we have the ability to know who has
performed an action, but this was previously not stored. Storing this
should help if we want to work out analytics based on a user service.

For the most part though this data should be pretty dull with mostly the
same accounts using the API.

### DIFF
--- a/db/migrate/20180116163006_store_signon_user_ids.rb
+++ b/db/migrate/20180116163006_store_signon_user_ids.rb
@@ -1,0 +1,9 @@
+class StoreSignonUserIds < ActiveRecord::Migration[5.1]
+  def change
+    add_column :content_changes,    :signon_user_uid, :string
+    add_column :subscriber_lists,   :signon_user_uid, :string
+    add_column :subscriptions,      :signon_user_uid, :string
+    add_column :subscribers,        :signon_user_uid, :string
+    add_column :delivery_attempts,  :signon_user_uid, :string
+  end
+end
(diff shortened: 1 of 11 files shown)


db/migrate/20180116163006_store_signon_user_ids.rb
```diff
@@ -0,0 +1,9 @@
+class StoreSignonUserIds < ActiveRecord::Migration[5.1]
+  def change
+    add_column :content_changes,    :signon_user_uid, :string
```
thomasleese: I wonder if we should take the opportunity now to clear these tables and add this column in as `null: true`?

kevindew: Sorry @thomasleese I missed that you commented days ago. I'm assuming typo'd `null: false`also.

Seems a reasonable step to require it to not be null. I feel like adding the clearing out of tables into this PR overloads things a bit though. Are we going to be planning to do a wipe pre-launch? we could write a story for that nowish and flag setting these as not null to it.

kevindew:
@gpeng yup, if everything goes to plan the data in here will be very dull. It only really becomes of any interest if something out of the ordinary happens and we need to work out why.

db/migrate/20180116163006_store_signon_user_ids.rb
thomasleese: Yeah, okay, if we add a card to do that later then that's fine.

PR #361, opened 2018-01-18
Add frequency to subscriptions

thomasleese:
This will be used to set whether a subscription is a daily or weekly digest, or if the emails should be sent immediately.

[Trello Card](https://trello.com/c/xNqETkOc/349-add-type-information-to-subscription)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171219104253) do
+ActiveRecord::Schema.define(version: 20180118085957) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -113,6 +113,7 @@ ActiveRecord::Schema.define(version: 20171219104253) do
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.uuid "uuid", null: false
+    t.integer "frequency", default: 0, null: false
     t.index ["subscriber_id", "subscriber_list_id"], name: "index_subscriptions_on_subscriber_id_and_subscriber_list_id", unique: true
     t.index ["subscriber_id"], name: "index_subscriptions_on_subscriber_id"
     t.index ["subscriber_list_id"], name: "index_subscriptions_on_subscriber_list_id"
--- a/app/models/content_change.rb
+++ b/app/models/content_change.rb
@@ -1,7 +1,7 @@
 class ContentChange < ApplicationRecord
   include SymbolizeJSON
 
-  enum priority: %i(low high)
+  enum priority: { low: 0, high: 1 }
 
   def mark_processed!
     update!(processed_at: Time.now)
--- a/app/models/delivery_attempt.rb
+++ b/app/models/delivery_attempt.rb
@@ -3,8 +3,8 @@ class DeliveryAttempt < ApplicationRecord
 
   validates :email, :status, :provider, presence: true
 
-  enum status: %i(sending delivered permanent_failure temporary_failure technical_failure)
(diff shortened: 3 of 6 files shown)


thomasleese:
Yeah same 😢

db/schema.rb
```diff
@@ -113,6 +113,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.uuid "uuid", null: false
+    t.integer "frequency", default: 0, null: false
```
bevanloon: Does anyone think it's worth having a table for frequency? I'm just wondering if we risk propagating magic numbers around the data that add a cognitive load to future support work on the app. I know we are highly unlikely to ever change from the three frequencies we have, but we are completely coupling the meaning of this piece of data to our ruby code.

thomasleese: I'm not opposed to this idea, would you imagine the table contains the information about the different types of frequencies (so the code is incredibly generic and can be used to generate the frontend), perhaps with a field like `group_days` which would be set to 1 for daily, 7 for weekly and `null` for immediate, or would it just be a table with names?

thomasleese: (If it's just a table with names, I think it only moves the coupling down a level rather than fixing the problem of coupling, in which case I'm not convinced it's worth it.)

kevindew: It is a shame that we lose the meaning by having integer, but I don't table would be the right way to solve it as that adds different additional load (both technical through joins and cognitive through extra table) and does lead to us fighting the Rails way somewhat.

I imagine we're caught in a state where a postgres enum is ideal fit - but is a total pain for Rails (from my experience), a string would be easier to understand (not sure how well that plays with Rails) and integer is following the Rails docs so least likely to have surprises/bugs.

bevanloon: I was thinking about just representing the current enum in a seeded lookup table - three rows, immediate(ly?), daily, weekly. Potentially a uuid as well for `email-alert-frontend` to use.

I was then picturing querying this table through `gds-api-adapters` in `email-alert-frontend` so we can generate the radio buttons for the form that passes data through to `email-alert-api`. That way we don't have to translate the frequency coming through from the frontend and also if we add or change frequency, the frontend code wouldn't need to change.

thomasleese: I'm not sure if it makes sense for the frontend to be tied to a table here because we may want custom layouts and things - and it might be easier for content designers to be able to change the text in the frontend repo rather than in the database.

Plus if we're going to need to add or change a frequency, we'll end up changing a significant amount of code anyway such that changing the frontend will probably be a very small part of it.

bevanloon: @kevindew sorry, stupid question from a non-Railsy person - how does an extra table fight the Rails way?

Good point about extra load to join to the table, but it will be a very small table. I like having the meaning of the data self contained outside of application code, but maybe that's just me.

bevanloon: Fair point about content designers. I think having a single source of truth for that data is worth the trade off.

Anyway, seems we have more ➕ s for the enums than ➖ s . So go for it.

Oh - only other question I have: is it good practice to pin the enums to their integer values when they are defined a-la `frequency: {immediately: 0, daily: 1, weekly: 2}` (to avoid accidental re-ording/insertion)?

thomasleese: We can probably pin them now since there's no reason not to.

thomasleese: @bevanloon Pinned them in https://github.com/alphagov/email-alert-api/pull/361/commits/46ded4ee02d6fe6c42d1258f36cfe2e851d22d23

kevindew: Pinning sounds good 👍 

@bevanloon I'd suggest it's not Railsy because they provide this enum functionality to fit this scenario so would be somewhat unexpected to use a join instead.

PR #365, opened 2018-01-19
Add digest schedulers

gpeng:
As part of the implementation of digest emails within email-alert-api we need a way of scheduling and scoping their creation. This PR adds a `DailyDigestSchedulerService` which when called creates a `DigestRange` object. This object models the time scope of the digest ie the span within which notifications should be generated for any matching content changes.

To reduce complexity in deployment the service has been implemented to allow multiple calls without initiated duplicate runs of digest emails. This allows it to be called from all host machines so we can schedule it simply with cron.

This class will be expanded in future PRs to enqueue jobs to create the digest emails.

Marked WIP as I'm adding the weekly version and wanted to get some early feedback and not block the next bit but if we're happy with this it can be merged and I'll update in a separate PR.

[Trello](https://trello.com/c/WPFnHqCJ/533-create-digest-schedulers)

### DIFF
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -1,0 +1,37 @@
+class DigestRun < ApplicationRecord
+  validates :starts_at, :ends_at, :date, :range, presence: true
+  before_validation :set_range_dates, on: :create
+  validate :ends_at_is_in_the_past
+
+  enum range: { daily: 0, weekly: 1 }
+
+  DAILY = "daily".freeze
+  WEEKLY = "weekly".freeze
+
+private
+
+  def set_range_dates
+    self.starts_at = configured_starts_at
+    self.ends_at = configured_ends_at
+  end
+
+  def ends_at_is_in_the_past
+    errors.add(:ends_at, "must be in the past") if ends_at >= Time.now
+  end
+
+  def configured_starts_at
+    Time.parse("#{digest_range_hour}:00", starts_at_date)
+  end
+
+  def configured_ends_at
+    Time.parse("#{digest_range_hour}:00", Time.now)
+  end
+
+  def starts_at_date
+    daily? ? 1.day.ago : 1.week.ago
+  end
+
+  def digest_range_hour
+    ENV.fetch("DIGEST_RANGE_HOUR", 8).to_i
+  end
+end
--- a/app/services/daily_digest_scheduler_service.rb
+++ b/app/services/daily_digest_scheduler_service.rb
@@ -1,27 +1,5 @@
 class DailyDigestSchedulerService
+  include DigestSchedulerService
   LOCK_NAME = "daily_digest_scheduler".freeze
-
-  def self.call(*args)
-    new.call(*args)
-  end
-
-  def call
-    run_with_advisory_lock do
-      digest_run = DigestRun.find_or_initialize_by(
-        date: Date.current, range: DigestRun::DAILY
-      )
-      return if digest_run.persisted?
-      digest_run.save!
-    end
-
-    #enqueue the digest creation workers TBC
-  end
-
-private
-
-  def run_with_advisory_lock
-    DigestRun.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
-      yield
-    end
-  end
+  RANGE = DigestRun::DAILY.freeze
 end
--- a/app/services/daily_digest_scheduler_service.rb
+++ b/app/services/daily_digest_scheduler_service.rb
@@ -1,0 +1,27 @@
+class DailyDigestSchedulerService
+  LOCK_NAME = "daily_digest_scheduler".freeze
+
+  def self.call(*args)
+    new.call(*args)
+  end
+
+  def call
+    run_with_advisory_lock do
+      digest_run = DigestRun.find_or_initialize_by(
+        date: Date.current, range: DigestRun::DAILY
+      )
+      return if digest_run.persisted?
+      digest_run.save!
+    end
+
+    #enqueue the digest creation workers TBC
+  end
+
+private
+
+  def run_with_advisory_lock
+    DigestRun.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
+      yield
+    end
+  end
+end
--- a/app/services/digest_scheduler_service.rb
+++ b/app/services/digest_scheduler_service.rb
@@ -1,0 +1,29 @@
+module DigestSchedulerService
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def call(*args)
+      new.call(*args)
+    end
+  end
+
+  def call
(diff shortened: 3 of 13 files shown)


app/services/daily_digest_scheduler_service.rb
```diff
@@ -0,0 +1,47 @@
...
+  attr_accessor :starts_at, :ends_at
+
+  LOCK_NAME = "daily_digest_scheduler".freeze
+
+  def self.call(*args)
+    new.call(*args)
+  end
+
+  def initialize
+    @digest_range_hour = ENV["DIGEST_RANGE_HOUR"] || 8
```
thomasleese: I think you'll have to do `ENV.fetch("DIGEST_RANGE_HOUR", "8").to_i` to make sure it's an integer.

config/schedule.rb
```diff
@@ -0,0 +1,24 @@
...
+#   command "/usr/bin/some_great_command"
+#   runner "MyModel.some_method"
+#   rake "some:great:rake:task"
+# end
+#
+# every 4.days do
+#   runner "AnotherModel.prune_old_records"
+# end
+
+every 1.day, at: "8:30 am" do
```
thomasleese: Is there some way we can get this to retry if the service fails for whatever reason? Or perhaps that doesn't matter because we can run the service again manually if we see an entry in Sentry?

app/services/daily_digest_scheduler_service.rb
gpeng: It gets stringificatorised in the `configured_(starts|ends)_at` methods anyway so it probably doesn't matter much. I've updated it in 46a59e7 as having a predictable type is a noble cause 😄 I've also fixed it so as it's easier on @kevindew's eyes 🤣

app/services/daily_digest_scheduler_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    digest_range = nil
+
+    run_with_advisory_lock do
+      digest_range = DigestRange.find_or_initialize_by(
+        starts_at: starts_at, ends_at: ends_at
+      )
+      return if digest_range.persisted?
+    end
+
+    digest_range.save!
```
bevanloon: Is there any risk of a clash here? The `save` runs outside of the lock, which means there's a point where we've released the lock, but no record is persisted and another box could, in theory execute `find_or_initialize_by` before we `save`? Or is that not how the advisory lock stuff works?

I also vaguely wondered how we feel about a unique constraint at the database level which would add an extra level of security against duplicate rows?

config/schedule.rb
gpeng: Aye... I originally planned to run it in a sidekiq worker so as we would get the retries. I decided not to as:

* it will get called 3 times anyway from each backend box
* I thought we'd probably want to know and intervene for now rather than have retries queueing up in sidekiq
* it takes a bit of load of the queues (and inversely doesn't have to wait if the queue is busy)

All that said, I'm happy to run it in sidekiq if we decide that would be better

app/models/digest_run.rb
```diff
@@ -0,0 +1,37 @@
+class DigestRun < ApplicationRecord
+  validates :starts_at, :ends_at, :date, :range, presence: true
+  before_validation :set_range_dates, on: :create
+  validate :ends_at_is_in_the_past
+
+  enum range: %i(daily weekly)
```
thomasleese: We might to define the enum as `{ daily: 0, weekly: 1 }` here.

app/services/digest_scheduler_service.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  class_methods do
+    def call(*args)
+      new.call(*args)
+    end
+  end
+
+  def call
+    run_with_advisory_lock do
+      digest_run = DigestRun.find_or_initialize_by(
```
thomasleese: I don't know if this is related to this PR, but if someone were to accidentally (or intentionally to deal with an incident?) call the `WeeklyDigestSchedulerService` one day after it had ran, it would end up creating a new `DigestRun` starting one day after rather than finding the existing one suitable for that week. I don't know if this is a concern of this class, or something more higher up?

db/migrate/20180118134516_create_digest_runs.rb
```diff
@@ -0,0 +1,12 @@
+class CreateDigestRuns < ActiveRecord::Migration[5.1]
+  def change
+    create_table :digest_runs do |t|
+      t.date :date, null: false
+      t.datetime :starts_at, null: false
+      t.datetime :ends_at, null: false
```
thomasleese: I wonder if it's better to use a word like `until` here to suggest that it's a `<` not a `<=` relation.

app/models/digest_run.rb
gpeng: 0 and 1 would be the default values assigned anyway wouldn't they? I noticed you defined them explicitly in a PR the other day. Any other reason? I've only done that before where you need values that don't correspond to the array index.

app/services/digest_scheduler_service.rb
gpeng: I think we were talking about doing a spike into what we would do in those instances so I've just made it as simple as possible here and not worried about it.

db/migrate/20180118134516_create_digest_runs.rb
gpeng: I'm easy either way... I'll change unless anyone else has any opinions.

app/models/digest_run.rb
bevanloon: I always saw it as a measure to stop someone accidentally adding a new one in between and breaking the meaning of the data in the database.

gpeng: Fair enough. Fixed in 737ccd1

db/migrate/20180118134516_create_digest_runs.rb
gpeng: Actually... no I won't. `until` is a reserved word (as was `end` when I originally named them `start` and `end` 🤕 )

thomasleese: Hmm, oh yeah... I'll leave it as it is, we can see if it's confusing.

spec/units/models/digest_run_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      expect {
+        described_class.create(
+          attributes_for(:digest_run)
+        )
+      }.to change { DigestRun.count }.from(0).to(1)
+    end
+
+    context "daily" do
+      context "with no environment vars set" do
+        context "daily" do
```
bevanloon: do we need both `daily`s here?

gpeng: oops... Fixed in 269464d

PR #368, opened 2018-01-23
ADR-002 Digests

kevindew:
To view: https://github.com/alphagov/email-alert-api/blob/133fe46731c7e595ba2b2f0e4fdacd7384f04ad4/doc/arch/adr-002-digest-mvp.md


doc/arch/adr-002-digest-mvp.md
```diff
@@ -0,0 +1,187 @@
...
+  - The end time for the digest period
+
+##### Should
+
+- Store a timestamp for when all the emails were created for the digest, to act
+  as indicator it is complete.
+
+##### May
+
+- Have a unique index for digest date and period, however this would limit
```
thomasleese: I imagined a re-run could re-use the record since I imagine the `completed_at` wouldn't be set if it was being re-run.

doc/arch/adr-002-digest-mvp.md
```diff
@@ -0,0 +1,187 @@
...
+- Mark a `DigestRunSubscriber` as completed
+- Mark `DigestRun` as completed if all `DigestRunSubscriber` are complete for
+  the `DigestRun`
+
+##### May
+
+- Liaise with a `DigestGeneratorService` to do the work if the code for the
+  worker becomes complicated or should be more re-usable
+
+#### `ContentChangeQuery`
```
thomasleese: Not sure if this matters, I called this `ContentChangeMatcher` in my WIP implementation as it nicely fits with the `SubscriptionMatcher` class that already exists.

doc/arch/adr-002-digest-mvp.md
```diff
@@ -0,0 +1,187 @@
...
+
+For a given [`DigestRunSubscriber`](#digestrunsubscriber) this determines which,
```
thomasleese: I thought it might be good to pass in the subscriptions to this class, since they're easy to get from the `DigestRunSubscriber`, and it means the `ContentChangeMatcher` and `SubscriptionMatcher` classes are exact reverses of each other.

doc/arch/adr-002-digest-mvp.md
```diff
@@ -0,0 +1,187 @@
...
+if any, `ContentChange` entities apply and will make up the contents for the
+Email.
+
+##### Should
+
+- Return which subscriptions are associated with a `ContentChange`. It may be
```
thomasleese: Do you mean this would return the `ContentChanges` rather than the subscriptions?

doc/arch/adr-002-digest-mvp.md
gpeng: There isn't currently a unique index on this. Nothing to stop us adding one obvs but just flagging that.

doc/arch/adr-002-digest-mvp.md
kevindew: Sure, makes sense that it matches the naming of that. I reckon both should be suffixed with `Query` though so it's pretty clear which directory they belong in.

doc/arch/adr-002-digest-mvp.md
kevindew: I just had a chat with @gpeng about this sort of stuff and hadn't properly considered your comment.

We were wrestling with how we'd know which subscriptions are associated with a content change. We were toying with:

`ContentChangeMatcher.call(digest_run_subscriber)` which returns something like
```
{ 
  1 => [content_change, content_change],
  2 => [content_change],
}
```
Where 1 and 2 are subscription ids

In theory we could do this differently with more steps in the worker class like
```
subscriptions = digest_run_subscriber.subscriber.subscriptions
grouped_content_changes = subscriptions.map do |s| 
  {
     subscription: s,
     content_changes: ContentChangeMatcher.call(s.subscriber_list)
  }
end
```

But I don't know for sure what structure we want to go into the `DigestEmailRenderer` so will probably leave this with options in this doc.

doc/arch/adr-002-digest-mvp.md
kevindew: I thought we'd want to know both but of course if we adapt it to accept subscription it wouldn't need this.

PR #369, opened 2018-01-23
Store matched content changes

thomasleese:
This PR introduces a new table to store content changes match with their subscriber lists when they come straight into the system. This allows the workers down the line to make use of this "cached" data, and enables the creation of a query that matches subscribers with content changes useful for digests.

[Trello Card](https://trello.com/c/NRw0zkGc/534-content-changes-for-subscription)

### DIFF
--- a/spec/units/services/notification_handler_service_spec.rb
+++ b/spec/units/services/notification_handler_service_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe NotificationHandlerService do
     }
   }
 
-  let(:subscriber_list) do
+  let!(:subscriber_list) do
     create(:subscriber_list, tags: { topics: ["oil-and-gas/licensing"] })
   end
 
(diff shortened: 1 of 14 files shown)


thomasleese:
@gpeng I've extracted the matched content change generation into a separate service class.

Yeah, I'm also expecting it to affect performance in that way. The `SubscriptionMatcher` query was determined to be a fairly expensive query while I was load testing.

spec/units/services/notification_handler_service_spec.rb
```diff
@@ -44,11 +44,15 @@
...
+      expect { described_class.call(params: params) }
+        .to change { MatchedContentChange.count }.by(1)
+    end
+
     it "enqueues the content change to be processed by the subscription content worker" do
-      allow(ContentChange).to receive(:create!).and_return(double(id: 1))
       expect(SubscriptionContentWorker)
         .to receive(:perform_async)
-        .with(1)
+        .with(kind_of(Integer))
```
gpeng: I think there's value in having the explicit value in here to prove we're actually passing the content change id and not something else by mistake. We do have higher level test coverage of this though...

thomasleese: Agreed, I thought I needed this, but I've fixed it now so it's similar to how it was.

PR #371, opened 2018-01-23
Add SubscriptionContentChangeQuery

thomasleese:
The purpose of this class will be to query which content changes
should be associated with a subscriber and a digest run. It will
return the content changes grouped by subscription and subscriber
list.

[Trello Card](https://trello.com/c/NRw0zkGc/534-content-changes-for-subscription)

### DIFF
--- a/app/queries/subscription_content_change_query.rb
+++ b/app/queries/subscription_content_change_query.rb
@@ -1,0 +1,49 @@
+class SubscriptionContentChangeQuery
+  def initialize(subscriber:, digest_run:)
+    @subscriber = subscriber
+    @digest_run = digest_run
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    present
+  end
+
+  private_class_method :new
+
+private
+
+  attr_reader :subscriber, :digest_run
+
+  Result = Struct.new(:subscription_id, :subscription_uuid, :subscriber_list_title, :content_changes)
+
+  def present
+    grouped_content_changes.map do |(subscription_id, subscription_uuid, subscriber_list_title), content_changes|
+      Result.new(
+        subscription_id,
+        subscription_uuid,
+        subscriber_list_title,
+        content_changes,
+      )
+    end
+  end
+
+  def grouped_content_changes
+    content_changes.group_by do |record|
+      [record["subscription_id"], record["subscription_uuid"], record["subscriber_list_title"]]
+    end
+  end
+
+  def content_changes
+    ContentChange
+      .select("content_changes.*", "subscriptions.id AS subscription_id", "subscriber_lists.title AS subscriber_list_title", "subscriptions.uuid AS subscription_uuid")
+      .joins(matched_content_changes: { subscriber_list: { subscriptions: :subscriber } })
+      .where(subscribers: { id: subscriber.id })
+      .where("content_changes.created_at >= ?", digest_run.starts_at)
+      .where("content_changes.created_at < ?", digest_run.ends_at)
+      .order("subscriber_list_title ASC", "content_changes.title ASC")
+  end
+end
--- a/app/models/content_change.rb
+++ b/app/models/content_change.rb
@@ -1,7 +1,7 @@
 class ContentChange < ApplicationRecord
   include SymbolizeJSON
 
-  has_many :matched_subscriber_lists, through: :matched_content_changes
+  has_many :matched_content_changes
 
   enum priority: { low: 0, high: 1 }
 
--- a/app/models/subscriber_list.rb
+++ b/app/models/subscriber_list.rb
@@ -10,7 +10,7 @@ class SubscriberList < ApplicationRecord
 
   has_many :subscriptions
   has_many :subscribers, through: :subscriptions
-  has_many :matched_content_changes, through: :matched_content_changes
+  has_many :matched_content_changes
 
   def self.build_from(params:, gov_delivery_id:)
     new(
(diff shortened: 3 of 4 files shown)


app/queries/content_change_matcher.rb
```diff
@@ -0,0 +1,7 @@
+class ContentChangeMatcher
+  def self.call(subscription:)
+    ContentChange.where(
```
gpeng: I think this works at the moment but I wonder if we'll need to update this to return something here that will expose which `SubscriberList` the `ContentChange`s relate to. Depends how we structure the email and implement the unsubscribe stuff.

thomasleese: Yeah, I agree, we might need to return something more useful, but I wasn't sure what that would be yet.

gpeng: No, me neither... This works for a minimal implementation though so I'd go with it for now.

app/queries/content_change_matcher.rb
```diff
@@ -0,0 +1,7 @@
+class ContentChangeMatcher
+  def self.call(subscription:)
```
gpeng: Just been thinking a bit further about this and I'm not sure it will work as is. Won't this return all the content changes that ever matched the subscriber lists? Also we might need to call this with an array of subscriptions if we're building one digest per subscriber.

thomasleese: It will return all content changes, I assumed it would be up to the caller to limit the results by the date range they want for the digest. 

I believe the decision has been made to stick with building one digest per subscription rather than per subscriber, so we shouldn't need to pass in an array.

app/queries/subscription_content_change_query.rb
```diff
@@ -0,0 +1,53 @@
...
+    @digest_run = digest_run
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    present
+  end
```
kevindew: might want to also have a `private_class_method :new` if self.call is the only route into this class but I know it's something we don't do a lot (or at all) at gov.uk

app/queries/subscription_content_change_query.rb
```diff
@@ -0,0 +1,53 @@
...
+        subscriber_list_title,
+        content_changes.map { |attrs| hydrate_content_change(attrs) },
+      )
+    end
+  end
+
+  def hydrate_content_change(attrs)
+    ContentChange.new(
+      attrs.except("subscription_id", "subscription_uuid", "subscriber_list_title")
+    )
```
kevindew: This seems a bit unorthadox - I can see why you've done it though. I think there might be a way around it though. I'll just have a quick tinker.

app/queries/subscription_content_change_query.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def grouped_content_changes
+    content_changes.as_json.group_by do |record|
```
kevindew: Are we effectively taking ContentChange objects here, converting them to a hash (through as_json) and then rehydrating them?

kevindew:
So had a little bit of a tinker my end, and now feel I understand more how you've arrived at the solution.

The hydrate approach does sit a bit odd with me. It’s probably not actually a big deal in practice, but it does create a slightly weird model that thinks it’s not persisted and will blow up if you try change any data on it. I’m not sure if it might be better to drop the .as_json part and just deal with having these ContentChanges with extra fields eg: `ContentChange.select("content_changes.*", "subscriptions.id AS subscription_id", "subscriber_lists.title AS subscriber_list_title", "subscriptions.uuid AS subscription_uuid").joins(matched_content_changes: { subscriber_list: { subscriptions: :subscriber } }).first["subscription_uuid”]` I think they’d be more orthodox.

Also some things to consider (that probably aren’t biggies) are just the ordering of the data for the digest, which you might want to slot in.

Couple of alternative approaches here: 

Simplest way to me, but at an expense of queries would be: 

- grab a users subscriptions ordered by title
- for each subscription run query to grab content changes

This one would be easiest route to apply caching as the latter query would be identical results for each user that has same subscription.

Otherwise could break it into two queries where we have one that gets subscription info and ContentChange data eg `Subscription.joins(subscriber_list: :matched_content_changes).group(:id, :"subscriber_lists.title").pluck(:id, :"subscriber_lists.title", :uuid, "ARRAY_AGG(matched_content_changes.content_change_id)”)` and then use the ContentChange id array to grab them all in one query - and would mean we wouldn't need to grab same ContentChange out multiple times if it appeared in multiple subscriptions.

Does feel like there's not one clear win route, just a balance of number of queries and number of wtfs.

thomasleese:
Thanks @kevindew I think I've addressed your comments here!

I might have misinterpreted what you were saying about the ordering, but I didn't need to make more than one query as I could use `.order` with the extra fields we're pulling out.

app/queries/subscription_content_change_query.rb
```diff
@@ -0,0 +1,51 @@
...
+  def grouped_content_changes
+    content_changes.group_by do |record|
+      [record["subscription_id"], record["subscription_uuid"], record["subscriber_list_title"]]
+    end
+  end
+
+  def content_changes
+    ContentChange
+      .select("content_changes.*", "subscriptions.id AS subscription_id", "subscriber_lists.title AS subscriber_list_title", "subscriptions.uuid AS subscription_uuid")
+      .from("content_changes")
```
kevindew: You don't need this method

app/queries/subscription_content_change_query.rb
```diff
@@ -0,0 +1,51 @@
...
+      .joins(matched_content_changes: { subscriber_list: { subscriptions: :subscriber } })
+      .where(subscribers: { id: subscriber.id })
+      .where("content_changes.created_at >= ?", digest_run.starts_at)
+      .where("content_changes.created_at < ?", digest_run.ends_at)
+      .order("subscriber_list_title ASC")
```
kevindew: I think this would be a little clearer as "subscriber_list_title ASC, title ASC" and skip the call below.

PR #374, opened 2018-01-24
Refactor various classes to match digest architecture

thomasleese:
This PR renamed and refactors various classes related to digest so that it better follows our architecture design. Probably easier to review the individual commits for more details.


app/presenters/unsubscribe_link_presenter.rb
```diff
@@ -17,7 +17,11 @@ def call
 private
 
   def url
-    PublicUrlService.url_for(base_path: "/email/unsubscribe/#{uuid}?title=#{escaped_title}")
+    PublicUrlService.url_for(base_path: base_path)
+  end
+
+  def base_path
+    "/email/unsubscribe/#{uuid}?title=#{escaped_title}"
```
kevindew: I'd dare suggest that having a method for this adds unnecessary indirection. I'd maybe suggest we did url as:

```
escaped_title = ERB::Util.url_encode(title)
base_path =  "/email/unsubscribe/#{uuid}?title=#{escaped_title}"
PublicUrlService.url_for(base_path: base_path)
```

and ditched the other two methods

app/workers/immediate_email_generation_worker.rb
```diff
@@ -0,0 +1,7 @@
+class ImmediateEmailGenerationWorker
+  include Sidekiq::Worker
+
+  def perform
+    ImmediateEmailGenerationService.call
```
kevindew: Not something new here, but I wonder whether we might want to ditch that service and just have a fatter worker? Maybe I'm missing what we're gaining.

app/services/immediate_email_generation_service.rb
```diff
@@ -54,21 +54,14 @@ def subscription_contents
...
+  def import_emails(subscription_contents)
     ImmediateEmailBuilder.call(
-      subscriber: subscription_content.subscription.subscriber,
-      content_change: subscription_content.content_change
+      subscription_contents.map do |subscription_content|
+        {
+          subscriber: subscription_content.subscription.subscriber,
+          content_change: subscription_content.content_change,
+        }
+      end
```
kevindew: I'd suggest that it might be more readable to have the map result assigned to a variable and then passed in to the call, gets a little confusing within the method arguments.

app/builders/immediate_email_builder.rb
```diff
@@ -1,43 +1,57 @@
...
     new(*args).call
   end
 
   def call
-    {
-      address: subscriber.address,
-      subject: subject,
-      body: body
-    }
+    Email.import!(records)
```
kevindew: I notice this creates the emails, whereas I think the digest one merely builds one. Might it make sense for the digest one to also create for consistency?

spec/units/builders/immediate_email_builder_spec.rb
```diff
@@ -19,14 +19,18 @@
...
-    subject(:email) { described_class.call(subscriber: subscriber, content_change: content_change) }
+    subject(:email_import) do
+      described_class.call([{ subscriber: subscriber, content_change: content_change }])
+    end
+
+    let(:email) { Email.find(email_import.ids.first) }
 
     it "returns an email hash" do
-      expect(email).to be_a(Hash)
+      expect(email_import.ids.count).to eq(1)
```
kevindew: Does this test still assert correctly to the it?

app/builders/immediate_email_builder.rb
```diff
@@ -1,43 +1,57 @@
 class ImmediateEmailBuilder
-  def initialize(subscriber:, content_change:)
-    @subscriber = subscriber
-    @content_change = content_change
+  def initialize(subscriber_content_changes)
+    @subscriber_content_changes = subscriber_content_changes
```
kevindew: Probably overkill but as I understand this is a an array of hashes right? I wonder if it'd be consistent with the digest equivalent to have it be an array of structs (that might not even be how that works - ignore me if I've got the scenarios wrong)

thomasleese: The digest equivalent only makes one email, so it doesn't need an array, it uses keyword arguments instead. So in a way this implement is more consistent with digests.

We might want to revisit this if we decide that the digest equivalent needs to also create things in bulk.

app/workers/immediate_email_generation_worker.rb
thomasleese: Agreed - and I'll rebase this off #379 and do the same for that.

thomasleese:
Thanks @kevindew I believe I've addressed all your comments now!

PR #376, opened 2018-01-25
Add digest generation and sending

gpeng:
This PR adds digest generation functionality that will be initiatiated by the `DigestScheduler`s (soon to be renamed to `Initiator`s)

It also extracts presentation behaviour for `ContentChange` and `UnsubscribeLink` which will be incorporated into immediate email generation in a future PR.

We'll add a feature test for this whole process in another PR once we have implemented `DigestSubscriberQuery` and hooked it up to the initiator.

Paired with @thomasleese 

[Trello](https://trello.com/c/t1tltKn6/535-build-digestbuilder-worker)

### DIFF
--- a/app/builders/digest_email_builder.rb
+++ b/app/builders/digest_email_builder.rb
@@ -1,0 +1,63 @@
+class DigestEmailBuilder
+  def initialize(subscriber:, digest_run:, subscription_content_change_results:)
+    @subscriber = subscriber
+    @digest_run = digest_run
+    @results = subscription_content_change_results
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    Email.new(
+      subject: subject,
+      body: body,
+      address: subscriber.address
+    )
+  end
+
+  private_class_method :new
+
+private
+
+  attr_reader :subscriber, :digest_run, :results
+
+  def body
+    results.map { |result| presented_result(result) }.join("\n&nbsp;\n\n")
+  end
+
+  def presented_result(result)
+    <<~RESULT
+      ##{result.subscriber_list_title}
+
+      #{presented_content_changes(result.content_changes)}
+      ---
+
+      #{unsubscribe_link(result)}
+    RESULT
+  end
+
+  def subject
+    if digest_run.daily?
+      "GOV.UK Daily Update"
+    else
+      "GOV.UK Weekly Update"
+    end
+  end
+
+  def presented_content_changes(content_changes)
+    changes = content_changes.map do |content_change|
+      ContentChangePresenter.call(content_change)
+    end
+
+    changes.join("\n---\n\n")
+  end
+
+  def unsubscribe_link(result)
+    UnsubscribeLinkPresenter.call(
+      uuid: result.subscription_uuid,
+      title: result.subscriber_list_title
+    )
+  end
+end
--- a/app/presenters/content_change_presenter.rb
+++ b/app/presenters/content_change_presenter.rb
@@ -1,0 +1,37 @@
+class ContentChangePresenter
+  EMAIL_DATE_FORMAT = "%I:%M %P on %-d %B %Y".freeze
+
+  def initialize(content_change)
+    @content_change = content_change
+  end
+
+  def self.call(content_change)
+    new(content_change).call
+  end
+
+  def call
+    <<~BODY
+      [#{title}](#{content_url})
+
+      #{change_note}: #{description}
+
+      Updated at #{public_updated_at}
+    BODY
+  end
+
+  private_class_method :new
+
+private
+
+  attr_reader :content_change
+
+  delegate :title, :description, :change_note, to: :content_change
+
+  def content_url
+    PublicUrlService.content_url(base_path: content_change.base_path)
+  end
+
+  def public_updated_at
+    content_change.public_updated_at.strftime(EMAIL_DATE_FORMAT)
+  end
+end
--- a/app/workers/digest_email_generation_worker.rb
+++ b/app/workers/digest_email_generation_worker.rb
@@ -1,0 +1,61 @@
+class DigestEmailGenerationWorker
+  include Sidekiq::Worker
+
+  attr_reader :subscriber, :digest_run, :email
+
+  def perform(subscriber_id:, digest_run_id:)
+    @subscriber = Subscriber.find(subscriber_id)
+    @digest_run = DigestRun.find(digest_run_id)
+
+    generate_email_and_subscription_contents
+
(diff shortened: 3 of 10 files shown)


app/builders/digest_email_builder.rb
```diff
@@ -0,0 +1,65 @@
...
+    if digest_run.daily?
+      "GOV.UK Daily Update"
+    else
+      "GOV.UK Weekly Update"
+    end
+  end
+
+  def body
+    presented_results
+  end
```
kevindew: Do we need these to be public anymore?

app/presenters/content_change_presenter.rb
```diff
@@ -0,0 +1,45 @@
+class ContentChangePresenter
+  EMAIL_DATE_FORMAT = "%I:%M %P on %-d %B %Y".freeze
+
```
kevindew: Might want to add something here of `delegate :title, :description, :change_note to :content_change`

Or even have this class inherit from [SimpleDelegator](http://ruby-doc.org/stdlib-2.2.1/libdoc/delegate/rdoc/SimpleDelegator.html)

app/workers/digest_email_generation_worker.rb
```diff
@@ -0,0 +1,39 @@
+class DigestEmailGenerationWorker
+  include Sidekiq::Worker
+
+  def perform(subscriber_id:, digest_run_id:)
```
kevindew: This should be taken a DigestRunSubscriber, do they exist yet?

app/workers/digest_email_generation_worker.rb
```diff
@@ -0,0 +1,39 @@
...
+    Email.transaction do
+      email.save!
+
+      subscription_content_changes = results.flat_map do |result|
+        result.content_changes.map do |content_change|
+          {
+            email_id: email.id,
+            subscription_id: result.subscription_id,
+            content_change_id: content_change.id,
+            #TODO we need to set the type on this when there is one
```
kevindew: guess this is where we'd bung the digest_run_subscriber_id ?

app/builders/digest_email_builder.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def call
+    Email.new(
+      subject: subject,
+      body: body,
+      address: subscriber.address
+    )
+  end
+
+  def subject
```
thomasleese: I wonder if this and `body` should also be private?

app/builders/digest_email_builder.rb
```diff
@@ -0,0 +1,65 @@
...
+    if digest_run.daily?
+      "GOV.UK Daily Update"
+    else
+      "GOV.UK Weekly Update"
+    end
+  end
+
+  def body
```
thomasleese: If we make `body` private, I don't think we need the `presented_results` method.

app/presenters/content_change_presenter.rb
```diff
@@ -0,0 +1,45 @@
...
+  def initialize(content_change)
+    @content_change = content_change
+  end
+
+  def self.call(content_change)
```
thomasleese: Should we use `*args` here for consistency with the other classes, and in case we decide to change the method signature?

app/workers/digest_email_generation_worker.rb
thomasleese: I guess we don't actually need this since the `type` will be a dynamic scope with a `JOIN`.

app/workers/digest_email_generation_worker.rb
```diff
@@ -0,0 +1,39 @@
...
+          }
+        end
+      end
```
kevindew: I wonder if we want some private methods so that this method is a bit easier to digest, feels like the subscription_content_changes stuff is suitable for abstracting?

app/builders/digest_email_builder.rb
gpeng: Fixed in 299ca1d

app/builders/digest_email_builder.rb
gpeng: Fixed in 299ca1d

app/presenters/content_change_presenter.rb
gpeng: Fixed in 809c5d2

app/workers/digest_email_generation_worker.rb
gpeng: Fixed in dea0018

app/builders/digest_email_builder.rb
gpeng: Nope... fixed in 299ca1d

app/presenters/content_change_presenter.rb
gpeng: Good shout... fixed in b202f42

app/workers/digest_email_generation_worker.rb
gpeng: No, not yet. I'll do that in the next PR when we connect this up to the initiator thing

app/workers/digest_email_generation_worker.rb
gpeng: Yep... probably. Next PR's the charm. I've removed that comment now too as it's bobbins.

app/workers/digest_email_generation_worker.rb
gpeng: Yep... I'm going to refactor this. Bear with...

gpeng: I've had a hasty stab at splitting it up a bit in 2f24381. I'll have another look tomorrow.

app/presenters/content_change_presenter.rb
```diff
@@ -0,0 +1,37 @@
+class ContentChangePresenter
+  EMAIL_DATE_FORMAT = "%I:%M %P on %-d %B %Y".freeze
+
+  def initialize(content_change:)
```
kevindew: Oh super minor, but doesn't really feel like this needs keyword arguments. We end up with calling it looking like this `ContentChangePresenter.call(content_change: content_change)` which is a lot of content changes.

gpeng: I agree. I didn't use them originally but we (@thomasleese and I) discussed making it consistent with the other workers. Happy to change back if you prefer.

kevindew: I think it'd be better back. But happy for you and @thomasleese to decide.

thomasleese: I'm happy either way.

gpeng: Ok... i've fixed it in ff29953 and 3ef6219. Should be good to go now if one of you could reapprove.

PR #378, opened 2018-01-25
Digest deduplication

gpeng:
Remove duplicates within a `SubscriberList` group in digest emails.

[Trello](https://trello.com/c/t1tltKn6/535-build-digestbuilder-worker)

### DIFF
--- a/app/services/content_change_deduplicator_service.rb
+++ b/app/services/content_change_deduplicator_service.rb
@@ -1,0 +1,18 @@
+class ContentChangeDeduplicatorService
+  def self.call(*args)
+    new.call(*args)
+  end
+
+  def call(content_changes)
+    reverse_by_public_updated_at(content_changes)
+      .uniq(&:content_id)
+  end
+
+private
+
+  def reverse_by_public_updated_at(content_changes)
+    content_changes.sort do |cc_a, cc_b|
+      cc_b.public_updated_at <=> cc_a.public_updated_at
+    end
+  end
+end
--- a/spec/units/builders/digest_email_builder_spec.rb
+++ b/spec/units/builders/digest_email_builder_spec.rb
@@ -35,6 +35,10 @@ RSpec.describe DigestEmailBuilder do
     )
   }
 
+  def simulated_deduplification(content_changes)
+    content_changes.slice(0, content_changes.length - 1)
+  end
+
   it "returns an Email" do
     expect(email).to be_a(Email)
   end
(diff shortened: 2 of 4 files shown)


app/services/content_change_deduplicator_service.rb
```diff
@@ -0,0 +1,18 @@
+class ContentChangeDeduplicatorService
+  def self.call(*args)
+    new.call(*args)
+  end
+
+  def call(content_changes)
+    deduplicated = reverse_by_public_updated_at(content_changes)
```
thomasleese: I wonder if the separate methods are overkill here, I think this is just as readable:

```ruby
class ContentChangeDeduplicatorService
  def self.call(content_changes)
    content_changes
      .sort_by(&:public_updated_at).reverse
      .uniq(&:content_id)
      .sort_by(&:public_updated_at)
  end
end
```

app/services/content_change_deduplicator_service.rb
```diff
@@ -0,0 +1,18 @@
...
+      .uniq(&:content_id)
+
+    deduplicated.sort_by(&:public_updated_at)
```
thomasleese: Should we return the results ordered alphabetically?

spec/units/services/content_change_deduplicator_service_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  it "removes duplicates preserving only the latest" do
+    content_changes = [
+      build(
+        :content_change, id: 1,
+        public_updated_at: 10.day.ago,
+        content_id: "8e1fef7b-cd23-4dd2-84d0-61955203db61"
+      ),
+      latest_content_change = build(
+        :content_change,
+        id: 2,
```
thomasleese: I reckon we don't need to explicitly set the IDs here (saves a few lines 😄)

spec/units/builders/digest_email_builder_spec.rb
```diff
@@ -50,7 +50,23 @@
...
-    expect(ContentChangePresenter).to receive(:call).exactly(6).times
+    first_content_changes = subscription_content_change_results
+      .first
+      .content_changes
+
+    second_content_changes = subscription_content_change_results
+      .second
+      .content_changes
+
+    expect(ContentChangeDeduplicatorService).to receive(:call)
```
thomasleese: I wonder if we can extract these two calls into a method to make it slightly clearer about what it does?

app/services/content_change_deduplicator_service.rb
gpeng: I've changed this a bit but not quite how you suggested. I've done the reverse sort in one operation so I've left that as a separate method as it's unclear what it's doing otherwise. I've also remove the last sort as having spoken to @keithiopia offline they want to proceed with descending order of `public_updated_at` for now.

app/services/content_change_deduplicator_service.rb
gpeng: No 🤣

spec/units/services/content_change_deduplicator_service_spec.rb
gpeng: 👍

spec/units/builders/digest_email_builder_spec.rb
gpeng: fixed in 09c3dec

PR #379, opened 2018-01-26
Schedule digest workers with Sidekiq

thomasleese:
We're seeing a few problems with setting up Whenever so we've decided to switch to using Sidekiq.

The configuration for Whenever needs to be coupled to `govuk-puppet` which makes it harder to run the workers locally, and the application deployment needed updating to support Whenever.

[Trello Card](https://trello.com/c/VQYJdiVr/547-implement-digest-architecture)


config/sidekiq.yml
```diff
@@ -10,3 +10,9 @@
   email_generation:
     every: '5s'
     class: EmailGenerationWorker
+  daily_digest_initiator:
+    cron: '30 8 * * *'
+    class: DailyDigestInitiatorWorker
+  weekly_digest_initiator:
+    cron: '30 8 * * 6'
+    class: WeeklyDigestInitiatorWorker
```
kevindew: Might be useful to just have a comment about what these translate to in human terms as 6th day of week isn't too fast to grok. 

Also not really a discussion for here, but why'd we go for Saturday over Monday?

PR #381, opened 2018-01-26
Connect DigestInitiatorService

gpeng:
This PR hooks up the `DigestInitiatorService` to all of the other elements of the digest creation flow as per https://github.com/alphagov/email-alert-api/blob/master/doc/digests.png

[Trello](https://trello.com/c/VQYJdiVr/547-implement-digest-architecture)

### DIFF
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -15,6 +15,10 @@ class DigestRun < ApplicationRecord
     update_attributes!(completed_at: Time.now)
   end
 
+  def check_and_mark_complete!
+    mark_complete! unless has_incomplete_subscribers?
+  end
+
 private
 
   def starts_at=(value)
@@ -48,5 +52,9 @@ private
 
   def digest_range_hour
     ENV.fetch("DIGEST_RANGE_HOUR", 8).to_i
+  end
+
+  def has_incomplete_subscribers?
+    digest_run_subscribers.incomplete_for_run(id).exists?
   end
 end
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -11,6 +11,10 @@ class DigestRun < ApplicationRecord
   DAILY = "daily".freeze
   WEEKLY = "weekly".freeze
 
+  def mark_complete!
+    update_attributes!(completed_at: Time.now)
+  end
+
 private
 
   def starts_at=(value)
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -3,6 +3,9 @@ class DigestRun < ApplicationRecord
   before_validation :set_range_dates, on: :create
   validate :ends_at_is_in_the_past
 
+  has_many :digest_run_subscribers, dependent: :destroy
+  has_many :subscribers, through: :digest_run_subscribers
+
   enum range: { daily: 0, weekly: 1 }
 
   DAILY = "daily".freeze
--- a/app/models/digest_run_subscriber.rb
+++ b/app/models/digest_run_subscriber.rb
@@ -1,0 +1,9 @@
+class DigestRunSubscriber < ApplicationRecord
+  validates :digest_run_id, :subscriber_id, presence: true
+  belongs_to :digest_run
+  belongs_to :subscriber
+
+  def mark_complete!
+    update_attributes!(completed_at: Time.now)
+  end
+end
--- a/app/models/digest_run_subscriber.rb
+++ b/app/models/digest_run_subscriber.rb
@@ -3,6 +3,8 @@ class DigestRunSubscriber < ApplicationRecord
   belongs_to :digest_run
   belongs_to :subscriber
 
+  scope :incomplete_for_run, ->(digest_run_id) { where(digest_run_id: digest_run_id).where(completed_at: nil) }
+
   def mark_complete!
     update_attributes!(completed_at: Time.now)
   end
--- a/app/services/digest_initiator_service.rb
+++ b/app/services/digest_initiator_service.rb
@@ -8,36 +8,52 @@ class DigestInitiatorService
   end
 
   def call
-    digest_run = nil
+    digest_run = create_digest_run
+    return if digest_run.nil?
 
+    subscribers = DigestRunSubscriberQuery.call(digest_run: digest_run)
+
+    digest_run_subscriber_params = build_digest_run_subscriber_params(
+      digest_run,
+      subscribers
+    )
+
+    digest_run_subscriber_ids = import_digest_run_subscribers(
+      digest_run_subscriber_params
+    )
+
+    enqueue_jobs(digest_run_subscriber_ids)
+  end
+
+private
+
+  attr_reader :range
+
+  def create_digest_run
     run_with_advisory_lock do
       digest_run = DigestRun.find_or_initialize_by(
         date: Date.current, range: range
       )
       return if digest_run.persisted?
       digest_run.save!
+      digest_run
     end
+  end
 
-    #TODO think about this being retried
-    #
-    subscribers = DigestRunSubscriberQuery.call(digest_run: digest_run)
-
-    digest_run_subscriber_params = subscribers.map do |subscriber|
+  def build_digest_run_subscriber_params(digest_run, subscribers)
+    subscribers.map do |subscriber|
       {
         subscriber_id: subscriber.id,
         digest_run_id: digest_run.id
       }
     end
+  end
 
-    import_result = DigestRunSubscriber.import!(digest_run_subscriber_params)
-    import_result.ids.each do |digest_run_subscriber_id|
+  def enqueue_jobs(digest_run_subscriber_ids)
+    Array(digest_run_subscriber_ids).each do |digest_run_subscriber_id|
       DigestEmailGenerationWorker.perform_async(digest_run_subscriber_id)
     end
   end
-
-private
-
-  attr_reader :range
 
   def run_with_advisory_lock
     DigestRun.with_advisory_lock(lock_name, timeout_seconds: 0) do
@@ -47,5 +63,9 @@ private
 
   def lock_name
     "#{range}_digest_initiator"
+  end
+
+  def import_digest_run_subscribers(params)
+    DigestRunSubscriber.import!(params).ids
   end
 end
--- a/app/services/digest_initiator_service.rb
+++ b/app/services/digest_initiator_service.rb
@@ -29,7 +29,10 @@ class DigestInitiatorService
       }
     end
 
-    DigestRunSubscriber.import!(digest_run_subscriber_params)
+    import_result = DigestRunSubscriber.import!(digest_run_subscriber_params)
+    import_result.ids.each do |digest_run_subscriber_id|
+      DigestEmailGenerationWorker.perform_async(digest_run_subscriber_id)
+    end
   end
 
 private
--- a/app/services/digest_initiator_service.rb
+++ b/app/services/digest_initiator_service.rb
@@ -8,6 +8,8 @@ class DigestInitiatorService
   end
 
   def call
+    digest_run = nil
+
     run_with_advisory_lock do
       digest_run = DigestRun.find_or_initialize_by(
         date: Date.current, range: range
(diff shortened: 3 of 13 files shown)


app/services/digest_initiator_service.rb
```diff
@@ -16,7 +18,14 @@ def call
       digest_run.save!
     end
 
-    #enqueue the digest creation workers TBC
+    #TODO think about this being retried
+    #
+    subscribers = DigestRunSubscriberQuery.call(digest_run: digest_run)
+
+    subscribers.each do |subscriber|
+      digest_run_subscriber = DigestRunSubscriber.create(digest_run: digest_run, subscriber: subscriber)
```
thomasleese: Don't know if it matters at this point, but we may want to insert these in bulk and then call the worker using the `.ids` field of the result value from `DigestRunSubscriber.import!`.

app/workers/digest_email_generation_worker.rb
```diff
@@ -1,13 +1,17 @@
 class DigestEmailGenerationWorker
   include Sidekiq::Worker
 
-  def perform(subscriber_id:, digest_run_id:)
-    @subscriber = Subscriber.find(subscriber_id)
-    @digest_run = DigestRun.find(digest_run_id)
+  def perform(digest_run_subscriber_id)
```
thomasleese: Do we need to wrap this whole thing in a transaction in case it restarts? Or maybe at least the `generate_email_and_subscription_contents` method.

app/services/digest_initiator_service.rb
gpeng: This might be a bit of a premature optimisation but we've used the same pattern a lot so I've updated in 6e30b26.

app/workers/digest_email_generation_worker.rb
gpeng: That method did run in a transaction but I've expanded it a bit to wrap the queuing and marking complete bits. Added in ef4607c

app/services/digest_initiator_service.rb
```diff
@@ -8,6 +8,8 @@ def self.call(*args)
   end
 
   def call
+    digest_run = nil
```
kevindew: Rather than this being a distinct assignment can it be just returned from `run_with_advisory_lock` ?

Also what happens if lock is in progress, does this wait or isn't executed? just wondering if we need provisions for digest_run to be a nil value

app/services/digest_initiator_service.rb
```diff
@@ -16,7 +18,21 @@ def call
       digest_run.save!
     end
 
-    #enqueue the digest creation workers TBC
+    #TODO think about this being retried
```
kevindew: I don't think retry is too big a concern (if I understand correctly) as it'll only reach this point in an execution if there was not a DigestRun for date and range

app/services/digest_initiator_service.rb
```diff
@@ -8,6 +8,8 @@ def self.call(*args)
   end
 
   def call
+    digest_run = nil
+
     run_with_advisory_lock do
       digest_run = DigestRun.find_or_initialize_by(
         date: Date.current, range: range
```
kevindew: Is it prudent to set this date to be UTC for resolving any ambiguity there?

db/migrate/20180126144040_create_digest_run_subscribers.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :digest_run_subscribers do |t|
+      t.integer :digest_run_id, null: false
+      t.integer :subscriber_id, null: false
+      t.datetime :completed_at
+      t.timestamps
+    end
+
+    add_index :digest_run_subscribers, :digest_run_id
+    add_index :digest_run_subscribers, :subscriber_id
```
kevindew: Shall we have these as foreign_keys? and if so shall we have on_delete: cascades?

db/migrate/20180129081557_add_digest_run_subscriber_id_to_subscription_contents.rb
```diff
@@ -0,0 +1,6 @@
+class AddDigestRunSubscriberIdToSubscriptionContents < ActiveRecord::Migration[5.1]
+  def change
+    add_column :subscription_contents, :digest_run_subscriber_id, :integer
+    add_index :subscription_contents, :digest_run_subscriber_id
```
kevindew: Similar comment about foreign keys, I think we have those on other columns in this model

app/workers/digest_email_generation_worker.rb
```diff
@@ -1,27 +1,30 @@
...
+    @digest_run = digest_run_subscriber.digest_run
 
-    generate_email_and_subscription_contents
+    Email.transaction do
+      generate_email_and_subscription_contents
+
+      DeliveryRequestWorker.perform_async_with_priority(email.id, priority: :low)
 
-    DeliveryRequestWorker.perform_async_with_priority(email.id, priority: :low)
+      digest_run_subscriber.mark_completed!
```
kevindew: I think this is the point where we also want to mark a digest as completed if all digest runs are complete.

app/workers/digest_email_generation_worker.rb
```diff
@@ -1,27 +1,30 @@
...
+  def perform(digest_run_subscriber_id)
+    @digest_run_subscriber = DigestRunSubscriber.find(digest_run_subscriber_id)
+    @subscriber = digest_run_subscriber.subscriber
+    @digest_run = digest_run_subscriber.digest_run
 
-    generate_email_and_subscription_contents
+    Email.transaction do
+      generate_email_and_subscription_contents
+
+      DeliveryRequestWorker.perform_async_with_priority(email.id, priority: :low)
```
kevindew: I think we want to move this until after the transaction as the id won't exist until the transaction is committed (if I've understood this correctly)

app/services/digest_initiator_service.rb
gpeng: It blocks until it can get the lock and then runs. I've refactored this a bit and returned nil (which in turn bails out of the worker) if the `DigestRun` already exists.

app/workers/digest_email_generation_worker.rb
gpeng: Yep... Fixed

app/workers/digest_email_generation_worker.rb
gpeng: Added :+1:

db/migrate/20180126144040_create_digest_run_subscribers.rb
gpeng: Yep, updated.

db/migrate/20180129081557_add_digest_run_subscriber_id_to_subscription_contents.rb
gpeng: I've made this a foreign key. Not sure we want the dependent destroy on this one. What do you think?

app/models/digest_run_subscriber.rb
```diff
@@ -0,0 +1,11 @@
+class DigestRunSubscriber < ApplicationRecord
+  validates :digest_run_id, :subscriber_id, presence: true
+  belongs_to :digest_run
+  belongs_to :subscriber
+
+  scope :incomplete_for_run, ->(digest_run_id) { where(digest_run_id: digest_run_id).where(completed_at: nil) }
+
+  def mark_completed!
```
thomasleese: This is called `mark_completed!` but we called the one on `DigestRun` `mark_complete!`, I reckon it would be good to call them the same. I don't mind which though.

app/services/digest_initiator_service.rb
```diff
@@ -8,20 +8,50 @@ def self.call(*args)
...
+
+  attr_reader :range
+
+  def create_digest_run
     run_with_advisory_lock do
       digest_run = DigestRun.find_or_initialize_by(
         date: Date.current, range: range
       )
-      return if digest_run.persisted?
+      return nil if digest_run.persisted?
```
thomasleese: I suspect we don't need an explicit `nil` here.

app/services/digest_initiator_service.rb
```diff
@@ -8,20 +8,50 @@ def self.call(*args)
...
+
+    subscribers = DigestRunSubscriberQuery.call(digest_run: digest_run)
+
+    digest_run_subscriber_params = build_digest_run_subscriber_params(
+      digest_run,
+      subscribers
+    )
+
+    enqueue_jobs(
+      DigestRunSubscriber.import!(digest_run_subscriber_params).ids
```
thomasleese: I wonder if we should make this line more of an explicit step rather than as part of arguments to a call, might make it slightly clearer.

app/workers/digest_email_generation_worker.rb
```diff
@@ -50,4 +55,8 @@ def subscriber_content_changes
       digest_run: digest_run
     )
   end
+
+  def update_digest_run
+    digest_run.mark_complete! unless DigestRunSubscriber.incomplete_for_run(digest_run.id).exists?
```
thomasleese: I might make `DigestRunSubscriber.incomplete_for_run(digest_run.id).exists?` into a method on the `DigestRun` class, so you can do something like `digest_run.has_incomplete_subscribers?`.

thomasleese: I agree it is a bit of an odd coupling, maybe `mark_complete!` could check whether the `DigestRunSubscriber`s are all marked as complete and if not raise an exception, and then have a method like you suggest `mark_complete_if_able_to` which ignores that exception.

db/migrate/20180129081557_add_digest_run_subscriber_id_to_subscription_contents.rb
```diff
@@ -0,0 +1,6 @@
+class AddDigestRunSubscriberIdToSubscriptionContents < ActiveRecord::Migration[5.1]
+  def change
+    add_column :subscription_contents, :digest_run_subscriber_id, :integer
+    add_foreign_key :subscription_contents, :digest_run_subscribers
```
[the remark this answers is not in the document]
gpeng: Shall we remove the foreign key constraint then and go 'full Rails'?

app/workers/digest_email_generation_worker.rb
gpeng: I've added a variation on this that encapsulates the whole 'complete' thing in `DigestRun`. I didn't go the exception route as we'd be expecting a lot of the time so it was more control of flow than exceptional. We could add maybe add a validation if you think we should guard it a bit more.

app/models/digest_run.rb
```diff
@@ -3,11 +3,18 @@ class DigestRun < ApplicationRecord
   before_validation :set_range_dates, on: :create
   validate :ends_at_is_in_the_past
 
+  has_many :digest_run_subscribers, dependent: :destroy
```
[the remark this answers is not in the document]
gpeng: I've added that. I think I prefer keeping these sort of rules in the app where you can see 'em but in the interest of consistency an' all that :)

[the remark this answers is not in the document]
kevindew: @gpeng this is where we differ GP 😉I'm all about the db constraints and find the rails ones dangerous and confusing (ala callbacks)

db/schema.rb
```diff
@@ -157,9 +166,12 @@
   end
 
   add_foreign_key "delivery_attempts", "emails", on_delete: :cascade
+  add_foreign_key "digest_run_subscribers", "digest_runs"
+  add_foreign_key "digest_run_subscribers", "subscribers"
```
kevindew: I've just been having a chat with @thomasleese about how these should be set up cascade wise (and this was what I was going to chat to you about it so it might be easier to pick up over a hangout).

So we think that it should be possible to delete DigestRun rows and without them DigestRunSubscriber no longer makes sense so they should be cascaded, and the corresponding subscription_contents should also be removed as they don't make sense without the digest run. Which would indicate: `add_foreign_key "digest_run_subscribers", "digest_runs"` should have an `on_delete: :cascade`

I'd guess we also want to remove digest_run_subscriber entry if the subscriber is removed which would be another `on_delete: :cascade`

Also looks like the email `on_delete: :nullify` is wrong (not your PR of course) as null email value means something. I think resolving this though should be part of working out an email archiving strategy.

app/models/digest_run.rb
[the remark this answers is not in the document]
gpeng: Yep... I know you're quite 'passionate' about them, that's why I didn't labour the point 😄

db/schema.rb
gpeng: ok... I think we're good now.

app/models/digest_run.rb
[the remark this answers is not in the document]
kevindew: haha I do love data integrity - even before I met Publishing API

PR #382, opened 2018-01-29
Configure email delivery priorities for digest and immediate emails

thomasleese:
This PR introduces a new Sidekiq queue dedicated to digest emails and updated the `DeliveryRequestWorker` to support sending immediate emails with a normal or high priority, or sending digest emails all with the same priority.

Both kinds of immediate emails will be sent out with a higher priority than digest emails.

Following on from this, the next part will be to update Travel Advice Publisher and Specialist Publisher to set the content changes as high priority.

[Trello Card](https://trello.com/c/mHXNdskh/550-setup-delivery-priorities)

### DIFF
--- a/app/workers/delivery_request_worker.rb
+++ b/app/workers/delivery_request_worker.rb
@@ -13,6 +13,10 @@ class DeliveryRequestWorker
     end
   end
 
+  def self.queue_for_digest
+    :delivery_digest
+  end
+
   sidekiq_options retry: 3, queue: queue_for_immediate(:normal)
 
   sidekiq_retry_in do |count|
@@ -34,6 +38,10 @@ class DeliveryRequestWorker
   def self.perform_async_for_immediate(*args, priority: :normal)
     queue = queue_for_immediate(priority)
     set(queue: queue).perform_async(*args, queue)
+  end
+
+  def self.perform_async_for_digest(*args)
+    set(queue: queue_for_digest).perform_async(*args, queue_for_digest)
   end
 
   def check_rate_limit!
--- a/app/workers/delivery_request_worker.rb
+++ b/app/workers/delivery_request_worker.rb
@@ -13,10 +13,6 @@ class DeliveryRequestWorker
     end
   end
 
-  def self.queue_for_digest
-    :delivery_digest
-  end
-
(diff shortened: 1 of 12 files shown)


app/workers/delivery_request_worker.rb
```diff
@@ -23,23 +27,21 @@ def self.queue_for_priority(priority)
...
     check_rate_limit!
     increment_rate_limiter
     DeliveryRequestService.call(email: @email)
   end
 
-  def self.perform_async_with_priority(*args, priority:)
-    @priority = priority
-    set(queue: queue_for_priority(priority))
-      .perform_async(*args)
+  def self.perform_async_for_immediate(*args, priority: :normal)
```
gpeng: Do we want to have this worker knowing about immediate/digest etc? I wonder if it would be better to have a more generic `perform_async_in_queue` (like Whitehall 🎉 ) and then take this logic out of here.

app/workers/digest_email_generation_worker.rb
```diff
@@ -7,7 +7,7 @@ def perform(subscriber_id:, digest_run_id:)
 
     generate_email_and_subscription_contents
 
-    DeliveryRequestWorker.perform_async_with_priority(email.id, priority: :low)
+    DeliveryRequestWorker.perform_async_for_digest(email.id)
```
gpeng: This already knows about digests so it might be better placed to either specify the queue or use another thing?

app/workers/delivery_request_worker.rb
thomasleese: I think I like these convenience methods as it helps make the existence of the queues more explicit. Particularly if we're having to manually send out emails on the console, these methods might make things easier to understand.

gpeng: I get the explicit bit but it does make this worker need to know about the types of emails we're handling and what their priorities should be rather than just delivering an email. We could make it explicit elsewhere.

thomasleese: I've pushed a few commits now which gets rid of those specific methods, I agree it makes sense for the `DeliveryRequestWorker` to not know about digest and immediate emails.

PR #385, opened 2018-01-30
Fake Data

thomasleese:
This PR introduces a rake task that can be used to generate fake testing data that is similar to a working production system. This allows us to better test the system before we go live.

The task took about 40 minutes to import all the fake data on my local dev VM.

[Trello Card](https://trello.com/c/vvfhCzJo/548-create-a-script-to-generate-fake-data-for-testing)


lib/fake_data.rb
```diff
@@ -0,0 +1,115 @@
...
+        loop do
+          SubscriberList.order("RANDOM()").pluck(:id).each do |id|
+            yielder << id
+          end
+        end
+      end
+    end
+  end
+
+  def uuids
```
gpeng: Do we need this?  On 69 we could just do `count.times.map { "email+#{SecureRandom.uuid}@blah.com"}` couldn't we?

thomasleese: It also gets used on this line: https://github.com/alphagov/email-alert-api/pull/385/files#diff-83292a82e43497f335fecce8e354102bR86

gpeng: Oh hang on... I see you use it again later.

lib/fake_data.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def existing_test_subscribers
+    Subscriber.where("address LIKE 'success+%@simulator.amazonses.com'")
+  end
+
+  def has_test_subscribers?
+    existing_test_subscribers.exists?
+  end
+
+  SubscriptionStat = Struct.new(:no_subscriptions, :no_subscribers)
```
gpeng: I think `number_of_` would be clearer here.

thomasleese: Fixed in 356b7a9

lib/fake_data.rb
```diff
@@ -0,0 +1,115 @@
...
+        end
+      end
+    end
+  end
+
+  def subscriber_list_ids
+    @subscriber_list_ids ||= begin
+      Enumerator.new do |yielder|
+        loop do
+          SubscriberList.order("RANDOM()").pluck(:id).each do |id|
```
kevindew: Should we grab all of this once and then pick a random one (no idea if this is part of performance issues mind)

lib/fake_subscription_data.csv
```diff
@@ -0,0 +1,414 @@
+1,296512
```
kevindew: Is there any chance the commit for this could explain what this data is and where it came from?

lib/tasks/fake_data.rake
```diff
@@ -0,0 +1,13 @@
+require "fake_data"
+
+namespace :fake_data do
+  desc "Insert fake data for testing the system"
+  task :insert, [] => :environment do
```
kevindew: Do we want any ability to pass in number we want to create? not sure if we want to be able to start small and then have to scope to experiment with an even bigger number?

lib/fake_subscription_data.csv
thomasleese: Agreed, added to 02ee34a.

lib/fake_data.rb
thomasleese: Not a performance issue, but it makes sense, I've added that in 40a2b5b.

lib/tasks/fake_data.rake
thomasleese: Agreed, I've added this in 226c296 allowing the user to specify a proportion of data.

kevindew: ooh a proportion is clever!

lib/fake_data.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def self.delete
+    new.delete
+  end
+
+  def insert(proportion)
+    if has_test_subscribers?
+      puts "There is already test data in the system."
+      puts "Run rake fake_data:delete first."
+      return
```
kevindew: Quite minor but probably want to raise here so it has an exit code other than 0. Otherwise may look successful running it when it's not.

thomasleese: Good point! Fixed and rebased!

PR #390, opened 2018-01-31
Reorganise tests

gpeng:
Moves unit test folders out of `units` into the root spec directory and helpers into support. This is more of a 'rails' standard structure and means that tooling within editors will work (jumping from class to spec etc).

The `features` and `integration` folders are untouched. This commit also updates the README in the top level `spec` folder to outline the changes.


gpeng:
Rebased from master to fix a couple of conflicts

PR #392, opened 2018-01-31
Only one unsubscribe link

thomasleese:
Currently immediate emails include an unsubscribe link for everything the subscriber is subscribed too. This can lead to very large and confusing looking emails, it makes more sense to only show the unsubscribe link of the subscription that caused the email to be sent.

Due to the courtesy emails, we have to let `ImmediateEmailBuilder` support being given either a `subscription ` or a `subscriber`.

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -37,31 +37,34 @@ private
                    subscription_content_change.fetch(:subscriber)
                  end
 
-    [subscriber.address, subject(content_change), body(subscriber, content_change)]
+    [subscriber.address, subject(content_change), body(content_change, subscription)]
   end
 
   def subject(content_change)
     "GOV.UK Update - #{content_change.title}"
   end
 
-  def body(subscriber, content_change)
-    <<~BODY
-      #{presented_content_change(content_change)}
-      ---
+  def body(content_change, subscription)
+    if subscription
+      <<~BODY
+        #{presented_content_change(content_change)}
+        ---
 
-      #{unsubscribe_links(subscriber)}
-    BODY
+        #{presented_unsubscribe_link(subscription)}
+      BODY
+    else
+      presented_content_change(content_change)
+    end
   end
 
   def presented_content_change(content_change)
     ContentChangePresenter.call(content_change)
   end
 
-  def unsubscribe_links(subscriber)
-    links = subscriber.subscriptions.map do |subscription|
-      UnsubscribeLinkPresenter.call(uuid: subscription.uuid, title: subscription.subscriber_list.title)
-    end
-
-    links.join("\n\n")
+  def presented_unsubscribe_link(subscription)
+    UnsubscribeLinkPresenter.call(
+      uuid: subscription.uuid,
+      title: subscription.subscriber_list.title,
+    )
   end
 end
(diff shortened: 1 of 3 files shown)


kevindew:
Wasn't the intention here to have links to each subscription that triggered this email? i.e. if you were subscribed to environment and population and a population email was sent you'd have links to both to allow unsubscribing?

Does look like this was implemented with a bug given this: https://github.com/alphagov/email-alert-api/pull/392/files#diff-82ce03e7a7d42444a4612921816ee07eL55 but just wonder if this solution impedes on a product decision

(for what it's worth I'd be a fan of just a single unsubscribe link as agree multiple is confusing)

rubenarakelyan:
Yeah, the intent is to fix the bug that shows links for everything a user is subscribed to, even if most of those have nothing to do with the email being received. But I don't see how we can choose only one link if the email was triggered as a result of multiple matches - is there something I'm misunderstanding here?

thomasleese:
Having looked through the code, I don't think we currently do anything to deduplicate emails such that a user will only get one email if a content change matches with more than one of their subscriptions. 

It feels out of scope for this PR to fix that (since this was just meant to change the fact that we're seeing emails with hundreds of unsubscribe links in them), so I think we should do an investigation in a separate card. I reckon it's best to, for the moment, leave it as only showing one unsubscribe link for each email as then at least the users will be able to see clearly which subscription caused which email if they receive multiple emails about the same content change.

Does that sound okay?

rubenarakelyan:
Got it, yeah that sounds good. We should definitely not forget about de-duplicating immediate emails too!

app/builders/immediate_email_builder.rb
```diff
@@ -15,47 +15,56 @@ def call
...
   end
 
-  def single_email_record(subscriber_content_change)
-    subscriber = subscriber_content_change.fetch(:subscriber)
-    content_change = subscriber_content_change.fetch(:content_change)
+  def single_email_record(subscription_content_change)
+    content_change = subscription_content_change.fetch(:content_change)
+    subscription = subscription_content_change[:subscription]
+
+    subscriber = if subscription
```
bevanloon: When might we not have a `subscription` in the `subscription_content_change`?

Also, I wonder if/whether it's a good idea to take advantage of passing a default to `.fetch` rather than the `if` statement? Maybe it would look a bit unwieldy given the length of `subscription_content_change`.

thomasleese: If we're sending courtesy emails, we don't have a subscription available, only a subscriber (the courtesy email address subscriber).

thomasleese:
@bevanloon Yeah, the commit message is a bit unclear, I will reword. But basically, we need to be able to specify the subscriptions on the subscriber and the subscriber on the subscriptions, and we can't do that with doubles as they both depend on each other.

app/builders/immediate_email_builder.rb
thomasleese: The reason I chose to not use `fetch` with a default argument is that it means a user could override the `subscriber` if they pass a hash that includes both a `subscription` and a `subscriber` which doesn't feel like something that should be possible to do.

I suppose we could catch that and raise an exception?

bevanloon: that sounds sensible enough. I would say leave it as it is then 👍

PR #394, opened 2018-02-01
Add `Frequency` class

gpeng:
Both `DigestRun` and `Subscription` have frequency enums. The string values for these were defined as constants on `DigestRun`. This commit abstracts them out into their own `Frequency` class to reuse everywhere.

Also updates the factories that have frequency traits and updates some specs to use the traits rather than specifying the frequency directly.

### DIFF
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -8,9 +8,6 @@ class DigestRun < ApplicationRecord
 
   enum range: { daily: 0, weekly: 1 }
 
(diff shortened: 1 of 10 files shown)


gpeng:
Ha ha... I knew you'd say that about the `module` bit. I decided it didn't really matter. Models seemed like the most appropriate place. It models frequency, just doesn't happen to store anything in the DB.

kevindew:
@gpeng Sounds like I need some new material!

PR #399, opened 2018-02-02
Use import with columns and records

thomasleese:
This is the most performant method of bulk insertion of records and it will use the least memory.

[Trello Card](https://trello.com/c/HTzGEXG7/569-investigate-increased-memory-usage-for-email-alert-api)

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -8,7 +8,7 @@ class ImmediateEmailBuilder
   end
 
   def call
-    Email.import!(records)
+    Email.import!(columns, records)
   end
 
   private_class_method :new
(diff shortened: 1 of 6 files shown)


thomasleese:
With a quick comparison locally of inserting 10,000 emails (I tested it a few times and got the same sorts of results each time):

## Hash method

```ruby
puts Benchmark.measure { Email.import!(hash_records) }
1.340000   0.000000   1.340000 (  1.507943)
```

## Array method

```ruby
puts Benchmark.measure { Email.import!(columns, array_records) }
1.240000   0.020000   1.260000 (  1.447746)
```

I agree it's not an enormous difference, but it helps to keep all our imports using the same method and I don't think there's a reason not to.

PR #404, opened 2018-02-05
Add daily digest feature spec

gpeng:
This PR adds a high level feature spec for the daily digest functionality.

I've intentionally left it 'long and procedural' as I think it reads better as a spec without jumping between method definitions where possible. Obviously, open to debate.

I've also had to dip into the DB at one point to retrieve the subscriptions to allow checking of unsubscribe links as we don't currently have an endpoint for that.  We should be adding one soon so I've left it as a TODO.

That does throw up whether this test is going into too much detail at this level. We could just check the right number of emails are sent to the right people here and go into more detail of the email content in an integration test. Happy to take that approach if we prefer it but I think this approach exercises a lot of the system, still runs quickly and gives a nice readable 'narrative' of what is supposed to happen.

Once we've agreed on this I'll add a weekly version in a subsequent PR.

[Trello](https://trello.com/c/b6ZxunyQ/589-create-high-level-test-that-creates-and-delivers-digests)

### DIFF
--- a/spec/features/create_and_deliver_digests_spec.rb
+++ b/spec/features/create_and_deliver_digests_spec.rb
@@ -104,6 +104,9 @@ RSpec.describe "creating and delivering digests", type: :request do
     subscriber_one_address = "test-one@example.com"
     subscriber_two_address = "test-two@example.com"
 
+    #subscriber that shouldn't receive a digest
+    non_digest_subscriber_address = "test-three@example.com"
+
     subscribe_to_subscribable(
       subscribable_one_id,
       address: subscriber_one_address,
@@ -120,6 +123,12 @@ RSpec.describe "creating and delivering digests", type: :request do
       subscribable_one_id,
       address: subscriber_two_address,
       frequency: Frequency::DAILY
+    )
+
+    subscribe_to_subscribable(
+      subscribable_one_id,
+      address: non_digest_subscriber_address,
+      frequency: Frequency::IMMEDIATELY
     )
 
     #publish two items to each list
--- a/spec/features/create_and_deliver_digests_spec.rb
+++ b/spec/features/create_and_deliver_digests_spec.rb
@@ -1,0 +1,195 @@
+require "rails_helper"
+
+RSpec.describe "creating and delivering digests", type: :request do
+  before do
+    Timecop.freeze "2017-01-02 10:00"
+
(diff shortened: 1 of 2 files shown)


spec/features/create_and_deliver_digests_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+      topics: [list_one_topic_id]
+    })
+
+    stub_govdelivery("TEST456")
+    create_subscribable(title: "Subscriber list two", links: {
+      topics: [list_two_topic_id]
+    })
+
+    lookup_subscribable("TEST123")
+    subscribable_one_id = JSON.parse(response.body)["subscribable"]["id"]
```
thomasleese: Could we make `lookup_subscribable` return the id so this line wasn't necessary?

spec/features/create_and_deliver_digests_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+        description: "Description one",
+        change_note: "Change note one",
+        public_updated_at: "2017-01-01 10:00:00",
+        links: {
+          topics: [list_one_topic_id]
+        }
+      )
+    end
+
+    Timecop.freeze "2017-01-01 10:00" do
```
thomasleese: The date here doesn't match the given `public_updated_at`, not sure if that matters though.

spec/features/create_and_deliver_digests_spec.rb
gpeng: :+1: I've added that.

spec/features/create_and_deliver_digests_spec.rb
gpeng: Fixed in  208d127

bevanloon:
This looks pretty good to me. 

I was wondering whether we need/care about testing a "subscription does not match" in this test or are we only trying to test out the happy path of call the DigestWorker and get an email. I think we have some unit tests around that so maybe it isn't really what we should be testing here?

I like the overt obviousness of the code for now. I might be tempted to move the actual email body stuff into some helper methods just because I think it's a little distracting to have giant heredocs in the middle. Feel free to ignore though.

gpeng:
Thanks. @bevanloon I've added a couple of extra commits implementing your suggestions if you could have another look.

PR #408, opened 2018-02-05
Add weekly digest feature spec

gpeng:
Follows on from https://github.com/alphagov/email-alert-api/pull/404 to add a feature spec that creates and sends a weekly digest email to two subscribers.

[Trello](https://trello.com/c/b6ZxunyQ/589-create-high-level-test-that-creates-and-delivers-digests)

### DIFF
--- a/spec/features/create_and_deliver_digests_spec.rb
+++ b/spec/features/create_and_deliver_digests_spec.rb
@@ -217,4 +217,208 @@ RSpec.describe "creating and delivering digests", type: :request do
     expect(first_digest_stub).to have_been_requested
     expect(second_digest_stub).to have_been_requested
   end
+
+  def first_expected_weekly_email_body(subscription_one, subscription_two)
+    <<~BODY
+      #Subscriber list one
+
+      [Title one](http://www.dev.gov.uk/base-path)
+
+      Change note one: Description one
+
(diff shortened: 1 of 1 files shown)


spec/features/create_and_deliver_digests_spec.rb
```diff
@@ -217,4 +217,208 @@ def second_expected_daily_email_body(subscription)
...
+
+    subscribe_to_subscribable(
+      subscribable_two_id,
+      address: non_weekly_digest_subscriber_address,
+      frequency: Frequency::DAILY
+    )
+
+    #publish two items to each list
+    Timecop.freeze "2016-12-28 10:00" do
+      create_content_change(
```
thomasleese: I wonder if we could make these into methods used across the daily and weekly digests? `create_content_change_1`?

I like that they're separate in each test right now, but it's also a lot of code duplication. What do you think?

gpeng: I don't mind duplication within the tests. The other issue is that the dates are different in the daily and the weekly tests so we'd have to pass them in if we extract a method. Happy to do that but I prefer it as it is.

thomasleese: Yeah, that's fine with me. I agree the code duplication is not too much of an issue.

PR #409, opened 2018-02-05
ENV vars for handling transition to notify

kevindew:
Trello: https://trello.com/c/CDYgpQDY/565-add-flags-to-enable-migration-from-govdelivery-to-notify

This adds a new ENV var (`DISABLE_GOVDELIVERY_EMAILS`) and documents an existing one (`USE_EMAIL_ALERT_FRONTEND_FOR_EMAIL_COLLECTION`) which will be used to transition to notify.

Follow up PRs are coming to:
- Set up these env vars in puppet
- Create a means to whitelist some people to receive notify emails during the transition but others don't (so we can have an imported database, but not be emailing them all)

### DIFF
--- a/doc/transition-from-govdelivery-to-notify.md
+++ b/doc/transition-from-govdelivery-to-notify.md
@@ -1,0 +1,18 @@
+# Transition from Govdelivery to Notify
+
+This document serves as documentation for the transition from Govdelivery to
+Notify as the means to send emails. The intention is that this stores
+information related to the migration and this app close to the codebase.
+It will be safe to remove this once the migration has been completed.
+
+## ENV vars
+
+### `DISABLE_GOVDELIVERY_EMAILS`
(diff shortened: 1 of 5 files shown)


doc/transition-from-govdelivery-to-notify.md
```diff
@@ -0,0 +1,26 @@
+# Transition from Govdelivery to Notify
+
+This document serves as documentation for the transition from Govdelivery to
+Notify as the means to send emails. The intention is that this stores
+information related to the migration and this app close to the codebase.
+It will be safe to remove this once the migration has been completed.
+
```
gpeng: An extra line has sneaked in 😱

kevindew: Whoops! thanks @gpeng

PR #410, opened 2018-02-05
Soft delete subscriptions

gpeng:
We currently 'hard' delete subscription records when a user unsubscribes to comply with our data retention policy. 

During the forthcoming migration between providers we don't want to do this as we will lose data should we need to rollback.

This PR implements 'soft deletion' on `Subscription` to preserve the data. Once we have successfully completed the migration we can remove the records and revert this PR.

[Trello](https://trello.com/c/aslPwUf1/566-temporarily-change-unsubscribes-to-soft-delete-only)

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -7,6 +7,7 @@ class SubscriptionsController < ApplicationController
 
     status = subscription.new_record? ? :created : :ok
 
+    subscription.deleted_at = nil
     subscription.frequency = frequency
     subscription.signon_user_uid = current_user.uid
     subscription.save!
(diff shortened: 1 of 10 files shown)


thomasleese:
I'm just wondering if we should deploy this closer to when we go live, since we're dealing with lots of fake data at the moment, it might be annoying if that stuff ends up hanging around when we think we're deleting it.

kevindew:
@thomasleese Yeah we have a growing situation around data and removal - cascade/analytic concerns, growing pile of emails in the db. I think rather than delay this we should try confront this soon. Maybe have a dev chat this week on the matters to go into work on it for next sprint.

gpeng:
Good point... I've marked this DNM for now and we can merge it as part of the switcheroo.

gpeng:
Discussed offline... We're merging this now and handling the soft deleted stuff to reduce the amount of new things we introduce during the migration.

PR #414, opened 2018-02-06
Strip markdown

gpeng:
We don't want to allow markdown through to Notify from publishing app change notes or descriptions. This PR utilises the redcarpet gem to strip markdown within the content change presenter.

The gem seems to add a `\n` at the end of its output which I've also stripped here.

[Trello](https://trello.com/c/fHzSedir/567-strip-markdown-from-incoming-content)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem 'govuk_app_config', '~> 1.2'
 gem 'nokogiri', '~> 1.8'
 gem 'notifications-ruby-client', '~> 2.6'
 gem 'plek', '~> 2.1'
+gem 'redcarpet', '~> 3.4'
 
 gem 'govuk_sidekiq', '~> 3.0'
 gem 'ratelimit', '~> 1.0'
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -14,6 +14,7 @@ gem 'govuk_app_config', '~> 1.2'
 gem 'nokogiri', '~> 1.8'
 gem 'notifications-ruby-client', '~> 2.6'
 gem 'plek', '~> 2.1'
+gem 'redcarpet'
```
thomasleese: Should we specify a version here? `~> 3.4`?

gpeng: I was going to do that but didn't we decide pinning was pointless in our new dependabot world?

thomasleese: Yeah, it is. Just makes it inconsistent with the others... I'm fine either way :man_shrugging:

gpeng: Cool... I think I'll leave it. We should unpin as we upgrade versions I suppose if that's what we have decided to do.

kevindew: Still useful for bundle updates, dependabot will also maintain the semvar pinning too.

gpeng: Agh... I'll pin it then.

PR #417, opened 2018-02-08
Remove load testing and fake data code

thomasleese:
This PR removes the old load testing and fake data code.

They're both no longer being actively used and therefore can very quickly become out of date. If a user wanted to run a load test or import fake data again they would likely have to make some code changes, and reverting this PR would only be a very small part of that.


thomasleese:
🎉

screen shot 2018-02-08 at 12 33 49

PR #420, opened 2018-02-09
Ignore duplicate content changes

gpeng:
In the event that `notifications#create` times out the client apps may retry the request despite it having already made it through.

This PR returns a conflict if there is already a `ContentChange` stored with the same `base_path`, `content_id` and `public_updated_at`. Assuming we agree these are a reasonable indicator of duplicate-ish-ness we should add this as a validation on `ContentChange` and a DB constraint but while we need to also prevent Govdelivery calls this is easier to handle in the controller for now. 

We'll also need to clean up the data before adding a DB constraint as there are already duplicates in there.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -1,5 +1,7 @@
 class NotificationsController < ApplicationController
   def create
+    return render_conflict if content_change_exists?
+
     NotificationWorker.perform_async(notification_params)
 
     NotificationHandlerService.call(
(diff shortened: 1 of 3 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -1,5 +1,7 @@
 class NotificationsController < ApplicationController
   def create
+    render_conflict && return if content_change_exists?
```
kevindew: might be cleaner as `return render_conflict if content_change_exists?`

kevindew:
Probably already on your radar but we'll need to make sure email-alert-service can handle this gracefully.

PR #422, opened 2018-02-12
Speed up GovDelivery import task

thomasleese:
This PR reworks the GovDelivery CSV import task with the intention of speeding it up but has ended up changing how it works. Some of the key changes include:

- Default to daily frequency when the digest data is unavailable (this may be contentious, but there are 100k subscription entries missing in the digest data and some of those include people who are subscribed to hundreds of subscriptions)
- Import the data in bulk (this means it's no longer possible to top up the data later)
- Topics that doesn't exist get ignored (since they won't be receiving emails anyway as the topic doesn't exist in production)

In terms of review, it's probably easier not to consider what's changed, but to consider the final code of the importer.

Running the dry run should help to identify any problems with this process and we can iterate on it in the future.

Importing all of the data onto my dev VM took over 12 hours before, and it took just under an hour after running with this.

[Trello Card](https://trello.com/c/Qe1wUlW8/603-speed-up-data-import)

### DIFF
--- a/lib/import_govdelivery_csv.rb
+++ b/lib/import_govdelivery_csv.rb
@@ -25,6 +25,17 @@ private
 
   attr_reader :subscriptions_csv_path, :digests_csv_path, :fake_import
 
+  def address_from_row(row)
+    address = row.fetch("DESTINATION")
+
+    if fake_import
+      hashed_address = Digest::SHA1.hexdigest(address)
+      "success+#{hashed_address}@simulator.amazonses.com"
+    else
+      address
+    end
+  end
+
   def import_row(row)
     subscriber = find_or_create_subscriber(row)
     subscribable = find_subscribable(row)
@@ -36,15 +47,7 @@ private
   end
 
   def find_or_create_subscriber(row)
-    Subscriber.find_or_create_by!(address: fetch_address_for_row(row))
-  end
-
-  def fetch_address_for_row(row)
-    address = row.fetch("DESTINATION")
-
-    return "success+#{Digest::SHA1.hexdigest(address)}@simulator.amazonses.com" if fake_import
-
-    address
+    Subscriber.find_or_create_by!(address: address_from_row(row))
   end
 
   def find_subscribable(row)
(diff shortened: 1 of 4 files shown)


lib/import_govdelivery_csv.rb
```diff
@@ -53,56 +49,89 @@ def import_subscribers
...
   def subscriber_for_row(row)
-    Subscriber.find_by!(address: address_from_row(row))
+    address = address_from_row(row)
+    all_subscribers.fetch(address)
+  end
+
+  def all_subscribables
+    @all_subscribables ||= begin
+      puts "Loading all subscribables from database..."
+      SubscriberList.all.each_with_object({}) do |subscribable, hash|
```
kevindew: Not a biggie but I think you can do this with `SubscriberList.all.index_by(&:gov_delivery_id)`

lib/import_govdelivery_csv.rb
```diff
@@ -53,56 +49,89 @@ def import_subscribers
...
+
+    puts "#{count} subscriptions imported!"
 
-  def find_or_create_subscription(subscriber, subscribable, frequency)
-    subscriber.subscriptions.find_or_create_by!(
-      subscriber_list: subscribable,
-      frequency: frequency,
-    )
+    puts "Unable to match #{@failed_topics.count} topics:"
+    p @failed_topics
```
kevindew: might want to attr_reader this for consistency.

lib/import_govdelivery_csv.rb
```diff
@@ -104,7 +104,7 @@ def import_subscriptions
 
         next unless subscribable
 
-        frequency = digest_frequencies.fetch(subscriber.address, Frequency::IMMEDIATELY)
+        frequency = digest_frequencies.fetch(subscriber.address, Frequency::DAILY)
```
kevindew: Seems reasonable, might cause surprises so should probably flag this with Martin if not already done so.

lib/import_govdelivery_csv.rb
```diff
@@ -98,7 +98,9 @@ def import_subscriptions
 
     records = CSV
       .foreach(subscriptions_csv_path, headers: true, encoding: "WINDOWS-1252")
-      .map do |row|
+      .each_with_index.map do |row, i|
```
kevindew: might want to pass in an argument of 1 to each_with_index (as offset) as I think it might all end up being off by one. Not really a biggie mind.

PR #423, opened 2018-02-12
Speed up status update handling

thomasleese:
Currently a normal status update will require two queries, one lookup and one update. Instead we can rework this so that only special status updates (permenant failures, ones that happen a lot less often) require the lookup and update query separately.

Also we've added an index on the `reference` field on delivery attempts since we'll be using them a lot in the status updates.

### DIFF
--- a/app/services/status_update_service.rb
+++ b/app/services/status_update_service.rb
@@ -10,14 +10,16 @@ class StatusUpdateService
   end
 
   def call
-    delivery_attempt.update!(
+    updates = DeliveryAttempt.where(reference: reference).update_all(
       status: status.underscore,
       signon_user_uid: user&.uid,
     )
 
-    if delivery_attempt.permanent_failure? && subscriber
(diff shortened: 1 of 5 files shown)


app/services/status_update_service.rb
```diff
@@ -33,12 +35,16 @@ def call
...
+    status == "permanent-failure"
+  end
+
+  def temporary_failure?
+    status == "temporary-failure"
   end
 
   def email
-    @email ||= delivery_attempt.email
+    @email ||= DeliveryAttempt.find_by!(reference: reference).email
```
kevindew: This seems to unnecessarily initialise delivery attempt. Can we do it more like `Email.joins(:delivery_attempts).find_by("delivery_attemps.reference": reference)`?

thomasleese: Yep, that's a lot better. https://github.com/alphagov/email-alert-api/pull/424

PR #426, opened 2018-02-13
Update email presentation

keithiopia:
These small changes bring the immediate and digest emails more into line 
with the examples we've got mocked up in Notify. 
Comes from: https://trello.com/c/aTBCNJTb/604-iterate-the-templated-content-sent-in-emails

### DIFF
--- a/spec/features/create_and_deliver_digests_spec.rb
+++ b/spec/features/create_and_deliver_digests_spec.rb
@@ -18,17 +18,17 @@ RSpec.describe "creating and delivering digests", type: :request do
 
       [Title one](http://www.dev.gov.uk/base-path)
 
-      Change note one: Description one
+      Description one
 
-      Updated at 10:00 am on 1 January 2017
+      10:00 am on 1 January 2017: Change note one
 
       ---
 
       [Title two](http://www.dev.gov.uk/base-path)
 
-      Change note two: Description two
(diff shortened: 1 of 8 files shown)


spec/features/create_and_deliver_digests_spec.rb
```diff
@@ -18,17 +18,17 @@ def first_expected_daily_email_body(subscription_one, subscription_two)
...
 
       ---
 
       [Title two](http://www.dev.gov.uk/base-path)
 
-      Change note two: Description two
+      Description two
 
-      Updated at 09:00 am on 1 January 2017
+       9:00am, 1 January 2017: Change note two
```
thomasleese: I suspect it doesn't matter when the email gets rendered, but should we strip the date output to avoid this leading space?

keithiopia: Stupid Ruby. Why would it replace a digit with a space?? If it doesn't render in the email then I guess it doesn't matter.

bevanloon: I'll leave this up to Keith. I'm not bothered either way.

bevanloon:
@thomasleese I've tidied up the commits a little - would you mind casting your eye over it again when you get a moment please?

PR #428, opened 2018-02-14
ADR 003 - Initial data retention strategy

kevindew:
https://trello.com/c/fbx5sg0k/595-implement-a-data-retention-policy-for-email-data

This builds on the spike from @gpeng and discussions with various team members about use cases/concerns/feasibility of data retention.

Hopefully it's all a sound plan and it's just my grammar that fails me - but very happy to debate bits of it that may seem problematic.

[Pretty rendered version](https://github.com/alphagov/email-alert-api/blob/60064d2b3c27d278255c0584aa2747c9ed1d9a36/doc/arch/adr-003-data-retention.md)

### DIFF
--- a/doc/arch/adr-003-data-retention.md
+++ b/doc/arch/adr-003-data-retention.md
@@ -150,6 +150,11 @@ Both ContentChange and SubscriptionList should not require changes to them in
 order to allow us to build up a full analytics picture. However they would
 require an intention not to delete either of them.
 
+Despite this a small change is suggested for ContentChange to switch the
+primary key type from being an auto-incrementing number to a uuid. This is to
+allow this to be sent out as part of analytics tracking without revealing state
+or ordering information of the underlying system.
+
 As a SubscriberList can be updated the only way we can determine the historical
 number of ContentChanges with accuracy is to have the ContentChanges
 available (to perform the look up between them).
@@ -163,6 +168,7 @@ SubscriptionLists don't currently have the ability to be deleted by apps, and
 it's likely if they gained that ability that would need to be some form of
 archiving so it could still be communicated to users that the list they had a
 subscription to is no longer available.
+
 
 #### Subscriber and Subscription
 
@@ -236,6 +242,7 @@ numerical id.
 
 #### ContentChange
 
+- Primary key should change to a uuid
 - Should not be deleted, as part of analytics picture
 
 #### SubscriberList
--- a/doc/arch/adr-003-data-retention.md
+++ b/doc/arch/adr-003-data-retention.md
@@ -1,0 +1,287 @@
+# Decision Record: Initial Data Retention Strategy
+
+## Introduction
+
+In February 2018, prior to switching from Govdelivery to Notify we wished to
+consider our approach to data retention in such a way to enable data driven
+decisions in the future.
+
+The aim in this work is to make suggestions that enable a comparable amount of
+data to what is currently used by Govdelivery and for that data to be
+flexible. This is balanced against the constraint that Email Alert API does
+produce a lot of data and this should be culled regularly to not use too much
+disk space without justification. It also intends to do this with respect to
+the constraints of GDPR by not storing personally identifiable data for
+longer than is needed.
+
+The goal being that we can complete this prior to the change to Notify so that
+we avoid a data black hole where we do not have the data to answer common
+data queries.
+
+## Specifics
+
+### Common data use scenarios
+
+Based on a chat with Paul (our performance analyst) we were demonstrated to a
+number of scenarios of data look ups and usage. As this was only a short session
+this list is likely incomplete.
+
+#### Data that is currently looked up
+
+- Number of people subscribed to list
+  - Given a date range: how many people subscribed in that range; how many
+    unscribed; total number of subscribers during that time period.
+- Activity Level of list
+  - Given a date range: how many emails were sent out to that list; how many
+    bulletins a list has associated with it; how many emails a list has
+    genertaed; whether lists have never produced an email.
+- Overall totals
+  - Counts of how many emails we have sent over a specific period of time or
+    total (mostly used internally)
+
+#### Support queries
+
+- People have been unsubscribed and don’t know why (Govdelivery tracks a user
+  unsubscribed but not which lists)
+- Why a user was unsubscribed - this can happen automatically or a user
+  performs the action.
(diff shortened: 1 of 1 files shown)


doc/arch/adr-003-data-retention.md
```diff
@@ -0,0 +1,287 @@
...
+for us to have a primary key that is the uuid and then not have a second
+numerical id.
+
+### Summary of database changes
+
+#### Email
+
+- Change the id type from a serial to a uuid
+- Add an archived_at timestamp
+- Would be removed from database after 14 days
```
thomasleese: I suspect we're going to need to look into increasing disk space for 14 days, since we already had to clear out the table within about a week of running at 70%.

doc/arch/adr-003-data-retention.md
```diff
@@ -0,0 +1,287 @@
...
+The first of these processes is an archiving one which should happen within
+24 hours of an email reaching a final state (either sent or failed). When this
+has occurred the row should be marked that as archived (suggestion is for a
+`archived_at` timestamp) so it not as archived again and is known to be safe
+to delete later.
+
+The archiving would not involve any personally identifiable data and would
+only include the subject field from the email. As an example:
+
+```
```
thomasleese: Small thing, but might be nice to mark this as JSON for the syntax highlighting... or maybe the comments make it look ugly?

doc/arch/adr-003-data-retention.md
kevindew: Ah sure - do you reckon there is a time period that is acceptable where we don't have to up the disk space? As we could start at that and then increase that time if we learn from support queries we need to store them for longer.

If not, we probably should get a card in to up the disk space regardless.

thomasleese: To be honest, I think we should get a card in about increasing disk space because at the moment I imagine we'll have to start deleting emails only after about 3 days.

doc/arch/adr-003-data-retention.md
kevindew: Yeah just tried it and it looked minging sadly. It always makes me sad how JSON doesn't allow comments. I will however provide a rendered link to this ADR in the description

doc/arch/adr-003-data-retention.md
kevindew: Cool I'll get a card in - I think we usually go up by about 64 GB that seem about reasonable?

thomasleese: Seems reasonable to me!

doc/arch/adr-003-data-retention.md
```diff
@@ -0,0 +1,287 @@
...
+much firmer understanding of the analytics usage and needs.
+
+SubscriptionLists don't currently have the ability to be deleted by apps, and
+it's likely if they gained that ability that would need to be some form of
+archiving so it could still be communicated to users that the list they had a
+subscription to is no longer available.
+
+#### Subscriber and Subscription
+
+Currently when a subscription is ended or we discover an email address is not
```
gpeng: We don't actually delete the subscriber currently I don't think, we just nullify their address. We're also currently only 'soft deleting' the subscriptions which we intended to remove but it sounds like we probably want to keep that. So... I don't think we need some of the changes proposed here.

doc/arch/adr-003-data-retention.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### ContentChange
+
+- Should not be deleted, as part of analytics picture
+
+#### SubscriberList
+
+- Should not be deleted, as part of analytics picture
+
+#### Subscriber
```
gpeng: As above... I think we already have this pretty much. We could probably infer `deactivated_at` from the soft deleted subscriptions but depending on how we're using it it could be easier to have it on this table as you suggest.

doc/arch/adr-003-data-retention.md
```diff
@@ -0,0 +1,287 @@
...
+
+- These should not be deleted
+- Add a `deactivated_at` timestamp field
+- Support a null value for address
+- Should not be deleted, as part of analytics picture
+
+#### Subscription
```
gpeng: What's the distinction between `ended_at` and `active` and are either (or both) of these already represented by `deleted_at`?

doc/arch/adr-003-data-retention.md
```diff
@@ -0,0 +1,287 @@
...
+or at least until we determine an approach to archiving them.
+
+#### ContentChange and SubscriptionList
+
+Both ContentChange and SubscriptionList should not require changes to them in
+order to allow us to build up a full analytics picture. However they would
+require an intention not to delete either of them.
+
+As a SubscriberList can be updated the only way we can determine the historical
+number of ContentChanges with accuracty is to have the ContentChanges
```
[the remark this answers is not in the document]
kevindew: Thanks 👍

doc/arch/adr-003-data-retention.md
```diff
@@ -0,0 +1,287 @@
...
+As DigestRuns are low on data usage they can effectively be stored indefinitely
+or at least until we determine an approach to archiving them.
+
+#### ContentChange and SubscriptionList
+
+Both ContentChange and SubscriptionList should not require changes to them in
+order to allow us to build up a full analytics picture. However they would
+require an intention not to delete either of them.
+
+As a SubscriberList can be updated the only way we can determine the historical
```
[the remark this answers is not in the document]
kevindew: Yeah, that seems an essential progression for user experience. I'll update the text.

doc/arch/adr-003-data-retention.md
kevindew: Active would be for the purpose of a unique constraint to prevent the same subscriber to have the same subscription multiple times. 

Ended_at serves same purpose as deleted_at but was just an attempt to make the temporary more permanent. I haven't phrased it as that though I kinda just ignored deleted_at. I'll rephrase that as a name change since `deleted_at` might be a bit confusing if we're keeping it around forever.

doc/arch/adr-003-data-retention.md
kevindew: Yup you're right. I'll adjust the text.

kevindew:
@gpeng As I'm kind I did the amends as a separate commit so you don't have to read through the whole thing again. Let me know what you think.

PR #429, opened 2018-02-15
Deduplicate immediate emails

gpeng:
Where a subscriber is subscribed to multiple lists that match a content change we are currently sending them multiple emails.

e.g

```
subscriber subscribes to "All publications"
same subscriber subscribes to "Publications from DfE"

DfE publishes a publication

subscriber receives two emails
```

~~We were using a query that returned `SubscriptionContent` objects that have nil email_id and then in a batch operation, creating `Email` objects for each and then building a SQL string from the returned email ids that updates the `SubscriptionContent` records in another single command.~~

~~This commit changes this to use a query that returns `Subscribers` and create one `Email` per `ContentChange` per subscriber. There is some complication as we then need to update all of the 
`SubscriptionContent` records with a single `Email#id`. This complexity seems unavoidable on account of us having to batch the processes to achieve the performance we need.~~

UPDATE: 17:15 19/2/2018 Totally reworked this based on discussions. 

We were using a query that returned `SubscriptionContent` objects that
have nil email_id and then in a batch operation, create `Email` objects
for each and then build a SQL from the returned email ids that updates
the `SubscriptionContent`records in another single command.

This commit changes this to use a query that returns `Subscribers` that
require an email and a separate query to return `SubscriptionContent`
grouped by `ContentChange`. These queries have been separated to avoid
loading all `Subscription` per `Subscriber` when trying to eager load
them. We then create one `Email` per `ContentChange`. There is some
further complication as we then need to update all of the `SubscriptionContent`
records with a single `Email#id`. This complexity seems unavoidable on account
of us having to batch the processes to achieve the performance we need.

To reduce the memory that is consumed during the process (hopefully, not tested yet but the existing implementation is struggling for RAM) this commit also introduces caching of the `ContentChange` objects. Previously we loaded a `ContentChange` on each `SubscriptionContent` when in reality
we are probably only dealing with a small number of them (or probably one).

This significantly changes the immediate email generation process so we'll need to load test again before merging.

[Trello](https://trello.com/c/bAiuj8JZ/601-de-duplicate-immediate-emails)

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -1,6 +1,6 @@
 class ImmediateEmailBuilder
-  def initialize(subscription_content_changes)
-    @subscription_content_changes = subscription_content_changes
+  def initialize(recipients_and_content)
+    @recipients_and_content = recipients_and_content
   end
 
   def self.call(*args)
@@ -15,11 +15,15 @@ class ImmediateEmailBuilder
 
 private
 
-  attr_reader :subscription_content_changes
+  attr_reader :recipients_and_content
 
   def records
-    subscription_content_changes.map do |subscription_content_change|
-      single_email_record(subscription_content_change)
+    recipients_and_content.map do |recipient_and_content|
+      [
+        recipient_and_content.fetch(:address),
+        subject(recipient_and_content.fetch(:content_change)),
+        body(recipient_and_content.fetch(:content_change), recipient_and_content.fetch(:subscriptions))
+      ]
     end
   end
 
@@ -27,33 +31,20 @@ private
     %i(address subject body)
   end
 
-  def single_email_record(subscription_content_change)
-    content_change = subscription_content_change.fetch(:content_change)
-    subscription = subscription_content_change[:subscription]
-
-    subscriber = if subscription
-                   subscription.subscriber
-                 else
-                   subscription_content_change.fetch(:subscriber)
-                 end
-
-    [subscriber.address, subject(content_change), body(content_change, subscription)]
-  end
-
   def subject(content_change)
     "GOV.UK update - #{content_change.title}"
   end
 
-  def body(content_change, subscription)
-    if subscription
+  def body(content_change, subscriptions)
+    if Array(subscriptions).empty?
+      presented_content_change(content_change)
+    else
       <<~BODY
         #{presented_content_change(content_change)}
         ---
 
-        #{presented_unsubscribe_link(subscription)}
+        #{presented_unsubscribe_links(subscriptions)}
       BODY
-    else
-      presented_content_change(content_change)
     end
   end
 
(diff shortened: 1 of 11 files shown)


app/builders/immediate_email_builder.rb
```diff
@@ -15,56 +15,51 @@ def call
...
 
-  attr_reader :subscription_content_changes
+  attr_reader :params
 
   def records
-    subscription_content_changes.map do |subscription_content_change|
-      single_email_record(subscription_content_change)
+    params.map do |param|
+      [
+        param[:address],
```
thomasleese: I would use `.fetch` on these lines to ensure the key exists.

app/workers/immediate_email_generation_worker.rb
```diff
@@ -60,22 +70,43 @@ def queue_for_priority(priority)
...
+    #   content_change => [subscription_content]
+    # }
+    #
+    # which maps to one key per email created
+    #
+    subscribers.inject({}) do |params, subscriber|
+      params.merge!(
+        subscriber.unprocessed_subscription_contents.group_by(&:content_change)
+      )
+      params
```
thomasleese: I suspect you don't need this as the `merge!` will return the hash.

app/builders/immediate_email_builder.rb
gpeng: Fixed in c7acb34

app/workers/immediate_email_generation_worker.rb
gpeng: Fixed in c7acb34

app/queries/subscribers_for_immediate_email_query.rb
```diff
@@ -0,0 +1,10 @@
+class SubscribersForImmediateEmailQuery
+  def self.call
+    Subscriber
+      .joins(subscriptions: { subscription_contents: :content_change })
+      .includes(subscriptions: { unprocessed_subscription_contents: :content_change })
+      .where(subscriptions: { subscription_contents: { email_id: nil } })
+      .where.not(address: nil)
+      .distinct
```
kevindew: I had a play with my pointless query non-optimisation.

```
Subscriber.where(
  SubscriptionContent
    .joins(:subscription)
    .where(email_id: nil)
    .where("subscriptions.subscriber_id = subscribers.id")
    .exists
).count
```

app/queries/subscribers_for_immediate_email_query.rb
```diff
@@ -0,0 +1,13 @@
+class SubscribersForImmediateEmailQuery
+  def self.call
+    Subscriber
+      .where(
+        SubscriptionContent
```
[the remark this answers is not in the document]
gpeng: Fixed in 36f8551

app/builders/immediate_email_builder.rb
```diff
@@ -1,6 +1,6 @@
 class ImmediateEmailBuilder
-  def initialize(subscription_content_changes)
-    @subscription_content_changes = subscription_content_changes
+  def initialize(params)
```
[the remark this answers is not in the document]
gpeng: Fixed in da29f68

app/queries/unprocessed_subscription_contents_by_subscriber_query.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    transform_results(subscription_contents)
+  end
+
+  private_class_method :new
+
+private
+
+  def transform_results(subscription_contents)
+    subscription_contents.inject({}) do |results, subscription_content|
```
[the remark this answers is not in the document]
gpeng: Fixed in c461b47

PR #438, opened 2018-02-16
Add digest_run_subscriber_id index

gpeng:
We don't join on this currently within the app but it's helpful in some diagnostic queries.

### DIFF
--- a/db/migrate/20180216100846_add_subscription_contents_index.rb
+++ b/db/migrate/20180216100846_add_subscription_contents_index.rb
@@ -1,0 +1,5 @@
+class AddSubscriptionContentsIndex < ActiveRecord::Migration[5.1]
+  def change
+    add_index :subscription_contents, :digest_run_subscriber_id
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20180216100846_add_subscription_contents_index.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubscriptionContentsIndex < ActiveRecord::Migration[5.1]
+  def change
+    add_index :subscription_contents, :digest_run_subscriber_id
```
thomasleese: Do you know how long this takes and if we have to add the index concurrently?

gpeng: No... But I do have a DB backup now so I'll find out.

gpeng: It was 0.1218s on my machine with 262751 `SubscriptionContent` records.

thomasleese: That seems fine to me.

PR #442, opened 2018-02-20
Various delivery request improvements

thomasleese:
This PR introduces a number of improvements to the delivery request service and delivery attempts.

Firstly, this makes the delivery attempt reference a UUID and stores it like that in the database. This should enable an ever so slightly performance improvement, and reduced disk storage size.

Secondly, this enables an option that allows only emails with an address in the whitelist to be sent, while all others are ignored. This allows us to reduce the number of emails being sent out in staging and integration.

[Trello Card](https://trello.com/c/Q7xZvdUN/630-only-send-emails-to-the-whitelist)

### DIFF
--- a/app/services/delivery_request_service.rb
+++ b/app/services/delivery_request_service.rb
@@ -22,27 +22,31 @@ class DeliveryRequestService
     reference = SecureRandom.uuid
     address = determine_address(email, reference)
 
-    MetricsService.email_send_request(provider_name) do
-      provider.call(
-        address: address,
-        subject: subject,
-        body: email.body,
+    return if address.nil?
+
+    begin
+      MetricsService.email_send_request(provider_name) do
+        provider.call(
+          address: address,
+          subject: subject,
+          body: email.body,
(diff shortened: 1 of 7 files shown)


app/services/delivery_request_service.rb
```diff
@@ -18,42 +18,39 @@ def self.call(*args)
...
-    reference = generate_reference(email)
+    reference = SecureRandom.uuid
     address = determine_address(email, reference)
 
-    MetricsService.email_send_request(provider_name) do
-      provider.call(
-        address: address,
-        subject: subject,
-        body: email.body,
+    return if address.nil?
```
bevanloon: This is interesting - previously, if we got here with a `nil` email, I think we would have ended up with an error in Sentry; Notify would have errd, we would have caught a the `ProviderError` and then attempted to write a DeliveryAttempt with a nil email, which would fail the validation on the model.  I'm not sure that was ever a conscious decision when we wrote the code or an accidental side effect.

PR #449, opened 2018-02-22
Store digest run subscribers count

thomasleese:
This stores the count of all the digest run subscribers created for a particular digest run as we our archive policy will be to delete old digest run subscribers over time and these keeps a useful statistic around.


thomasleese:
@kevindew Agreed... I've amended the PR!

PR #452, opened 2018-02-23
Add ability to ignore notify email addresses for testing

1pretz1:
Ignores notify email addresses containing '@notifications.service.gov.uk',
meaning no subsriber will be created but a successful sign up message will still
appear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,11 @@ and it will create a new subscription between the email address and the
 subscriber list. It will respond with a `201 Created` if it's a new
 subscription or a `200 OK` if the subscription already exists.
 
+### Using test email addresses for signup
+
+Using any email address that ends with '@notifications.service.gov.uk'
+will not create a subscriber or a subscription, however will return a `201 Created`.
+
 ### healthcheck API
 
 A queue health check endpoint is available at /healthcheck
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,5 +1,7 @@
 class SubscriptionsController < ApplicationController
   def create
+    return render json: { id: 0 }, status: :created if smoke_test_address?
+
     subscription = Subscription.find_or_initialize_by(
       subscriber: subscriber,
       subscriber_list: subscribable,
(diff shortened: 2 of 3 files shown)


1pretz1:
Also altered 'db/seeds.rb' as the original when running tests threw a syntax error.

app/controllers/subscriptions_controller.rb
```diff
@@ -1,5 +1,7 @@
 class SubscriptionsController < ApplicationController
   def create
+    return render json: {}, status: status if notify_address
```
kevindew: At this point status will be undefined (ruby probably hoists it to be a nil) probably be better if this was set to `:created`.

The json returned should also return the same response as the rest of the app as this would break any apps expecting an id in the response. I'd go with an id of 0.

app/controllers/subscriptions_controller.rb
```diff
@@ -11,15 +13,17 @@ def create
     subscription.frequency = frequency
     subscription.signon_user_uid = current_user.uid
     subscription.save!
-
     render json: { id: subscription.id }, status: status
   end
 
 private
 
+  def notify_address
```
kevindew: To follow ruby convention this method should end with a question mark

app/controllers/subscriptions_controller.rb
```diff
@@ -11,15 +13,17 @@ def create
...
+    address.include?("@notifications.service.gov.uk")
```
kevindew: Might be better to use `address.end_with?("@notifications.service.gov.uk")`

db/seeds.rb
```diff
@@ -1,9 +1,11 @@
...
-  name: "Test user",
-  permissions: %w[signin status_updates],
-  organisation_content_id: gds_organisation_id,
-)
+  User.create!(
+    name: "Test user",
+    permissions: %w[signin status_updates],
+    organisation_content_id: gds_organisation_id,
+  )
+end
```
kevindew: What's the error you get here? just tried on mine and it worked ok.

app/controllers/subscriptions_controller.rb
kevindew: Also might be better to call this smoke_test_address? be good to have clues that this is just for testing purposes.

README.md
```diff
@@ -54,6 +54,11 @@ sidekiq-monitoring for email-alert-api uses 3089
 * Run `RAILS_ENV=test bundle exec rake db:setup` to load the database
 * Run `bundle exec rspec` to run the tests
 
+### Using test email addresses for signup
+
+Using any email address that ends with '@notifications.service.gov.uk'
+will not create a subscriber or a subscription, however will return a 201.
+
```
kevindew: This sounds accurate but I think it would make more sense being in the context of the application endpoints section as this only applies to that particular endpoint.

PR #453, opened 2018-02-23
Fix delivery request worker

thomasleese:
Retry was out of scope of the retry mechanism. Move the retry code in to scope to fix the errors.

[Trello Card](https://trello.com/c/H9cStV98/640-retry-mechanism-for-dealing-with-a-rate-limit-doesnt-work)

### DIFF
--- a/app/workers/delivery_request_worker.rb
+++ b/app/workers/delivery_request_worker.rb
@@ -8,7 +8,12 @@ class DeliveryRequestWorker
   end
 
   sidekiq_retries_exhausted do |msg, _e|
-    reschedule_job if msg["error_class"] == "RatelimitExceededError"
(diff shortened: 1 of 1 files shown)


thomasleese:
Potentially, but the hard part is that it only runs after Sidekiq has stopped retrying it itself, I'm not sure how we can trigger that.

gpeng:
Oh yeah... Can we configure Sidekiq to retry 0 times in the test? We're getting into the realms of testing totally different behaviour than we're implementing at that point I suppose.

gpeng:
We're struggling to find a worthwhile way of testing this and it's already broken in staging/production so I'm going to merge for now and we can discuss further if need be.

PR #454, opened 2018-02-23
Add more email address validation

rubenarakelyan:
This commit adds more validation for email addresses. They must have:
* Exactly one “@“ sign
* At least one “.” in the domain or an IP address in “[]”
* No commas

Trello: https://trello.com/c/dywIBaJ5/639-add-more-validation-on-email-addresses

### DIFF
--- a/app/validators/email_address_validator.rb
+++ b/app/validators/email_address_validator.rb
@@ -1,0 +1,44 @@
+class EmailAddressValidator < ActiveModel::Validator
+  def validate(record)
+    email_address = record.address
+    unless email_address.nil? || valid_email_address?(email_address)
+      record.errors.add(:address, 'is not an email address')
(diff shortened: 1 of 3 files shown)


app/validators/email_address_validator.rb
```diff
@@ -0,0 +1,39 @@
...
+  def validate(record)
+    email_address = record.address
+    unless email_address.nil? || valid_email_address?(email_address)
+      record.errors.add(:address, 'is not an email address')
+    end
+  end
+
+  private
+
+  def valid_email_address?(email_address)
```
thomasleese: We should also check it doesn't contain a new line character.

rubenarakelyan: Done

PR #458, opened 2018-02-26
Add very primitive sanity check to govdelivery import

bevanloon:
A very primitive test against our govdelivery csv to make sure we pass
them in the correct order.

### DIFF
--- a/lib/import_govdelivery_csv.rb
+++ b/lib/import_govdelivery_csv.rb
@@ -16,6 +16,7 @@ class ImportGovdeliveryCsv
     check_encoding_is_windows_1252
     get_user_confirmation
 
+    sanity_check_csv
     import_subscribers
     import_subscriptions
   end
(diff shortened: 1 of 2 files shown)


lib/import_govdelivery_csv.rb
```diff
@@ -26,6 +27,11 @@ def call
 
   DEFAULT_DIGEST_FREQUENCY = Frequency::IMMEDIATELY
 
+  def sanity_check_csv
+    headers = CSV.open(subscriptions_csv_path, 'r') {|c| c.first}
+    raise RuntimeError.new("Your csv files do not seem to be correct.") if headers.map(&:downcase).include?("subscriber_type")
```
thomasleese: You don't need the `RuntimeError.new` here, if you `raise` a string you'll get a `RuntimeError`.

thomasleese: I think it might be worth doing a sanity check on both the files. I would check that the subscriptions file has a `TOPIC_CODE` and I would check that the digests file has a `DIGEST_FOR`.

PR #459, opened 2018-02-26
Add a migration to make various primary keys UUIDs

thomasleese:
This changes the primary keys on delivery attempts, content changes, emails and subscription contents into UUIDs. This is because UUIDs fit better for these records, especially in terms of archiving the data. We will also replace the reference field on delivery attempts to use this primary key.

Depends on https://github.com/alphagov/govuk-puppet/pull/7289.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -17,6 +17,7 @@ development:
 test:
   <<: *default
   database: email-alert-api_test
+  url: <%= ENV["TEST_DATABASE_URL"] %>
 
 production:
   <<: *default
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,5 +2,6 @@
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+  govuk.setEnvar("TEST_DATABASE_URL", "postgresql://email-alert-api:email-alert-api@localhost/email-alert-api_test")
   govuk.buildProject()
 }
--- a/app/workers/immediate_email_generation_worker.rb
+++ b/app/workers/immediate_email_generation_worker.rb
@@ -29,7 +29,9 @@ private
 
       email_ids = import_emails(subscribers, subscription_contents).ids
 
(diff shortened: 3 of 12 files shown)


kevindew:
Do we need an enable extension in migration too?

thomasleese:
@kevindew The apps don't have permissions to create extensions on their databases, so even if we had it in the migration it wouldn't work if the extension wasn't enabled. I suppose we could add it anyway as an early warning check.

kevindew:
@thomasleese Sure does it blow up if the extension is already enabled and it tries to enable it?

thomasleese:
@kevindew `enable_extension` does a `CREATE EXTENSION IF NOT EXISTS`, so no :)

bevanloon:
Does `.first` still work as expected when we move to UUIDs?

thomasleese:
@bevanloon Not really, we could introduce some default scopes for some models based on dates, but that will come in a later PR.

bevanloon:
@thomasleese sounds good. So I'm just wondering how that affects things like https://github.com/alphagov/email-alert-api/blob/master/app/services/metrics_service.rb#L50?

Might be an idea at some point to tidy up the migrations as well since a few of them rely on `.first`

thomasleese:
@bevanloon Hmm yeah, that's a weird one because I believe that's really only checking for "one or nil" since the check above it is making sure it's not a digest, I think that whole section of code could be made clearer.

config/database.yml
```diff
@@ -16,6 +16,9 @@ development:
 
 test:
   <<: *default
+  host: localhost
+  username: email-alert-api
+  password: email-alert-api
```
kevindew: Is this for CI? I wonder if it's better if we set this via DATABASE_URL env var? eg how we do this: https://github.com/alphagov/whitehall/blob/master/config/database.yml#L18 and set it in in the jenkinsfile (otherwise this may make it a bit of a pain to run tests locally)

thomasleese:
@kevindew It's possible we will see errors, although the whole migration takes only a few milliseconds. It won't cause anything fatal as GovDelivery is completely separated from the GOV.UK Notify system.

PR #460, opened 2018-02-26
Add a tally to the subscription import

bevanloon:
Having a tally of total subscriptions imported is a nice to have and
gives a rough indication of how much longer the import will take.

[Trello](https://trello.com/c/HX9boFIl/652-dry-run-todos)


bevanloon:
@kevindew - rejigged this.
Also half-inched your "time remaining" function 👍

PR #461, opened 2018-02-27
Marking an email as completed (finished sending)

kevindew:
Trello: https://trello.com/c/R14kXWU9/633-set-up-rudimentary-email-archive-system

This introduces a new field to email (finished_sending_at) which is used as a means to determine that the sending process for an Email has finished and then can be used as a means to determine which emails are archivable or not (the alternative to this is a rather grim query that pulls in latest delivery attempts for an email on a join.

I'm not particularly attached to the name `finished_sending_at` so suggestions for a better one are welcome. It seemed clearer than using terminology like "completed_at"

Along the way this also takes on a few issues: 

- Pseudo delivery attempts were being left as marked as sending forever as they won't have a status update
- There was the ability to get a status update for same delivery attempt twice which could cause incorrect data. 
- Fixes some concealed exceptions as part of feature testing.

### DIFF
--- a/app/services/status_update_service.rb
+++ b/app/services/status_update_service.rb
@@ -3,6 +3,7 @@ class StatusUpdateService
     @reference = reference
     @status = status
     @user = user
+    @delivery_attempt = find_delivery_attempt(reference)
   end
 
   def self.call(*args)
@@ -10,16 +11,22 @@ class StatusUpdateService
   end
 
   def call
-    updates = DeliveryAttempt.where(reference: reference).update_all(
-      status: status.underscore,
-      signon_user_uid: user&.uid,
-    )
+    begin
+      delivery_attempt.update!(
+        status: status.underscore,
+        signon_user_uid: user&.uid,
+      )
+    rescue ArgumentError
+      # This is because Rails doesn't currently do validations for enums
+      # see: https://github.com/rails/rails/issues/13971
+      error = "'#{status}' is not a supported status"
+      GovukError.notify(error)
+      raise DeliveryAttemptInvalidStatusError, error
+    end
 
-    raise ActiveRecord::RecordNotFound unless updates == 1
-
-    if permanent_failure? && subscriber
+    if delivery_attempt.permanent_failure? && subscriber
       UnsubscribeService.subscriber!(subscriber)
-    elsif temporary_failure?
+    elsif delivery_attempt.temporary_failure?
       DeliveryRequestWorker.perform_in(15.minutes, email.id, :default)
     end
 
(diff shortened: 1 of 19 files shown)


app/services/status_update_service.rb
```diff
@@ -3,23 +3,32 @@ def initialize(reference:, status:, user: nil)
     @reference = reference
     @status = status
     @user = user
+    @delivery_attempt = find_delivery_attempt(reference)
```
thomasleese: This class originally did look up the delivery attempt, but we removed it as part of speeding up this endpoint when we couldn't cope with the number of requests coming in from Notify, so this endpoint made only one query (the `UPDATE`(. This was before we added the database index so it may not be a problem anymore, but worth noting.

db/migrate/20180226101630_index_finished_sending_at_on_emails.rb
```diff
@@ -0,0 +1,7 @@
+class IndexFinishedSendingAtOnEmails < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :emails, [:finished_sending_at], algorithm: :concurrently
```
thomasleese: No need for the index to be in an array here.

app/services/status_update_service.rb
kevindew: @thomasleese yeah we have an email to update as well so I figured it was more prudent to grab email and delivery attempt together. Could probably at a push switch over to be just an update on each without the selects but lose the ability to do the preconditions. 

I realise I should always add a lock for update to it if I want to really stomp out the risk of conflicts.

db/migrate/20180226101630_index_finished_sending_at_on_emails.rb
kevindew: Cool thanks

kevindew:
@thomasleese are you happy for me to merge this? or do you think we should do any testing or any performance affects first?

app/services/status_update_service.rb
```diff
@@ -7,7 +7,7 @@ def initialize(reference:, status:, user: nil)
   end
 
   def self.call(*args)
-    new(*args).call
+    DeliveryAttempt.transaction { new(*args).call }
```
thomasleese: If we put the transaction here, I think we should put in a `private_class_method :new` to avoid anyone accidentally doing `.new(...).call`.

thomasleese:
@kevindew We can deploy to staging and then run the status updates load test against that and check it copes.

app/services/status_update_service.rb
kevindew: Think there is already one.

thomasleese: Oh so there is... oops

PR #463, opened 2018-02-28
Stop using old UUID fields

thomasleese:
This stops using the custom UUID fields on delivery attempts and subscriptions and instead uses their IDs as they are now UUIDs.

Depends on #461.

### DIFF
--- a/app/controllers/unsubscribe_controller.rb
+++ b/app/controllers/unsubscribe_controller.rb
@@ -6,10 +6,10 @@ class UnsubscribeController < ApplicationController
 private
 
   def subscription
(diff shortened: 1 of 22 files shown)


app/controllers/unsubscribe_controller.rb
```diff
@@ -6,10 +6,10 @@ def unsubscribe
 private
 
   def subscription
-    Subscription.not_deleted.find_by!(uuid: uuid)
+    Subscription.not_deleted.find_by!(id: id)
```
kevindew: another `find!` one

PR #464, opened 2018-02-28
Remove old UUID fields

thomasleese:
This removes the old UUID fields on delivery attempts and subscriptions as we can now just use their IDs.

Depends on #463.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -19,7 +19,6 @@ FactoryBot.define do
     email
     status :sending
     provider :notify
-    reference { SecureRandom.uuid }
   end
 
   factory :digest_run do
--- a/spec/integration/status_updates_spec.rb
+++ b/spec/integration/status_updates_spec.rb
@@ -1,9 +1,6 @@
 RSpec.describe "Receiving a status update", type: :request do
-  let(:reference) { "b6589b2b-8f8e-457b-9ddf-237b62438ad1" }
-
-  let!(:delivery_attempt) do
-    create(:delivery_attempt, id: reference, status: "sending")
-  end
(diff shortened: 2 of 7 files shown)


app/services/status_update_service.rb
```diff
@@ -54,7 +54,7 @@ def find_delivery_attempt(reference)
                 .includes(:email)
                 .joins(:email)
                 .lock
-                .find_by!(reference: reference)
+                .find_by!(id: reference)
```
kevindew: You can just switch this to `find!(reference)`

spec/factories.rb
```diff
@@ -16,6 +16,7 @@
   end
 
   factory :delivery_attempt do
+    id { SecureRandom.uuid }
```
kevindew: I'm not sure you need this - unless we're relying on this existing for some `build` methods?

spec/integration/status_updates_spec.rb
```diff
@@ -2,7 +2,7 @@
   let(:reference) { "b6589b2b-8f8e-457b-9ddf-237b62438ad1" }
 
   let!(:delivery_attempt) do
-    create(:delivery_attempt, reference: reference, status: "sending")
+    create(:delivery_attempt, id: reference, status: "sending")
```
kevindew: I'm not sure exactly why we've got a hardcoded uuid above so I may be misunderstanding something but could this just be a `let!(:delivery_attempt) { create(:delivery_attempt, status: "sending") }` and then reference is replaced with `delivery_attempt.id` ?

thomasleese: Yep, agreed that seems unnecessary... I've fixed that now!

PR #470, opened 2018-02-28
Subscriber Deactivation

thomasleese:
This PR deactivates subscribers instead of nullifying them when they unsubscribe, and then after 28 days it will nullify them.

It introduces the following states on subscribers with the appropriate validations and methods to change between them:

- **activated**: The subscriber wishes to receive emails.
- **deactivated**: The subscriber has asked to be unsubscribe from everything, they will be deleted 28 days after they unsubscribed.
- **deactivated and nullified (called nullified)**: There once was a subscriber but we now know nothing about them.

[Trello Card](https://trello.com/c/ElK95Sge/634-subscriber-deactivation)

### DIFF
--- a/app/models/subscriber.rb
+++ b/app/models/subscriber.rb
@@ -11,10 +11,11 @@ class Subscriber < ApplicationRecord
   has_many :digest_run_subscribers, dependent: :destroy
   has_many :digest_runs, through: :digest_run_subscribers
 
+  scope :nullified, -> { where(address: nil) }
   scope :deactivated, -> { where.not(deactivated_at: nil) }
   scope :activated, -> { where(deactivated_at: nil) }
 
-  def nullify_address!
+  def nullify!
     raise "Already nullified." if address.nil?
     raise "Must be deactivated first." if deactivated_at.nil?
 
--- a/app/models/subscriber.rb
+++ b/app/models/subscriber.rb
@@ -4,12 +4,41 @@ class Subscriber < ApplicationRecord
     validates_uniqueness_of :address, case_sensitive: false
   end
 
+  validate :not_nullified_and_activated
+
   has_many :subscriptions, -> { not_deleted }
   has_many :subscriber_lists, through: :subscriptions
   has_many :digest_run_subscribers, dependent: :destroy
   has_many :digest_runs, through: :digest_run_subscribers
 
+  scope :deactivated, -> { where.not(deactivated_at: nil) }
+  scope :activated, -> { where(deactivated_at: nil) }
+
   def nullify_address!
+    raise "Already nullified." if address.nil?
+    raise "Must be deactivated first." if deactivated_at.nil?
+
     update!(address: nil)
+  end
+
+  def deactivate!(datetime: nil)
+    raise "Already deactivated." unless deactivated_at.nil?
+
+    update!(deactivated_at: datetime || Time.now)
+  end
+
+  def activate!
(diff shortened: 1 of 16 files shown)


app/models/subscriber.rb
```diff
@@ -4,12 +4,41 @@ class Subscriber < ApplicationRecord
...
+  scope :activated, -> { where(deactivated_at: nil) }
+
   def nullify_address!
+    raise "Already nullified." if address.nil?
+    raise "Must be deactivated first." if deactivated_at.nil?
+
     update!(address: nil)
   end
+
+  def deactivate!(datetime: nil)
```
kevindew: Should these be bang methods? They aren't following convention of bang being a dangerous alternative to a non bang version.

I notice they're following the nullify_address! convention but not sure if that itself is an outlier.

app/workers/nullify_deactivated_subscribers_worker.rb
```diff
@@ -0,0 +1,16 @@
+class NullifyDeactivatedSubscribersWorker
+  include Sidekiq::Worker
+
+  def perform
+    subscribers.find_each(&:nullify!)
```
kevindew: Might just want to do this in an `update_all` though can't imagine this would see sufficient load to cause any pain?

app/workers/nullify_deactivated_subscribers_worker.rb
```diff
@@ -0,0 +1,16 @@
+class NullifyDeactivatedSubscribersWorker
+  include Sidekiq::Worker
```
kevindew: Do we need this in a queue? as it feels like this should be lower priority that some of the things below default priority?

I've same problem with email archiving. We could have a cleanup queue for all these sort of tasks.

Oh another point I just realised is this could go a bit funky with multiple workers trying it.

app/models/subscriber.rb
thomasleese: Yeah, I was following the convention of the old `nullify_address!` method.

I would also argue that since these are calling an `update!` method, it's also following some sort of Rails convention so should be left as bang methods.

app/workers/nullify_deactivated_subscribers_worker.rb
thomasleese: I liked being able to use the `nullify!` method here for clarity, and figured the load shouldn't cause any problems.

app/workers/nullify_deactivated_subscribers_worker.rb
thomasleese: Good point about multiple workers!

I agree another queue would make sense, shall I add one in this PR and you can rebase off that?

thomasleese: I've reworked this to put it on a `cleanup` queue, what do you think?

kevindew: That sounds good 👍

app/workers/nullify_deactivated_subscribers_worker.rb
kevindew: Sure thing.

app/models/subscriber.rb
kevindew: Sure, it does make sense but that is a tricky convention to follow I think Rails itself just follows the safe / danger one. 

Ideally this'd be something decided at linter level

PR #474, opened 2018-03-01
Append-only Subscriptions

thomasleese:
This PR makes it such that the subscriptions table will be append only, meaning that we have a record of what users used to be subscribed to.

This introduces a new `ended_reason` field to store why the subscription was ended, and a subscription is considered ended if both the `ended_reason` and `ended_at` fields are present. There is validation to make sure they are either both set or neither are set.

The `Subscription.active` scope can be used to look for currently active subscriptions.

[Trello Card](https://trello.com/c/RRhaDynE/635-append-only-subscription)

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -12,9 +12,12 @@ class SubscriptionsController < ApplicationController
     status = subscription.new_record? ? :created : :ok
 
     subscription.ended_at = nil
+    subscription.ended_reason = nil
+
     subscription.frequency = frequency
     subscription.signon_user_uid = current_user.uid
     subscription.source = subscription.new_record? ? :user_signed_up : :frequency_changed
+
     subscription.save!
 
     render json: { id: subscription.id }, status: status
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -2,24 +2,29 @@ class SubscriptionsController < ApplicationController
   def create
     return render json: { id: 0 }, status: :created if smoke_test_address?
 
-    subscription = Subscription.find_or_initialize_by(
-      subscriber: subscriber,
-      subscriber_list: subscribable,
-    )
+    existing_subscription = nil
+    subscription = nil
 
-    subscriber.activate! if subscriber.deactivated?
+    Subscription.transaction do
+      existing_subscription = Subscription.active.lock.find_by(
+        subscriber: subscriber,
+        subscriber_list: subscribable,
+      )
 
-    status = subscription.new_record? ? :created : :ok
+      existing_subscription.end(reason: :frequency_changed) if existing_subscription
 
-    subscription.ended_at = nil
-    subscription.ended_reason = nil
+      subscriber.activate! if subscriber.deactivated?
 
-    subscription.frequency = frequency
-    subscription.signon_user_uid = current_user.uid
-    subscription.source = subscription.new_record? ? :user_signed_up : :frequency_changed
+      subscription = Subscription.create!(
+        subscriber: subscriber,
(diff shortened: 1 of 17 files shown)


db/migrate/20180301153539_add_unique_index_on_active_subscriptions.rb
```diff
@@ -0,0 +1,10 @@
+class AddUniqueIndexOnActiveSubscriptions < ActiveRecord::Migration[5.1]
+  disable_ddl_transactions!
+
+  def up
+    add_index :subscriptions,
+      %w(subscriber_id subscriber_list_id),
+      unique: true,
+      where: "(ended_at IS NULL)"
```
kevindew: Nice 👍

app/models/subscription.rb
```diff
@@ -5,13 +5,34 @@ class Subscription < ApplicationRecord
...
+  validates_uniqueness_of :subscriber, scope: :subscriber_list, conditions: -> { active }
+  validate :both_ended_at_and_ended_reason_set
 
-  scope :not_deleted, -> { where(ended_at: nil) }
+  scope :active, -> { where(ended_at: nil) }
 
-  def destroy
-    update_attributes!(ended_at: Time.now)
+  def ended?
+    ended_at.present? && ended_reason.present?
```
kevindew: I'd not worry about sweating about whether ended_reason is set as well - it makes it all a little more complicated (and conflicts with the scope).

Whereas I think it's simpler if we just consider presence of ended_at to mean it's ended and we just do what we can to prevent the reason being null.

app/controllers/subscriptions_controller.rb
```diff
@@ -2,21 +2,24 @@ class SubscriptionsController < ApplicationController
   def create
     return render json: { id: 0 }, status: :created if smoke_test_address?
 
-    subscription = Subscription.find_or_initialize_by(
+    existing_subscription = Subscription.active.find_by(
```
kevindew: I wonder if you might want this in a lock for updates to prevent race conditions.

app/models/subscription.rb
```diff
@@ -5,13 +5,34 @@ class Subscription < ApplicationRecord
   has_many :subscription_contents
 
   enum frequency: { immediately: 0, daily: 1, weekly: 2 }
-  enum source: { user_signed_up: 0, frequency_changed: 1, imported: 2 }
+  enum source: { user_signed_up: 0, frequency_changed: 1, imported: 2 }, _prefix: true
+  enum ended_reason: { unsubscribed: 0, non_existant_email: 1, frequency_changed: 2 }, _prefix: :ended
```
kevindew: ooh didn't know about this _prefix option.

app/models/subscription.rb
thomasleese: Sounds fine to me, I've amended the PR with this.

app/controllers/subscriptions_controller.rb
thomasleese: Agreed, I've gone with a transaction though. Locking would be annoying with all the various different objects we've got, but a transaction will catch race conditions as the unique index on active subscriptions will only allow one transaction to succeed.

kevindew: It wouldn't actually be too tricky. You'd need to start a transaction and then add a lock to the Subscription find_by method chain.

The advantage you get over that than just DB locking is that it'll return the appropriate HTTP status code, whereas db locking you'll get a 500 (unless you go down the catching route)

thomasleese: Oh yeah, I forgot about the appropriate HTTP status codes... I figured we might have to lock the subscriber as well, but since we're in a transaction we don't need to. Amended the PR with the `lock` call!

PR #475, opened 2018-03-01
Add anti-spam and survey to templates

keithiopia:
Add "anti-spam" line and survey link to the immediate and digest email templates.
[Survey link Trello](https://trello.com/c/wa7ksgMD/653-add-basic-smartsurvey-link-into-email-footer)
[Spam line Trello](https://trello.com/c/SerqVQ2Y/643-add-anti-spam-footer-text)

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -32,7 +32,7 @@ private
   end
 
   def subject(content_change)
-    "GOV.UK update - #{content_change.title}"
+    "GOV.UK update – #{content_change.title}"
   end
 
   def body(content_change, subscriptions)
@@ -42,8 +42,12 @@ private
       <<~BODY
         #{presented_content_change(content_change)}
         ---
+        You’re getting this email because you subscribed to #{subscriptions.first.subscriber_list.title} updates on GOV.UK.
 
         #{presented_unsubscribe_links(subscriptions)}
+
+
+        ^Is this email useful? [Answer some questions to tell us more](https://www.smartsurvey.co.uk/s/govuk-email/?f=immediate).
       BODY
     end
   end
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -46,6 +46,7 @@ private
 
         #{presented_unsubscribe_links(subscriptions)}
 
+        &nbsp;
 
         ^Is this email useful? [Answer some questions to tell us more](https://www.smartsurvey.co.uk/s/govuk-email/?f=immediate).
       BODY
--- a/app/builders/digest_email_builder.rb
+++ b/app/builders/digest_email_builder.rb
@@ -24,10 +24,14 @@ private
   attr_reader :subscriber, :digest_run, :results
 
   def body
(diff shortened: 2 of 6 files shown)


app/builders/immediate_email_builder.rb
```diff
@@ -42,8 +42,12 @@ def body(content_change, subscriptions)
       <<~BODY
         #{presented_content_change(content_change)}
         ---
+        You’re getting this email because you subscribed to #{subscriptions.first.subscriber_list.title} updates on GOV.UK.
 
         #{presented_unsubscribe_links(subscriptions)}
+
```
thomasleese: Do we need this extra empty new line (and same for the other ones)? I don't think it will show up in the email.

keithiopia: The spacing would be useful – will the line breaks in the code get ignored?

thomasleese: Empty line breaks will be ignored, but you can do something like this to force a line break to appear:

```
blah blah

&nbsp;

blah blah
```

app/builders/immediate_email_builder.rb
```diff
@@ -46,6 +46,7 @@ def body(content_change, subscriptions)
 
         #{presented_unsubscribe_links(subscriptions)}
 
+        &nbsp;
```
cbaines: This is showing up in emails, which isn't exactly pleasing on the eye...

```
[Sri Lanka travel advice](https://www.gov.uk/foreign-travel-advice/sri-lanka)

 6:37am, 8 March 2018: Summary – Updated information on curfew imposed in the Kandy administrative region and related travel disruption, following a number of violent incidents – You’re
getting this email because you subscribed to Travel advice for all countries – travel advice updates on GOV.​UK.

Unsubscribe from [Travel advice for all countries – travel
advice](https://www.gov.uk/email/unsubscribe/0109de5d-5078-416b-a034-90777de8323a?title=Travel%20advice%20for%20all%20countries%20%20-%20travel%20advice)

&nbsp;

^Is this email useful? [Answer some questions to tell us more](https://www.smartsurvey.co.uk/s/govuk-email/?f=immediate).
```

Should this be an actual non breaking space, or is something going wrong with the rendering somewhere?

keithiopia: That's odd – I've not seen that rendered. Can you forward me the email you got, Chris?

thomasleese: The markdown is shown as in in text emails, if you view the emails as HTML they will show correctly.

thomasleese: If GOV.UK Notify support horizontal rules, we might be able to use `---` instead.

keithiopia: We're already using horizontal rules in the emails, though oddly can't spot that in the text version.
We tried some variations of the design using more than one horizontal rule but they all looked weird.

thomasleese: We could ask Notify to support a nice way of having a large break between paragraphs.

keithiopia: That would be a little more desirable. 
If viewing text only anyway, I feel that the nbsp is acceptable as a temporary visual aberration.

PR #476, opened 2018-03-01
Email archiving tasks

kevindew:
Trello: https://trello.com/c/R14kXWU9/633-set-up-rudimentary-email-archive-system

This sets up the appropriate extra data fields for email archiving as per [data retention strategy](https://github.com/alphagov/email-alert-api/blob/68fb5c40863114a18655c12db01c87e60eb5dc77/doc/arch/adr-003-data-retention.md) and the worker classes to perform the tasks.

I've set them to run every hour and I imagine they'll be pretty fast. My maths has them averaging at about 30,000 emails an hour and running locally (for a dataset of 30,000) I get:

```
irb(main):009:0> 30_000.times { FactoryBot.create(:subscription_content, :with_archivable_email) }
=> 30000
irb(main):011:0> { archivable_emails: Email.archivable.count, deleteable_emails: Email.deleteable.count, subscription_contents: SubscriptionContent.count, email_archives: EmailArchive.count }
=> {:archivable_emails=>30000, :deleteable_emails=>0, :subscription_contents=>30000, :email_archives=>0}
irb(main):012:0> EmailArchiveWorker.new.perform
{"@timestamp":"2018-03-01T16:41:58Z","@fields":{"pid":24478,"tid":"TID-gndlb4qq0","context":"","program_name":null,"worker":null},"@type":"sidekiq","@status":null,"@severity":"INFO","@run_time":null,"@message":"Archived 30000 emails in 12.77 seconds"}
=> true
irb(main):013:0> { archivable_emails: Email.archivable.count, deleteable_emails: Email.deleteable.count, subscription_contents: SubscriptionContent.count, email_archives: EmailArchive.count }
=> {:archivable_emails=>0, :deleteable_emails=>0, :subscription_contents=>30000, :email_archives=>30000}
irb(main):014:0> Email.update_all(finished_sending_at: 15.days.ago)
=> 30000
irb(main):015:0> { archivable_emails: Email.archivable.count, deleteable_emails: Email.deleteable.count, subscription_contents: SubscriptionContent.count, email_archives: EmailArchive.count }
=> {:archivable_emails=>0, :deleteable_emails=>30000, :subscription_contents=>30000, :email_archives=>30000}
irb(main):016:0> EmailDeletionWorker.new.perform
{"@timestamp":"2018-03-01T16:42:40Z","@fields":{"pid":24478,"tid":"TID-gndlb4qq0","context":"","program_name":null,"worker":null},"@type":"sidekiq","@status":null,"@severity":"INFO","@run_time":null,"@message":"Deleted 30000 emails in 7.18 seconds"}
=> true
irb(main):017:0> { archivable_emails: Email.archivable.count, deleteable_emails: Email.deleteable.count, subscription_contents: SubscriptionContent.count, email_archives: EmailArchive.count }
=> {:archivable_emails=>0, :deleteable_emails=>0, :subscription_contents=>30000, :email_archives=>30000}
irb(main):018:0> Email.count
=> 0
```

It is set to wait 14 days before deleting emails. Until we've upped the postgres disk space this might be a little optimistic, but hopefully that will be resolved soon - but we may still need to adjust this.

### DIFF
--- a/app/workers/email_archive_worker.rb
+++ b/app/workers/email_archive_worker.rb
@@ -1,0 +1,71 @@
+class EmailArchiveWorker
+  include Sidekiq::Worker
+
+  LOCK_NAME = "email_archive_worker".freeze
+
+  def perform
+    Email.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
+      EmailArchiveQuery.call.in_batches do |batch|
+        Email.transaction { archive_batch(batch) }
+      end
+    end
+  end
+
+private
+
+  def archive_batch(batch)
+    archived_at = Time.zone.now
+
+    import_batch(batch, archived_at)
+    Email.where(id: batch.pluck(:id)).update_all(archived_at: archived_at)
+  end
+
+  def import_batch(batch, archived_at)
+    to_import = batch.as_json.map { |e| build_email_archive(e, archived_at) }
+    columns = to_import.first.keys.map(&:to_s)
+    values = to_import.map(&:values)
+    EmailArchive.import!(columns, values)
+  end
+
+  def build_email_archive(email_data, archived_at)
+    content_change = build_content_change(email_data)
+
+    {
+      archived_at: archived_at,
+      content_change: content_change,
+      created_at: email_data.fetch("created_at"),
+      finished_sending_at: email_data.fetch("finished_sending_at"),
+      id: email_data.fetch("id"),
+      sent: email_data.fetch("sent"),
+      subject: email_data.fetch("subject"),
+      subscriber_id: build_subscriber_id(email_data),
+    }
+  end
+
+  def build_content_change(email_data)
+    return if email_data.fetch("content_change_ids").empty?
+
+    if email_data.fetch("digest_run_ids").count > 1
+      error = "Email with id: #{email_data['id']} is associated with "\
+        "multiple digest runs: #{email_data['digest_run_ids'].join(', ')}"
+      GovukError.notify(error)
+    end
+
+    {
+      content_change_ids: email_data.fetch("content_change_ids"),
+      digest_run_id: email_data.fetch("digest_run_ids").first,
+      subscription_ids: email_data.fetch("subscription_ids"),
+    }
+  end
+
+  def build_subscriber_id(email_data)
+    if email_data.fetch("subscriber_ids").count > 1
+      error = "Email with id: #{email_data['id']} is associated with "\
+        "multiple subscribers: #{email_data['subscribers'].join(', ')}"
+      GovukError.notify(error)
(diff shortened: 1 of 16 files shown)


app/workers/email_archive_worker.rb
```diff
@@ -0,0 +1,85 @@
...
+  LOCK_NAME = "email_archive_worker".freeze
+
+  def perform
+    Email.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
+      start_time = Time.zone.now
+      archived = 0
+
+      EmailArchiveQuery.call.in_batches do |batch|
+        Email.transaction do
+          archived += archive_batch(batch)
```
thomasleese: Might be tempted to call this `archived_count` as at first I thought this was storing each individual archived email.

app/workers/email_archive_worker.rb
```diff
@@ -0,0 +1,85 @@
...
+    EmailArchive.import(columns, values, validate: false)
+  end
+
+  def build_email_archive(email_data, archived_at)
+    content_change = build_content_change(email_data)
+
+    {
+      archived_at: archived_at,
+      content_change: content_change,
+      created_at: email_data["created_at"],
```
thomasleese: Is it worth using `.fetch` for a lot of these methods to error out early?

app/workers/email_archive_worker.rb
```diff
@@ -0,0 +1,71 @@
...
+      sent: email_data["sent"],
+      subject: email_data["subject"],
+      subscriber_id: build_subscriber_id(email_data),
+    }
+  end
+
+  def build_content_change(email_data)
+    return if email_data["content_change_ids"].empty?
+
+    if email_data["digest_run_ids"].count > 1
```
bevanloon: What circumstances do you think would cause us to have email associated with more than one digest run?

app/workers/email_archive_worker.rb
kevindew: Another good idea 👍

app/workers/email_archive_worker.rb
kevindew: Should be never. At an application level it should be impossible, but the means that we use to pull the data from the db means that it is an array that db wise could be multiple values. It's the same situation with Subscribers.

I was a bit unsure what to do with that possibility so @thomasleese and I had a chat through it and thought just use first value but moan to raven if we get anything weird (as that's probably a clue something weird is going on at an app level)

bevanloon: Cool thanks. I expect you've already wrestled with this, but the archive process feels like a bit on an anomalous place to be integrity checking our data. There's nowhere else to put it at the moment though.

PR #478, opened 2018-03-01
Migration to update SubscriberList titles

kevindew:
Trello: https://trello.com/c/zxbo8sHF/624-fix-the-subscription-titles-users-will-see-in-the-new-emails

This fixes a number of titles that have been noticed to be formatted
poorly.


kevindew:
@thomasleese oops thanks - need to get my migrations into order.

kevindew:
Marked as DNM as migration fails on actual database as there are duplicates in the CSV titles which cause it to break.

kevindew:
@thomasleese Thanks - it is not my day!

PR #483, opened 2018-03-02
Custom Travel Advice and Medical Safety Alerts

thomasleese:
This PR introduces custom emails for travel advice and medical safety alerts.

For travel advice emails, the description doesn't get shown. For medical safety alerts, we include a line about contacting MHRA directly.

[Travel Advice Trello Card](https://trello.com/c/2MgWX34y/642-change-what-we-put-in-travel-advice-updates) [Medical Safety Alert Trello Card](https://trello.com/c/f9gAAlay/627-add-an-extra-blurb-for-medical-device-alert-emails)

### DIFF
--- a/app/models/content_change.rb
+++ b/app/models/content_change.rb
@@ -8,4 +8,12 @@ class ContentChange < ApplicationRecord
   def mark_processed!
     update!(processed_at: Time.now)
   end
+
+  def is_travel_advice?
+    links.include?(:countries)
+  end
+
+  def is_medical_safety_alert?
+    tags.fetch(:format, []).include?("medical_safety_alert")
+  end
 end
--- a/app/presenters/content_change_presenter.rb
+++ b/app/presenters/content_change_presenter.rb
@@ -12,21 +12,12 @@ class ContentChangePresenter
   end
 
   def call
-    if include_description?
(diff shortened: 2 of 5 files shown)


kevindew:
Are we at risk of this including descriptions in digests with travel advice?

Also is it worth/possible making the travel advice change in travel advice itself?

thomasleese:
@kevindew The digests uses the same `ContentChangePresenter` so there shouldn't be any risk there.

We could include a field on content changes like `include_description`, but then the MHRA line will still be highly customised so I'm not convinced it's worth it given that only travel advice will be using that field.

kevindew:
@thomasleese 
> The digests uses the same ContentChangePresenter so there shouldn't be any risk there.

Great 👍 

> We could include a field on content changes like include_description, but then the MHRA line will still be highly customised so I'm not convinced it's worth it given that only travel advice will be using that field.

I mean that we just don't send a description from travel advice or create an abridged one on the fly. Basically the things we've introduced here are a debt that we ideally don't want in this app so just wondering what we can do to minimise pain they introduce.

thomasleese:
No longer necessary as we have found more generic solutions in the form of #488 and #486.

PR #490, opened 2018-03-07
Increase retry size healthcheck thresholds

thomasleese:
We rely on the retry set to perform our rate limiting so it's normal for these to be relatively high. These numbers are based on what we've seen happen in staging, but they may need tweaking in the future.

### DIFF
--- a/app/models/healthcheck/retry_size_healthcheck.rb
+++ b/app/models/healthcheck/retry_size_healthcheck.rb
@@ -25,11 +25,11 @@ class Healthcheck
     end
 
     def critical_size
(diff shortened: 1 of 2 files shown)


thomasleese:
I don't think we do anything, we can probably clear it out. I read that jobs in the dead queue will get cleared out after 6 months, but we can probably manually intervene earlier.

kevindew:
Yeah, it doesn't seem a big deal if it's things we can't fix but if it's things caught on rate limits then we should work out how they don't get lost in an abyss

thomasleese:
I'll merge this in now, we can continue discussion on Slack.

PR #496, opened 2018-03-08
Only deactivate a subscriber if not already

thomasleese:
We're seeing an issue with the status updates where we get back a permenant failure multiple times for the same subscriber leading to us deactiviting it multiple times which raises an exception.

[Sentry](https://sentry.io/govuk/app-email-alert-api/issues/483226884/)

### DIFF
--- a/app/services/unsubscribe_service.rb
+++ b/app/services/unsubscribe_service.rb
@@ -18,7 +18,7 @@ module UnsubscribeService
           subscription.end(reason: reason)
         end
 
-        if no_other_subscriptions?(subscriber, subscriptions)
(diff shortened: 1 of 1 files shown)


app/services/unsubscribe_service.rb
```diff
@@ -19,7 +19,7 @@ def unsubscribe!(subscriber, subscriptions, reason)
         end
 
         if no_other_subscriptions?(subscriber, subscriptions)
```
kevindew: might want to just put the condition in the if rather than two separate layers?

thomasleese: Sure, amended the commit with that.

PR #499, opened 2018-03-08
Add task to unsubscribe people

rubenarakelyan:
This commit adds a rake task that can be used to unsubscribe people from all emails. This will be used during the GovDelivery transition to unsubscribe people from the new system when they unsubscribe from the old system.

Trello: https://trello.com/c/bYeZh4Zd/681-make-a-rake-task-to-unsubscribe-people-in-bulk-from-a-csv

### DIFF
--- a/lib/tasks/unsubscribe.rake
+++ b/lib/tasks/unsubscribe.rake
@@ -1,0 +1,26 @@
+require 'csv'
+
+namespace :unsubscribe do
+  def unsubscribe(email_address)
+    subscriber = Subscriber.find_by(address: email_address)
+    if subscriber.nil?
+      puts "Subscriber #{email_address} not found"
+    else
+      puts "Unsubscribing #{email_address}"
+      UnsubscribeService.subscriber!(subscriber, :unsubscribed)
+    end
+  end
+
+  desc "Unsubscribe a single subscriber"
+  task :single, [:email_address] => :environment do |_t, args|
+    unsubscribe(args[:email_address])
+  end
+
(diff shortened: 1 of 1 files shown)


lib/tasks/unsubscribe.rake
```diff
@@ -0,0 +1,26 @@
+require 'csv'
+
+namespace :unsubscribe do
+  def unsubscribe!(email_address)
+    subscriber = Subscriber.find_by(address: email_address)
+    if subscriber.nil?
+      puts "Subscriber #{email_address} not found"
+    else
+      puts "Unsubscribing #{email_address}"
+      UnsubscribeService.subscriber!(subscriber, :unsubscribed) unless subscriber.nil?
```
thomasleese: We don't need the extra `unless ...` here.

lib/tasks/unsubscribe.rake
```diff
@@ -0,0 +1,26 @@
+require 'csv'
+
+namespace :unsubscribe do
+  def unsubscribe!(email_address)
```
thomasleese: In risk of starting to sound like @kevindew, I don't think this should be a `!` method.

lib/tasks/unsubscribe.rake
rubenarakelyan: Yeah I saw that as you were reviewing, a remnant of an earlier approach!

lib/tasks/unsubscribe.rake
rubenarakelyan: Righto, removed :)

PR #507, opened 2018-03-08
 Import subscription contents in batches grouped by subscriber

thomasleese:
We are seeing a problem where we are importing subscription contents in constant sized batches which could lead to not importing all the subscription contents for a particular subscriber in one go.

This can lead to users receiving duplicate emails for a content change since the deduplication relies on the assumption that all subscription contents are imported.

The alternative would be to import all subscription contents in one go, but that prevents us from being able to start sending emails as quick as possible as it can take a bit of time to generate the emails in the first place. It would also lead to a huge delivery queue which could use more memory.

[Trello Card](https://trello.com/c/LQ48zPbL/699-fix-any-post-deploy-bugs)

### DIFF
--- a/app/workers/subscription_content_worker.rb
+++ b/app/workers/subscription_content_worker.rb
@@ -19,11 +19,7 @@ private
         [content_change_id, subscription_id]
       end
 
-      begin
-        SubscriptionContent.import!(columns, records)
-      rescue StandardError => ex
-        Raven.capture_exception(ex, tags: { version: 2 })
-      end
+      SubscriptionContent.import!(columns, records)
 
       ImmediateEmailGenerationWorker.perform_async
     end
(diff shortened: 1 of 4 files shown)


app/workers/subscription_content_worker.rb
```diff
@@ -1,32 +1,48 @@
...
+    ImmediateEmailGenerationWorker.perform_async
+  end
+
+  def grouped_subscription_ids_per_subscriber(content_change)
+    subscriptions_for(content_change: content_change)
+      .pluck(:subscriber_id, :id)
+      .group_by(&:first)
+      .each_with_object({}) do |(subscriber_id, rows), hash|
+        hash[subscriber_id] = rows.map(&:last)
+      end
```
kevindew: You could possibly do more of this in postgres with a group(:subscriber_id) and pluck the array agg of the ids

thomasleese: Good idea!

thomasleese:
@kevindew I've addressed all your comments except the one about the infinite loop because, while I agree it is a problem, I think it is out of scope for this PR and I've put a card in planning for it. If you could take another look that would be great!

PR #508, opened 2018-03-09
Stop using GovDelivery IDs

thomasleese:
This stops us from using GovDelivery IDs as our topic IDs. This generates GovDelivery IDs by slugifying the title of the subscriber list.

The follow up PRs for this will be to remove the old GovDelivery client library code and cleanup references to GovDelivery in the tests, introduce a new `topic_id` or `slug` field which we can use instead of `gov_delivery_id` and then remove the old `gov_delivery_id` field.

[Trello Card](https://trello.com/c/15q8s623/683-change-subscriber-list-creation-process-to-not-rely-on-govdelivery)

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -1,10 +1,6 @@
 require "gov_delivery/client"
 
 class SubscriberListsController < ApplicationController
-  rescue_from GovDelivery::Client::TopicAlreadyExistsError do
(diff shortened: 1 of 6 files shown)


thomasleese:
@kevindew Good point, I have updated this to generate a different slug if already exists.

thomasleese:
@bevanloon We only support subscriptions for english version of content at the moment anyway, it would require a bigger change if we wanted to support different locales.

PR #512, opened 2018-03-12
Add sent_at and completed_at to DeliveryAttempt

1pretz1:
Email-alert-api now records when the email was delivered by notify(sent_at)
and when a delivery status update occurs(completed_at), e.g from temp-failure to
permanent failure. This information is recorded in the DeliveryAttempt
model. Also in the Email model, the 'finished_sending_at' column is now
populated with the 'sent_at' value from the DeliveryAttempt, instead of the 'updated_at' field.

### DIFF
--- a/app/services/status_update_service.rb
+++ b/app/services/status_update_service.rb
@@ -1,7 +1,9 @@
 class StatusUpdateService
-  def initialize(reference:, status:, user: nil)
+  def initialize(reference:, status:, completed_at:, sent_at:, user: nil)
     @reference = reference
     @status = status
+    @completed_at = completed_at
+    @sent_at = sent_at
     @user = user
     @delivery_attempt = find_delivery_attempt(reference)
   end
@@ -13,6 +15,8 @@ class StatusUpdateService
   def call
     begin
       delivery_attempt.update!(
+        sent_at: sent_at,
+        completed_at: completed_at,
         status: status.underscore,
         signon_user_uid: user&.uid,
       )
@@ -42,7 +46,7 @@ class StatusUpdateService
 
 private
 
-  attr_reader :delivery_attempt, :reference, :status, :user
+  attr_reader :delivery_attempt, :reference, :status, :user, :completed_at, :sent_at
   delegate :email, to: :delivery_attempt
 
   def subscriber
--- a/spec/features/technical_failure_spec.rb
+++ b/spec/features/technical_failure_spec.rb
@@ -15,8 +15,10 @@ RSpec.describe "Failing to deliver an email via Notify (technical failure)", typ
     expect(data.fetch(:status)).to eq("ok")
 
     reference = email_data.fetch(:reference)
-    send_status_update(reference, "technical-failure", expected_status: 204)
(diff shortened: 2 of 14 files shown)


app/services/status_update_service.rb
```diff
@@ -1,5 +1,7 @@
 class StatusUpdateService
-  def initialize(reference:, status:, user: nil)
+  def initialize(sent_at:, completed_at:, reference:, status:, user: nil)
```
thomasleese: Just personal opinion I guess, but I think `sent_at` and `completed_at` should appear after the `reference` and `status` arguments. I definitely think the `reference` should come first as it's the delivery attempt that this service class acts on.

spec/support/feature_helpers.rb
```diff
@@ -70,8 +70,8 @@ def create_content_change(overrides = {})
     expect(response.status).to eq(202)
   end
 
-  def send_status_update(reference, status, expected_status: 204)
-    params = { reference: reference, status: status }
+  def send_status_update(sent_at, completed_at, reference, status, expected_status: 204)
```
thomasleese: Similar to above, I would move these arguments after the `reference` and `status` arguments.

spec/services/status_update_service_spec.rb
```diff
@@ -2,24 +2,38 @@
...
     expect { status_update }
-      .to change { delivery_attempt.reload.status }
-      .to("delivered")
+    .to change { delivery_attempt.reload.status }
+    .to("delivered")
+  end
+
+  it "updates the completed_at field" do
+    expect { status_update }
+    .to change { delivery_attempt.reload.completed_at }
```
thomasleese: I think this `.to` lines should be indented in.

spec/services/status_update_service_spec.rb
```diff
@@ -2,24 +2,38 @@
   let(:reference) { "b6589b2b-8f8e-457b-9ddf-237b62438ad1" }
 
   let!(:delivery_attempt) do
-    create(:delivery_attempt, id: reference, status: "sending")
+    create(:delivery_attempt, id: reference, status: "sending", completed_at: nil, sent_at: nil)
```
thomasleese: This fields should default to `nil` so you don't need to specify them.

spec/services/delivery_request_service_spec.rb
```diff
@@ -125,13 +125,6 @@
         subject.call(email: email)
         expect(DeliveryAttempt.last).to be_delivered
       end
-
-      it "marks the email as being finished sending" do
```
thomasleese: Why do we need to remove this test?

spec/features/technical_failure_spec.rb
```diff
@@ -16,8 +16,10 @@
     expect(data.fetch(:status)).to eq("ok")
 
     reference = email_data.fetch(:reference)
-    send_status_update(reference, "technical-failure", expected_status: 204)
+    completed_at = Time.parse("2017-05-14T12:15:30.000000Z")
+    sent_at ||= completed_at
```
thomasleese: I think we can just use `=` here instead of `||=`.

PR #515, opened 2018-03-13
[WIP] Insert directly into subscription content

bevanloon:
I'm opening this so we can get some discussion on the approach because I've spent long enough on  a proof of concept. If we like the approach I'll fix it up and remove the unnecessary code paths.

Currently, to generate SubscriptionContent for immediate emails, we have https://github.com/alphagov/email-alert-api/blob/master/app/workers/subscription_content_worker.rb. Part of this worker's work (ahem) is to run this query https://github.com/alphagov/email-alert-api/blob/master/app/queries/content_change_immediate_subscription_query.rb. 

This basically finds all Subscriptions that match the current ContentChange with a frequency of `immediate` and then manually batches them up to insert them as SubscriptionContent items.

The proposed change is to use some sql to accomplish the same thing. Instead of having to write our own batch processor or complicate our code with grouping by subscriber. The idea is to run SubscriptionContentsImmediateInsert to populate the SubscriptionContent data, bypassing any ruby objects. Then continue as normal kicking off the ImmediateEmailGenerationWorker.

The advantage here is speed, and perhaps code readability (depending on your penchant for SQL). The disadvantage is we don't kick off the ImmediateEmailGenerationWorker until after the query has completed (currently we kick it off after every batch).

I've left some benchmark code in the specs for anyone to play with, but my results look like this for 1000 records:

Before:

```
#<Benchmark::Tms:0x000055a8211ab458
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=3.9506940889987163,
 @stime=0.11999999999999988,
 @total=3.0599999999999996,
 @utime=2.9399999999999995>
```

after:

```
#<Benchmark::Tms:0x00005598a099fd98
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=1.292155400995398,
 @stime=0.039999999999999813,
 @total=1.0200000000000002,
 @utime=0.9800000000000004>
```

### DIFF
--- a/app/queries/subscription_contents_immediate_insert.rb
+++ b/app/queries/subscription_contents_immediate_insert.rb
@@ -1,0 +1,45 @@
+class SubscriptionContentsImmediateInsert
+  def initialize(content_change_id)
+    @content_change_id = content_change_id
+  end
+
+  def self.call(content_change_id:)
+    new(content_change_id).call
+  end
+
+  def call
+    ActiveRecord::Base.connection.execute(
+      sql
+    )
+  end
+
+private
+
+  attr_reader :content_change_id
+
+  def sql
+    date = Time.zone.now
+    <<-SQL
+      INSERT INTO
+        subscription_contents(
+          subscription_id,
+          content_change_id,
+          created_at,
+          updated_at
+        )
+      SELECT
+        subscriptions.id,
+        uuid('#{content_change_id}'),
+        '#{date}',
+        '#{date}'
+      FROM
+        subscriptions
+        INNER JOIN subscriber_lists
+          ON subscriptions.subscriber_list_id = subscriber_lists.id
+        INNER JOIN matched_content_changes
+          ON subscriber_lists.id = matched_content_changes.subscriber_list_id
+      WHERE
+        matched_content_changes.content_change_id = uuid('#{content_change_id}')
+    SQL
+  end
+end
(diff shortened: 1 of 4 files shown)


bevanloon:
@thomasleese thanks - looking back at it, it sounded like the sql would take much longer than I would expect for a  roughly 30000 row insert (even given the joins). I've run both of these against a copy of the actual data for our large immediate subscription and this is what I got back

```
BATCH
#<Benchmark::Tms:0x0000556632036838
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=87.57642945699627,
 @stime=10.170000000000002,
 @total=68.93000000000005,
 @utime=58.76000000000005>

Batch SubscriptionContent row count: 34809

RAW
#<Benchmark::Tms:0x0000556632943548
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=2.7967737349972595,
 @stime=0.0,
 @total=0.009999999999990905,
 @utime=0.009999999999990905>

Raw SubscriptionContent row count: 34809
```

That looks more like what I would expect. Approximately 3 seconds executing just sql. I'm not sure why the benchmarking in the tests does not give the same results. Maybe there's some additional rspec overhead going on. Any thoughts?

thomasleese:
@bevanloon Hmm yeah, not sure what was going on there with the earlier benchmarks... but at 3 seconds to import all the subscription contents I definitely think we should pursue this method of creating them.

PR #522, opened 2018-03-13
Add API endpoint to list a user’s subscriptions

rubenarakelyan:
This commit adds an API endpoint at `GET /subscribers/:address/subscriptions` that lists a user’s active subscriptions given an email address.

Trello: https://trello.com/c/t9GWD8Rj/691-add-an-endpoint-to-email-alert-api-to-list-a-users-subscriptions

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,50 @@ The following fields are accepted on this endpoint: `subject`, `from_address_id`
 `document_type`, `content_id`, `public_updated_at`, `publishing_app`, `email_document_supertype`,
 `government_document_supertype`, `title`, `description`, `change_note`, `base_path`, `priority` and `footnote`.
 
+* `GET /subscribers/test@example.com/subscriptions` - gets a subscriber's subscriptions, in the form:
+
+```json
+{
+  "subscriber": {
+    "id": 1,
+    "address": "test@example.com",
+    "created_at": "Wed, 07 Mar 2018 17:04:28 UTC +00:00",
+    "updated_at": "Wed, 07 Mar 2018 17:04:28 UTC +00:00",
+    "signon_user_uid": null,
+    "deactivated_at": null
+  },
+  "subscriptions": [
+    {
+      "subscriber_id": 1,
+      "subscriber_list_id": 4232,
+      "created_at": "Wed, 07 Mar 2018 18:52:25 UTC +00:00",
+      "updated_at": "Wed, 07 Mar 2018 18:52:25 UTC +00:00",
+      "frequency": "daily",
+      "signon_user_uid": null,
+      "id": "476e1439-f5ba-4d7a-b4aa-1090563c5c36",
+      "source": "imported",
+      "ended_at": null,
+      "ended_reason": null,
+      "subscriber_list": {
+        "id": 4232,
+        "title": "All types of document about all topics by Foreign & Commonwealth Office",
+        "created_at": "Thu, 01 Aug 2013 12:53:34 UTC +00:00",
+        "updated_at": "Tue, 13 Mar 2018 15:11:29 UTC +00:00",
+        "document_type": "",
+        "tags": {},
+        "links": {
+          "organisations": ["9adfc4ed-9f6c-4976-a6d8-18d34356367c"]
+        },
+        "email_document_supertype": "",
+        "government_document_supertype": "",
+        "signon_user_uid": null,
+        "slug": "all-types-of-document-about-all-topics-by-foreign-commonwealth-office"
+      }
+    }
+  ]
+}
+```
+
 * `POST /subscriptions` with data:
 
 ```json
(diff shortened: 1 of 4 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -28,6 +28,20 @@ def create
...
+  def index
+    subscriber = find_subscriber
+
+    return render_not_found if subscriber.nil?
+
+    subscriptions = Subscription.active.
+      joins(:subscriber_list).
+      where(subscriber: subscriber).
+      order('subscriber_lists.title').
+      select(:id, :'subscriber_lists.title', :frequency).as_json
```
thomasleese: I wonder if we should use a `to_json` method on `Subscription` rather than selecting the fields like we do with `SubscriberList`.

app/controllers/subscriptions_controller.rb
```diff
@@ -36,14 +50,18 @@ def smoke_test_address?
...
+                      found = find_subscriber
                       found || Subscriber.create!(
                         address: address,
                         signon_user_uid: current_user.uid,
                       )
                     end
   end
 
+  def find_subscriber
+    @find_subscriber ||= Subscriber.where("LOWER(address) = ?", address.downcase).first
```
thomasleese: Probably not related to this PR, but I wonder if we should refactor this out into a class method of `Subscriber` as I could see it being useful for looking for subscribers in the console too.

app/controllers/subscriptions_controller.rb
```diff
@@ -28,6 +28,20 @@ def create
...
+
+    return render_not_found if subscriber.nil?
+
+    subscriptions = Subscription.active.
+      joins(:subscriber_list).
+      where(subscriber: subscriber).
+      order('subscriber_lists.title').
+      select(:id, :'subscriber_lists.title', :frequency).as_json
+
+    render json: { subscriber_address: address, subscriptions: subscriptions }
```
thomasleese: Instead of showing the address here, it might be good to just have a `subscriber: subscriber.to_json` then we can add extra fields easily if we want to.

app/controllers/subscriptions_controller.rb
rubenarakelyan: I've changed it so that we return the full subscriber and subscription details

app/controllers/subscriptions_controller.rb
rubenarakelyan: I've changed it to return the full subscriber

app/controllers/subscriptions_controller.rb
rubenarakelyan: Yes good idea

kevindew: Also if this was `Subscriber.find_by!("LOWER(address) = ?", address.downcase)` we'd not need to handle the nil case as Rails would raise the RecordNotFound exception that would automatically be converted to a 404.

app/controllers/subscriptions_controller.rb
```diff
@@ -28,6 +28,18 @@ def create
...
+    subscriber = find_subscriber
+
+    subscriptions = Subscription.active.
+      includes(:subscriber_list).
+      where(subscriber: subscriber).
+      order('subscriber_lists.title').
+      as_json(include: :subscriber_list)
+
+    render json: { subscriber: subscriber.as_json, subscriptions: subscriptions }
+  end
```
kevindew: This endpoint seems a touch unusual for an index. You'd expect that for an index you could traverse through all of the subscriptions in the system rather than a particular subscriber (and maybe you could then filter by address or addresses)

I'd suggest that since this all quite a specific case (and could hit growing pains if someone has a lot of subscriptions) that you might want to have this on it's own action within /subscriptions so index is free for more index like usages. 

Not sure quite what to name it as it's not super restful (I think closest rest approach is that we're actually doing a show on subscriber) but could be: `/subscriptions/subscriber/kevin@example.com`

app/controllers/subscriptions_controller.rb
kevindew: Oh and you do have find_by_address and find_by_address! already which could be overridden.

rubenarakelyan: I looked at using `find_by!` and then catching the exception in `subscriber` to create a new subscriber, but this then stops us from sending a 422 for an invalid email address for some reason. I'm going to defer changing this to a separate PR.

app/controllers/subscriptions_controller.rb
rubenarakelyan: I'm seeing this as an `index` of all of a user's subscriptions. To me it doesn't make sense to traverse all subscriptions in the system without at least a user filter. `show` would then be looking at a particular subscription. For other concepts like subscriber lists or subscribers, it makes sense to have an `index` that lists them all (with appropriate pagination of course).

kevindew: I think it does have users outside of just being a particular user. eg if you want to filter by a list and get all the subscriptions associated with it.

Can go with this but it's a bit of a RPC call wrapped onto a rest endpoint so is at risk of causing WTFs.

app/controllers/subscriptions_controller.rb
kevindew: Yeah I'd just go with two separate calls. Rather than having the same shared method can just have Subscriber.find_by_address(address) for the find/create and the bang variant for when you're handling a 404.

app/controllers/subscriptions_controller.rb
rubenarakelyan: I've given this some more thought and have come up with `GET /subscribers/test@example.com/subscriptions` so that we scope subscriptions to a subscriber. This will also allow us to have endpoints such as `POST /subscribers/test@example.com/address` to change a subscriber's email address.

PR #523, opened 2018-03-14
Isolate Sidkiq state changes in tests that don't adhere to the default `inline!`

bevanloon:
Properly isolate the Sidekiq state in spec/services/notification_handler_service_spec.rb and fix the subsequent failing tests. 

We set `Sidekiq::Testing.inline!` in `spec/rails_helper.rb`. In the `before` block in spec/services/notification_handler_service_spec.rb we set `Sidekiq::Testing.fake!`. Doing this means that `Sidekiq::Testing.inline?` returns `false` instead of our default `true`. This then sets the Sidekiq state for any subsequent tests, even in other files.

Since we are likely to make the assumption that `inline` is true when writing new tests, we should isolate any deviation - using an around block in `spec/services/notification_handler_service_spec.rb` accomplishes this.

### DIFF
--- a/spec/services/notification_handler_service_spec.rb
+++ b/spec/services/notification_handler_service_spec.rb
@@ -1,11 +1,10 @@
 RSpec.describe NotificationHandlerService do
-  before do
-    Timecop.freeze(Time.local(2017, 1, 1, 9))
-    Sidekiq::Testing.fake!
-  end
-
-  after do
-    Timecop.return
+  around(:example) do |example|
+    Timecop.freeze(Time.local(2017, 1, 1, 9)) do
+      Sidekiq::Testing.fake! do
+        example.run
+      end
+    end
   end
 
   let(:params) {
--- a/spec/services/status_update_service_spec.rb
+++ b/spec/services/status_update_service_spec.rb
@@ -53,8 +53,17 @@ RSpec.describe StatusUpdateService do
     end
 
     it "does not update the emails finished_sending_at timestamp" do
-      expect { status_update }
-        .to_not(change { delivery_attempt.reload.email.finished_sending_at })
+      # We set `inline!` in rails_helper which causes jobs to fire immediately.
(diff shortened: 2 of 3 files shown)


spec/services/status_update_service_spec.rb
```diff
@@ -53,8 +53,19 @@
...
 
     it "does not update the emails finished_sending_at timestamp" do
+      # We set `inline!` in rails_helper which causes jobs to fire immediately.
+      # Since DeliveryRequestWorker is fired on temporary_failure, this has the side effect of
+      # successfully sending the email and setting `finished_sending_at`.
+      # In reality, DeliveryRequestWorker is set to perform in 15 minutes time, so to mimic this
+      # we set `fake!` which pushes it on to an array instead. For the sake of this test
+      # we don't want it to perform since we are testing the state between the start of temporary
+      # failure and 15 minutes time when we try again.
+      Sidekiq::Testing.fake!
```
thomasleese: You can pass a block to this method so it reverts back automatically:

```
Sidekiq::Testing.fake! do
  expect { status_update }
    .to_not(change { delivery_attempt.reload.email.finished_sending_at })
end
```

bevanloon: 👍 fixed in 7456e1c

spec/services/notification_handler_service_spec.rb
```diff
@@ -6,6 +6,7 @@
 
   after do
     Timecop.return
+    Sidekiq::Testing.inline!
```
kevindew: Rather than a before and after could we put this all in an around block. Then we could do `Sidekiq::Testing.fake! { example.run }` (only learnt from this pr that it accepts a block 🎉 ) which would mean we wouldn't need to be in sync with rails_helper ?

bevanloon: Good idea - how does 5993053 look?

PR #525, opened 2018-03-14
Add rake task to change a subscriber’s email address

rubenarakelyan:
This commit adds a rake task to change a subscriber’s email address.

Trello: https://trello.com/c/KhYdhB2q/702-add-a-rake-task-to-change-a-subscribers-email-address

### DIFF
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -1,6 +1,17 @@
 require 'csv'
 
 namespace :manage do
+  def change_email_address(old_email_address:, new_email_address:)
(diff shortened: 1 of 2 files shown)


lib/tasks/manage.rake
```diff
@@ -1,6 +1,17 @@
 require 'csv'
 
 namespace :manage do
+  def change_email_address(old_email_address:, new_email_address:)
+    subscriber = Subscriber.find_by(address: old_email_address)
```
thomasleese: Do we need this to be case insensitive?

rubenarakelyan: Good point, changed

PR #526, opened 2018-03-14
Limit amount of emails copied to integration

1pretz1:
This query runs when data is getting copied to integration, currently
emails up to 14 days old are being transferred over from production, this query deletes all
emails that are older than 1 day old from the time of the query call.

### DIFF
--- a/lib/data_hygiene/anonymise_email_addresses.sql
+++ b/lib/data_hygiene/anonymise_email_addresses.sql
@@ -9,6 +9,10 @@
 # Please make changes to this script in email-alert-api where it is tested.
 # Then copy it to the env-sync-and-backup repository.
 
+# Deletes all email_archives that are older than 1 day old.
+DELETE FROM email_archives
+WHERE created_at < current_timestamp - interval '1 day';
+
 # Deletes all emails that are older than 1 day old.
 DELETE FROM emails
 WHERE created_at < current_timestamp - interval '1 day';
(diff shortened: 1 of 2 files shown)


lib/data_hygiene/anonymise_email_addresses.sql
```diff
@@ -27,6 +27,10 @@ SET address = CONCAT('anonymous-', a.id, '@example.com')
 FROM addresses a
 WHERE a.address = s.address;
 
+# Deletes all emails that are older than 1 day old.
+DELETE FROM emails
+WHERE created_at < current_timestamp - interval '1 day';
+
```
kevindew: This looks good - you'd want to move this to the start of this process as it does an insert based on all emails that could be a lot simpler with them deleted.

kevindew:
It'd be good to also do this against the email_archives table as that's another brute

lib/data_hygiene/anonymise_email_addresses.sql
1pretz1: Yes, good point! Done :)

1pretz1: Also added some tests for good measure

kevindew: Great stuff 👍 

I'm just trying it out (since I've been playing with db syncing) and it might also be prudent to have a db index on created_at for it as it's taking a super long time to run. I'm not really sure how much of a factor in it an index is though. I'll try and test it out.

1pretz1: Okay dokay, thanks! 💯

1pretz1:
& I've added the same delete query for email_archives

kevindew:
I checked with and without an index and they both took around 15 minutes each so doesn't seem to have any significant impact.

PR #530, opened 2018-03-15
Add API endpoint to change a user’s email address

rubenarakelyan:
This commit adds an API endpoint at `PATCH /subscribers/:address` that changes a user’s email address given an existing and new email address.

Trello: https://trello.com/c/1maT3NIT/692-add-an-endpoint-to-email-alert-api-to-change-a-subscribers-email-address

### DIFF
--- a/app/controllers/subscribers_controller.rb
+++ b/app/controllers/subscribers_controller.rb
@@ -9,10 +9,19 @@ class SubscribersController < ApplicationController
     render json: { subscriber: subscriber.as_json, subscriptions: subscriptions }
   end
 
+  def change_address
+    subscriber.update!(address: new_address)
+    render json: { subscriber: subscriber }
+  end
+
 private
 
   def subscriber
     @subscriber ||= Subscriber.find_by!("LOWER(address) = ?", address.downcase)
+  end
+
+  def new_address
+    subscriber_params.require(:new_address)
   end
 
   def address
@@ -20,6 +29,6 @@ private
   end
 
   def subscriber_params
-    params.permit(:address)
+    params.permit(:address, :new_address)
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/subscribers_controller.rb
```diff
@@ -9,17 +9,30 @@ def subscriptions
     render json: { subscriber: subscriber.as_json, subscriptions: subscriptions }
   end
 
+  def change_address
+    subscriber.address = new_address
+    if subscriber.save
+      render json: { subscriber: subscriber }
+    else
+      render json: { message: subscriber.errors.full_messages.to_sentence }, status: 422
```
thomasleese: Can we do `subscriber.update(address: new_address)` and leave the validation exception to show this message?

rubenarakelyan: Done!

config/routes.rb
```diff
@@ -13,6 +13,8 @@
 
     get "/subscribers/:address/subscriptions", to: "subscribers#subscriptions",
       constraints: { address: /.+@.+\..+/ }
+    post "/subscribers/:address/change-address", to: "subscribers#change_address",
```
thomasleese: I don't know how much we care about REST, but we could instead make this a `PATCH /subscribers/:address` endpoint and then the body will be replacement fields.

```
PATCH /subscribers/test@example.com

address=test2@example.com
```

config/routes.rb
```diff
@@ -13,6 +13,8 @@
...
+      constraints: { address: /.+@.+\..+/ }
```
thomasleese: I think it's possible to move this constraint into a block and then every route underneath it will have the same constraint applied to it.

config/routes.rb
rubenarakelyan: Yes good idea, done

config/routes.rb
rubenarakelyan: Yup works a treat

PR #536, opened 2018-03-16
Fix broken data

thomasleese:
This fixes some broken data we have in production related to missing some database restrictions on our columns.

This replaces #531 as the undeployable migration was reverted.


thomasleese:
@bevanloon I tested on staging, and it took a couple of minutes

PR #538, opened 2018-03-16
Subscriber ids on emails

thomasleese:
This PR introduces a new `subscriber_id` field to emails and sets it in the immediate email builder and
the digest email builder.

Replaces #527 as that was reverted. Unlike the original PR, this doesn't add the foreign key as it locks the table.

[Trello Card](https://trello.com/c/xEP0zV7E/686-store-subscriberid-with-email)

### DIFF
--- a/db/migrate/20180315080842_add_subscriber_id_to_emails.rb
+++ b/db/migrate/20180315080842_add_subscriber_id_to_emails.rb
@@ -1,5 +1,16 @@
 class AddSubscriberIdToEmails < ActiveRecord::Migration[5.1]
-  def change
+  def up
     add_reference :emails, :subscriber
+
+    execute %(
+      ALTER TABLE "emails"
+      ADD CONSTRAINT emails_subscriber_id_fk
+      FOREIGN KEY ("subscriber_id") REFERENCES "subscribers" ("id")
+      ON DELETE CASCADE NOT VALID
+    )
+  end
+
+  def down
+    remove_reference :emails, :subscriber
   end
 end
--- a/db/migrate/20180315080842_add_subscriber_id_to_emails.rb
+++ b/db/migrate/20180315080842_add_subscriber_id_to_emails.rb
@@ -1,0 +1,5 @@
+class AddSubscriberIdToEmails < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :emails, :subscriber
+  end
+end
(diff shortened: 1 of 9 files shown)


db/migrate/20180315080842_add_subscriber_id_to_emails.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubscriberIdToEmails < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :emails, :subscriber
```
bevanloon: in the reverted commit we've also added a `foreign_key` (https://github.com/alphagov/email-alert-api/pull/527/files#diff-5ada1ec3c318e9d3712d888134d92ec5R4) - any reason we aren't doing that here?

thomasleese: Oh yeah, I should clarify this difference in the PR. The reason is that adding the foreign key locks the tables for a few minutes so we can't deploy this easily during working hours.

bevanloon: Cool. I wonder if using `NOT VALID` on the creation of the foreign key would help speed that up?

thomasleese: I hadn't seen `NOT VALID` before! It looks like that would help, but it seems like there's no support for it in active record so we would need to use an `execute` statement in the migration.

thomasleese: I've pushed up a commit which creates the foreign key using `NOT VALID`.

bevanloon: awesome - I was thinking it might be a bit messy to include in this pr, but 👍

kevindew:
@thomasleese this breaks the test_email rake task as subscriber_list is a required argument.

lib/tasks/deliver.rake
```diff
@@ -1,15 +1,17 @@
 namespace :deliver do
-  def test_email(address)
+  def test_email(address, subscriber_id)
```
kevindew: This is the bit that causes the problem.

thomasleese: Oh oops, I'll raise a PR to make it optional.

thomasleese: https://github.com/alphagov/email-alert-api/pull/549

PR #540, opened 2018-03-16
Add API endpoint to change a subscription’s frequency

rubenarakelyan:
This commit adds an API endpoint at `PATCH /subscriptions/:subscription_id` that changes a subscription’s frequency given a subscription ID and a new frequency.

Trello: https://trello.com/c/BE9p0xVl/706-add-an-endpoint-to-email-alert-api-to-change-the-frequency-of-a-subscription


rubenarakelyan:
Comments encouraged for the rescuing and raising of exceptions for the enum validation.

app/controllers/subscriptions_controller.rb
```diff
@@ -28,6 +28,15 @@ def create
     render json: { id: subscription.id }, status: status
   end
 
+  def change_frequency
```
thomasleese: This endpoint and the endpoint above (`create`) are very similar. If a user "creates" a new subscription when they already have one, it is effectively the same as changing the frequency on an existing subscription. This works by "ending" the existing subscription and creating a new one with the `source` set as `:frequency_changed`. This allows us to have better historical analytics data.

I like the idea of a `PATCH` endpoint to make it more explicit that we're dealing with an existing subscription, and raising a 404 if the existing subscription doesn't exist, but I think it should work the same as the endpoint above: creating a new subscription record and ending the existing one.

app/controllers/subscriptions_controller.rb
```diff
@@ -28,6 +28,15 @@ def create
...
+    subscription.update!(frequency: frequency)
+    render json: { subscription: subscription }
+  rescue ArgumentError
+    # This happens if a frequency is provided that isn't included
+    # in the enum which is in the Subscription model
+    raise ActiveRecord::RecordInvalid
```
thomasleese: I'm not to fussed about handling the error in this way, it's the best we can do given that enums aren't 100% supported in terms of validation.

app/controllers/subscriptions_controller.rb
rubenarakelyan: That sounds better, changed

app/controllers/subscriptions_controller.rb
```diff
@@ -28,6 +28,33 @@ def create
...
+        subscriber: existing_subscription.subscriber,
+        subscriber_list: existing_subscription.subscriber_list,
+        frequency: frequency,
+        signon_user_uid: current_user.uid,
+        source: :frequency_changed
+      )
+    end
+
+    render json: { subscription: subscription }, status: :ok
+  rescue ArgumentError
```
thomasleese: I think we should wrap this around the `create!` in case we get an `ArgumentError` on a different line. Also, I wonder if a lot of this code can be extracted out into separate methods since it's so similar to the `create` method.

rubenarakelyan: I've changed the `ArgumentError` rescue block to be more specific. I'm struggling a little to identify things worth extracting into their own method since things are just different enough to make the methods nearly identical to what they're replacing IMO. Do you have any specific things you think could be extracted?

thomasleese: Hmm yeah, I had a quick go myself and there wasn't much benefit in it.

PR #541, opened 2018-03-16
Stop retrying after 24 hours

kevindew:
Trello: https://trello.com/c/Dgf0fbal/703-stop-retrying-for-temporary-failures-after-24-hours

Note: I'm putting this up at the end of Friday as I'm on leave Monday and didn't know if we were in a rush to get this out or not. It may need some additional thought alongside the review if there's any edge cases I've missed.

This sets up a new state on DeliveryAttempt for when they have had 24 hours of temporary_failure statuses from Notify and then ceases to retry. It is a final state. 

This isn't a particularly good solution to the problem though as it leaves us with the slightly thorny situation where we can have DeliveryAttempts that have returned as temporary_failure and are marked as different states in the db. This basically loads DeliveryAttempt with accumulative state which is far better served on Email.

I've done some work implementing this into state into email but it will require a couple of deploys so I've put this together as a faster approach to getting the situation looked at. Storing state on Email will resolve a few other complexities with archiving too.

### DIFF
--- a/app/models/delivery_attempt.rb
+++ b/app/models/delivery_attempt.rb
@@ -3,13 +3,28 @@ class DeliveryAttempt < ApplicationRecord
 
   validates :email, :status, :provider, presence: true
 
-  FINAL_STATUSES = %i[delivered permanent_failure].freeze
+  FINAL_STATUSES = %i[
+    delivered
+    permanent_failure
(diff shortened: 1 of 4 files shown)


kevindew:
I'm going to close this in favour of the alternative approach.

PR #543, opened 2018-03-20
Show subscription endpoint

thomasleese:
This will be used by email-alert-frontend to present the correct title of the subscription instead of taking it from the query string.

[Trello Card](https://trello.com/c/tQznyVfV/704-remove-the-need-to-specify-a-subscriber-list-name-on-the-unsubscribe-page)

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -28,6 +28,11 @@ class SubscriptionsController < ApplicationController
     render json: { id: subscription.id }, status: status
   end
 
+  def show
+    subscription = Subscription.find(subscription_params.require(:id))
+    render json: { subscription: subscription }
+  end
+
   def change_frequency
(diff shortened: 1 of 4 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -28,13 +28,18 @@ def create
     render json: { id: subscription.id }, status: status
   end
 
-  def change_frequency
+  def show
+    subscription = Subscription.find(subscription_params.require(:id))
+    render json: subscription.to_json
```
rubenarakelyan: I think we should aim for consistency here, since `update` returns `render json: { subscription: subscription }, status: :ok`. I'm happy with either one but they should be the same :)

thomasleese: Ah yep, I agree. I'll make it return everything under a `subscription` key like `update`.

PR #547, opened 2018-03-20
[WIP] Email status

kevindew:
This is a rough overview of the stuff to introduce state into emails of Email Alert API. I'm going to break this up into separate PRs and think through some bits a little more.

Thought I'd put it up as a PR if it raises any questions/discussions though.

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -20,6 +20,7 @@ private
   def records
     recipients_and_content.map do |recipient_and_content|
       [
+        Email.statuses["pending"],
         recipient_and_content.fetch(:address),
         subject(recipient_and_content.fetch(:content_change)),
         body(recipient_and_content.fetch(:content_change), recipient_and_content.fetch(:subscriptions)),
@@ -29,7 +30,7 @@ private
   end
 
   def columns
-    %i(address subject body subscriber_id)
+    %i(status address subject body subscriber_id)
   end
 
   def subject(content_change)
--- a/app/services/status_update_service.rb
+++ b/app/services/status_update_service.rb
@@ -21,7 +21,7 @@ class StatusUpdateService
         signon_user_uid: user&.uid,
       )
 
-      email.finish_sending(delivery_attempt) if delivery_attempt.has_final_status?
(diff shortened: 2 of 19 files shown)


app/builders/immediate_email_builder.rb
```diff
@@ -20,6 +20,7 @@ def call
   def records
     recipients_and_content.map do |recipient_and_content|
       [
+        Email.statuses["pending"],
```
thomasleese: I believe you can actually use the symbol `:pending` here.

thomasleese: Unless this is for performance?

app/services/status_update_service.rb
```diff
@@ -32,7 +32,7 @@ def call
 
     if delivery_attempt.permanent_failure? && subscriber
       UnsubscribeService.subscriber!(subscriber, :non_existant_email)
-    elsif delivery_attempt.temporary_failure?
+    elsif delivery_attempt.temporary_failure? && email.pending?
```
thomasleese: Why do we need this extra check here?

app/builders/immediate_email_builder.rb
kevindew: oh really? as in just `:pending` or `Email.statuses[:pending]`? Be cool if I can do the former.

app/services/status_update_service.rb
kevindew: Because they may have exhausted retries and email could be failed.

app/builders/immediate_email_builder.rb
thomasleese: The former, just `:pending`, I used that in the import script!

app/services/status_update_service.rb
thomasleese: Oh yes sorry I was getting confused and thinking we were dealing with permanent failures when doing the retries exhausted state.

app/builders/immediate_email_builder.rb
kevindew: Awesome, will try that out 👍

kevindew:
Closed as superseded by later PRs

PR #553, opened 2018-03-21
Add GA UTM params to all links in emails

rubenarakelyan:
This commit adds the UTM parameters used by Google Analytics to all links in emails sent from the system. This allows anonymous statistics to be collected.

Trello: https://trello.com/c/lO6tqY1H/696-add-tracking-to-links-within-email-alerts

### DIFF
--- a/app/presenters/content_change_presenter.rb
+++ b/app/presenters/content_change_presenter.rb
@@ -3,8 +3,9 @@ require 'redcarpet/render_strip'
 class ContentChangePresenter
   EMAIL_DATE_FORMAT = "%l:%M%P, %-d %B %Y".freeze
 
-  def initialize(content_change)
+  def initialize(content_change, frequency: "immediate")
     @content_change = content_change
+    @frequency = frequency
   end
 
   def self.call(*args)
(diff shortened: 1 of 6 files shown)


app/presenters/content_change_presenter.rb
```diff
@@ -29,7 +30,12 @@ def call
   delegate :title, :description, :change_note, :footnote, to: :content_change
 
   def content_url
-    PublicUrlService.url_for(base_path: content_change.base_path)
+    utm_source = content_change.id
+    utm_medium = "email"
+    utm_campaign = "govuk-notifications"
+    utm_content = @frequency
```
kevindew: It'd be a bit more consistent with this class and code base to have frequency behind an attr_reader

spec/factories.rb
```diff
@@ -1,5 +1,6 @@
 FactoryBot.define do
   factory :content_change do
+    id { SecureRandom.uuid }
```
kevindew: Is this needed for build? Since it's set automatically on create I'm just wondering if this was added to resolve an issue.

app/presenters/content_change_presenter.rb
```diff
@@ -29,7 +30,12 @@ def call
...
+    base_path = "#{content_change.base_path}?utm_source=#{utm_source}&utm_medium=#{utm_medium}&utm_campaign=#{utm_campaign}&utm_content=#{utm_content}"
```
kevindew: At a fundamental level I'm wondering if we want the email_id available which can be used to determine more things (but is not fun to use).

This might not matter until it's decided it's wanted.

app/presenters/content_change_presenter.rb
rubenarakelyan: Done

spec/factories.rb
rubenarakelyan: Removed

app/presenters/content_change_presenter.rb
rubenarakelyan: That would make the analytics traceable to a particular person though, which may not be what we want

kevindew: Sure thing.

PR #556, opened 2018-03-21
Migration to non null and default for Email Status

kevindew:
Do not merge until https://github.com/alphagov/email-alert-api/pull/551 is
deployed and run in production.

I'm not sure if I'll be able to actually run this migration on the live
database as it'll lock. May be ok without the null, if it's in the realm
of a few seconds though we should be ok as anything can retry.


rubenarakelyan:
Does this need a schema update?

kevindew:
@rubenarakelyan D'oh! 2 mins.

PR #557, opened 2018-03-21
Final bits of Email Status

kevindew:
Do not merge until https://github.com/alphagov/email-alert-api/pull/555 is
deployed and run in production

This switches the archiving/deleting process to use the new Email status field


rubenarakelyan:
Need the schema update from https://github.com/alphagov/email-alert-api/pull/556 here too

kevindew:
@rubenarakelyan this should be sorted now

PR #560, opened 2018-03-22
Create an auth token

kevindew:
Trello: https://trello.com/c/6YmlH3Mq/725-add-token-based-login-for-the-subscription-management-interface-to-email-alert-frontend

This adds an endpoint which can be posted to and will create a token that is encrypted with a shared secret containing the subscribers id. This can then be decrypted by Email Alert Frontend to authenticate the user.

The token is sent to a user via an email.

### DIFF
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -1,5 +1,5 @@
 class SubscribersAuthTokenController < ApplicationController
-  # @TODO - there is no validation here
+  before_action :validate_params
 
   def auth_token
     subscriber = find_or_create_subscriber(expected_params.require(:address))
@@ -41,5 +41,20 @@ private
 
   def expected_params
     params.permit(:address, :destination, :redirect)
+  end
+
+  def validate_params
+    ParamsValidator.new(expected_params).validate!
+  end
+
+  class ParamsValidator < OpenStruct
+    include ActiveModel::Validations
+
+    validates :address, presence: true
+    validates :address, email_address: true, allow_blank: true
+
+    validates :destination, presence: true
+    validates :destination, absolute_path: true, allow_blank: true
+    validates :redirect, absolute_path: true, allow_blank: true
   end
 end
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -1,0 +1,45 @@
+class SubscribersAuthTokenController < ApplicationController
+  # @TODO - there is no validation here
+
+  def auth_token
+    subscriber = find_or_create_subscriber(expected_params.require(:address))
+    token = generate_token(subscriber)
+    email = build_email(subscriber, token)
+
+    DeliveryRequestWorker
+      .perform_async_in_queue(email.id, queue: :delivery_immediate_high)
+
+    render json: { subscriber: subscriber }, status: :created
+  end
+
+private
+
+  def find_or_create_subscriber(address)
+    found = Subscriber.find_by_address(address)
+    found.activate! if found&.deactivated?
+    found || Subscriber.create!(
+      address: address,
+      signon_user_uid: current_user.uid,
+    )
+  end
+
+  def generate_token(subscriber)
+    AuthTokenGeneratorService.call(
+      subscriber,
+      redirect: expected_params[:redirect],
+      expiry: 1.week.from_now
+    )
+  end
+
+  def build_email(subscriber, token)
+    AuthEmailBuilder.call(
+      subscriber: subscriber,
+      destination: expected_params.require(:destination),
+      token: token,
+    )
+  end
+
+  def expected_params
+    params.permit(:address, :destination, :redirect)
+  end
+end
--- a/app/models/subscriber.rb
+++ b/app/models/subscriber.rb
@@ -1,6 +1,6 @@
 class Subscriber < ApplicationRecord
   with_options allow_nil: true do
-    validates_with EmailAddressValidator, fields: [:address]
+    validates :address, email_address: true
     validates_uniqueness_of :address, case_sensitive: false
   end
 
--- a/app/models/subscriber.rb
+++ b/app/models/subscriber.rb
@@ -58,6 +58,11 @@ class Subscriber < ApplicationRecord
     update!(address: nil)
   end
 
+  def as_json(options = {})
+    options[:except] ||= %i(signon_user_uid)
+    super(options)
+  end
+
 private
 
   def not_nullified_and_activated
--- a/app/services/auth_token_generator_service.rb
+++ b/app/services/auth_token_generator_service.rb
@@ -1,0 +1,18 @@
+class AuthTokenGeneratorService
+  def self.call(subscriber)
+    new.call(subscriber)
+  end
+
+  def call(subscriber)
+    data = { "data" => { "subscriber_id" => subscriber.id } }
+    JWT.encode(data, secret, "HS256")
+  end
+
+  private_class_method :new
+
+private
+
+  def secret
+    Rails.application.secrets.email_alert_auth_token
+  end
+end
--- a/app/services/auth_token_generator_service.rb
+++ b/app/services/auth_token_generator_service.rb
@@ -1,16 +1,28 @@
 class AuthTokenGeneratorService
-  def self.call(subscriber)
-    new.call(subscriber)
+  def self.call(*args)
+    new.call(*args)
   end
 
-  def call(subscriber)
-    data = { "data" => { "subscriber_id" => subscriber.id } }
+  def call(subscriber, redirect: nil, expiry: 1.week.from_now)
+    data = token_data(subscriber, redirect, expiry)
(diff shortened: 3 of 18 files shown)


app/controllers/subscribers_controller.rb
```diff
@@ -10,25 +11,42 @@ def subscriptions
...
+    subscriber = find_subscriber(subscriber_params.require(:address))
+    subscriber.update!(address: subscriber_params.require(:new_address))
     render json: { subscriber: subscriber }
   end
 
-private
-
-  def subscriber
-    @subscriber ||= Subscriber.find_by!("LOWER(address) = ?", address.downcase)
+  def auth_token
```
bevanloon: I'm not really picturing the flow of the aut_token through the api and frontend at the moment....why do we need it as a public endpoint? is the idea to auth token to `email-alert-frontend` when we get a new subscriber and as such create the session?

app/models/subscriber.rb
```diff
@@ -48,6 +48,11 @@ def nullify!
     update!(address: nil)
   end
 
+  def as_json(options = {})
+    options[:except] ||= %i(signon_user_uid)
```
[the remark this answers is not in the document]
kevindew: I'm not sure I understand too much. You can change this if you pass in other options like `only:` or set `except:` to be an empty array.

app/controllers/subscribers_controller.rb
kevindew: The endpoint here is only available internally and a rough sketch of the process is:

- You visit gov.uk/email/authenticate and are served a form
- You submit that form by posting an email address to /email/authenticate. This then
  - Sends a request to Email Alert API to create an auth token
  - That token is emailed to the user as part of a link
- The user is then given a message saying check email
- Then the email to login flow kicks in.

Does that help?

app/models/subscriber.rb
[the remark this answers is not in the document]
bevanloon: maybe I'm misunderstanding - if I pass in `except:` populated with something else, I essentially override our intention to exclude `signon_user_uid`?

I was wondering if we want to allow that behaviour or if we want to be really strict and never expose `signon_user_id` so something like:

```
options[:except] || = []
options[:except].concat(%i(signon_user_uid))
```

app/controllers/subscribers_controller.rb
bevanloon: 👍 yep thanks

app/models/subscriber.rb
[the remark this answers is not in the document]
kevindew: I think that'd be going a little overboard. I think really if we're wanting to be clear about our serialisation we should use a dedicated presenter. 

I don't really see this as something super strict though more a "that probably shouldn't be going through"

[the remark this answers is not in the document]
bevanloon: cool

app/services/auth_token_generator_service.rb
```diff
@@ -0,0 +1,30 @@
...
+
+private
+
+  def token_data(subscriber, redirect, expiry)
+    {
+      "data" => {
+        "subscriber_id" => subscriber.id,
+        "redirect" => redirect,
+      },
+      "exp" => expiry.to_i,
```
bevanloon: Do we want to consider validating that `expiry` is in the future? I can't see it being a big problem, and it is low risk given we are only calling it from one place, but we can, by the looks of it, set an expiry that's already past.

app/controllers/subscribers_auth_token_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    DeliveryRequestWorker
+      .perform_async_in_queue(email.id, queue: :delivery_immediate_high)
+
+    render json: { subscriber: subscriber }, status: :created
+  end
+
+private
+
+  def find_or_create_subscriber(address)
```
bevanloon: Does this mean if I visit the site to change my subscription and enter an email address that isn't currently in the system, it will create that email address and then send that address a magic link?

spec/integration/subscribers_auth_token_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    context "when we have an existing user" do
+      let!(:subscriber) { create(:subscriber, address: address) }
+
+      it "returns subscriber details" do
+        post path, params: params
+        expect(data[:subscriber][:id]).to eq(subscriber.id)
+      end
+    end
+
+    context "when we user we didn't previously know" do
```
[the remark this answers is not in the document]
kevindew: Oops - thanks for catching!

app/controllers/subscribers_auth_token_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    found = Subscriber.find_by_address(address)
+    found.activate! if found&.deactivated?
```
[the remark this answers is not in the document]
kevindew: Yup that's correct on both fronts.

app/services/auth_token_generator_service.rb
kevindew: Yeah it's quite possible, although this is only used internally - similarly a token with a very very short expiry time in the present is pretty pointless. I think it's ok that the onus is on the consumer if they send this service bad data.

app/controllers/subscribers_auth_token_controller.rb
kevindew: Yes, that's correct. Not exactly ideal but seems better than getting into the murky world of exposing whether we have subscribers or not by returning different status codes. Does make life a bit painful with activating and deactivating subscribers though :cry:

app/controllers/subscribers_auth_token_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def expected_params
+    params.permit(:address, :destination, :redirect)
+  end
+
+  def validate_params
+    ParamsValidator.new(expected_params).validate!
+  end
+
+  class ParamsValidator < OpenStruct
```
thomasleese: 👌

app/controllers/subscribers_auth_token_controller.rb
bevanloon: So what happens if someone decides to maliciously enter `n` number of email addresses that don't exist? I'm wondering how that affects our relationship with Notify and whether continually trying to send to non-existent email addresses affects Notify's relationship with Amazon at all?

I'm also a little worried that it really isn't the user's intent to be created in the database at this point in the journey. I wouldn't expect a password reset, for example, to create new details in a database if I use a different email address to the one I signed up with by mistake; and this feels like a similar process. Users wouldn't expect to have the email address stored from this operation. That feels a bit wrong to me. Could we not return a generic "If you're in our database, we've sent you link" to every request?

We also risk confusing and annoying users - if a user signs up to notifications with  `emailaddress1` and then enters `emailaddress2` at this point by mistake, they get an email to their second email address with a link to empty subscriptions.

Also, has anyone looked at this from a GDPR perspective? - I don't know much about it, but seem to remember that it calls for data minimisation.

A minor disadvantage is we potentially end up mis-representing the number of subscribers we have in any stats we create (admittedly outside of bad actors it will be a very minor percentage). But we also have a bunch of orphaned subscribers.

kevindew: Interesting points @bevanloon 👍 

> How does a potential malicious user who decides to enter n number of email addresses that don't exist affect our relationship with Notify, and by extension their relationship with their provider? (i.e. we'll try to send lots of emails to non-existent email addresses).

I'm not too concerned about this as this doesn't introduce this situation. You could do this far worse by signing up to alerts with n number of addresses for instance

> There's a risk here of confusing and annoying users - if a user signs up to notifications with emailaddress1 and then enters emailaddress2 at this point by mistake, they get an email to their second email address with a link to empty subscriptions.

that's true though I'd argue that this is actually less confusing though than a "If you're in our database, we've sent you link" scenario.

> Have we looked at this from a GDPR perspective? - I don't know much about it, but seem to remember that it calls for data minimisation, which I expect one could argue that this isn't.

Not in huge detail. I'd suggest it might be appropriate to that we tell him that their email address is stored.

Where this all gets a bit complicated though and problems emerge with the "If you're in our database, we've sent you link" approach is the whole deactivation process of a user. If you unsubscribe from stuff and then come back later you might find you've not got an account which could be very confusing if you previously had one.

Where I think the sweet spot in this is if we improved the deactivation/nullify process of a user so that every user who interacts with an email does get registered but the process of removing them is more automatic than it currently is.

bevanloon: Oh dear. We seem to have backed ourselves into a bit of a corner with all this.

Good point about the signup - that's always been like that though, so we've kept status quo, this is new attack surface for us.

I guess which process is more confusing should come out of user research at some point.

> Where this all gets a bit complicated though and problems emerge with the "If you're in our database, we've sent you link" approach is the whole deactivation process of a user. If you unsubscribe from stuff and then come back later you might find you've not got an account which could be very confusing if you previously had one.

isn't this only the case once you've been nullified though? If you've just been deactivated, we've still got your email address. I agree it is a bit confusing though.

PR #562, opened 2018-03-23
Convenience method tweaks

kevindew:
This PR is off the back of diagnosing some issues in the console and intends
to make the process easier with convenience methods and more explicit naming.

See commits for further details.

### DIFF
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -2,7 +2,7 @@ require 'csv'
 
 namespace :manage do
   def change_email_address(old_email_address:, new_email_address:)
-    subscriber = Subscriber.find_by("LOWER(address) = ?", old_email_address.downcase)
+    subscriber = Subscriber.find_by_address(old_email_address)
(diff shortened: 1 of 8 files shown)


lib/tasks/manage.rake
```diff
@@ -2,7 +2,7 @@ require 'csv'
 
 namespace :manage do
   def change_email_address(old_email_address:, new_email_address:)
-    subscriber = Subscriber.find_by("LOWER(address) = ?", old_email_address.downcase)
+    subscriber = Subscriber.find_by_address(old_email_address)
```
bevanloon: is it worth using `find_by!` and relying on the `RecordNotFound` error rather than rolling our own on the next line?

kevindew: Could well do. I assumed it was for a clearer error message, but didn't really ponder how useful/superfluous it is.

bevanloon: cool. no matter at this point anyway. the code works.

PR #563, opened 2018-03-23
Add API endpoint to unsubscribe a subscriber completely

rubenarakelyan:
This commit adds an API endpoint at `DELETE /subscribers/:subscriber_id` that unsubscribes a subscriber from all their subscriptions and marks their account as deactivated.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,6 +10,7 @@ Rails.application.routes.draw do
     resources :subscriptions, only: %i[create show update]
 
     patch "/subscribers/:id", to: "subscribers#change_address"
+    delete "/subscribers/:id", to: "unsubscribe#unsubscribe_all"
     get "/subscribers/:id/subscriptions", to: "subscribers#subscriptions"
 
     get "/healthcheck", to: "healthcheck#check"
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -15,5 +15,6 @@
     get "/healthcheck", to: "healthcheck#check"
 
     post "/unsubscribe/:id", to: "unsubscribe#unsubscribe"
+    post "/unsubscribe-all/:id", to: "unsubscribe#unsubscribe_all"
```
thomasleese: Not sure it matters now since we've already got one in production, but I think a better route structure for both of these would be:

```ruby
delete "/subscriptions/:id", to: "unsubscribe#unsubscribe"
delete "/subscriber/:id", to: "unsubscribe#unsubscribe_all"
```

rubenarakelyan: That does look nicer - I can make this one a `DELETE` on subscriber, but would that be OK in terms of diverging?

rubenarakelyan: Haha OK

rubenarakelyan: Done

PR #565, opened 2018-03-26
Add subscription management links to emails

1pretz1:
By clicking the 'View and manage your subscriptions link' in an email, the user will be redirected to a landing page 'email/authenticate', where a query string will also be passed in of the users email address. Keith made some changes, removing `nbsp;`, in favour of `\u00A0`. After a discussion with Kevin, decided to make the format standardised and convert the existing `&nbsp;` to `\u00A0` (Unicode).


keithiopia:
@1pretz1 the unicode substitution was to stop `&nbsp;` from appearing in plain text emails.

PR #580, opened 2018-04-05
Ignore null addresses when sanitising

bevanloon:
The sanitise script should not try to anonymise nullified Subscriber
records because this violates the not-null constraint on the addresses
table.

### DIFF
--- a/lib/data_hygiene/anonymise_email_addresses.sql
+++ b/lib/data_hygiene/anonymise_email_addresses.sql
@@ -21,8 +21,9 @@ WHERE created_at < current_timestamp - interval '1 day';
 CREATE TABLE addresses (id SERIAL, address VARCHAR NOT NULL);
 
 # Copy all email addresses into the table.
+# Ignore nulled out subscriber addresses.
 INSERT INTO addresses (address)
-  SELECT address FROM subscribers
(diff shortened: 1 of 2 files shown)


lib/data_hygiene/anonymise_email_addresses.sql
```diff
@@ -21,8 +21,9 @@ WHERE created_at < current_timestamp - interval '1 day';
 CREATE TABLE addresses (id SERIAL, address VARCHAR NOT NULL);
 
 # Copy all email addresses into the table.
+# Ignore nulled out subscriber addresses.
 INSERT INTO addresses (address)
-  SELECT address FROM subscribers
+  SELECT address FROM subscribers WHERE address IS NOT NULL
```
thomasleese: Will this mean that for subscribers that are nullified, we will see `Email`s with non-anonymised addresses?

bevanloon: I don't think so - we aren't joining on `subscribers`  when we anonymise `addresses` in the `emails` table.


```
UNION DISTINCT
SELECT address FROM emails
```

will take care of any addresses in `emails` table.

thomasleese: Ah yes, good point!

PR #582, opened 2018-04-06
Create worker to mark digest runs as complete

boffbowsh:
After every digest email is sent, the worker checks if it was the last
email to be sent for that digest run. If it is, it marks the run as
complete. The check required is expensive, between 2 to 4 seconds in
the database while under load. We've been unable to make Postgres use
an index to speed up the operation.

This commit adds a new worker that finds all runs that haven't yet been
marked as completed, checks if all the subscribers are marked as
completed then uses the most recent completion time as the run's
completion time. Running this every minute will still give us accurate
statistics and will greatly reduce the load on Postgres.

### DIFF
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -6,10 +6,13 @@ class DigestRun < ApplicationRecord
   has_many :digest_run_subscribers, dependent: :destroy
   has_many :subscribers, through: :digest_run_subscribers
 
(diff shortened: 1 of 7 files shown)


boffbowsh:
I tried many different types of query and I just couldn't get Postgres to use either index (compound and the `digest_run_id` by itself)

PR #586, opened 2018-04-11
Bulk emails

thomasleese:
This adds the ability to bulk send out an email to many subscriber lists.

Once this is deployed, there will be a custom Jenkins task set up to run this rake task with a more friendly user interface.

[Trello Card](https://trello.com/c/e90wvivv/49-build-tooling-for-bulk-contacting-email-subscribers-for-a-given-subscriber-list-2)

### DIFF
--- a/lib/tasks/bulk.rake
+++ b/lib/tasks/bulk.rake
@@ -1,0 +1,13 @@
+namespace :bulk do
+  desc "Send a bulk email to many subscriber lists"
+  task :email, %i(subject body) => :environment do |_t, args|
+    subscriber_list_ids = args.extras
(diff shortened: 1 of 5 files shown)


lib/tasks/bulk.rake
```diff
@@ -0,0 +1,13 @@
+namespace :bulk do
+  desc "Send a bulk email to many subscriber lists"
+  task :email, %i(subject body) => :environment do |_t, args|
+    subscriber_list_ids = args.extras
```
kevindew: Have you experimented with how this might get populated? I wonder if we're planning to do pre-processing on this or plan to have something through that we process into the list?

PR #593, opened 2018-05-02
Update Land registry titles

deborahchua:
In email subscription alerts about Land Registry, there is some inconsistency where some titles are displayed as 'Land registry' or 'HM Land Registry'.

This migration finds all titles containing 'Land Registry' and replaces them with 'HM Land Registry', and skips those which are already correct.

Trello card: [Land Registry title fields in email-alert-api SubscriptionList are inconsistent - make them all HM Land Registry (2)](https://trello.com/c/AB2qHyp0/116-land-registry-title-fields-in-email-alert-api-subscriptionlist-are-inconsistent-make-them-all-hm-land-registry-2)

Before:
land_registry_subscriptionlist_title

### DIFF
--- a/db/migrate/20180501155601_update_land_registry_titles.rb
+++ b/db/migrate/20180501155601_update_land_registry_titles.rb
@@ -1,0 +1,10 @@
+class UpdateLandRegistryTitles < ActiveRecord::Migration[5.2]
+  def change
+    list = SubscriberList.where("title LIKE '%Land Registry%' AND title NOT LIKE '%HM Land Registry%'")
+
+    list.each do |item|
+      item.title.sub! 'Land Registry', 'HM Land Registry'
+      item.save!
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180321125113) do
+ActiveRecord::Schema.define(version: 2018_05_01_155601) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"


db/migrate/20180501155601_update_land_registry_titles.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateLandRegistryTitles < ActiveRecord::Migration[5.2]
+  def change
+    list = SubscriberList.where("title like '%Land Registry%'")
+
+    list.each do |item|
+      next if item.title.include?("HM")
```
barrucadu: I'd change that to `"HM Land Registry"` to be more explicit.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180321125113) do
+ActiveRecord::Schema.define(version: 2018_05_01_155601) do
```
barrucadu: How come the date format has changed?

deborahchua: I wondered about this too, I'm not sure either!

db/migrate/20180501155601_update_land_registry_titles.rb
barrucadu: Or maybe the where to something like `title like '%Land Registry%' and title not like '%HM Land Registry%'`.

deborahchua: I didn't realise I could do that - thanks @barrucadu !

db/migrate/20180501155601_update_land_registry_titles.rb
```diff
@@ -0,0 +1,10 @@
+class UpdateLandRegistryTitles < ActiveRecord::Migration[5.2]
+  def change
+    list = SubscriberList.where("title like '%Land Registry%' and title not like '%HM Land Registry%'")
```
thomasleese: I like to make the SQL part (i.e. `like` and `not like`) in capitals, but that's just personal preference really.

PR #595, opened 2018-05-04
Add an export of EU "Living in" taxons

thomasleese:
We need to send the FCO statistics on the EU "Living in" taxons regularly. Unfortunately, there's currently no way for them to self serve this data so we have to add code to do this for them.

We're going to merge this feature now so we can give the results to FCO, [but in the near future we will move this into Content Tagger](https://trello.com/c/3fxFsB93/360-add-subscriber-list-count-stats-to-taxons-in-content-tagger) and remove these Rake tasks from here.

[Trello Card](https://trello.com/c/KI1ebOiL/114-get-email-subscriptions-figures-for-different-subscriber-lists-2)

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -1,24 +1,12 @@
-require 'csv'
-
 namespace :export do
-  def present_subscriber_list(list_id)
-    list = SubscriberList.find(list_id)
-    { subscriber_list_id: list_id, title: list.title, count: list.subscribers.count }
-  rescue ActiveRecord::RecordNotFound
-    warn "could not fetch record for #{list_id}"
-  end
-
   desc "Export the number of subscribers for a collection of lists as a csv, accepts multiple arguments"
   task :csv, [:subscriber_list_id] => :environment do |_, args|
-    CSV($stdout, headers: %i[subscriber_list_id title count], write_headers: true) do |csv|
-      rows = args.to_a.map { |list_id| present_subscriber_list(list_id) }
-      rows.compact.each { |row| csv << row }
-    end
+    DataExporter.new.export_csv(args.to_a)
   end
 
   desc "Export the number of subscribers for a list"
   task :count, [:subscriber_list_id] => :environment do |_, args|
-    list = present_subscriber_list(args[:subscriber_list_id])
+    list = DataExporter.new.present_subscriber_list(args[:subscriber_list_id])
     puts list[:count] unless list.nil?
   end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/export.rake
```diff
@@ -1,24 +1,11 @@
...
-  end
-
-  desc "Export the number of subscribers for a collection of lists as a csv, accepts multiple arguments"
-  task :csv, [:subscriber_list_id] => :environment do |_, args|
-    CSV($stdout, headers: %i[subscriber_list_id title count], write_headers: true) do |csv|
-      rows = args.to_a.map { |list_id| present_subscriber_list(list_id) }
-      rows.compact.each { |row| csv << row }
-    end
+  desc "Export the number of subscriptions for a collection of lists as a csv, accepts multiple arguments"
+  task :csv_from_ids, [] => :environment do |_, args|
```
elliotcm: This can just be `task csv_from_ids: :environment do`

lib/tasks/export.rake
```diff
@@ -1,24 +1,11 @@
...
-    warn "could not fetch record for #{list_id}"
-  end
-
-  desc "Export the number of subscribers for a collection of lists as a csv, accepts multiple arguments"
-  task :csv, [:subscriber_list_id] => :environment do |_, args|
-    CSV($stdout, headers: %i[subscriber_list_id title count], write_headers: true) do |csv|
-      rows = args.to_a.map { |list_id| present_subscriber_list(list_id) }
-      rows.compact.each { |row| csv << row }
-    end
+  desc "Export the number of subscriptions for a collection of lists as a csv, accepts multiple arguments"
```
elliotcm: If I'm understanding this right could we change this to say "accepts multiple list IDs" or something?  Took me some double-checking to be sure what args to give.

lib/tasks/export.rake
```diff
@@ -1,24 +1,11 @@
...
+  task :csv_from_ids, [] => :environment do |_, args|
+    DataExporter.new.export_csv_from_ids(args.to_a)
   end
 
-  desc "Export the number of subscribers for a list"
-  task :count, [:subscriber_list_id] => :environment do |_, args|
-    list = present_subscriber_list(args[:subscriber_list_id])
-    puts list[:count] unless list.nil?
+  desc "Export the number of subscriptions for the 'Living in' taxons for EU countries"
+  task :csv_from_living_in_eu, [] => :environment do
```
elliotcm: Same as above for the environment call.

PR #597, opened 2018-05-10
Handle capitalisation differences in anonymisation

cbaines:
For email addresses. There can be duplicates in the emails table when
you normalise the capitalisation, this breaks creating the index as
it's a unique index on the lower form of the address. To fix this,
just use the unchanged address for the index.

### DIFF
--- a/lib/data_hygiene/anonymise_email_addresses.sql
+++ b/lib/data_hygiene/anonymise_email_addresses.sql
@@ -28,7 +28,7 @@ INSERT INTO addresses (address)
   SELECT address FROM emails;
 
 # Index the table so we can efficiently lookup addresses.
-CREATE UNIQUE INDEX addresses_index ON addresses (lower(address));
+CREATE UNIQUE INDEX addresses_index ON addresses (address);
 
 # Set subscribers.address from the auto-incremented id in addresses table.
 UPDATE subscribers s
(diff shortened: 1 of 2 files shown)


kevindew:
Hello - so we purposefully have the lower index so we don't have multiple accounts for different cases of the same address. As per your test foo@example.com and Foo@example.com are the same address so we want the same account.

Do you have details of the issue that led to this?

cbaines:
> Actually, we shouldn't need to do this because the addresses come from the subscribers table which should already not allow emails with multiple cases from being in there.

You're right that the `subscribers` table doesn't, but the addresses come from both the `subscribers` and `emails` tables, and the `emails` table does contain duplicates when you use the lower form of the addresses.

kevindew:
I hadn't quite realised this was the sanitisation script when I was first looking at this - sorry /facepalm

Yeah so I see how this could be a problem and this approach seems like one of the least painful ways to solve it.

Do we get any performance boosts from using a unique index rather than a normal index too? Since we don't actually care if it's unique I wonder if that's just a nagging point where things can break.

cbaines:
> Do we get any performance boosts from using a unique index rather than a normal index too? Since we don't actually care if it's unique I wonder if that's just a nagging point where things can break.

This is the first time I've seen this, but I'm not entirely sure what allowing multiple rows with the same address would mean for the `UPDATE` parts.

From looking at the docs, it cautions against such circumstances:

```
When a FROM clause is present, what essentially happens is that the
target table is joined to the tables mentioned in the from_list, and
each output row of the join represents an update operation for the
target table. When using FROM you should ensure that the join produces
at most one output row for each row to be modified. In other words, a
target row shouldn't join to more than one row from the other
table(s). If it does, then only one of the join rows will be used to
update the target row, but which one will be used is not readily
predictable.
```
https://www.postgresql.org/docs/9.3/static/sql-update.html

kevindew:
Sure that makes sense - thanks for passing in the documentation :)

PR #598, opened 2018-05-10
Add more countries to the living in Europe exporter

thomasleese:
FCO have asked that a few more countries be added to the list and it is no longer specific to EU countries but also across Europe.

This follows on from #595 which added the data exporter, and once the ability to self-serve this from content tagger has been added we can remove this code.

[Trello Card](https://trello.com/c/3fxFsB93/360-add-subscriber-list-count-stats-to-taxons-in-content-tagger)

### DIFF
--- a/lib/data_exporter.rb
+++ b/lib/data_exporter.rb
@@ -1,13 +1,14 @@
 class DataExporter
   CSV_HEADERS = %i(id title count).freeze
 
-  EU_COUNTRIES = %w(
+  EUROPEAN_COUNTRIES = %w(
     austria belgium bulgaria croatia cyprus czech-republic denmark estonia finland france germany greece hungary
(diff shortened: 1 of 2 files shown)


lib/data_exporter.rb
```diff
@@ -1,13 +1,14 @@
 class DataExporter
   CSV_HEADERS = %i(id title count).freeze
 
-  EU_COUNTRIES = %w(
+  EUROPEAN_COUNTRIES = %w(
     austria belgium bulgaria croatia cyprus czech-republic denmark estonia finland france germany greece hungary
     ireland italy latvia lithuania luxembourg malta netherlands poland portugal slovakia slovenia spain sweeden
```
deborahchua: Is Sweden intentionally meant to be spelt like `sweeden` here?

thomasleese: Nope! Good catch, I'll raise a PR for it now.

PR #604, opened 2018-05-31
Expose active subscription count in subscriber list json

barrucadu:
[Trello card](https://trello.com/c/3fxFsB93/201-add-subscriber-list-count-stats-to-taxons-in-content-tagger)


bevanloon:
We probably want to update `gds-api-adapters` as well.....e.g. https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/email_alert_api.rb#L277 and https://github.com/alphagov/gds-api-adapters/blob/master/test/email_alert_api_test.rb#L208

PR #608, opened 2018-06-07
Add ability to export subscription counts up to a certain date

thomasleese:
This allows us to see the active subscription count for a particular subscriber list on a certain date.

[Trello Card](https://trello.com/c/aCTaSZCS/208-get-email-subscription-figures-russia-travel-advice)


thomasleese:
https://github.com/alphagov/email-alert-api/blob/master/db/schema.rb#L170 Should ensure that situation doesn’t happen.

PR #612, opened 2018-06-18
Remove uniq validation for title in SubscriberList

1pretz1:
https://trello.com/c/C9Dr0Acy/167-investigate-a-solution-for-certain-taxons-signup-errors

Signing up for email alerts for certain taxon's throws an error due to the uniq
constraint on title in the SubscriberList model. This validation was
needed when Email was using GovDelivery, but now serves no purpose.

The removal of this validation will allow users to sign up for email
alerts for taxons which share the same title as other legacy taxonomies
instead of returning an error.

### DIFF
--- a/spec/integration/create_subscriber_list_spec.rb
+++ b/spec/integration/create_subscriber_list_spec.rb
@@ -80,11 +80,11 @@ RSpec.describe "Creating a subscriber list", type: :request do
       expect(response.status).to eq(422)
     end
 
-    it "returns an error if creating the same topic" do
-      create_subscriber_list(title: "oil and gas", tags: { topics: ["oil-and-gas/licensing"] })
-      create_subscriber_list(title: "oil and gas", tags: { topics: ["oil-and-gas/licensing"] })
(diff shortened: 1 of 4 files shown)


spec/integration/create_subscriber_list_spec.rb
```diff
@@ -80,11 +80,11 @@
       expect(response.status).to eq(422)
     end
 
-    it "returns an error if creating the same topic" do
+    it "successfully creates two SubscriberList objects with the same title" do
       create_subscriber_list(title: "oil and gas", tags: { topics: ["oil-and-gas/licensing"] })
       create_subscriber_list(title: "oil and gas", tags: { topics: ["oil-and-gas/licensing"] })
```
cbaines: I'd perhaps change the test here to reflect a more reasonable scenario, e.g. having the 2nd subscriber list be a policy area subscription for example?

1pretz1: Agreed 👍

PR #613, opened 2018-06-18
Add dockerfile

kevindew:
Trello: https://trello.com/c/U9nNDFrH/222-decide-on-whether-e2e-tests-should-include-email-alert-api-following-on-from-content-tagger-work

This will allow this app to be part of [publishing-e2e-tests](https://github.com/alphagov/publishing-e2e-tests)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM ruby:2.5.1
+RUN apt-get update -qq && apt-get install -y \
+  build-essential \
+  libpq-dev \
+  libxml2-dev \
+  libxslt1-dev
+RUN apt-get clean
+RUN gem install foreman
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ruby:2.5.1
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential libpq-dev libxml2-dev libxslt1-dev && apt-get clean
```
thomasleese: I think it might be better to split the upgrade line and the installation line into two runs because this will allow Docker to cache the upgraded image and we can use it as a base when building other images (assuming the RUN line is the same string exactly).

PR #619, opened 2018-06-26
Add a healthcheck for subscription content processing time

barrucadu:
This also creates an index on `SubscriptionContent.created_at`, as we query that field.

Tests are failing because they're not running the migrations.

---

[Trello card](https://trello.com/c/yGEsHaY9/253-email-alert-api-healthcheck-should-monitor-more-critical-areas)

### DIFF
--- a/app/models/healthcheck/subscription_content_healthcheck.rb
+++ b/app/models/healthcheck/subscription_content_healthcheck.rb
@@ -1,0 +1,50 @@
+module Healthcheck
+  class SubscriptionContentHealthcheck
+    def name
+      :subscription_content
+    end
+
+    def status
+      if count_subscription_contents(critical_latency).positive?
+        :critical
+      elsif count_subscription_contents(warning_latency).positive?
+        :warning
+      else
+        :ok
+      end
+    end
+
+    def details
+      {
+        critical: count_subscription_contents(critical_latency),
+        warning: count_subscription_contents(warning_latency),
+      }
+    end
+
+  private
+
+    def count_subscription_contents(age)
+      # The `merge(Subscription.active)` check is because there is a
+      # race condition in email generation: if someone unsubscribes
+      # after the `ContentChange` has been processed but before the
+      # generated `SubscriptionContent`s have been, then those
+      # `SubscriptionContent`s will never get an email associated with
+      # them - this is the correct behaviour, we don't want to email
+      # people who have unsubscribed.
+      SubscriptionContent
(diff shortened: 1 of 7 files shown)


app/models/healthcheck/subscription_content_healthcheck.rb
```diff
@@ -0,0 +1,50 @@
...
+      # after the `ContentChange` has been processed but before the
+      # generated `SubscriptionContent`s have been, then those
+      # `SubscriptionContent`s will never get an email associated with
+      # them - this is the correct behaviour, we don't want to email
+      # people who have unsubscribed.
+      SubscriptionContent
+        .where("subscription_contents.created_at < ?", age.ago)
+        .where(email: nil)
+        .joins(:subscription)
+        .where("subscriptions.ended_at": nil)
```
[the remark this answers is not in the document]
barrucadu: I've now changed this, but couldn't double check it because integration is a tad broken.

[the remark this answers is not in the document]
thomasleese: Cool, let me know when it's tested and I'll approve! 👍

barrucadu:
@thomasleese I've checked the query on integration, and it seems right.  One thing does confuse me though: I enabled the healthcheck in 54f15bd without adding it to the test output (as in https://github.com/alphagov/email-alert-api/commit/dbab295651506d65cc153761d14326e7305fa47c#diff-bc9465c23bcd50fdfafc52817405e509), so I don't really know why the tests are passing.

thomasleese:
@barrucadu ah, that would be because the test checks that the hash is `include`d not that it is `exact`ly the same.

config/routes.rb
```diff
@@ -1,4 +1,4 @@
-Rails.application.routes.draw do
+Rails.application.routes.draw do # rubocop:disable Metrics/BlockLength
```
thomasleese: A better alternative to this might be to do something like https://github.com/alphagov/collections-publisher/commit/5579586fee81eda19639bbde2dc21f0002f16806 but for `routes.rb`.

barrucadu: That makes it fail with the same cop on `spec/models/healthcheck/subscription_content_healthcheck.rb` for some reason.  If I also exclude that file, a bunch more `spec` files start to fail.

thomasleese: Oh... that's weird. Fine with this comment then 👍

barrucadu:
I think this is done now.

PR #621, opened 2018-06-27
Temporarily disable subscription content healthcheck

thomasleese:
The index added in 0e4be2ac95b19eb23c54dd2850538a3353c1cb6c is taking too long to run during a deploy and times out. So I have to deploy the app without migrations first and then run the migrations manually afterwards.

I want to temporarily disable this healthcheck so it doesn't try and run the query without the index which would be very slow.


elliotcm:
Should we just delete these lines and revert when you want it back?  Always a bit wary of leaving commented out stuff.

PR #627, opened 2018-07-02
Archive emails to S3

kevindew:
Trello: https://trello.com/c/FM3MCuAl/255-investigate-email-alert-api-database-size-and-push-email-archives-to-s3

This PR sets up archiving emails to S3 when the EmailArchiveWorker runs and creates a rake job that can do archiving. 

Further details can be found in individual commits.

### DIFF
--- a/app/services/s3_email_archive_service.rb
+++ b/app/services/s3_email_archive_service.rb
@@ -18,14 +18,14 @@ private
       # we group by date in this way to create partitions for s3/athena
       # these are grouped in case dates span more than one day
       Date.parse(
-        item.fetch(:finished_sending_at)
+        item.fetch(:finished_sending_at_utc)
       ).strftime("year=%Y/month=%m/date=%d")
     end
   end
 
   def send_to_s3(prefix, records)
-    records = records.sort_by { |r| r.fetch(:finished_sending_at) }
-    last_time = records.last[:finished_sending_at]
+    records = records.sort_by { |r| r.fetch(:finished_sending_at_utc) }
+    last_time = records.last[:finished_sending_at_utc]
     obj = bucket.object(object_name(prefix, last_time))
     obj.put(
       body: object_body(records),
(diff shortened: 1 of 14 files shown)


app/workers/email_archive_worker.rb
```diff
@@ -24,47 +24,34 @@ def perform
 
   def archive_batch(batch)
     archived_at = Time.zone.now
+    exported_at = nil
 
-    import_batch(batch, archived_at)
-    Email.where(id: batch.pluck(:id)).update_all(archived_at: archived_at)
-  end
+    if ENV["EMAIL_ARCHIVE_S3_ENABLED"]
```
thomasleese: Might be better to do `.include?` here since `EMAIL_ARCHIVE_S3_ENABLED` with a value of `""` might be confusing.

app/services/s3_email_archive_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def bucket
+    @bucket ||= begin
+                  s3 = Aws::S3::Resource.new
+                  s3.bucket(ENV.fetch("EMAIL_ARCHIVE_S3_BUCKET"))
+                end
+  end
+
+  def object_name(prefix, last_time)
+    uuid = SecureRandom.uuid
```
thomasleese: If this fails half way through, can we end up in a state where half the archives are uploaded, and then the transaction rolls back and when it runs again we get everything uploaded so end up with double the archives on half of the dates?

I haven't thought about it too much, but can we build this `uuid` from some sort of hash of the batch to avoid this situation?

I realise that in most cases, only one days worth of emails will be archived so this won't be a problem though.

app/workers/email_archive_worker.rb
kevindew: Good point, I'll switch to that.

app/services/s3_email_archive_service.rb
kevindew: Yeah this is a problem we have with every upload to S3 - if we think it's failed but it actually succeeded we may end up duplicating the data. The only way I can think around this is having a very predictable batching so that we can generate a reliable hash but this seems difficult to do relative to the risk involved.

PR #632, opened 2018-07-09
Analytics documentation and breaking up readme

kevindew:
Trello: https://trello.com/c/FM3MCuAl/255-investigate-email-alert-api-database-size-and-push-email-archives-to-s3

This creates an an analytics documentation page which explains how to find data with AWS athena and then breaks up the readme so that it's easier to bring parts of this into developer-docs and make the readme less of a monster.


doc/analytics.md
```diff
@@ -0,0 +1,172 @@
...
+developer docs. To access the production data you will need to use the
+`govuk-infrastructure-production` account, once there you can head to
+[athena](https://eu-west-1.console.aws.amazon.com/athena) and select the
+`email_alert_api_archive` database.
+
+The data stored in Athena is the email id; when the sending process completed;
+whether the email sent; the email subject of the email; the associations to
+content change and digests; and timestamps for when the email was created and
+archived. The data is arranged by partitions of the date when the email
+finished sending - **It is vastly cheaper and faster to query with partitions**.
```
boffbowsh: I think some expansion of what partitions means in practice would be useful, something like "You should always have a `WHERE` clause that limits the query to at most 7 days unless you know what you're doing". And possibly explain why it's so much cheaper and faster.

kevindew:
@boffbowsh no worries, I've just updated this now. Thanks for reviewing.

PR #639, opened 2018-07-18
Change the batching approach in EmailArchiveWorker

cbaines:
Don't use in_batches, as this sorts by the id (which is a UUID) which
adds unnecessary overhead. As the emails are archived, the
EmailArchiveQuery will ignore them when fetching the next batch, so
the approach can be simplified to just fetch batches in a loop, and
then terminate after the final batch is processed.

### DIFF
--- a/app/workers/email_archive_worker.rb
+++ b/app/workers/email_archive_worker.rb
@@ -4,6 +4,7 @@ class EmailArchiveWorker
   sidekiq_options queue: :cleanup
 
   LOCK_NAME = "email_archive_worker".freeze
+  BATCH_SIZE = 1000
 
   def perform
     return unless ENV.include?("EMAIL_ARCHIVE_S3_ENABLED")
@@ -11,10 +12,13 @@ class EmailArchiveWorker
       start_time = Time.zone.now
       archived_count = 0
 
-      EmailArchiveQuery.call.in_batches do |batch|
-        Email.transaction do
-          archived_count += archive_batch(batch.as_json)
-        end
+      loop do
+        batch = EmailArchiveQuery.call.limit(BATCH_SIZE).as_json
+        break if batch.empty?
+
+        archived_count += archive_batch(batch)
+
+        break if batch.length < BATCH_SIZE
       end
 
       log_complete(archived_count, start_time, Time.zone.now)


app/workers/email_archive_worker.rb
```diff
@@ -4,17 +4,20 @@ class EmailArchiveWorker
...
       archived_count = 0
 
-      EmailArchiveQuery.call.in_batches do |batch|
-        Email.transaction do
-          archived_count += archive_batch(batch.as_json)
-        end
+      loop do
+        batch = EmailArchiveQuery.call.limit(BATCH_SIZE).as_json
+
+        archived_count += archive_batch(batch)
```
kevindew: I think we might need something here (or in the `archive_batch` method) to deal with the scenario that the query returns 0 results.

app/workers/email_archive_worker.rb
```diff
@@ -4,17 +4,20 @@ class EmailArchiveWorker
...
     Email.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
       start_time = Time.zone.now
       archived_count = 0
 
-      EmailArchiveQuery.call.in_batches do |batch|
-        Email.transaction do
-          archived_count += archive_batch(batch.as_json)
-        end
+      loop do
+        batch = EmailArchiveQuery.call.limit(BATCH_SIZE).as_json
```
kevindew: Something that would be nice to do (if it didn't impact performance too much) is to order this by finished_sending_at ASC as that would produce more consistent S3 files. 

Not a must have naturally since we didn't have it before when using `in_batches`

cbaines:
> This seems sound. Have you benchmarked the difference in query performance at all?

Only on the individual queries, not overall.

> that seems to be just a query for id, whereas this generates a pretty meaty query with various array agg joins. Which makes me wonder if a second query is being executed unnecessarily here and also confuses me a bit that the meatier query isn't the thing throwing performance red flags.

So, peeking in to `in_batches` [1], it'll fetch the ids, which is the query we're seeing, and then pass a relation that filters for those ids through to the block. This is both the expensive bit, and is also avoided by the approach here.

It's expensive as the results are ordered by the id, compare these two queries [2]. My guess why the ordering performs so poorly is that there can be lots of emails that need archiving, so PostgreSQL may need to look through 100,000 emails to try and find the 1000 with the lowest ids. The actual fields in the EmailArchiveQuery are comparatively easy to select, as in the worst case it requires finding 1000 records by id, compared to considering say 100,000 records.

1: https://github.com/rails/rails/blob/a481603480ba0db44c85ab1ef64f3c0a34c16457/activerecord/lib/active_record/relation/batches.rb#L229-L230

2: 
```
email-alert-api_production=# EXPLAIN ANALYSE SELECT  "emails"."id" FROM "emails" WHERE "emails"."archived_at" IS NULL AND "emails"."status" != 0 LIMIT 1000;
                                                                       QUERY PLAN                                                                       
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.56..3897.09 rows=1000 width=16) (actual time=0.029..1.635 rows=1000 loops=1)
   ->  Index Scan using index_emails_on_archived_at on emails  (cost=0.56..703693.63 rows=180595 width=16) (actual time=0.028..1.566 rows=1000 loops=1)
         Index Cond: (archived_at IS NULL)
         Filter: (status <> 0)
 Total runtime: 1.675 ms
(5 rows)
```
```
email-alert-api_production=# EXPLAIN ANALYSE SELECT  "emails"."id" FROM "emails" WHERE "emails"."archived_at" IS NULL AND "emails"."status" != 0 ORDER BY "emails"."id" ASC LIMIT 1000;
                                                                   QUERY PLAN                                                                   
------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.56..185371.84 rows=1000 width=16) (actual time=328.684..51402.914 rows=1000 loops=1)
   ->  Index Scan using emails_pkey on emails  (cost=0.56..33477126.90 rows=180595 width=16) (actual time=328.682..51402.268 rows=1000 loops=1)
         Filter: ((archived_at IS NULL) AND (status <> 0))
         Rows Removed by Filter: 64076
 Total runtime: 51403.256 ms
(5 rows)
```

app/workers/email_archive_worker.rb
cbaines: Good spot, I've added another `break`. Let me know what you think?

app/workers/email_archive_worker.rb
cbaines: > Something that would be nice to do (if it didn't impact performance too much) is to order this by finished_sending_at ASC as that would produce more consistent S3 files.

Ordering within the batch should be fine in terms of performance, but I don't know if that's useful. But I think that ordering overall is too expensive, this is just based on my trial query not yet finishing...

kevindew: Cool - yeah not necessary to order in the batch (that actually happens later at archive to s3 part if I remember correctly.

cbaines:
> It surprises me how much they differ since their similar indexes and (if I remember right) PostgreSQL stores the id index in order confused

Looking at `iotop, I think it might be I/O related due to the size of the table. Adding an ordering requires much more I/O, even using indexes it seems.

PR #640, opened 2018-07-19
Remove in_batches from the EmailDeletionWorker

cbaines:
Email uses a UUID field as the id, and in_batches orders the results
by the id field. This is very expensive, and unnecessary.

Just delete the rows directly, to make this faster, and less expensive
for PostgreSQL to handle.


cbaines:
Very similar to https://github.com/alphagov/email-alert-api/pull/639 just a different worker.

PR #641, opened 2018-07-23
Add marked_as_spam field to emails

binaryberry:
This will be set when a user marks an email as spam. We are including this as
part of the fields that will be archived to s3 so we can get the list of emails
that have been marked as spam most often.

Paired with @thomasleese 

https://trello.com/c/DRM82ETh/328-auto-unsubscribe-email-alert-api-users-who-report-spam


app/presenters/email_archive_presenter.rb
```diff
@@ -14,6 +14,7 @@ def for_s3(record, archived_at)
       created_at_utc: record.fetch("created_at").utc.strftime(S3_DATETIME_FORMAT),
       finished_sending_at_utc: record.fetch("finished_sending_at").utc.strftime(S3_DATETIME_FORMAT),
       id: record.fetch("id"),
+      marked_as_spam: true,
```
rubenarakelyan: Should this be hardcoded as `true`?

thomasleese: Oops... it should not be!

binaryberry: Well spotted - now amended to the more sensible `record.fetch("marked_as_spam")`.

I've also changed the presenter test to expect false as a value for `marked_as_spam` as I assume we are testing the default non-spam case here.

binaryberry:
@thomasleese it's not slow 

```
14:16:38 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] Migrating to AddMarkedAsSpamToEmail (20180723091112)
14:16:38 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] == 20180723091112 AddMarkedAsSpamToEmail: migrating ===========================
14:16:38 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] -- add_column(:emails, :marked_as_spam, :boolean, {:default=>false})
14:18:18 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] -> 100.6973s
14:18:18 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] == 20180723091112 AddMarkedAsSpamToEmail: migrated (100.6974s) ================
14:18:18 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] 
14:18:19     command finished in 102964ms
```
https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/5839/console

thomasleese:
@binaryberry Yeah, I agree with @cbaines. I wonder if we can get it to run faster if we remove the `default: false` and leave the field as `null` for the current records. We can always fix that later once the field is there.

binaryberry:
@thomasleese and @cbaines with default:nil it's down to three seconds
```
09:29:55 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] Migrating to AddMarkedAsSpamToEmail (20180723091112)
09:29:55 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] == 20180723091112 AddMarkedAsSpamToEmail: migrating ===========================
09:29:55 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] -- add_column(:emails, :marked_as_spam, :boolean, {:default=>nil})
09:29:56 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] -> 1.0081s
09:29:56 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] == 20180723091112 AddMarkedAsSpamToEmail: migrated (1.0082s) ==================
09:29:56 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] 
09:29:56     command finished in 3094ms
```
And if we follow by another migration that sets the column value to false, it's another 3 seconds.

```
09:48:44 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] Migrating to SetMarkedAsSpamDefaultValueToFalse (20180724093446)
09:48:44 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] == 20180724093446 SetMarkedAsSpamDefaultValueToFalse: migrating ===============
09:48:44 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] -- change_column_default(:emails, :marked_as_spam, false)
09:48:45 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] -> 1.0179s
09:48:45 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] == 20180724093446 SetMarkedAsSpamDefaultValueToFalse: migrated (1.0180s) ======
09:48:45 *** [err :: ip-10-1-4-107.eu-west-1.compute.internal] 
09:48:45     command finished in 3069ms
```

cbaines:
> Does that approach work for you? Any issues you can think of that might arise from doing both these migrations in the same PR?

Looks like the actual change takes around 1 second in each case, which seems fine, and I don't see any issue with doing them one after the other.

Note though that all the existing entries in the table will still have NULL as the value.

thomasleese:
Yeah, that looks good to me!

I’m fine with leaving the existing records as NULL, since we don’t actually know whether or not they would have been marked as spam so it’s better to leave it in this ambiguous state. In a couple of weeks once all the emails have been archived, we can set the field as NOT NULL and it should be pretty quick since none of the values will be NULL.

binaryberry:
Thanks! And true. I just tried to see how long it would take to change all the existing values in the table to false (`Email.update_all("marked_as_spam=false")`), and that's nearly 3 minutes, so probably best not to do it unless we need to!

PR #642, opened 2018-07-23
Unsubscribe users who marked us as spam

binaryberry:
This adds an endpoint that Notify will call when our users have marked one of our emails as spam. Calling that endpoint will unsubscribe the user from all our subscription lists.

Paired with @thomasleese 
https://trello.com/c/DRM82ETh/328-auto-unsubscribe-email-alert-api-users-who-report-spam

### DIFF
--- a/app/services/unsubscribe_service.rb
+++ b/app/services/unsubscribe_service.rb
@@ -8,13 +8,21 @@ module UnsubscribeService
       unsubscribe!(subscription.subscriber, [subscription], reason)
     end
 
+    def spam_report!(delivery_attempt)
+      subscriber_id = delivery_attempt.email.subscriber_id
+      subscriber = Subscriber.find(subscriber_id)
+      unsubscribe!(subscriber, subscriber.active_subscriptions, :marked_as_spam, delivery_attempt.email)
+    end
+
   private
 
-    def unsubscribe!(subscriber, subscriptions, reason)
+    def unsubscribe!(subscriber, subscriptions, reason, email = nil)
       ActiveRecord::Base.transaction do
         subscriptions.each do |subscription|
           subscription.end(reason: reason)
         end
+
+        email.update!(marked_as_spam: true) if email && reason == :marked_as_spam
 
         if !subscriber.deactivated? && no_other_subscriptions?(subscriber, subscriptions)
           subscriber.deactivate!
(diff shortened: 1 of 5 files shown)


app/presenters/email_archive_presenter.rb
```diff
@@ -14,6 +14,7 @@ def for_s3(record, archived_at)
       created_at_utc: record.fetch("created_at").utc.strftime(S3_DATETIME_FORMAT),
       finished_sending_at_utc: record.fetch("finished_sending_at").utc.strftime(S3_DATETIME_FORMAT),
       id: record.fetch("id"),
+      marked_as_spam: true,
```
bevanloon: I'm not very familiar with our archiving - does this mean every email sent to S3 has `marked_as_spam` set to `true` or is this hash merged and overridden somewhere else?

app/services/unsubscribe_service.rb
```diff
@@ -8,14 +8,22 @@ def subscription!(subscription, reason)
...
   private
 
-    def unsubscribe!(subscriber, subscriptions, reason)
+    def unsubscribe!(subscriber, subscriptions, reason, email = nil)
       ActiveRecord::Base.transaction do
         subscriptions.each do |subscription|
           subscription.end(reason: reason)
         end
 
+        email.update!(marked_as_spam: true) if email
```
bevanloon: This feels a little non-obvious to me - with the diff in front of me I can see that we only pass the `email` in if we have a spam report, but I wonder if your future selves will notice that or if we'll just get confused. Once we've added the extra argument in, we might be tempted to use it in the future and wouldn't expect the presence of an `email` to trigger a `marked_as_spam` from the method definition alone. Could we make the `if` more explicit perhaps with an `&& reason == :marked_as_spam` or somesuch?

app/presenters/email_archive_presenter.rb
bevanloon: sorry just noticed that you've fixed this in the PR this depends on. Probably just need a rebase.

binaryberry: No worries, thanks for spotting it! Done now!

app/services/unsubscribe_service.rb
binaryberry: Sure, that sounds clearer. Amended now.

PR #644, opened 2018-07-25
Slightly increase subscription content healthcheck thresholds

thomasleese:
We often see this go critical in the morning while it's sending out the digests and the 9:30 scheduled publishings. It seems fine to increase the threshold a bit since there isn't anything wrong but we're seeing a critical alert.


barrucadu:
Perhaps there needs to be thinking about what latency we guarantee (if any?) for travel advice emails.

PR #645, opened 2018-07-26
Export subscriber lists by slug

steventux:
https://trello.com/c/8HwDgibA/340-make-it-easier-to-get-subscriber-numbers-for-a-given-email-list-slug

Add rake task to export active subscriber counts by subscription list slug(s).

#### Integration benchmark

```
lists = SubscriberList.take(50)

irb(main):031:0> puts Benchmark.measure { SubscriberList.where(slug: lists.map(&:slug)) }
  0.001285   0.000000   0.001285 (  0.000288)
=> nil
irb(main):032:0> puts Benchmark.measure { SubscriberList.where(slug: lists.map(&:slug)) }
  0.000758   0.000000   0.000758 (  0.000283)
=> nil
```

versus ID based query...

```
irb(main):033:0> puts Benchmark.measure { SubscriberList.where(slug: lists.map(&:id)) }
  0.000204   0.000000   0.000204 (  0.000275)
=> nil
irb(main):034:0> puts Benchmark.measure { SubscriberList.where(slug: lists.map(&:id)) }
  0.000198   0.000072   0.000270 (  0.000266)
=> nil

```

### DIFF
--- a/lib/data_exporter.rb
+++ b/lib/data_exporter.rb
@@ -13,6 +13,10 @@ class DataExporter
     export_csv(living_in_europe_subscriber_lists)
   end
 
+  def export_csv_from_slugs(slugs)
+    export_csv(SubscriberList.where(slug: slugs))
+  end
+
 private
 
   CSV_HEADERS = %i(id title count).freeze
--- a/lib/data_exporter.rb
+++ b/lib/data_exporter.rb
@@ -9,12 +9,16 @@ class DataExporter
     export_csv(SubscriberList.where(id: ids), at: date)
   end
 
-  def export_csv_from_living_in_europe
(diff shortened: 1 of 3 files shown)


lib/data_exporter.rb
```diff
@@ -13,6 +13,10 @@ def export_csv_from_living_in_europe
     export_csv(living_in_europe_subscriber_lists)
   end
 
+  def export_csv_from_slugs(slugs)
```
thomasleese: I wonder if it's worth, just in case, having a similar `export_csv_from_slugs_at` method for getting results at a particular date - it wouldn't surprise me if we need that in the future.

PR #655, opened 2018-08-10
Add unpublish message feature

1pretz1:
Now when a message is received by the email-alert-api that a taxon has been unpublished, the users who are subscribed to that taxon will receive an email notifying them that their subscription has ended. 

Currently only taxon's are being processed if they've been unpublished but the plan is to also add this feature for legacy taxons like Policy Areas and Specialist Sectors etc.

(Emails wont actually be sent as is, only logged)

Trello;
https://trello.com/c/fPAplaaV/67-send-notifications-when-unpublishing-taxons

### DIFF
--- a/app/builders/unpublish_email_builder.rb
+++ b/app/builders/unpublish_email_builder.rb
@@ -1,0 +1,40 @@
+class UnpublishEmailBuilder
+  def self.call(*args)
+    new.call(*args)
+  end
+
+  def call(emails)
+    ids = Email.import!(email_records(emails)).ids
+    Email.where(id: ids)
+  end
+
+private
+
+  def email_records(emails)
+    emails.map do |email|
+      {
+        address: email.fetch(:address),
+        subject: email.fetch(:subject),
+        body: body(email.fetch(:subject), email.fetch(:address)),
+        subscriber_id: email.fetch(:subscriber_id)
+      }
+    end
+  end
+
+  def body(title, address)
+    <<~BODY
+      Your subscription to ‘#{title}’ no longer exists, as a result you will no longer receive emails
+      about this subject.
+
+      #{presented_manage_subscriptions_links(address)}
+
+      &nbsp;
+
+      ^Is this email useful? [Answer some questions to tell us more](https://www.smartsurvey.co.uk/s/govuk-email/?f=immediate).
+    BODY
+  end
+
+  def presented_manage_subscriptions_links(address)
+    ManageSubscriptionsLinkPresenter.call(address: address)
+  end
+end
--- a/app/controllers/unpublish_messages_controller.rb
+++ b/app/controllers/unpublish_messages_controller.rb
@@ -1,0 +1,16 @@
+class UnpublishMessagesController < ApplicationController
+  def create
+    UnpublishHandlerService.call(
+      unpublishing_params[:content_id]
+    )
+
+    render json: { message: "Unpublish message queued for sending" }, status: 202
+  end
+
+private
+
+  def unpublishing_params
+    permitted_params = params.permit!.to_h
+    permitted_params.slice(:content_id)
+  end
+end
--- a/app/services/unpublish_handler_service.rb
+++ b/app/services/unpublish_handler_service.rb
@@ -30,7 +30,7 @@ private
 
   def unsubscribe(subscriber_lists)
     subscriber_lists.each do |subscriber_list|
-      UnsubscribeService.subscriber_list!(subscriber_list, :unpublished)
+      UnsubscribeSubscriberListWorker.perform_async(subscriber_list.id, :unpublished)
     end
   end
 
--- a/app/services/unpublish_handler_service.rb
+++ b/app/services/unpublish_handler_service.rb
@@ -1,0 +1,102 @@
+class UnpublishHandlerService
+  def self.call(*args)
+    new.call(*args)
+  end
+
+  def call(content_id)
+    lists = subscriber_list(content_id)
+    taxon_subscriber_lists, other_subscriber_lists = split_subscriber_lists(lists)
+
+    taxon_emails = build_emails(taxon_subscriber_lists)
+    emails = UnpublishEmailBuilder.call(taxon_emails + courtesy_emails(taxon_emails))
+
+    queue_delivery_request_workers(emails)
+
+    log_taxon_emails(emails)
+    log_non_taxon_lists(other_subscriber_lists)
+
+    unsubscribe(taxon_subscriber_lists)
+  end
+
+private
+
+  def split_subscriber_lists(lists)
+    list_groupings = lists.group_by do |list|
+      list.links.has_key?(:taxon_tree) ? :taxon : :other
+    end
+
+    [list_groupings.fetch(:taxon, []), list_groupings.fetch(:other, [])]
+  end
+
+  def unsubscribe(subscriber_lists)
+    subscriber_lists.each do |subscriber_list|
+      UnsubscribeService.subscriber_list!(subscriber_list, :unpublished)
+    end
+  end
+
+  def queue_delivery_request_workers(emails)
+    emails.each do |email|
+      DeliveryRequestWorker.perform_async_in_queue(
+        email.id, queue: :delivery_immediate
+      )
+    end
+  end
+
+  # For this query to return the content id has to be wrapped in a double quote blame psql 9.3
+  def subscriber_list(content_id)
+    SubscriberList
+      .where(":id IN (SELECT json_array_elements((json_each(links)).value)::text)", id: "\"#{content_id}\"")
+      .includes(:subscribers)
+  end
+
+  def build_emails(subscriber_lists)
+    subscriber_lists.flat_map do |subscriber_list|
+      subscriber_list.subscribers.activated.map do |subscriber|
+        {
+          subject: subscriber_list.title,
+          address: subscriber.address,
+          subscriber_id: subscriber.id
+        }
+      end
+    end
+  end
+
+  def courtesy_emails(taxon_emails)
+    return [] if taxon_emails.empty?
+    Subscriber.where(address: Email::COURTESY_EMAIL).map do |subscriber|
+      {
+        subject: taxon_emails.first.fetch(:subject),
+        address: subscriber.address,
+        subscriber_id: subscriber.id
+      }
+    end
+  end
+
+  def log_taxon_emails(emails)
+    emails.each do |email|
+      Rails.logger.info(<<-INFO.strip_heredoc)
+        ----
+        Created Email:
+        id: #{email.id}
+        subject: #{email.subject}
+        body: #{email.body}
+        subscriber_id: #{email.subscriber_id}
+        ----
+      INFO
+    end
+  end
+
+  def log_non_taxon_lists(subscriber_lists)
+    subscriber_lists.each do |list|
+      Rails.logger.info(<<-INFO.strip_heredoc)
+        ++++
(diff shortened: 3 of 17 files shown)


config/routes.rb
```diff
@@ -1,10 +1,14 @@
-Rails.application.routes.draw do # rubocop:disable Metrics/BlockLength
+# rubocop:disable Metrics/BlockLength
```
cbaines: I think it would be best to disable this here https://github.com/alphagov/email-alert-api/blob/master/.rubocop.yml as that avoids the distraction of the comments in the file.

Whitehall might serve as a good example https://github.com/alphagov/whitehall/blob/master/.rubocop.yml

spec/builders/unpublish_email_builder_spec.rb
```diff
@@ -0,0 +1,65 @@
+# rubocop:disable Layout/EmptyLinesAroundArguments
```
cbaines: Same thing for this https://github.com/alphagov/email-alert-api/pull/655/review_comment/create#r209238087

app/controllers/unpublish_messages_controller.rb
```diff
@@ -0,0 +1,20 @@
+class UnpublishMessagesController < ApplicationController
+  def create
+    UnpublishHandlerService.call(
+      unpublishing_params[:content_id]
+    )
+
+    render json: { message: "Unpublish message queued for sending" }, status: 202
+  end
+
+  def index
```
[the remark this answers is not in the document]
1pretz1: Just following the pattern of the `NotificationsController`, i can remove it however and just have the create action.

config/routes.rb
1pretz1: Sure, thanks for the example!

spec/builders/unpublish_email_builder_spec.rb
1pretz1: 👍

app/controllers/unpublish_messages_controller.rb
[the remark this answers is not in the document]
koetsier: think so. had a look at the notification controller and the relevant commit for the index  says:

>Make deprecated notification endpoint a no-op  
>These used to talk to GovDelivery but now they won't return anything useful.

So I don't think it is necessary in this  case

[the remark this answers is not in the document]
1pretz1: Okay dokay, i've removed it and the corresponding route!

spec/builders/digest_email_builder_spec.rb
```diff
@@ -132,5 +130,3 @@ def simulated_deduplification(content_changes)
     end
   end
 end
-
-# rubocop:enable Layout/EmptyLinesAroundArguments
```
cbaines: :+1:

spec/services/unpublish_handler_service_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        .to change { Email.count }.by(2)
+      end
+
+     # it "increases the size of the queue in the DeliverRequestWorker" do
+     #   Sidekiq::Testing.fake! do
+     #     DeliveryRequestWorker.jobs.clear
+     #     described_class.call(@content_id)
+     #   end
+     #   expect(DeliveryRequestWorker.jobs.size).to eq(2)
+     # end
```
cbaines: Could we mark this as pending somehow [1], rather than commenting it out?

1: https://relishapp.com/rspec/rspec-core/v/2-0/docs/pending/pending-examples

app/services/unpublish_handler_service.rb
```diff
@@ -0,0 +1,89 @@
...
+        }
+      end
+    end
+  end
+
+  # For this query to return the content id has to be wrapped in a double quote blame psql 9.3
+  def subscriber_lists
+    @subscriber_lists ||= SubscriberList
+    .where(":id IN (SELECT json_array_elements((json_each(links)).value)::text)", id: "\"#{content_id}\"")
+    .includes(:subscribers).to_a
```
cbaines: This layout might be a little more readable.
```ruby
    @subscriber_lists ||= SubscriberList.where(
      ":id IN (SELECT json_array_elements((json_each(links)).value)::text)", id: "\"#{content_id}\""
    ).includes(
      :subscribers
    ).to_a
```

Or maybe just indent the lines after SubscriberList by two more spaces?

app/services/unpublish_handler_service.rb
```diff
@@ -0,0 +1,89 @@
...
+        id: #{email.id}
+        subject: #{email.subject}
+        body: #{email.body}
+        subscriber_id: #{email.subscriber_id}
+        ----
+      INFO
+    end
+  end
+
+  def log_none_taxon_lists(subscriber_lists)
```
cbaines: I'm not sure what "none" means here, were you thinking of "non"?

app/services/unpublish_handler_service.rb
```diff
@@ -0,0 +1,89 @@
...
+    subscriber_lists.each do |list|
+      logger.info(<<-INFO.strip_heredoc)
+        ++++
+        Email not created but matching a SubscriberList (Policy area etc):
```
cbaines: Topics (taxons) use subscriber lists too, maybe say "Not sending notification about non-Topic SubscriberList".

spec/services/unpublish_handler_service_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      it "doesn't create email if subscriber is deactivated" do
+        expect { described_class.call(@content_id) }.to_not(change { Email.count })
+      end
+    end
+
+    context "email is not created if there is no SubscriberList" do
+      it "doesn't build an email if there is no associated SubscriberList" do
+        content_id_two = "content-id"
```
cbaines: Maybe `"non-existent-content-id"` to make it clear (assuming that is what's going on).

app/services/unpublish_handler_service.rb
```diff
@@ -0,0 +1,89 @@
...
+  def initialize(content_id)
+    @content_id = content_id
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    if email_params.empty?
```
koetsier: email_params is also called in line 16 and could be memoized

app/services/unpublish_handler_service.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def queue_delivery_request_workers(email_ids)
+    email_ids.each do |email_id|
+      DeliveryRequestWorker.perform_async_in_queue(
+        email_id, queue: :delivery_immediate
+      )
+    end
+  end
+
+  def email_params
```
koetsier: I think sometimes it is clearer to explicitly pass in subscriber_lists so that you can see at the calling method that this depends on subscriber_lists. Same with some of the other methods

app/services/unpublish_handler_service.rb
koetsier: You could also make UnpublishEmailBuilder deal with an empty array

app/builders/unpublish_email_builder.rb
```diff
@@ -0,0 +1,65 @@
...
+    BODY
+  end
+
+  def presented_manage_subscriptions_links(address)
+    ManageSubscriptionsLinkPresenter.call(address: address)
+  end
+
+  def courtesy_email
+    addresses = [
+      "govuk-email-courtesy-copies@digital.cabinet-office.gov.uk",
```
koetsier: can we make this a constant somewhere? I believe it is used elsewhere too

spec/services/unpublish_handler_service_spec.rb
1pretz1: Yup, that makes it clearer!

spec/services/unpublish_handler_service_spec.rb
1pretz1: Yep sure, thats interesting, didn't know you could do that.

app/services/unpublish_handler_service.rb
1pretz1: I've indented and moved the `.to_a` to a separate line which i think looks the best!

app/services/unpublish_handler_service.rb
1pretz1: Yep, got the two mixed up! -.-

app/services/unpublish_handler_service.rb
1pretz1: 👍

app/services/unpublish_handler_service.rb
1pretz1: Yep i could memoize email_params, however because I memoized `subscriber_lists` I assumed it would be okay to leave.

However i'll explore making `UnpublishEmailBuilder` handle an empty array instead, which will make this problem obsolete and probably make the code look nicer!

app/services/unpublish_handler_service.rb
1pretz1: I've done this for all the other methods, however i think for `email_params` it would look messy as i call `UnpublishEmailBuilder.call(email_params)` ? 

But noted that i should pass arguments more often.

koetsier: no I meant def ```email_params(subscriber_lists)``` so you know what stuff this method depends on

1pretz1: Yep i know, i meant that because i call the method from `UnpublishEmailBuilder.call(email_params)`, it looks a little messy to me to have `UnpublishEmailBuilder.call(email_params(subscriber_lists))` but if you think it would make things clearer then i'm not opposed to changing it :)

1pretz1: I've added the arguments.

app/services/unsubscribe_service.rb
```diff
@@ -1,37 +1,52 @@
...
-      unsubscribe!(subscriber, subscriber.active_subscriptions, :marked_as_spam, delivery_attempt.email)
+      ActiveRecord::Base.transaction do
+        unsubscribe_subscriptions!(subscriber.active_subscriptions, :marked_as_spam, delivery_attempt.email)
+        unsubscribe_subscriber!(subscriber)
+      end
     end
 
-  private
-
-    def unsubscribe!(subscriber, subscriptions, reason, email = nil)
```
cbaines: I'm a bit unclear what part of this commit is refactoring, and what purpose the refactoring is for. It would be useful to have some more information in the commit message, or maybe either separate the refactoring out in to a different commit.

cbaines:
I've pulled out a rubocop change in `config/routes.rb` in to the `.rubocop.yml`, just to make the `config/routes.rb` file a bit cleaner.

app/services/unsubscribe_service.rb
```diff
@@ -1,37 +1,52 @@
...
+    def subscriber_list!(list, reason)
       ActiveRecord::Base.transaction do
-        subscriptions.each do |subscription|
-          subscription.end(reason: reason)
+        subscriptions = list.subscriptions
+        unsubscribe_subscriptions!(subscriptions, reason)
+        Subscriber.where(subscriptions: subscriptions).each do |subscriber|
```
cbaines: Given this is happening in the request (as far as I can see), I don't know how this will cope with lists with many subscriptions. The largest I know of for the legacy things is ~16000, and I'd worry that doing 16000 queries inside the request won't work.

Not something we necessarily need to solve now, but if I'm right, it might be worth improving soon so that this will work with large lists.

koetsier: You're right - made this call asynchronous

PR #657, opened 2018-08-16
Increase subscription contents healthcheck thresholds around 09:30

barrucadu:
We frequently see this healthcheck go critical in the morning, because there's a lot of scheduled publishing at 09:30.  Rather than increase the thresholds unconditionally, we can just relax them around this time.

---

[Trello card](https://trello.com/c/fxDUVoSk/401-stop-email-alert-api-going-critical-every-morning)

### DIFF
--- a/app/models/healthcheck/subscription_contents.rb
+++ b/app/models/healthcheck/subscription_contents.rb
@@ -55,11 +55,22 @@ module Healthcheck
     end
 
     def critical_latency
-      10.minutes
+      is_scheduled_publishing_time? ? 20.minutes : 10.minutes
     end
 
     def warning_latency
-      5.minutes
+      is_scheduled_publishing_time? ? 15.minutes : 5.minutes
+    end
+
+    # There's a lot of scheduled publishing at 09:30 UK time, so we
+    # want larger thresholds around then.
+    def is_scheduled_publishing_time?
+      @is_scheduled_publishing_time ||= begin
+        now = Time.zone.now
+        begun = Time.zone.parse("09:30") <= now
+        ended = Time.zone.parse("10:30") <= now
+        begun && !ended
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/healthcheck/subscription_contents.rb
```diff
@@ -55,11 +57,20 @@ def count_subscription_contents(age)
...
+      is_scheduled_publishing_time? ? 15.minutes : 5.minutes
+    end
+
+    # There's a lot of scheduled publishing at 09:30 UK time, so we
+    # want larger thresholds around then.
+    def is_scheduled_publishing_time?
+      now = Time.zone.now
+      begun = now.hour > 9 || (now.hour == 9 && now.min >= 30)
+      ended = now.hour > 10 || (now.hour == 10 && now.min >= 30)
+      begun && !ended
```
thomasleese: I think it's possible to do something like this which might be more readable:

```ruby
Time.zone.parse("09:30") >= now
Time.zone.parse("10:30") < now
```

app/models/healthcheck/subscription_contents.rb
```diff
@@ -1,3 +1,5 @@
+require 'tzinfo'
```
thomasleese: I don't think we need to include this as Rails will do that for us.

spec/models/healthcheck/subscription_contents_spec.rb
```diff
@@ -14,27 +14,71 @@
     specify { expect(subject.message).to match(/1 created over 600 seconds ago/) }
   end
 
-  context "when a subscription content was created 10 second ago" do
+  context "between 09:30 and 10:30" do
     before do
-      create(:subscription_content, created_at: 10.seconds.ago)
+      Timecop.freeze("10:00", Time.zone.now)
```
thomasleese: You can do something like this to make it slightly more succinct:

```ruby
around do |example|
  Timecop.freeze("10:00") { example.run }
end
```

barrucadu:
@thomasleese Having some difficulty with the tests - is it perhaps using the same instance for all of the tests?

barrucadu:
Ah, it's because of the frozen time.

PR #664, opened 2018-08-22
Add redirect title to unpublishing message

koetsier:
When unpublishing a Taxon, we need to display the title of the redirect included in the 
unpublishing message.  The unpublishing message has been updated to use the title.

Trello: https://trello.com/c/gPSdi81i/59-xl-write-email-template-to-notify-users-that-an-email-subscription-has-ended-because-a-taxon-has-been-unpublished

### DIFF
--- a/app/services/public_url_service.rb
+++ b/app/services/public_url_service.rb
@@ -18,7 +18,7 @@ module PublicUrlService
     end
 
     def redirect_url(path:)
-      "#{website_root}/#{path}"
+      File.join(website_root, path)
     end
 
   private
--- a/app/builders/unpublish_email_builder.rb
+++ b/app/builders/unpublish_email_builder.rb
@@ -21,9 +21,11 @@ private
     end
   end
 
-  def body(title, address, _)
+  def body(title, address, redirect)
     <<~BODY
       You were subscribed to emails about '#{title}'. This topic no longer exists so you won't get any more emails about it.
+
+      You can subscribe to '#{redirect.title}' instead: (#{redirect.url})
 
       #{presented_manage_subscriptions_links(address)}
     BODY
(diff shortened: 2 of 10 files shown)


app/models/redirect.rb
```diff
@@ -0,0 +1,20 @@
+class Redirect
```
thomasleese: I'd be tempted to call this class `ContentItem` since it doesn't necessarily represent a `Redirect` and is generic enough that it would work for any piece of content.

I'm also not sure about putting it in `app/models` because it's not an active record model, I would probably put it in `app/services` but that's just my personal preference.

app/models/redirect.rb
```diff
@@ -0,0 +1,20 @@
...
+  DEFAULT = ''.freeze
+  attr_reader :path
+
+  def initialize(path)
+    @path = path
+  end
+
+  def title
+    @title ||= begin
+      Services.content_store.content_item(@path).to_h['title'] || DEFAULT
```
thomasleese: Do you need `.to_h` here? Could you instead do `.fetch("title", DEFAULT)` perhaps?

app/models/redirect.rb
koetsier: Good point. I renamed the class to ContentItem

ContentItem isn't what I would call a service though. I would normally have put it in a 'lib' directory or something generic but that doesn't exist here. The model 'Frequency' is already in the models directory and that is not an AR model so I didn't think ContentItem was too out of place in 'models'

app/models/redirect.rb
koetsier: The thing is that the hash can contain ```{title: nil}``` and so this that would return nil. Otherwise I'd have to use ```.fetch('title', DEFAULT) || DEFAULT``` which isn't great.

app/services/public_url_service.rb
```diff
@@ -18,7 +18,7 @@ def authenticate_url(address:)
     end
 
     def redirect_url(path:)
-      "#{website_root}/#{path}"
+      File.join(website_root, path)
```
cbaines: Won't this break on Microsoft Windows, or any OS that doesn't use `/` as the file separator?

PR #673, opened 2018-08-29
Send emails when unpublishing policy(areas) WIP

koetsier:
When unpublishing a policy or policy_area all subscribers to these content items should receive an email including an alternative to subscribe to.

Trello: https://trello.com/c/dxQwG2LO/133-m-send-emails-when-unpublishing-policies


app/queries/content_changes_for_email_query.rb
```diff
@@ -1,9 +1,10 @@
 class ContentChangesForEmailQuery
   def self.call(email)
-    subscription_content = SubscriptionContent
-      .where("subscription_contents.content_change_id = content_changes.id")
-      .where(email: email)
-      .exists
+    subscription_content = SubscriptionContent.exists?(
+      subscription_contents:
+          { content_change_id: 'content_changes.id' },
```
cbaines: The indentation is a bit odd here.

app/services/unpublish_handler_service.rb
```diff
@@ -9,18 +9,17 @@ def call(content_id, redirect)
     case type
     when :taxon_tree
       unsubscribe(subscriber_lists, redirect, taxon_template)
+    when :policy_areas
+      unsubscribe(subscriber_lists, redirect, policy_area_template)
+    when :policies
+      unsubscribe(subscriber_lists, redirect, policy_template)
     else
-      unsubscribe_other(subscriber_lists)
+      log_non_taxon_lists(subscriber_lists)
```
cbaines: This method name looks like it needs tweaking now.

app/queries/content_changes_for_email_query.rb
koetsier: fixed

app/services/unpublish_handler_service.rb
koetsier: fixed

PR #676, opened 2018-08-30
Fix soon-to-be deprecated use of exists

koetsier:
Using arel 'exists' shows the following message
DEPRECATION WARNING: Delegating exists to arel is deprecated and will be
removed in Rails 6.0.

### DIFF
--- a/app/queries/content_changes_for_email_query.rb
+++ b/app/queries/content_changes_for_email_query.rb
@@ -1,9 +1,8 @@
 class ContentChangesForEmailQuery
   def self.call(email)
-    subscription_content = SubscriptionContent
-      .where("subscription_contents.content_change_id = content_changes.id")
-      .where(email: email)
-      .exists
-    ContentChange.where(subscription_content)
+    content_change_ids = SubscriptionContent.
+        where(email: email).
+        select(:content_change_id)
+    ContentChange.where(id: content_change_ids)
   end
 end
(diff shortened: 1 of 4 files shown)


app/queries/content_changes_for_email_query.rb
```diff
@@ -1,9 +1,7 @@
 class ContentChangesForEmailQuery
   def self.call(email)
     subscription_content = SubscriptionContent
-      .where("subscription_contents.content_change_id = content_changes.id")
-      .where(email: email)
-      .exists
-    ContentChange.where(subscription_content)
+      .where(email: email).select(:content_change_id)
+    ContentChange.where(id: subscription_content)
```
cbaines: Perhaps inline in the SubscriptionContent part of the query? Or, update the variable name to make sense, now that it's some `content_change_id`s.

app/queries/subscribers_for_immediate_email_query.rb
```diff
@@ -4,20 +4,7 @@ def self.call
...
-
-private
-
-  def unprocessed_subscription_contents_exist_for_subscribers
-    SubscriptionContent
-      .joins(:subscription)
-      .where(email_id: nil)
-      .where("subscriptions.subscriber_id = subscribers.id")
-      .exists
+    subscriber_ids = Subscription.joins(:subscription_contents).where(subscription_contents: { email_id: nil }).select(:subscriber_id)
```
cbaines: It would be good to format this like it was, with the different parts of the query on their own lines, avoid having one really one line.

PR #689, opened 2018-09-10
Track unpublish emails [WIP]

cbaines:
These changes should allow us to track the emails sent out around unpublishing things people subscribe to (e.g. Topic Taxonomy Topics).


cbaines:
This still needs a bit more work, I think there are some broken tests, and it needs some checking regarding the performance and the generated SQL.

PR #696, opened 2018-09-20
Bulk unpublish policies

cbaines:


app/services/bulk_unsubscribe_service.rb
```diff
@@ -0,0 +1,89 @@
...
+        [
+          subscription.subscriber_list.title,
+          content_ids_and_replacements.fetch(relevant_content_id)
+        ]
+      end
+
+      email = send_email_to_subscriber(subscriber, subscription_details, index)
+
+      subscriptions.each do |subscription|
+        subscription.update!(
```
koetsier: You could maybe use UnsubscribeService.subscription! here.
It doesn't yet handle the case of unsubscribing with :unpublished but  that is easily added. And you could update Subscription#end as well to reflect this new use case.

app/services/bulk_unsubscribe_service.rb
```diff
@@ -0,0 +1,133 @@
...
+        ]
+      end
+
+      subscription_details = subscription_details.sort_by do |(title, _replacement)|
+        title
+      end
+
+      email = nil
+      ActiveRecord::Base.transaction do
+        email = send_email_to_subscriber(
```
koetsier: Should this be in this transaction? If the Transaction fails, the BulkUnsubscribeEmailBuilder will have already been called and the email queued.

app/services/bulk_unsubscribe_service.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    subscriptions_to_end = Subscription
+                             .where(subscriber_list_id: affected_subscriber_list_ids)
+                             .includes(:subscriber)
+                             .active
+
+    subscriptions_to_end
+      .group_by(&:subscriber)
+      .sort_by { |_subscriber, subscriptions| subscriptions.length }
+      .each_with_index do |(subscriber, subscriptions), index|
```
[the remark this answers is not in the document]
cbaines: It's used when deciding to send a courtesy copy email: https://github.com/alphagov/email-alert-api/pull/696/files#diff-1ad1e303fbe22f51ea316a35427c5584R51

app/services/bulk_unsubscribe_service.rb
```diff
@@ -0,0 +1,89 @@
...
+        )
+      end
+    end
+
+    SubscriberDeactivationWorker.perform_async(
+      subscriptions_to_end.map(&:subscriber_id).uniq
+    )
+  end
+
+  def self.send_email_to_subscriber(subscriber, subscription_details, index)
```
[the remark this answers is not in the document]
cbaines: Is this an issue with an individual commit? I think I switched it around to a boolean at some point, but maybe the commits are a bit confusing.

app/services/bulk_unsubscribe_service.rb
cbaines: Using Subscription#end sounds good, but I'm not too convinced about using UnsubscribeService.subscription! here, mostly because passing through the email might be a bit awkward, as it already has some special behaviour about spam tracking.

app/models/content_item.rb
```diff
@@ -2,19 +2,40 @@ class ContentItem
...
+
   def url
     @url ||= PublicUrlService.absolute_url(path: @path)
   end
+
+  def content_store_data
+    @content_store_data ||= begin
+      response = Services.content_store.content_item(@path).to_h
+
+      unless response['base_path'] == @path
```
koetsier: could add a spec for this case

cbaines: Sure, I've added one here 25932d7a0b1c4203a2d1b0ecba5d643b5656b032 . Also, despite seeing this work yesterday, I actually had to tweak the implementation a bit (adding `.new`)...

app/services/bulk_unsubscribe_service.rb
```diff
@@ -72,6 +74,29 @@ def self.send_email_to_subscriber(subscriber, subscription_details, index)
       queue: :delivery_immediate
     )
 
+    # Send a sample of the emails to the courtesy copy group. As
+    # we're iterating over subscribers here, sending each email
+    # would be excessive.
+    if (index % 500).zero?
```
koetsier: Instead of using 'index' here, maybe it is better to make this a boolean 'send_courtesy_copy?' and then compute that when calling this service. I think that would make the behaviour a bit clearer

app/services/bulk_unsubscribe_service.rb
koetsier: That is a private method. the public one doesn't do spam tracking.

I'm just a bit worried that we end up with multiple paths to unsubscribe a subscription, which makes changing the behaviour of unsubscribing a subscription more difficult

app/services/bulk_unsubscribe_service.rb
cbaines: I did switch to that approach eventually, but I've got back and switched around the commits to avoid this intermediate state with the index, so hopefully the commit history is a bit neater now.

app/services/bulk_unsubscribe_service.rb
cbaines: > I'm just a bit worried that we end up with multiple paths to unsubscribe a subscription, which makes changing the behaviour of unsubscribing a subscription more difficult

Sure, I've pushed a few commits now which get the BulkUnsubscribeService to use the same code path through the UnsubscribeService. How does that look?

PR #711, opened 2018-10-19
Rename education and skill funding agency subscriber list

steventux:
https://trello.com/c/qqw5RbKb/556-change-this-email-signup-from-education-and-education-and-skills-funding-agency-to-education-and-skills-funding-agency

@rubenarakelyan could you advise please? Not sure if this is the correct approach, it seems better to have the slug and title in sync but we could just update the title according to the trello story.


rubenarakelyan:
Yeah I think this is fine. The whitehall translation from feed URL to a subscriber list uses tags rather than the slug - the slug is returned by email-alert-api when whitehall queries it for the correct subscriber list to direct users to.

steventux:
It looks like there's already a subscriber list at this slug in production so migrating subscribers and removing this list would probably be what we need here.

PR #731, opened 2018-11-21
Inject tags for content related to business readiness

thomasleese:
This injects tags for content changes which are related to business readiness as configured in the CSV file.

Ideally this wouldn't live in email-alert-api, it would instead be part of the content items (or at the very least email-alert-service), but because of specialist-publisher and travel-advice-publisher not going through email-alert-service, we want to catch all cases here.

The effort required to add this feature to all publishing apps, the publishing-api and rummager is too much effort for the time scale of this project so the solution is to put the code here.


spec/lib/business_readiness/content_change_injector_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        )
+      end
+    end
+
+    context "as nil" do
+      let(:existing_tags) do
+        { "sector_business_area" => nil }
+      end
+
+      it "should have both the old and new values" do
```
barrucadu: Should this be more like `it "should not change" do`?

thomasleese: I've changed the string to make it clearer. I think we do want the tags from the CSV file to be injected in this case it was really just a test to check we handle `nil` without throwing an exception which might prevent emails from being sent out (I don't think we'll actually ever see a situation where `nil` is passed like this, for the moment anyway).

PR #732, opened 2018-11-22
Read facet configuration from a shared definition

steventux:
https://trello.com/c/UhV3PPfO/87-metadata-tagger-should-use-finder-facet-definitions

Updates to facet names and values is shared between dependent applications (temporarily these are Rummager and Email Alert API) until a central process for assigning metadata is in place.
Use a shared facet definition from file to save hardcoding facet names and values, this DRYs up the update process.

### DIFF
--- a/lib/business_readiness/loader.rb
+++ b/lib/business_readiness/loader.rb
@@ -3,7 +3,7 @@ require "csv"
 module BusinessReadiness
   class Loader
     def initialize(filename, facets_filename)
-      @rows = CSV.read(filename)
+      @rows = CSV.read(filename, converters: lambda { |v| v || "" })
       @facets = {}
 
       if File.exist?(facets_filename)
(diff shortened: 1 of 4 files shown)


lib/business_readiness/loader.rb
```diff
@@ -30,142 +36,23 @@ def tags_for_row(row)
...
-        "employ_eu_citizens" => all_employ_eu_citizens,
-        "doing_business_in_the_eu" => all_doing_business_in_the_eu,
-        "regulations_and_standards" => all_regulations_and_standards,
-        "personal_data" => all_personal_data,
-        "intellectual_property" => all_intellectual_property,
-        "receiving_eu_funding" => all_receiving_eu_funding,
-        "public_sector_procurement" => all_public_sector_procurement
-      }
+      tags = {}
+      facets.each do |facet|
```
thomasleese: Here you could do `facets.each_with_object({}) do |facet, tags|` to avoid the variable definition and return below.

PR #746, opened 2018-12-13
Add an index on the status column in emails table

brucebolt:
Trying to resend just the failed emails is painfully slow as the where query on the status column takes an age to process.  Adding an index to speed this up.

### DIFF
--- a/db/migrate/20181213113655_add_index_to_email_status.rb
+++ b/db/migrate/20181213113655_add_index_to_email_status.rb
@@ -1,0 +1,5 @@
+class AddIndexToEmailStatus < ActiveRecord::Migration[5.2]
+  def change
+    add_index :emails, :status
+  end
+end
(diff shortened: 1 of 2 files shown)


bevanloon:
I think we had an index on status before which we removed https://github.com/alphagov/email-alert-api/pull/646 

I wonder whether it's a major inconvenience on the occasions when we do have to investigate. We are mostly looking at specific content changes when we investigate something in email-alert-api which should allow us to reduce the number of items we are looking for considerably.

Any thoughts @thomasleese ?

db/migrate/20181213113655_add_index_to_email_status.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToEmailStatus < ActiveRecord::Migration[5.2]
+  def change
+    add_index :emails, :status
```
thomasleese: If we do want to add this index, we should [add it concurrently](https://robots.thoughtbot.com/how-to-create-postgres-indexes-concurrently-in) to avoid locking the table.

brucebolt:
I'll close this PR.  We can review the example queries to put into the dev docs separately.

PR #754, opened 2018-12-27
Add rake task to present email status report for content change

edwardkerry:
Given a Content Change ID, the new Rake Task
`report:content_change_email_status_count[':ID']`
will print a short report with number of sent, pending and failed emails

```
    -------------------------------------------
    Email status counts for Content Change 0016983a-216f-4543-b88c-bb1fd5bdd8be
    -------------------------------------------

    Sent emails: 3035

    Pending emails: 61

    Failed emails: 0

    -------------------------------------------
```


lib/tasks/content_change_email.rake
```diff
@@ -0,0 +1,7 @@
+namespace :content_change_email do
```
thomasleese: I think it might be worth putting this inside a `reports` namespace.

lib/content_change_email_status_count.rb
```diff
@@ -0,0 +1,47 @@
+class ContentChangeEmailStatusCount
```
thomasleese: Maybe we can put this in a `lib/reports` module to future reports together. We could then also move the `UnpublishingReport` into that folder.

lib/content_change_email_status_count.rb
```diff
@@ -0,0 +1,47 @@
...
+      Failed emails: #{@email_status_counts[:failed]}
+
+      -------------------------------------------
+    TEXT
+  end
+
+private
+
+  def email_statuses
+    @content_change.subscription_contents.map do |sub_contents|
```
thomasleese: I suspect this will lead to many queries on certain content changes, perhaps we could do it in stages, something like below to save on the number of queries:

```ruby
subscription_contents_ids = @content_change.subscription_contents.pluck(:id)
email_ids = SubscriptionContent.where(id: subscription_contents_ids).pluck(:email_id)
Email.where(id: email_ids).pluck(:status)
```

We should have indexes on all those ID fields so it should be reasonably fast.

thomasleese: In fact, just thinking you could even do the count query this way:

```
Email.where(id: email_ids).group(:status).count
```

edwardkerry: Thanks @thomasleese, have updated the queries

lib/tasks/content_change_email.rake
edwardkerry: 👍

lib/reports/content_change_email_status_count.rb
```diff
@@ -0,0 +1,49 @@
...
+        -------------------------------------------
+      TEXT
+    end
+
+  private
+
+    def email_statuses
+      subscription_contents_ids = @content_change.subscription_contents.pluck(:id)
+      email_ids = SubscriptionContent.where(id: subscription_contents_ids).pluck(:email_id)
+      Email.where(id: email_ids).pluck(:status)
```
thomasleese: I think it might be worth doing `Email.where(id: email_ids).group(:status).count` here which I think will get you the same result as `email_status_counts` currently does.

edwardkerry: Very nice, thanks

PR #760, opened 2019-01-03
Add Rake task to resend failed emails

edwardkerry:
Given an Array of email ids, the new Rake task
`bundle exec rake deliver:resend_failed_emails[<email_one_id>,<email_two_id>...]`
will check if each email has a failed status and, if so, attempt to resend it.

### DIFF
--- a/lib/tasks/deliver.rake
+++ b/lib/tasks/deliver.rake
@@ -20,4 +20,14 @@ namespace :deliver do
     email = test_email(args[:test_email_address])
     DeliveryRequestWorker.perform_async_in_queue(email.id, queue: :delivery_immediate)
   end
+
+  desc "Re-send failed emails by email ids"
+  task resend_failed_emails: [:environment] do |_, args|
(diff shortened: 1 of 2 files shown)


lib/tasks/deliver.rake
```diff
@@ -20,4 +20,14 @@ namespace :deliver do
     email = test_email(args[:test_email_address])
     DeliveryRequestWorker.perform_async_in_queue(email.id, queue: :delivery_immediate)
   end
+
+  desc "Re-send failed emails by email ids"
+  task resend_failed_emails: [:environment] do |_, args|
+    failed_ids = Email.where(id: args.to_a, status: 'failed').pluck(:id)
+
+    email_ids.each do |email_id|
```
thomasleese: This should be `failed_ids.each`.

thomasleese: Sorted!

edwardkerry: you're too quick!

PR #766, opened 2019-01-08
Add rake task to get notification report from Notify

deborahchua:
This rake task allows us to directly query Notify's API and find
out information about an email's notification, should the callbacks
fail.

It uses Notify's [multiple notifications method](https://docs.notifications.service.gov.uk/ruby.html#get-the-status-of-multiple-messages), using the
`reference` field to query for an email notification.
We are unable to use Notify's [single notification method](https://docs.notifications.service.gov.uk/ruby.html#get-the-status-of-one-message-method) as it
requires the `notification_id`, which we don't store.

The `reference` is the `DeliveryAttempt.id` and is unique to each
email.

Example output when a notification is found:
```
Query Notify for emails with the reference ref_123
-------------------------------------------
Notification ID: f163deaf-2d3f-4ec6-98fc-f23fa511518f
Status: delivered
created_at: 2019-01-29 11:12:30 UTC
sent_at: 2019-01-29 11:12:40 UTC
completed_at: 2019-01-29 11:12:52 UTC
```

Example output when a notification is not found:
```
Query Notify for emails with the reference PPP
No results found, empty collection returned
```

Example output when the request fails:
```
Query Notify for emails with the reference ref_123
Returns request error 400, message: [{"error"=>"ValidationError",
"message"=>"bad status is not one of [created, sending, sent,
delivered, pending, failed, technical-failure, temporary-failure,
permanent-failure, accepted, received]"}]
```

A rework of original branch https://github.com/alphagov/email-alert-api/compare/query-notify-api

Trello card: https://trello.com/c/lUfUspdh/18-easily-query-the-govuk-notify-api-from-email-alert-api-for-specific-emails

### DIFF
--- a/lib/reports/notifications_from_notify.rb
+++ b/lib/reports/notifications_from_notify.rb
@@ -22,7 +22,7 @@ module Reports
       )
 
       if response.is_a?(Notifications::Client::NotificationsCollection)
-        if response.collection.count == 0
+        if response.collection.count.zero?
           puts "No results found, empty collection returned"
         else
           response.collection.each do |notification|
--- a/lib/reports/notifications_from_notify.rb
+++ b/lib/reports/notifications_from_notify.rb
@@ -22,16 +22,22 @@ module Reports
       )
 
       if response.is_a?(Notifications::Client::NotificationsCollection)
-        response.collection.each do |notification|
-          puts <<~TEXT
-            -------------------------------------------
-            Notification ID: #{notification.id}
-            Status: #{notification.status}
-            created_at: #{notification.created_at}
-            sent_at: #{notification.sent_at}
-            completed_at: #{notification.completed_at}
-          TEXT
+        if response.collection.count == 0
+          puts "No results found, empty collection returned"
+        else
+          response.collection.each do |notification|
+            puts <<~TEXT
+              -------------------------------------------
+              Notification ID: #{notification.id}
+              Status: #{notification.status}
+              created_at: #{notification.created_at}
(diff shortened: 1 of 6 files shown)


deborahchua:
@thomasleese I noticed this was merged and probably released soon https://github.com/alphagov/notifications-ruby-client/pull/63
So I'll hold off on this for now, and can try to make the appropriate changes once it's released.

However currently testing this PR on `integration` still doesn't return anything when the request fails:
https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/13866/console

lib/tasks/notify_report.rake
```diff
@@ -0,0 +1,6 @@
+namespace :report do
+  desc "Query the Notify API for email(s) by reference"
+  task :get_notifications_from_notify, [:reference] => :environment do |_t, args|
```
thomasleese: I wonder if we could have another task which takes an email ID, finds each delivery attempt for it and prints the report for each one of those.

lib/reports/notifications_from_notify.rb
```diff
@@ -0,0 +1,50 @@
+module Reports
+  class NotificationsFromNotify
+    def initialize(config: EmailAlertAPI.config.notify)
+      api_key = config.fetch(:api_key)
+      base_url = config.fetch(:base_url)
+
+      @client = Notifications::Client.new(api_key, base_url)
```
thomasleese: I wonder if we can move this into EmailAlertApi.config so we can reuse the same initialisation in the NotifyProvider.

PR #784, opened 2019-02-06
Enable rejection of a content_purpose_supergroup

bilbof:
https://trello.com/c/ylw7DDQa/329-enable-email-alert-api-to-reject-a-content-purpose-supergroup

This lets one set reject_content_purpose_supergroup as a parameter when creating and looking up subscriber lists.

This parameter, if set, will enable an email subscription to not receive emails when documents in certain supergroups are changed.

The main use case for this is the ability to subscribe to all content that isn't 'other' content. 

We need to update gds api adapters for this to have an effect.

*Tangential:*

The intent is for email subscriptions to be able to provide alerts when anything that would appear in a search query is changed. Adding a new column here is easy to understand, but is a little crude. It might be worth creating an RFC for a longer term solution. I expect a nice solution would be to use rummager for matching documents to subscriber lists. Perhaps we could store subscriber lists in rummager, and query rummager for subscriber lists using a content change. It would be similar to the current behaviour but might enable us to have a one-to-one match between email subscriptions and the search queries they are intended to mimic.

### DIFF
--- a/app/models/subscriber_list.rb
+++ b/app/models/subscriber_list.rb
@@ -6,6 +6,7 @@ class SubscriberList < ApplicationRecord
   validate :tag_values_are_valid
   validate :link_values_are_valid
   validate :content_purpose_supergroup_is_valid
+  validate :reject_content_purpose_supergroup_is_valid
 
   validates :title, presence: true
   validates_uniqueness_of :slug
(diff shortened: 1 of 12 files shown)


app/models/subscriber_list.rb
```diff
@@ -6,6 +6,7 @@ class SubscriberList < ApplicationRecord
   validate :tag_values_are_valid
```
koetsier: Could you add a few unit tests?

bilbof: Added some.

PR #795, opened 2019-02-25
[WIP]Add rake task to update subscriber list

deborahchua:
This rake task updates a `SubscriberList`'s title and slug.

This is required as part of [renaming a country](https://docs.publishing.service.gov.uk/manual/rename-a-country.html), so it's useful to
automate this job.

Trello card: https://trello.com/c/ySMq2bcm/786-slug-and-tag-changes-country-name-change-macedonia-to-north-macedonia

PR #799, opened 2019-02-26
Add rake task to find subscription lists by title

deborahchua:
This rake task will search for `SubscriberLists` using a partial
or whole title match, and will return the `title` and `slug` for
each one.

Example output:
```
Found 3 subscriber lists containing 'Macedonia'
=============================
title: Publications related to Macedonia
slug: publications-related-to-macedonia
=============================
title: Macedonia
slug: macedonia
=============================
title: North Macedonia - travel advice
slug: north-macedonia-travel-advice
```

This is to be used to assist with running the rake task
`manage:update_subscriber_list[slug,new_title,new_slug]` (https://github.com/alphagov/email-alert-api/pull/795) as part
of jobs such as [renaming a county](https://docs.publishing.service.gov.uk/manual/rename-a-country.html).

### DIFF
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -66,6 +66,20 @@ namespace :manage do
     puts "#{subscribers.count} active subscribers moved from #{from_slug} to #{to_slug}"
   end
 
+  def find_subscriber_list_by_title(title:)
+    subscriber_lists = SubscriberList.where("title ILIKE '%#{title}%'")
+
+    raise "Cannot find any subscriber lists with title containing `#{title}`" if subscriber_lists.nil?
+
+    puts "Found #{subscriber_lists.count} subscriber lists containing '#{title}'"
+
+    subscriber_lists.each do |subscriber_list|
+      puts "============================="
+      puts "title: #{subscriber_list.title}"
+      puts "slug: #{subscriber_list.slug}"
+    end
+  end
+
   def update_subscriber_list(slug:, new_title:, new_slug:)
     subscriber_list = SubscriberList.find_by(slug: slug)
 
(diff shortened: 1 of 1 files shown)


lib/tasks/manage.rake
```diff
@@ -66,6 +66,20 @@ namespace :manage do
     puts "#{subscribers.count} active subscribers moved from #{from_slug} to #{to_slug}"
   end
 
+  def find_subscriber_list_by_title(title:)
+    subscriber_lists = SubscriberList.where("title LIKE '%#{title}%'")
```
AlanGabbianelli: Do we want this to be case sensitive? If not, we could use `ILIKE` instead of `LIKE`.

deborahchua: Good point thanks! Have updated this now 😄

lib/tasks/manage.rake
```diff
@@ -66,6 +66,20 @@ namespace :manage do
     puts "#{subscribers.count} active subscribers moved from #{from_slug} to #{to_slug}"
   end
 
+  def find_subscriber_list_by_title(title:)
+    subscriber_lists = SubscriberList.where("title ILIKE '%#{title}%'")
```
bevanloon: I know there's a very small risk here because it's a rake task, but would it be worth mitigating sql injection anyway? Would something like 

```
SubscriberList.where("title ILIKE ?", "%#{title}%")
```

work?

deborahchua: Thanks @bevanloon I forgot about this risk! I've tested the change and updated in https://github.com/alphagov/email-alert-api/pull/799/commits/863c8405212ed00fea3e4ca96b230718368be423

PR #800, opened 2019-02-27
Rename subscribable to subscriber list

thomasleese:
At one point we were planning to rename all instances of subscriber lists with subscribables however these never happened and it's confusing to have the same thing named differently. Instead we've decided to stick with subscriber list.

This is a backwards compatible change as I've left in the references to subscribable from any external interfaces, but internally this is all renamed to subscriber list.

[Trello Card](https://trello.com/c/5qd2mgF0/76-subscriberlist-and-subscribable-are-used-for-the-same-thing-in-email-alert-api)


kevindew:
yay 🥇

PR #803, opened 2019-02-28
Add rake task to move business sector subscriptions

vanitabarrett:
Trello: https://trello.com/c/pH0tEctw/242-migrate-all-existing-business-finder-email-subscriptions

**This PR was originally to add a rake task to move business sector subscriptions as a result of renaming sectors. The scope has now changed, see below:**

We have decided to remove the ability for users to select facets when signing up to email updates on the business finder. Instead, any user wishing to get email updates will have to sign up to the same email subscription as everyone else and they will receive updates for all the content in the finder.

Once this change has been done, we will need to move the old business finder email subscriptions to the new one we create. This rake task should help us do that.

### DIFF
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -23,6 +23,48 @@ namespace :manage do
     end
   end
 
+  def move_to_new_generic_business_finder_emails
+    email_subject = "Changes to GOV.UK email alerts"
+    email_utm_parameters = {
+      'utm_source' => email_subject,
+      'utm_medium' => 'email',
+      'utm_campaign' => 'govuk-subscription-ended'
+      }
+    email_redirect = EmailTemplateContext.new.add_utm('https://gov.uk/email/manage', email_utm_parameters)
+
+    bulk_business_move_template = <<~BODY.freeze
+      Hello,
+
+      You've subscribed to get email alerts about EU Exit guidance for your business.
+
+      GOV.UK is changing the type of email alerts you get to make sure you find out when any EU Exit business guidance is added or updated.
+
+      You can [change how often you get updates or unsubscribe](#{email_redirect}) from GOV.UK email alerts.
+
+      Thanks,
+
+      GOV.UK
+    BODY
+
+    old_business_finder_slug_prefix = "find-eu-exit-guidance-for-your-business-with".freeze
+    to_slug = "find-eu-exit-guidance-for-your-business-appear-in-find-eu-exit-guidance-business-finder"
+
+    all_business_subscriber_lists = SubscriberList.where("slug LIKE '#{old_business_finder_slug_prefix}%'")
+    all_active_business_subscriber_lists = all_business_subscriber_lists.reject { |subscriber_list| subscriber_list.active_subscriptions_count.zero? }
+
+    puts "#{all_active_business_subscriber_lists.map(&:subscribers).flatten.count} subscribers found"
+
+    BulkEmailSenderService.call(
+      subject: email_subject,
+      body: bulk_business_move_template,
+      subscriber_lists: all_active_business_subscriber_lists,
+    )
+
+    all_active_business_subscriber_lists.each do |subscriber_list|
+      move_all_subscribers(from_slug: subscriber_list.slug, to_slug: to_slug)
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


thomasleese:
Unfortunately, you will need to do other things here to get it to work:

The `slug` of a subscriber list isn't actually that important when it comes to matching subscribers with content changes - the `tags` and the `links` (plus `document_type`, etc) are most important here.

What happens when you try and subscribe to a finder (once you've selected the facets you want) is finder-frontend asks email-alert-api for a subscriber list that matches all the `tags` that are selected. email-alert-api will then either return one that matches, or create a new one (automatically generating an appropriate title and slug) if there isn't one that already exists. Perhaps confusingly, finder-frontend then passes that subscriber list into email-alert-frontend using the slug (https://www.gov.uk/email/subscriptions/new?topic_id=<INSERT SLUG HERE>) which renders the standard email sign up flow for all subscriptions.

There are effectively two ways of managing a change in the facets here:

- You can change the `tags` of the subscriber list to correctly match with the new facets. Apart from that nothing needs to change, the subscribers will still be subscribed to the same subscriber list but with the new tags it will match correctly in finder-frontend. You don't even need to change the slug or title for it to work, but you probably should.

- You can forcibly create a new subscriber list by being the first person to subscribe with the new facets and then move people across into the newly created subscriber list using the existing rake tasks. This has the benefit of preserving history, and there being a record of a new subscriber list being created and an old one being deleted, but it's more work if you have lots of facets that have changed.

vanitabarrett:
@thomasleese Thanks Tom, that was really helpful! The scope of this has changed somewhat, in that we're now changing the way emails work and removing the ability to select facets. Am I right in thinking that in this case, as long as we create a subscriber list to this new subscription first, we can now just do a simple move using slugs (PR updated accordingly)?

thomasleese:
@vanitabarrett Yep, that sounds about right to me!

lib/tasks/manage.rake
```diff
@@ -23,6 +23,31 @@ namespace :manage do
...
   end
 
+  def move_to_new_generic_business_finder_emails
+    bulk_business_move_template = <<~BODY.freeze
+      We're updating your subscription to business content related to EU Exit on GOV.UK.
+    BODY
+    business_finder_slug_prefix = "find-eu-exit-guidance-for-your-business-with-sector".freeze
+
+    ## TODO Replace the to_slug with what it will actually be!
+    to_slug = "find-eu-exit-guidance-for-your-business-appear-in-find-eu-exit-guidance-business-finder"
```
leenagupte: This is the right slug 🙂

lib/tasks/manage.rake
```diff
@@ -23,6 +23,31 @@ namespace :manage do
     end
   end
 
+  def move_to_new_generic_business_finder_emails
+    bulk_business_move_template = <<~BODY.freeze
+      We're updating your subscription to business content related to EU Exit on GOV.UK.
+    BODY
+    business_finder_slug_prefix = "find-eu-exit-guidance-for-your-business-with-sector".freeze
```
leenagupte: Nitpick: You could rename this to `old....` so that it's clearer that it's the one being replaced.

vanitabarrett:
@thomasleese @leenagupte Can one of you please do a quick re-review - the logic around sending the emails out has changed slightly due to changing the email body (which includes a link)

lib/tasks/manage.rake
```diff
@@ -23,6 +23,50 @@ namespace :manage do
...
+    email_redirect = EmailTemplateContext.new.add_utm('https://gov.uk/email/manage', email_utm_parameters)
+
+    bulk_business_move_template = <<~BODY.freeze
+      Hello,
+
+      You've subscribed to get email alerts about EU Exit guidance for your business.
+
+      GOV.UK is changing the type of email alerts you get to make sure you find out when any EU Exit business guidance is added or updated.
+
+      You can [change how often you get updates or unsubscribe](<%= email_redirect %>) from GOV.UK email alerts.
```
thomasleese: Could we use `#{email_redirect}` here instead of needing to use `ERB`?

vanitabarrett: Done!

PR #809, opened 2019-03-13
Remove reject_content_purpose_supergroup field from subscriber lists

bilbof:
This field was used to filter out / reject specified supergroups.
The field is no longer needed (introduced in https://github.com/alphagov/email-alert-api/pull/784),
so this removes it.

It looks like just one active subscription used this (it wasnt
seen by the public), so can be removed without issues.

https://trello.com/c/45eoS6mj/491-remove-rejectcontentpurposesupergroup-functionality-from-email-alert-api-m-%F0%9F%98%AD

I've added a db migration to remove this column from the subscriber list here: https://github.com/alphagov/email-alert-api/compare/RemoveRejectContentPurposeSupergroupFromSubscriberList?expand=1


bilbof:
Email delivery seems OK. I'll merge this on Monday.

PR #815, opened 2019-03-20
Change title for new business subscription

vanitabarrett:
Due to the override we're doing for the EU Exit business finder email subscriptions, the default title is "Find EU Exit guidance for your business appear in find eu exit guidance business finder".

This migration renames the title to "EU Exit guidance for your business or organisation"

**This has been tested on integration. The title is successfully renamed and the new title appears in the signup process. I couldn't get emails sent to me to access the manage interface, but I assume this will be changed there too."


thomasleese:
You can use https://github.com/alphagov/email-alert-api/blob/7f8824809d685fecfd5945b2910e4d9ff3152da5/lib/tasks/manage.rake#L156 to rename a subscriber list which might be easier than having to do a deploy with a migration.

PR #818, opened 2019-03-22
Add setup troubleshooting to README

leenagupte:


README.md
```diff
@@ -50,6 +50,27 @@ subscribers to those tags via the external services.
...
+/var/govuk/email-alert-api/db/schema.rb:13:in `<main>'
+/usr/lib/rbenv/versions/2.6.1/bin/bundle:23:in `load'
+/usr/lib/rbenv/versions/2.6.1/bin/bundle:23:in `<main>'
+```
+
+Run the following to be able to run psql as a superuser:
+
+```
+sudo -u postgres psql postgres
+ALTER USER vagrant WITH SUPERUSER;
```
cbaines: I thought govuk-puppet was meant to do something about this [1]... maybe running `govuk_puppet` would work? I wouldn't be surprised if govuk-puppet would revert that role change.

1: https://github.com/alphagov/govuk-puppet/blob/4c27016f72e5b2a661393c0cb148d11b94599d95/modules/govuk/manifests/apps/email_alert_api/db.pp#L38

leenagupte: Hmmm. I didn't try that. I'll update the docs.

thomasleese:
We actually already had some docs on this https://github.com/alphagov/email-alert-api/blob/master/doc/troubleshooting.md#fixing-pginsufficientprivilege-error-in-the-development-vm, but it looks like perhaps it's a bit hidden away and we might need to find a better way of making it easier to spot.

Also, the solution in that file was similar to your initial solution, if we can just run `govuk_puppet` and it fixes it, we could update those docs as well.

leenagupte:
@thomasalanlee I didn't see those! I'll just add a link to the troubleshooting docs from the main README.

leenagupte:
There are already trouble-shooting docs that cover this error.

PR #819, opened 2019-03-22
Remove business sectors from email-alert-api

leenagupte:
Trello: https://trello.com/c/ZufSHOpe

Following https://trello.com/c/0rDl6oiK/232-simplify-email-alerts-for-business-finder and https://trello.com/c/pH0tEctw/242-migrate-all-existing-business-finder-email-subscriptions, email alerts are no longer driven from the different facets for the business finder, they only require `appear_in_find_eu_exit_guidance_business_finder`.

This means we can remove https://github.com/alphagov/email-alert-api/blob/daf1251e777733dc2a8dee45e4f3b62d17d34043/spec/lib/business_readiness/examples/facet_config.yml and the associated logic.


vanitabarrett:
LGTM! Do we want to remove the rake task that moves old business subscriptions to the new one?

leenagupte:
@vanitabarrett I've removed the migration task too.

vanitabarrett:
@leenagupte I think you need to remove the rake task associated with it too, not just the method it calls

bevanloon:
Does this mean you can get rid of https://github.com/alphagov/govuk-app-deployment/blob/fb3f9b40ecaf92f54a92dce290f245606060011f/email-alert-api/config/deploy.rb#L5 as well?

leenagupte:
> @leenagupte I think you need to remove the rake task associated with it too, not just the method it calls

Done! 🤦‍♀️

leenagupte:
> Does this mean you can get rid of https://github.com/alphagov/govuk-app-deployment/blob/fb3f9b40ecaf92f54a92dce290f245606060011f/email-alert-api/config/deploy.rb#L5 as well?

No, we still need the CSV file to know which content items to add the `appear_in_find_eu_exit_guidance_business_finder` tag to.

PR #832, opened 2019-04-02
Update subscription count docs for clarity

thomasleese:
The queries weren't clear on what they were returning and developers were using them incorrectly. They should now be more accurate.

[Trello Card](https://trello.com/c/bCMwB74D/912-email-rake-task-returns-suspicious-subscriber-numbers)

### DIFF
--- a/doc/analytics.md
+++ b/doc/analytics.md
@@ -28,14 +28,16 @@ endpoints or rake tasks.
 
 Here are some example queries to pull out particular insights.
 
-### How many subscribers does a list have at a particular point in time
+### How many subscribers does a list have
 
 ```ruby
-Subscription.where(
-    "created_at > ? AND (ended_at IS NULL OR ended_at <= ?)",
-    "2018-06-01",
-    "2018-07-01"
-  ).where(subscriber_list_id: 4194).count
+SubscriberList.find(4194).active_subscriptions_count
+```
+
+### How many subscribers did a list have at a particular point in time
+
+```ruby
+Subscription.active_on("2018-06-01").where(subscriber_list_id: 4194).count
 ```
 
 ### Lists with most new subscriptions in a time frame
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -12,6 +12,10 @@ class Subscription < ApplicationRecord
 
   scope :active, -> { where(ended_at: nil) }
   scope :ended, -> { where.not(ended_at: nil) }
+  scope :active_on, ->(date) do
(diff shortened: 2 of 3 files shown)


vanitabarrett:
I was just looking at this today and was really confused by the numbers. This looks similar to what I did to fix it locally 👍

doc/analytics.md
```diff
@@ -28,14 +28,20 @@ endpoints or rake tasks.
...
+### How many subscribers did a list have at a particular point in time
+
+```ruby
+date = "2018-6-01"
 Subscription.where(
-    "created_at > ? AND (ended_at IS NULL OR ended_at <= ?)",
-    "2018-06-01",
-    "2018-07-01"
-  ).where(subscriber_list_id: 4194).count
+  "created_at < ? AND (ended_at IS NULL OR ended_at <= ?)",
```
elliotcm: I'm a bit confused here.  Does `ended_at` mean when they unsubscribed?  If so shouldn't this query be `created_at < ? AND (ended_at IS NULL OR ended_at > ?)`?

Might also be worth using named interpolations to make it clearer you're using the same date, at which point you wouldn't need the extra variable either:
```
User.where(["name = :name and email = :email", { name: "Joe", email: "joe@example.com" }])
# SELECT * FROM users WHERE name = 'Joe' AND email = 'joe@example.com';
```

Could we also turn this into a method/scope on a SubscriberList instance to further reduce confusion?

doc/analytics.md
```diff
@@ -28,14 +28,20 @@ endpoints or rake tasks.
 
 Here are some example queries to pull out particular insights.
 
-### How many subscribers does a list have at a particular point in time
+### How many subscribers does a list have
 
 ```ruby
+Subscription.active.where(subscriber_list_id: 4194).count
```
elliotcm: There's already a method for this:
```suggestion
SubscriberList.find(4194).active_subscriptions_count
```

doc/analytics.md
vanitabarrett: Should this be OR? Just running it locally and getting smaller numbers that expected

vanitabarrett: Yes, I think it needs changing as I'm only getting a very small number of subscribers when I run it

thomasleese: Oh you're right, it should be `>`. I'll look at creating a scope as well.

thomasleese: @vanitabarrett @elliotcm Thanks for the review! I've pushed up a fix with a scope, what do you think?

PR #846, opened 2019-05-02
Move subscriptions to new subscriber lists

koetsier:
We will soon remove the 'content_purpose_supergroup' field in
the SubscriberLists. The content_purpose_supergroup is now
embedded in the 'tags' field.

This commit introduces rake tasks to get rid of any subscriber
lists with a value set in content_purpose_supergroup field.

This process is split into two tasks. The first moves all subscriptions
from the 'legacy' subscriber_lists to the new lists. The second removes
the now unused subscriber lists and moves all matched_content_changes
to their new counterparts. This is done so we avoid most
synchronisation issues that may arise.

trello: https://trello.com/c/NPTRaB1e/618-remove-supergroup-attribute-from-subscriber-lists-in-email-alert-api

### DIFF
--- a/app/services/remove_cps_service.rb
+++ b/app/services/remove_cps_service.rb
@@ -1,0 +1,85 @@
+class RemoveCpsService
+  def delete_subscriber_lists
+    subscriber_lists = fetch_subscriber_lists
+
+    subscriptions = Subscription.where(subscriptions: { subscriber_list: subscriber_lists })
+    raise "Can't have any subscribers left" unless subscriptions.empty?
+
+    count = subscriber_lists.count
+    subscriber_lists.each_with_index do |subscriber_list, index|
+      SubscriberList.transaction do
+        matching_list = matching_subscriber_list(subscriber_list)
+        move_matched_content_changes(subscriber_list, matching_list)
+        subscriber_list.destroy
+      end
+      puts "Done #{index} out of #{count} subscriber lists"
+    end
+  end
+
+  def deactivate_subscriber_lists
+    subscriber_lists = fetch_subscriber_lists
+    count = subscriber_lists.count
+    subscriber_lists.each_with_index do |subscriber_list, index|
+      SubscriberList.transaction do
+        matching_list = matching_subscriber_list(subscriber_list)
+        if matching_list.nil?
+          content_purpose_supergroup_to_tags(subscriber_list)
+        else
+          move_subscriptions(subscriber_list, matching_list)
+        end
+      end
+      puts "Done #{index} out of #{count} subscriber lists"
+    end
+  end
+
+private
+
+  def fetch_subscriber_lists
+    SubscriberList.where('content_purpose_supergroup IS NOT NULL')
+  end
+
+  def active_subscribers(subscriber_list)
+    Subscriber.joins(:subscriptions).where(subscriptions: { subscriber_list: subscriber_list, ended_at: nil })
+  end
+
+  def move_subscriptions(from_subscriber_list, to_subscriber_list)
+    subscriptions = from_subscriber_list.subscriptions
+    subscribers = active_subscribers(to_subscriber_list)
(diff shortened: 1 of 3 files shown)


app/services/remove_cps_service.rb
```diff
@@ -0,0 +1,58 @@
...
+    subscribers = active_subscribers(to_subscriber_list)
+    subscriptions.each do |subscription|
+      if subscription.active? && subscribers.include?(subscription.subscriber)
+        subscription.end(reason: 'subscriber_list_changed')
+      end
+      subscription.update_attribute(:subscriber_list, to_subscriber_list)
+    end
+  end
+
+  def move_matched_content_changes(from_subscriber_list, to_subscriber_list)
```
barrucadu: What happens if a new content change comes in while this function is running?

app/services/remove_cps_service.rb
```diff
@@ -0,0 +1,58 @@
+class RemoveCpsService
+  def call
```
barrucadu: Can this whole method be wrapped in a transaction?

koetsier: good point

app/services/remove_cps_service.rb
koetsier: That is also a good point.

I changed the code so that the process is now in two steps. One deactivates the legacy subscriber lists by moving all the subscriptions. Afterwards, there should not be new matched_content_changes that affect these lists.

The second rake task removes the superfluous lists and move the associated matched_content_changes.

app/services/remove_cps_service.rb
koetsier: wrapped it in a transaction

lib/tasks/manage.rake
```diff
@@ -1,6 +1,14 @@
...
 
 namespace :manage do
+  def deactivate_legacy_subscriber_lists
+    RemoveCpsService.new.deactivate_subscriber_lists
+  end
+
+  def delete_subscriber_lists
+    RemoveCpsService.new.delete_subscriber_lists
+  end
+
```
bilbof: Which rake task uses these new methods?

koetsier: oops.. fixed

spec/services/remove_cps_service_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      before :each do
+        @subscriber_list = FactoryBot.create(:subscriber_list_with_subscribers,
+                                             tags: tags,
+                                             content_purpose_supergroup: content_purpose_supergroup,
+                                             subscriber_count: 2)
+        @duplicate_subscriber_list = FactoryBot.create(:subscriber_list, tags: updated_tags, content_purpose_supergroup: nil)
+      end
+      it 'adds the subscriptions to the duplicate' do
+        expect { RemoveCpsService.new.deactivate_subscriber_lists }.to change {
+          @duplicate_subscriber_list.reload.subscribers.count
```
bilbof: Might also be worth checking that the subscribers have been moved off @subscriber_list, or do we not mind about this?

app/services/remove_cps_service.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def move_matched_content_changes(from_subscriber_list, to_subscriber_list)
+    matched_content_changes = MatchedContentChange.where(subscriber_list: from_subscriber_list)
+    matched_content_changes.each do |matched_content_change|
+      matched_content_change.update(subscriber_list: to_subscriber_list)
+    end
+  end
+
+  def content_purpose_supergroup_to_tags(subscriber_list)
+    subscriber_list.update(tags: updated_tags(subscriber_list),
```
bilbof: Do you need to use `update!` here to save the change?

bilbof:
Did you test this on integration already? Might be worth running it for a few (e.g. do `.take(10)` and see if anything looks weird).

PR #848, opened 2019-05-03
Removes all references to content_purpose_supergroup field 

koetsier:
The content_purpose_supergroup field is not used anymore. filters
are now contained in the 'tags' field instead. This commit removes
this field and all associated code.

see also: https://github.com/alphagov/gds-api-adapters/pull/893 and
https://github.com/alphagov/email-alert-api/pull/846 

trello: https://trello.com/c/NPTRaB1e/618-remove-supergroup-attribute-from-subscriber-lists-in-email-alert-api


db/schema.rb
```diff
@@ -139,6 +138,7 @@
     t.uuid "subscription_id", null: false
     t.uuid "content_change_id", null: false
     t.index ["content_change_id"], name: "index_subscription_contents_on_content_change_id"
+    t.index ["created_at"], name: "index_subscription_contents_on_created_at"
```
barrucadu: There's a migration [removing this index][1], can you remove it from this diff?

[1]: https://github.com/alphagov/email-alert-api/blob/298b36a3948b84c3fb7a05504ed64bd6689e66c0/db/migrate/20180727115914_drop_unused_indexes.rb#L8

koetsier: done

PR #853, opened 2019-05-14
Add rake task to update business finder subscriptions

AgaDufrat:
Since rolling out changes to subscribing users to business finder updates
(https://github.com/alphagov/finder-frontend/commit/bac5bf269ea30070af929ce4fd9c48a3c7fff80a),
the title for new subscriptions has become a long string, rather than "EU Exit
guidance for your business or organisation".

The rake task:
- moves Business Finder subscriptions to point to the primary SubscriberLists (id 18160)
- removes the Business Finder SubscriberLists which now don't have any Subscriptions
- removes `{"facet_groups"=>{"any"=>["52435175-82ed-4a04-adef-74c0199d0f46"]}}` from links of the SubscriberLists that still have Subscriptions. The Subscriptions cannot be deleted because we want to keep a record of subscription history

To run the rake task `bundle exec rake manage:update_business_finder_subscriptions`

[Trello card](https://trello.com/c/vueh4v07/98-fix-bug-in-business-finder-email-signup)

### DIFF
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -95,6 +95,48 @@ namespace :manage do
     end
   end
 
+  def update_business_finder_subscriptions
+    facet_group_content_id = "52435175-82ed-4a04-adef-74c0199d0f46".freeze
+
+    Subscription.transaction do
+      primary_business_finder_subscriber_list = SubscriberList.where(title: "EU Exit guidance for your business or organisation")
+      primary_business_finder_subscriber_list_id = primary_business_finder_subscriber_list.ids[0]
+
+      other_business_finder_subscriber_lists = SubscriberList.find_by_links_value(facet_group_content_id) - primary_business_finder_subscriber_list
+      other_business_finder_subscriber_list_ids = other_business_finder_subscriber_lists.map(&:id)
+
+      business_finder_subscriptions = Subscription.where(subscriber_list_id: other_business_finder_subscriber_list_ids)
+      subscriber_ids_that_moved_to_primary_subscriber_list = Subscription.where(subscriber_id: business_finder_subscriptions.map(&:subscriber_id)).where(subscriber_list_id: primary_business_finder_subscriber_list_id).map(&:subscriber_id)
+      subscriptions_to_move = business_finder_subscriptions.where.not(subscriber_id: subscriber_ids_that_moved_to_primary_subscriber_list)
+
+      subscriber_lists_to_change = SubscriberList.where(id: subscriptions_to_move.map(&:subscriber_list_id)).map(&:id)
+
+      subscriptions_to_move.each do |subscription|
+        old_subscriber_list = subscription.subscriber_list_id
+        subscription.subscriber_list_id = primary_business_finder_subscriber_list_id
+        subscription.save!
+        puts "Moved subscription #{subscription.id} from Subscriber List #{old_subscriber_list} to #{subscription.subscriber_list_id}"
+      end
+
+      # If Subscriber already moved to the primary, old Subscriptions remain under that Subscriber List and cannot be deleted
+      # Delete Subscriber Lists that have no Subscriptions
+      subscriber_lists_to_delete = []
+      subscriber_lists_to_change.each { |id|
+        subscriber_lists_to_delete << id if Subscription.where(subscriber_list_id: id) == []
+      }
+      SubscriberList.destroy(subscriber_lists_to_delete)
+
+      # If Subscriber List cannot be deleted, remove the links
+      subscriber_lists_to_remove_links = other_business_finder_subscriber_lists - subscriber_lists_to_delete
+
+      subscriber_lists_to_remove_links.each do |list|
+        list.links = {}
+        list.save!
+        puts "Removed facet_group links from Subscriber List #{list.id}"
+      end
+    end
+  end
+
   desc "Change the email address of a subscriber"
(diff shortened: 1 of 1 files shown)


lib/tasks/manage.rake
```diff
@@ -95,6 +95,26 @@ namespace :manage do
     end
   end
 
+  def update_business_finder_subscriptions
+    business_finder_facet_group_content_id = "52435175-82ed-4a04-adef-74c0199d0f46".to_s.freeze
```
ChrisBAshton: `.to_s` is not necessary here, it's already a string.

lib/tasks/manage.rake
```diff
@@ -95,6 +95,26 @@ namespace :manage do
...
+
+    primary_business_finder_subscriber_list = SubscriberList.where(title: "EU Exit guidance for your business or organisation")
+
+    business_finder_subscriber_lists = SubscriberList.find_by_links_value(business_finder_facet_group_content_id) - primary_business_finder_subscriber_list
+    business_finder_subscriber_list_ids = business_finder_subscriber_lists.each(&:id).to_a
```
ChrisBAshton: I _think_ this can be rewritten as:

```ruby
    business_finder_subscriber_list_ids = business_finder_subscriber_lists.map { &:id }
```

...but I might be wrong!

lib/tasks/manage.rake
```diff
@@ -156,4 +176,9 @@ namespace :manage do
   task :update_subscriber_list, %i[slug new_title new_slug] => :environment do |_t, args|
     update_subscriber_list(slug: args[:slug], new_title: args[:new_title], new_slug: args[:new_slug])
   end
+
+  desc "Update business finder subscriptions"
+  task update_business_finder_subscriptions: :environment do |_t|
```
ChrisBAshton: What do we use `:environment` for here? Does this rake task apply to all environments..?

lib/tasks/manage.rake
AgaDufrat: I had hard time with linter here, which wouldn't let me interpolate. 
When calling `find_by_links_value` the content_id has to be wrapped in a double quote so I want the string to be `"52435175-82ed-4a04-adef-74c0199d0f46"` not just `52435175-82ed-4a04-adef-74c0199d0f46`, if that makes sense. I could't think of a better way of doing it and keeping the linter happy. I will rename the variable to make it clearer.

lib/tasks/manage.rake
AgaDufrat: ah! I have to use 'map' because I want it to return a new array with ids

lib/tasks/manage.rake
AgaDufrat: From my understanding it loads the Rails environment and let's me use the existing classes (I'm using SubscriberList, Subscription) and methods

lib/tasks/manage.rake
ChrisBAshton: Oh ok, makes sense 👍 In that case I'd maybe add a comment to the code saying the string needs to have double-quotes on it.

Could that be done with something like this...?

```suggestion
    business_finder_facet_group_content_id = "\"52435175-82ed-4a04-adef-74c0199d0f46\"".freeze
```

lib/tasks/manage.rake
ChrisBAshton: `map` does return a new array 👍

lib/tasks/manage.rake
brucebolt: The other option would be to put the string here:
```
business_finder_facet_group_content_id = "52435175-82ed-4a04-adef-74c0199d0f46"
```

Then modify the value when invoking `find_by_links_value` (either with `to_s` or by escaping the quotes), e.g.
```
business_finder_subscriber_lists = SubscriberList.find_by_links_value(formatted_facet_group_content_id.to_s.freeze) - primary_business_finder_subscriber_list
```

It doesn't make much difference here since the value is only being used once, but this would be cleaner if the value for `business_finder_facet_group_content_id` were being reused elsewhere..

oscarwyatt:
My only thought is that it might be worth wrapping everything in a transaction and making all saves/updates use the bang method so if one update goes awry we'll know and can figure out what went wrong and ensure they all get transitioned

lib/tasks/manage.rake
```diff
@@ -95,6 +95,48 @@ namespace :manage do
...
+      subscriber_lists_to_change = SubscriberList.where(id: subscriptions_to_move.map(&:subscriber_list_id)).map(&:id)
+
+      subscriptions_to_move.each do |subscription|
+        old_subscriber_list = subscription.subscriber_list_id
+        subscription.subscriber_list_id = primary_business_finder_subscriber_list_id
+        subscription.save!
+        puts "Moved subscription #{subscription.id} from Subscriber List #{old_subscriber_list} to #{subscription.subscriber_list_id}"
+      end
+
+      # If Subscriber already moved to the primary, old Subscriptions remain under that Subscriber List and cannot be delete
```
brucebolt:
```suggestion
      # If Subscriber already moved to the primary, old Subscriptions remain under that Subscriber List and cannot be deleted
```

AgaDufrat: 🤦‍♀

brucebolt: To be fair, there was a typo in my comment about your typo in your comment until I quickly edited it 🙈

PR #855, opened 2019-05-15
Remove remove_content_purpose_supergroup_from_subscriber_list migration

koetsier:
This migration cannot be simply deployed in production because
Rails caches these columns and accessing them will cause errors
when hot-deploying.


koetsier:
> Should the schema file also change as a result of this?

True - reverted that too

PR #856, opened 2019-05-15
Adds 'find latest matching subscription' query

ChrisBAshton:
In order to facilitiate Trello card: https://trello.com/c/VlnvU2Dl
Depended on by: https://github.com/alphagov/gds-api-adapters/pull/928

## What

This PR:

-  adds a `/subscriptions/:id/latest` route which calls a new query, `FindLatestMatchingSubscription`. This query fetches the most *recent* subscription that matches the given subscription, i.e. has the same subscriber, subscription list and frequency.
- tests the route
- tests the logic added in `FindLatestMatchingSubscription`

## Why

It's intended to be used to detect unsubscribe requests from old subscription emails that the user may have clicked on, as this is the same subscription conceptually. In most cases this will simply return the original subscription, as a newer one won't exist.

Where there are multiple subscriptions that match the parameters, the query will return the one with the most recent `created_at`. This should be safe given:

- When a user changes their subscription through frequency change
- ...and when a user changes through unsubscribing and resubscribing
- ...a new subscription is created every time, with a more recent 'created_at'.
- In other words, it's not possible for an older record that has become inactive to become active again. It will always be a new subscription.

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -63,7 +63,8 @@ class SubscriptionsController < ApplicationController
   end
 
   def latest_matching
-    render json: { todo: true, comment: "Need to check that this method is correctly called!" }, status: ok
+    subscription = Subscription.find(subscription_params.require(:id))
+    render json: { subscription: FindLatestMatchingSubscription.call(subscription) }, status: ok
   end
 
 private
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -64,7 +64,7 @@ class SubscriptionsController < ApplicationController
 
   def latest_matching
     subscription = Subscription.find(subscription_params.require(:id))
-    render json: { subscription: FindLatestMatchingSubscription.call(subscription) }, status: ok
+    render json: { subscription: FindLatestMatchingSubscription.call(subscription) }, status: :ok
   end
 
 private
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -62,6 +62,10 @@ class SubscriptionsController < ApplicationController
     render json: { subscription: subscription }, status: :ok
   end
 
+  def latest_matching
+    render json: { todo: true, comment: "Need to check that this method is correctly called!" }, status: ok
+  end
+
 private
 
   def smoke_test_address?
--- a/app/queries/find_latest_matching_subscription.rb
+++ b/app/queries/find_latest_matching_subscription.rb
@@ -1,9 +1,11 @@
 class FindLatestMatchingSubscription
-  def self.call
-    new.call
-  end
-
-  def call(subscription)
-    # TODO - find the latest subscription that matches the given subscription's subscription_list and email address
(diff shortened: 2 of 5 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -62,6 +62,11 @@ def update
     render json: { subscription: subscription }, status: :ok
   end
 
+  def get_latest_matching
```
oscarwyatt: I think it'd be a nice thing if this matches the url route and _probably_ oughtn't include 'get' as the method doesn't need to include the HTTP verb (I can't find any style guidelines on controller method names so these are just my opinions). I'd go with making them all `latest_matching` as that's clearest 😃

app/queries/find_latest_matching_subscription.rb
```diff
@@ -0,0 +1,12 @@
+class FindLatestMatchingSubscription
+  def self.call(original_subscription)
+    subscriber_list_id = original_subscription.subscriber_list_id
+    subscriber_id = original_subscription.subscriber_id
+    frequency = original_subscription.frequency
+    latest_subscription = Subscription.
+        where(subscriber_list_id: subscriber_list_id, subscriber_id: subscriber_id, frequency: frequency).
+        order("created_at DESC")
+        .first
+    latest_subscription
```
oscarwyatt: Out of interest, is it possible that there could be no matches? If so, what should we do if this is the case?
Additionally (on a super pedantic note) you could just have the query (ie `Subscription.where...` as the last line and not both copying it into a variable and then returning it but if you have  a good readability reason then I can acquiesce

ChrisBAshton: Thanks, have fixed up the return value 👍 

There _should_ always be a match, because we're passing a valid subscription to the method, so when we do the database grab we should at the very least be getting back that same subscription! But to be ultra safe, I could wrap it in a `begin/rescue` and return the original subscription...?

What do you think? Felt to me like we'd be covering a case that will never happen (it's hard to even write a test for that!) but I'm happy to have a go 😄

oscarwyatt: Cool. I was just clarifying - I think it's fine to leave as is, I forgot that subscriptions don't get deleted, just marked as ended. 👍

app/queries/find_latest_matching_subscription.rb
```diff
@@ -0,0 +1,11 @@
+class FindLatestMatchingSubscription
+  def self.call(original_subscription)
+    subscriber_list_id = original_subscription.subscriber_list_id
+    subscriber_id = original_subscription.subscriber_id
+    frequency = original_subscription.frequency
+    Subscription.
+      where(subscriber_list_id: subscriber_list_id, subscriber_id: subscriber_id, frequency: frequency).
```
brucebolt: Do you need to check here whether the subscription is still active?

ChrisBAshton: Nope, just want the latest subscription that matches the email address, subscriber list and frequency of the original subscription. Can be active or inactive.

If inactive, we'll display "You're already unsubscribed".

PR #862, opened 2019-05-20
Removes frequency from FindLatestMatchingSubscription query

ChrisBAshton:
Fixes a logic error introduced in #856.
Depended on by https://github.com/alphagov/email-alert-frontend/pull/474
See Trello https://trello.com/c/VlnvU2Dl. 

## What

Removes frequency from the `FindLatestMatchingSubscription` check, and updates the tests to describe the behaviour we now expect.

## Why

We always need this to retrieve the _latest_ subscription, regardless of frequency, otherwise we're limited in what we can offer the user.

For example, a user who changes frequency from Daily to Weekly and then clicks Unsubscribe from an old Daily email would continue to see "You're already unsubscribed", as `/latest` would currently return the expired Daily subscription rather than the active Weekly subscription.

In this case, what we would actually like to do is redirect the user to manage their subscriptions, rather than display a potentially confusing message. This is not possible without removing frequency from the query here.

We can then implement the logic we want here: https://github.com/alphagov/email-alert-frontend/pull/474

### DIFF
--- a/app/queries/find_latest_matching_subscription.rb
+++ b/app/queries/find_latest_matching_subscription.rb
@@ -2,10 +2,9 @@ class FindLatestMatchingSubscription
   def self.call(original_subscription)
     subscriber_list_id = original_subscription.subscriber_list_id
     subscriber_id = original_subscription.subscriber_id
-    frequency = original_subscription.frequency
-    Subscription.
-      where(subscriber_list_id: subscriber_list_id, subscriber_id: subscriber_id, frequency: frequency).
-      order("created_at DESC")
+    Subscription
+      .where(subscriber_list_id: subscriber_list_id, subscriber_id: subscriber_id)
+      .order("created_at DESC")
       .first
   end
 end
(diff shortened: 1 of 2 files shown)


app/queries/find_latest_matching_subscription.rb
```diff
@@ -2,9 +2,8 @@ class FindLatestMatchingSubscription
   def self.call(original_subscription)
     subscriber_list_id = original_subscription.subscriber_list_id
     subscriber_id = original_subscription.subscriber_id
-    frequency = original_subscription.frequency
     Subscription.
-      where(subscriber_list_id: subscriber_list_id, subscriber_id: subscriber_id, frequency: frequency).
+      where(subscriber_list_id: subscriber_list_id, subscriber_id: subscriber_id).
```
brucebolt: Would it be possible to make the chaining dots consistent over the multiple lines (i.e. at the beginning of the following line)?

PR #865, opened 2019-05-23
[do-not-merge] Add migration to add type field to subscriber_lists

oscarwyatt:
So we can use Single Type inheritance for subscriber_lists

Trello: https://trello.com/c/y6ZhETsr/130-implement-a-joined-subscriber-list-model-in-email-alert-api-recommended-%F0%9F%8D%90

Joint effort with @AgaDufrat


leenagupte:
The commit is replicated in PR #866, so can this one be closed?

PR #867, opened 2019-05-23
Adds or joined facet subscriber list

oscarwyatt:
Will match a content change against this type of subscriber list
when at least one facet matches.
Adds in new parameter for subscriber_lists#create 'join_facets_with'
which defaults to and (if not included) but if a value of 'or' is
supplied it will create an OrJoinedFacetSubscriberList and will
match the resulting subscriber list with an or join on the facets

Trello: https://trello.com/c/y6ZhETsr/130-implement-a-joined-subscriber-list-model-in-email-alert-api-recommended-%F0%9F%8D%90

Joint effort with @AgaDufrat

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -39,7 +39,13 @@ private
       title: title,
       slug: slug,
       signon_user_uid: current_user.uid,
+      type: type
     )
+  end
+
+  def type
+    combine_mode = params[:combine_mode].present? && params[:combine_mode] == "or"
+    combine_mode ? 'OrJoinedFacetSubscriberList' : ''
   end
 
   def convert_legacy_params(link_or_tags)
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -39,7 +39,6 @@ private
       title: title,
       slug: slug,
       signon_user_uid: current_user.uid,
-      type: type
     )
   end
 
@@ -56,7 +55,7 @@ private
 
   def find_exact_query_params
     permitted_params = params.permit!.to_h
-    {
+    exact_query_params = {
       tags: convert_legacy_params(permitted_params.fetch(:tags, {})),
       links: convert_legacy_params(permitted_params.fetch(:links, {})),
       document_type: permitted_params.fetch(:document_type, ""),
@@ -64,6 +63,8 @@ private
       government_document_supertype: permitted_params.fetch(:government_document_supertype, ""),
       slug: params[:gov_delivery_id],
     }
+    exact_query_params[:type] = type if params[:combine_mode].present?
+    exact_query_params
   end
 
   def slugify(title)
--- a/app/models/subscriber_list.rb
+++ b/app/models/subscriber_list.rb
@@ -44,7 +44,7 @@ class SubscriberList < ApplicationRecord
   def to_json(options = {})
     options[:except] ||= %i{signon_user_uid}
     options[:methods] ||= %i{subscription_url gov_delivery_id active_subscriptions_count}
-    super(options)
+    convert_to_subscriber_list_json(as_json(options))
   end
 
   def is_travel_advice?
@@ -75,5 +75,17 @@ private
         %i[all any].include?(operator) && values.is_a?(Array)
       end
     end
+  end
+
+  def convert_to_subscriber_list_json(subscriber_list_hash)
+    # If it is already correct, just return
+    return ActiveSupport::JSON.encode(subscriber_list_hash) if subscriber_list_hash.has_key?("subscriber_list")
+
+    # to_json and as_json will return a hash with the root key as
+    # or_joined_facet_subscriber_list, for that type which
+    # will make breaking changes.
+    # Instead, just change the root key to 'subscriber_list'
+    result_with_super_table_type = { "subscriber_list" => subscriber_list_hash.delete(type.underscore) }
+    ActiveSupport::JSON.encode(result_with_super_table_type)
   end
 end
--- a/app/queries/matched_for_notification.rb
+++ b/app/queries/matched_for_notification.rb
@@ -7,46 +7,70 @@ class MatchedForNotification
   end
 
   # Filter out subscriber lists:
-  # - where the keys from the supplied hash are a superset of the keys in the specified query_field AND
+  # - where the keys from the supplied hash:
+  #   - Have a superset of the keys in the specified query_field for subscriber_lists
+  #   - Have an overlap with  the keys in the specified query_field for or_joined_facet_subscriber_lists
+  # - AND
   # - If the operator is 'any':
   #     the key values in the subscriber list have at least one corresponding match in the supplied hash.
   # - If the operator is 'all'
   #   all key values in the subscriber list have corresponding matches in the supplied hash.
-  # Note that this means that not all keys from the the supplied hash are required to be matched.
   def call(content_item_tags_or_links)
     return [] unless content_item_tags_or_links.present?
 
     content_item_tags_or_links = content_item_tags_or_links.deep_symbolize_keys
 
     only_contains_keys(content_item_tags_or_links.keys).select do |subscriber_list|
-      subscriber_list_tags_or_links = subscriber_list.send(@query_field) # send ensures the keys are symbols
-
-      subscriber_list_tags_or_links.keys.all? do |key|
-        content_item_values = Array(content_item_tags_or_links[key])
-        any_values = subscriber_list_tags_or_links[key].fetch(:any, [])
-        all_values = subscriber_list_tags_or_links[key].fetch(:all, [])
-
-        (all_values - content_item_values).empty? &&
-          (any_values.empty? || (any_values & content_item_values).any?)
-      end
+      filter_tags_or_links(subscriber_list, content_item_tags_or_links)
     end
   end
 
 private
 
-  # Return all SubscriberLists which have a subset of keys from those requested.
+  def filter_tags_or_links(subscriber_list, content_item_tags_or_links)
+    operator = subscriber_list.type == 'OrJoinedFacetSubscriberList' ? :any? : :all?
+    subscriber_list_tags_or_links = subscriber_list.send(@query_field) # send ensures the keys are symbols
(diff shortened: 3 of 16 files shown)


db/migrate/20190522114510_migrate_subscriber_lists_to_and_joined_facet_subscriber_lists.rb
```diff
@@ -0,0 +1,5 @@
+class MigrateSubscriberListsToAndJoinedFacetSubscriberLists < ActiveRecord::Migration[5.2]
+  def change
+    SubscriberList.where(type: nil).update_all(type: 'AndJoinedFacetSubscriberList')
```
leenagupte: Is it really necessary to do this? Isn't it less risky to just create a new Subscriber List for "ORs" rather than migrating all of the old ones as well?

db/schema.rb
```diff
@@ -99,11 +99,27 @@
...
     t.index ["subscriber_list_id"], name: "index_matched_content_changes_on_subscriber_list_id"
   end
 
+  create_table "or_joined_subscriber_list_subscriber_lists", force: :cascade do |t|
+    t.integer "subscriber_list_id"
+    t.integer "or_joined_subscriber_list_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["id", "or_joined_subscriber_list_id"], name: "idx_or_joined_sub_list_sub_list_on_or_joined_sub_list_id"
+    t.index ["id", "subscriber_list_id"], name: "idx_or_joined_sub_list_sub_list_on_sub_list_id"
```
ChrisBAshton: These names read as having one too many `sub_list`s - is this something we have control over or is it enforced by Rails?

app/models/and_joined_facet_subscriber_list.rb
```diff
@@ -0,0 +1,3 @@
+class AndJoinedFacetSubscriberList < SubscriberList
```
ChrisBAshton: Is this class necessary? I thought we were just doing Or-Joined lists? PR description doesn't mention the And-Joined list either 🤔

app/queries/matched_for_notification.rb
```diff
@@ -7,46 +7,70 @@ def initialize(query_field: :tags, scope: SubscriberList)
...
-      end
+      filter_tags_or_links(subscriber_list, content_item_tags_or_links)
     end
   end
 
 private
 
-  # Return all SubscriberLists which have a subset of keys from those requested.
+  def filter_tags_or_links(subscriber_list, content_item_tags_or_links)
+    operator = subscriber_list.type == 'OrJoinedFacetSubscriberList' ? :any? : :all?
```
ChrisBAshton: This is nice though 😄

spec/integration/create_subscriber_list_spec.rb
```diff
@@ -24,7 +24,7 @@
       end
 
       it "creates another subscriber list with a different slug" do
-        create_subscriber_list(title: "oil and gas", tags: { topics: { any: ["oil-and-gas/licensing"] } })
+        create_and_joined_facet_subscriber_list(title: "oil and gas", tags: { topics: { any: ["oil-and-gas/licensing"] } })
```
ChrisBAshton: A little confused as to why we're making an AndJoined list out of _just_ 'oil-and-gas' - isn't that one topic?

spec/integration/create_subscriber_list_spec.rb
```diff
@@ -101,22 +101,22 @@
...
       )
 
       expect(response.status).to eq(422)
     end
 
     it "successfully creates two SubscriberList objects with the same title" do
-      create_subscriber_list(title: "oil and gas", links: { taxons: { any: ["oil-and-gas"] } })
-      create_subscriber_list(title: "oil and gas", links: { policies: { any: ["oil-and-gas/licensing"] } })
+      create_and_joined_facet_subscriber_list(title: "oil and gas", links: { taxons: { any: ["oil-and-gas"] } })
+      create_and_joined_facet_subscriber_list(title: "oil and gas", links: { policies: { any: ["oil-and-gas/licensing"] } })
```
ChrisBAshton: I'm a little nervous even suggesting it, but... would the method name be clearer if we renamed it to: `create_AND_joined_facet_subscriber_list`?

I know that goes against the standard rules of the language but it currently reads as "create and joined", which is kind of like "created and joined", and takes a bit of rethinking to actually understand what it's doing. Whereas with the capitals it hopefully reads as "create an AND-joined list" a bit more clearly.

Feel free to burn this idea with fire.

spec/models/subscriber_list_spec.rb
```diff
@@ -120,4 +120,78 @@
...
+                      document_type
+                      tags
+                      links
+                      email_document_supertype
+                      government_document_supertype
+                      slug
+                      subscription_url
+                      gov_delivery_id
+                      active_subscriptions_count),
+                   parsed_json["subscriber_list"].keys
```
ChrisBAshton: could we extract these out into a single method, as they're the same each time?

```ruby
assert_subscriber_list(parsed_json)

def assert_subscriber_list(parsed_json)
      assert_equal %w(id
                      title
                      created_at
                      updated_at
                      document_type
                      tags
                      links
                      email_document_supertype
                      government_document_supertype
                      slug
                      subscription_url
                      gov_delivery_id
                      active_subscriptions_count),
                   parsed_json["subscriber_list"].keys
end
```

spec/queries/matched_for_notification_spec.rb
```diff
@@ -1,73 +1,177 @@
...
+      @subscriber_list6 = create(:subscriber_list, links: { topics: { all: ["uuid-345", "uuid-456"], any: ["uuid-567", "uuid-678"] } })
+
+      @subscriber_list7 = create(:subscriber_list, links: { topics:
+                                                                                                  {
+                                                                                                      all: ["uuid-345", "uuid-456"]
+                                                                                                  },
+                                                                                              policies:
+                                                                                                  {
+                                                                                                      all: ["uuid-567", "uuid-678"]
+                                                                                                  } })
```
ChrisBAshton: The formatting seems quite off here - can it be fixed up? Difficult to read.

spec/queries/matched_for_notification_spec.rb
```diff
@@ -1,73 +1,177 @@
...
+        expect(lists).to eq([@or_joined_facet_subscriber_list1])
+      end
+
+      it 'will not match an OrJoinedSubscriberList with multiple facets where no facets match' do
+        lists = execute_query(field: :tags, query_hash: { format: %w[employment_tribunal_decision] })
+        expect(lists).to eq([])
+      end
+
+      it 'will not match an OrJoinedSubscriberList if the query match only matches one of just one facet that has an all' do
+        lists = execute_query(field: :tags, query_hash: { organisations: %w[organisation_1_list_2] })
```
ChrisBAshton: Think these tests would be easier to read if they followed the Arrange, Act, Assert model. Currently the 'Arrange' all happens at the top and you have to scroll up and down to figure out what's going on.

I also wonder if a helper method might make the 'Arrange' parts easier to read.

That way we can do away with the naming that doesn't really mean anything ("organisation 1 list 2")

db/schema.rb
oscarwyatt: fortunately, this table wasnt actually necessary so it's gone

db/schema.rb
```diff
@@ -99,11 +99,27 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.uuid "content_change_id", null: false
+    t.bigint "or_joined_subscriber_list_id"
```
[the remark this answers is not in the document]
oscarwyatt: Whoops, this wasn't meant to be here

app/models/and_joined_facet_subscriber_list.rb
oscarwyatt: I _thought_ it might make it clearer which lists were doing what but happy to remove it

spec/models/subscriber_list_spec.rb
oscarwyatt: 👍

spec/queries/matched_for_notification_spec.rb
oscarwyatt: Yup. Moral of the story: the linter can fix offences, but don't expect it to make things easier to read!

spec/models/subscriber_list_spec.rb
leenagupte: I think refactoring out the set-up is ok, but I think that having a shared expectation makes the test harder to read. 
Perhaps just refactor the long list of keys in to shared variable?

spec/queries/matched_for_notification_spec.rb
oscarwyatt: Agreed, this was hard to read when I started and now it's worse. Will have a think about ways to improve it

app/models/subscriber_list.rb
```diff
@@ -76,4 +76,16 @@ def valid_subscriber_criteria(link_or_tags)
       end
     end
   end
+
+  def convert_to_subscriber_list_json(subscriber_list_hash)
+    # If it is already correct, just return
+    return ActiveSupport::JSON.encode(subscriber_list_hash) if subscriber_list_hash.has_key?("subscriber_list")
+
+    # to_json and as_json will return a hash with the root key as
+    # or_joined_facet_subscriber_list or and_joined_facet_subscriber_list
```
leenagupte: I think this comment is out of date now.

app/controllers/subscriber_lists_controller.rb
```diff
@@ -39,9 +39,15 @@ def subscriber_list_params
       title: title,
       slug: slug,
       signon_user_uid: current_user.uid,
+      type: type
     )
   end
 
+  def type
+    combine_mode = params[:combine_mode].present? && params[:combine_mode] == "or"
+    (combine_mode ? 'OrJoinedFacetSubscriberList' : '')
```
brucebolt: Are the parentheses needed here?  (I don't recall seeing them in the last iteration of this PR)

oscarwyatt: I got paranoid but removed it and the tests still pass, good spot

PR #868, opened 2019-05-23
Add missing tests for any/all behaviour in email alerts

brucebolt:
Before implementing the `OrJoinedSubscriberList` model, we need to be confident that tests are in place to ensure the new queries do not affect queries for the existing `SubscriberList`.

This adds additional tests where the `all` behaviour was not properly being tested.  We only ensured that the `SubscriberList` was found when all values were present.  There was no test for ensuring the `SubscriberList` was not found when not all the topics are present, therefore `all` could have been implemented as `any` and the tests would still have passed.  This PR adds those tests.

Additionally, this is adding a test where `all`/`any` is combined across multiple tags, which not previously tested.

Trello card: https://trello.com/c/Qt6sEJ2o/129-implement-review-baseline-tests-for-finder-email-alerts-recommended-%F0%9F%8D%9

### DIFF
--- a/spec/queries/matched_for_notification_spec.rb
+++ b/spec/queries/matched_for_notification_spec.rb
@@ -59,15 +59,31 @@ RSpec.describe MatchedForNotification do
       expect(lists).to include(@list5)
     end
 
+    it 'does not find subscriber list for all topics when not all topics present' do
+      lists = execute_query(field: :links, query_hash: { topics: ["uuid-234", "other2"] })
+      expect(lists).not_to include(@list5)
+    end
+
     it 'finds subscriber lists matching any and all topics' do
       lists = execute_query(field: :links, query_hash: { topics: ["uuid-345", "uuid-678", "uuid-456"] })
       expect(lists).to include(@list6)
+    end
+
+    it 'does not find subscriber list for mixture of all and any topics when not all topics present' do
+      lists = execute_query(field: :links, query_hash: { topics: ["uuid-345", "uuid-678"] })
+      expect(lists).not_to include(@list6)
     end
 
     it 'finds subscriber lists matching a mix of all topics and policies' do
       lists = execute_query(field: :links, query_hash: { topics: ["uuid-345", "uuid-456", "other1"],
                                                          policies: ["uuid-567", "uuid-678", "other2"] })
       expect(lists).to include(@list7)
+    end
+
+    it 'does not find subscriber list for mix of all topics and policies when not all policies present' do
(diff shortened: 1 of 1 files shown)


spec/queries/matched_for_notification_spec.rb
```diff
@@ -59,17 +59,33 @@ def execute_query(field:, query_hash:)
       expect(lists).to include(@list5)
     end
 
+    it 'does not find subscriber list for all topics when not all topics present' do
+      lists = execute_query(field: :links, query_hash: { topics: ["uuid-234", "other-2"] })
```
ChrisBAshton:
```suggestion
      lists = execute_query(field: :links, query_hash: { topics: ["uuid-234", "some-random-uuid"] })
```

spec/queries/matched_for_notification_spec.rb
```diff
@@ -59,17 +59,33 @@ def execute_query(field:, query_hash:)
...
       lists = execute_query(field: :links, query_hash: { topics: ["uuid-345", "uuid-456", "other1"],
                                                          policies: ["uuid-567", "uuid-678", "other2"] })
       expect(lists).to include(@list7)
     end
 
+    it 'does not find subscriber list for mix of all topics and policies when not all policies present' do
+      lists = execute_query(field: :links, query_hash: { topics: ["uuid-345", "uuid-456", "other1"],
+                                                         policies: ["uuid-678", "other2"] })
+      expect(lists).not_to include(@list7)
+    end
```
ChrisBAshton: Good to have these tests 👍 but I'll make the same suggestion as I did on Oscar & Aga's PR: https://github.com/alphagov/email-alert-api/pull/867#discussion_r287263976

> Think these tests would be easier to read if they followed the Arrange, Act, Assert model. Currently the 'Arrange' all happens at the top and you have to scroll up and down to figure out what's going on.

spec/queries/matched_for_notification_spec.rb
```diff
@@ -29,6 +29,15 @@
                                                    {
                                                      all: ["uuid-567", "uuid-678"]
                                                    } })
+
+      @list8 = create(:subscriber_list, links: { topics:
```
ChrisBAshton: These lists could be renamed, e.g.
instead of `@list7` and `@list8`
could have `@list_of_all_topics_and_all_policies` and `@list_of_all_topics_and_any_policies`

spec/queries/matched_for_notification_spec.rb
brucebolt: Maybe this could come with a later refactor?  Some of the lists are used in multiple tests, so we'd have to rethink the way the tests are broken down into smaller units.

PR #888, opened 2019-06-11
Introduce new internal_failure status

thomasleese:
This is used to allow us to make a distinction between two kinds of `technical_failure`s. Right now `technical_failure` covers any kind of unrecoverable/unknown error we experience while sending an email.

With this change, we have the following meanings:

- **`technical_failure`** is a valid status response from our provider (Notify) when it's unable to send an email for an unspecified reason. It's also the response we use when Notify gives us back some sort of error response such as HTTP error code `500`.

- **`internal_failure`** is set when we receive an exception from our provider or in our code that wasn't handled - implying that the failure condition was unexpected. This is likely to tell us about problems in our code rather than in a failure with Notify. This was the case in a recent incident where we were unable to start a connection to Notify but we were reporting this failure as `technical_failure` whereas now it would return a `internal_failure`.

[Trello Card](https://trello.com/c/x7qVo3IM/1051-improve-email-alert-api-code-to-differentiate-between-notify-errors-and-errors-in-our-code) · [Incident Report](https://docs.google.com/document/d/1NyxVSFzQnBSOAlVbeRagVW1SjXdkJHjrKz2xnWYhwR0/edit#)

### DIFF
--- a/app/models/delivery_attempt.rb
+++ b/app/models/delivery_attempt.rb
@@ -5,7 +5,7 @@ class DeliveryAttempt < ApplicationRecord
 
   FINAL_STATUSES = %i[delivered permanent_failure].freeze
 
-  enum status: { sending: 0, delivered: 1, permanent_failure: 2, temporary_failure: 3, technical_failure: 4 }
+  enum status: { sending: 0, delivered: 1, permanent_failure: 2, temporary_failure: 3, technical_failure: 4, internal_failure: 5 }
   enum provider: { pseudo: 0, notify: 1 }
 
   def has_final_status?
--- a/app/models/healthcheck/delivery_status.rb
+++ b/app/models/healthcheck/delivery_status.rb
@@ -1,0 +1,46 @@
+module Healthcheck
+  class DeliveryStatus < GovukHealthcheck::ThresholdCheck
+    def value
+      return 0 if total.zero?
+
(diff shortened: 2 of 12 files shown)


thomasleese:
@kevindew Yeah... deciding on the name took most of the time in writing this PR...

My first thought was actually to reverse them so `provider_failure` meant a failure in the provider (like you thought at first glance) and `technical_failure` was an issue in our code. However, as you say, the reason I decided against that was because our statuses are so closely tied to the statuses from Notify. If someone read the Notify documentation they would see the `technical_failure` status response and if they saw that in our code it would be easy to assume that stands for the same thing.

Thinking about it now, perhaps a better name is `internal_failure` since `provider_failure` does have an implication that it's related to the provider, whereas we're really trying to expose issues in our code (granted this would also cover unhandled errors in the Notify gem, but it'll be hard to make that distinction).

`provider_communication_failure` is a bit wordy, however, since we're really just doing a `rescue StandardError` it's entirely possible that it'll get caught under errors which aren't strictly related to communication with the provider.

kevindew:
`internal_failure` sounds a good option 👍 

I think code errors within the notify gem still count as internal failures as they're caused by the local ruby code rather than the external API.

thomasleese:
@kevindew Cool, I've rebased this PR changing the name to `internal_failure`.

PR #893, opened 2019-06-25
Rake task for updating subscriber list facet values & titles

ChrisBAshton:
We recently split and merged some facet values for the EU Exit Business
Readiness Finder (see [alphagov/content-tagger#941](https://github.com/alphagov/content-tagger/pull/941)). We now need to migrate users who had subscribed to the old facets.

This PR adds a rake task to perform the migration for us, and should be removed once the task has been run. Tests have been written to have confidence that the task will work.

Technical approach:

> Iterate through the affected subscriber lists and replace any facet values
> that need replacing with a newer one. Then remove old facet values and write
> only the unique values that remain. Finally, take those updated facet values
> to update the title.

Trello card: https://trello.com/c/SlZdVUbe/.

### DIFF
--- a/lib/tasks/eu_exit_business_finder.rake
+++ b/lib/tasks/eu_exit_business_finder.rake
@@ -1,0 +1,83 @@
+namespace :eu_exit_business_finder do
+  task update_subscriber_list_titles_and_facet_values: [:environment] do
+    subscriber_lists = SubscriberList
+      .where("links::text != '{}'")
+      .where("ARRAY(SELECT json_object_keys(links)) <@ Array[:keys]", keys: [:facet_values])
+    update_facet_values(subscriber_lists)
+  end
+
+  # Iterate through the affected subscriber lists and if we encounter any
+  # facet values that need replacing with a newer one, we insert the new one
+  # to the array (inserted 'in place' to retain original order). Then remove
+  # the old facet values in bulk and write only the unique values that remain.
+  def update_facet_values(subscriber_lists)
+    subscriber_lists.each do |list|
+      links = list.links[:facet_values][:any]
+      updated_links = links.map { |original_value|
+        EuExitFacetMigrationConfig::facet_values_to_replace[original_value] || original_value
+      }.flatten
+      updated_links = (updated_links - EuExitFacetMigrationConfig::facet_values_to_remove).uniq
+      list.write_attribute('links', facet_values: { any: updated_links })
+      list.write_attribute('title', title_from_facets(updated_links))
+      list.save
+    end
+  end
+
+  def title_from_facets(facet_values)
+    subscription_title_prefix = "EU Exit guidance for your business in the following #{'category'.pluralize(facet_values.count)}: "
+    titles = facet_values.map do |facet_value|
+      if EuExitFacetMigrationConfig.facet_value_label_overrides[facet_value].present?
+        "'#{EuExitFacetMigrationConfig.facet_value_label_overrides[facet_value]}'"
+      else
+        content_item_title = GdsApi.publishing_api_v2.get_content(facet_value).to_h['title']
+        "'#{content_item_title}'"
+      end
+    end
+    subscription_title_prefix + titles.join(', ')
+  end
+end
+
+module EuExitFacetMigrationConfig
+  def self.facet_values_to_replace
+    {
+      "9f3476e1-8ff0-455d-a14e-003236b2797c" => ["53f9ce4c-7cbb-447f-bdf1-a9b022896d3a"],
+      "b4e507df-f067-4749-9468-3de120775216" => ["24fd50fa-6619-46ca-96cd-8ce90fa076ce"],
+      "356f46a0-17d3-4ba4-8952-8c02244f904" => ["18c3892e-8a0e-4884-906e-5938380eceee"],
+      "356f46a0-17d3-4ba4-8952-8c02244f9045" => ["18c3892e-8a0e-4884-906e-5938380eceee"],
+      # "34a6edd0-46ea-4a76-ae80-8c96709d4f5" is worth noting as
+      # it's the result of a merge between two facets. The
+      # find-and-replace works the same way regardless.
+      "1e3e8abd-135d-4844-afa8-5c51df3d3c57" => ["34a6edd0-46ea-4a76-ae80-8c96709d4f59"],
+      "c1d8057c-76bf-431c-9ac8-6281a4b7b9ca" => ["34a6edd0-46ea-4a76-ae80-8c96709d4f59"],
+      # "94b3cfe2-af89-4744-b8d7-7fc79edcbc85" is a special case
+      # as it's been split into two facets. We should assume the
+      # user wants to remain subscribed to both.
+      "94b3cfe2-af89-4744-b8d7-7fc79edcbc85" => [
+        "9d54c591-f5ca-4d0c-a484-12d5591987cb",
+        "afd45eef-743a-417d-9245-3eab8322116d"
+      ],
+      # to fix a missing character issue introduced in
+      # https://github.com/alphagov/search-api/commit/8fce3db218821dfc7a2b50dbd1b03984b1ec41b1
+      "7620da7a-0427-4b3c-9498-db9dc25209b" => ["7620da7a-0427-4b3c-9498-db9dc25209b0"]
+    }
+  end
+
+  def self.facet_values_to_remove
+    [
+      "7536c0c4-fb41-43f4-a2c4-08f4fa9f5427",
+      "5faa1741-fc55-4110-b342-de92f6324118",
+      "14cf2a68-3297-44d3-ba01-a4426845b1b8",
+      "040649fc-4e2c-4028-b846-77fe3eebd1f7",
+      "94b3cfe2-af89-4744-b8d7-7fc79edcbc85"
+    ]
+  end
+
+  def self.facet_value_label_overrides
+    {
+      "5476f0c7-d029-459b-8a17-196374ae3366" => "Employing EU citizens",
+      "bbdbda71-b1ec-46b8-a5b8-931d933288e9" => "Employing non-EU citizens",
+      "f165dc7c-7cef-446a-bdfd-8a1ca685d091" => "Public sector procurement - civil government contracts",
+      "33fc20d7-6a45-40c9-b31f-e4678f962ff1" => "Public sector procurement - defence contracts"
+    }
+  end
+end
--- a/spec/lib/tasks/eu_exit_business_finder_spec.rb
+++ b/spec/lib/tasks/eu_exit_business_finder_spec.rb
@@ -1,0 +1,179 @@
+require 'rake'
+require 'rails_helper'
+require 'gds_api/test_helpers/publishing_api_v2'
+
+RSpec.describe 'eu_exit_business_finder:update_subscriber_list_titles_and_facet_values' do
+  include GdsApi::TestHelpers::PublishingApiV2
+
+  before :each do
+    Rails.application.load_tasks
+  end
+
+  def mock_facet_values_to_remove(values)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_values_to_remove).and_return(values)
+  end
+
+  def mock_facet_values_to_replace(replacements)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_values_to_replace).and_return(replacements)
+  end
+
+  def mock_facet_value_label_overrides(replacements)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_value_label_overrides).and_return(replacements)
+  end
+
+  let(:facet_values) { [] }
+  let(:mock_behaviour!) {}
+  let(:create_list_and_execute_rake!) {
+    create(:subscriber_list, links: {
+      facet_values: {
+        any: facet_values
+      }
+    })
+    Rake::Task['eu_exit_business_finder:update_subscriber_list_titles_and_facet_values'].execute
+  }
+
+  describe "updating the facet values" do
+    subject {
+      stub_publishing_api_has_item("content_id" => "111-new")
+      stub_publishing_api_has_item("content_id" => "222-new")
+      stub_publishing_api_has_item("content_id" => "123-keep-me")
+      mock_behaviour!
+      create_list_and_execute_rake!
+      SubscriberList.last.links[:facet_values][:any]
+    }
+
+    context "given values that should not be removed" do
+      let(:facet_values) { ["123-keep-me"] }
+
+      it { is_expected.to eql ["123-keep-me"] }
+    end
+
+    context "given only values to remove" do
+      let(:facet_values) { ["000-remove-me"] }
+      let(:mock_behaviour!) { mock_facet_values_to_remove(["000-remove-me"]) }
+
+      it { is_expected.to eql [] }
+    end
+
+    context "given a mixture of values to keep and remove" do
+      let(:facet_values) { ["123-keep-me", "000-remove-me"] }
+      let(:mock_behaviour!) { mock_facet_values_to_remove(["000-remove-me"]) }
+
+      it { is_expected.to eql ["123-keep-me"] }
+    end
+
+    context "given values to replace" do
+      let(:facet_values) { ["000-old"] }
+      let(:mock_behaviour!) {
+        mock_facet_values_to_replace("000-old" => ["111-new"])
+      }
+
+      it { is_expected.to eql ["111-new"] }
+    end
+
+    context "given a mixture of values to replace, remove and keep" do
+      let(:facet_values) { ["000-old", "000-remove-me", "123-keep-me"] }
+      let(:mock_behaviour!) {
+        mock_facet_values_to_remove(["000-remove-me", "000-old"])
+        mock_facet_values_to_replace("000-old" => ["111-new"])
+      }
+
+      it { is_expected.to eql ["111-new", "123-keep-me"] }
+    end
+
+    context "given facet replacements which could lead to duplicates" do
+      let(:facet_values) { ["000-old", "001-also-old"] }
+      let(:mock_behaviour!) {
+        mock_facet_values_to_replace(
+          "000-old" => "111-new",
+          "001-also-old" => "111-new",
+        )
+      }
+
+      it { is_expected.to eql ["111-new"] } # `111-new` should only appear once
+    end
+
+    context "given a facet value which has been split into multiple new facets" do
+      let(:facet_values) { ["000-old"] }
+      let(:mock_behaviour!) {
+        mock_facet_values_to_replace(
+          "000-old" => ["111-new", "222-new"]
+        )
+      }
+
+      it { is_expected.to eql ["111-new", "222-new"] }
+    end
+  end
+
+  describe "updating the title" do
+    subject {
+      stub_publishing_api_has_item(
+        "content_id": "000-content-id-of-food-drink-tobacco-facet",
+        "title": "Food, drink and tobacco (retail and wholesale)",
+      )
+      stub_publishing_api_has_item(
+        "content_id": "111-content-id-of-ports-airports",
+        "title": "Ports and airports",
+      )
+      stub_publishing_api_has_item(
+        "content_id": "777-content-id-of-employing-eu-citizens",
+        "title": "EU citizens",
+      )
+      stub_publishing_api_has_item(
+        "content_id": "444-content-id-of-eu-funding",
+        "title": "EU funding",
+      )
+      mock_behaviour!
+      create_list_and_execute_rake!
+      SubscriberList.last.title
+    }
+
+    context "build title from single facet value" do
+      let(:facet_values) { ["000-content-id-of-food-drink-tobacco-facet"] }
+
+      it {
+        is_expected.to eql "EU Exit guidance for your business in the following category: 'Food, drink and tobacco (retail and wholesale)'"
+      }
+    end
+
+    context "build title from multiple facet values" do
+      let(:facet_values) { ["000-content-id-of-food-drink-tobacco-facet", "111-content-id-of-ports-airports"] }
+
+      it {
+        is_expected.to eql "EU Exit guidance for your business in the following categories: 'Food, drink and tobacco (retail and wholesale)', 'Ports and airports'"
+      }
+    end
+
+    context "build title from facet value whose title has been overridden" do
+      let(:facet_values) { ["000-content-id-of-food-drink-tobacco-facet"] }
+      let(:mock_behaviour!) {
+        mock_facet_value_label_overrides(
+          "000-content-id-of-food-drink-tobacco-facet" => "Special category",
+        )
+      }
+
+      it { is_expected.to eql "EU Exit guidance for your business in the following category: 'Special category'" }
+    end
+
+    context "build title from a mixture of facet values and overridden facet titles" do
+      let(:facet_values) {
+        [
+          "000-content-id-of-food-drink-tobacco-facet",
+          "111-content-id-of-ports-airports",
+          "777-content-id-of-employing-eu-citizens",
+          "444-content-id-of-eu-funding",
+        ]
+      }
+      let(:mock_behaviour!) {
+        mock_facet_value_label_overrides(
+          "777-content-id-of-employing-eu-citizens" => "Employing EU citizens",
+          "000-content-id-of-food-drink-tobacco-facet" => "Special category",
+        )
+      }
+
+      it {
+        is_expected.to eql "EU Exit guidance for your business in the following categories: 'Special category', 'Ports and airports', 'Employing EU citizens', 'EU funding'"
+      }
+    end
+  end
+end


spec/lib/tasks/fixtures/find-eu-exit-guidance-business.json
```diff
@@ -0,0 +1 @@
...ath":"/api/content/find-eu-exit-guidance-business","withdrawn":false,"content_id":"42ce66de-04f3-4192-bf31-8394538e0734","locale":"en","api_url":"https://www.gov.uk/api/content/find-eu-exit-guidance-business","web_url":"https://www.gov.uk/find-eu-exit-guidance-business","links":{}}]},"description":"This is the information that has been published so far for your business to prepare for EU exit. You can change what information you get using the checkboxes. Come back to this page regularly or sign up to receive emails when new information is published.","details":{"beta":false,"document_noun":"publication","summary":"This is the information that has been published so far for your business to prepare for EU exit. You can change what information you get using the checkboxes. Come back to this page regularly or sign up to receive emails when new information is published.","canonical_link":true,"sort":[{"name":"Topic","key":"topic","default":true},{"name":"Most viewed","key":"-popularity"},{"name":"Relevance","key":"-relevance"},{"name":"Most recent","key":"-public_timestamp"},{"name":"A to Z","key":"title"}],"filter":{"facet_groups":["52435175-82ed-4a04-adef-74c0199d0f46"]},"facets":[]}}
```
brucebolt: Would it be possible to prettify this JSON, so it's not all on one line?

Also, I wonder whether the whole content item is required, or whether a small amount of test data (even with fake facet names) could be added?

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,106 @@
...
+      updated_links = updated_links - EuExitFacetMigrationConfig::facet_values_to_remove
+      updated_links = updated_links.uniq
+      list.write_attribute('links', facet_values: { any: updated_links })
+      list.write_attribute('title', title_from_facets(updated_links))
+      list.save
+    end
+  end
+
+  # Title is constructed from finder content item, so we have to grab the JSON.
+  def title_from_facets(links)
```
brucebolt: Is there any possibility to use `EmailAlertListTitleBuilder` instead of copying much of the code into this rake task?  If this rake task ever needs to be used again, it would save future developers having to check the code and hard-coded text matches whatever is in `EmailAlertListTitleBuilder`.

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,106 @@
...
+    json_response = JSON.parse(Net::HTTP.get(URI('https://www.gov.uk/api/content/find-eu-exit-guidance-business')))
```
brucebolt: Would it be possible to use `gds-api-adapters` to get this from publishing-api?  It would prevent any issues with egress rules on machines in production.

spec/lib/tasks/fixtures/find-eu-exit-guidance-business.json
ChrisBAshton: Hmmm I _could_ delete everything apart from `['links']['facet_group'][0]['links']['facets']` but would have to keep that hierarchy in order for the code to work on production. I envisage us removing this code once we've run the rake task once, so it didn't seem worth the effort to chop it around, plus this gives me confidence it will work against the production API and I won't have accidentally mangled it. I'd suggest leaving it un-prettified for the same reason.

lib/tasks/eu_exit_business_finder.rake
ChrisBAshton: I agree, this was my first instinct & I copied EmailAlertListTitleBuilder to try to use that, but found I was doing so much data manipulation to get the content item JSON in the structure EmailAlertListTitleBuilder that it was a better decision to roll out a pared down version myself.

The code that generates the email title from finder-frontend conflates three things: all the possible facet values, and the user's facet value choices, and the groupings of the facet values inside different facets. So we'd have a structure like:

```json
filter: {
  "sector_business_area" => [
    "content-id-of-a-facet-the-user-selected",
    "content-id-of-another-facet-the-user-selected"
  ],
  "buying_selling_in_the_eu" => [
    "yet-another",
  ] // etc
}
```

The facet groupings were required for the email alert frontend but what it all comes down to in email subscriber list is keys for the chosen facet-values:

```
[
  "content-id-of-a-facet-the-user-selected",
  "content-id-of-another-facet-the-user-selected"
  "yet-another",
]
```

So the solution for the rake task is simplest when we just take that array of content IDs and find the corresponding `title`s in the content item, rather than try and map it to its facet group (which has no bearing on the outcome here) just to shoehorn it into a local copy of EmailAlertListTitleBuilder.

lib/tasks/eu_exit_business_finder.rake
ChrisBAshton: Good shout. 👍 Shall I use [publishing_api_v2](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api_v2.rb)?

brucebolt: Yes, that should have a suitable method to get content

ChrisBAshton: Any reason why publishing API and not content API? I thought the former was for writing and the latter was for retrieving...?

brucebolt: Content API is a service that's available to the public for accessing data.  As far as I'm aware, it's not actually used by any of our internal services.

ChrisBAshton: Our docs seem to suggest they do (if front-end apps count as internal services):

> This API is used by front-end apps but is also exposed externally at /api/content/<path>, such as https://www.gov.uk/api/content/take-pet-abroad

https://docs.publishing.service.gov.uk/apis/content-store.html#reading-and-writing-to-the-store

Also could you explain "It would prevent any issues with egress rules on machines in production"? I don't know what that means :)

brucebolt: >Also could you explain "It would prevent any issues with egress rules on machines in production"? I don't know what that means :)
I believe some machines in AWS have rules about what ports and IP addresses access into (ingress) and access out of (egress) machines is allowed, for security reasons.  It may be that the production machines can't make a HTTP request to an external site (I'm not certain).

lib/tasks/eu_exit_business_finder.rake
ChrisBAshton: I've now added this explanation into the commit message (and fixed up the linting issues).

Still hoping to refactor the `title_from_facets` method before it's truly 'review ready'.

lib/tasks/eu_exit_business_finder.rake
ChrisBAshton: I've just learned two words today - thanks for that!

Not exactly sure what I'm doing here - replacing the line with: 

```ruby
json_response = JSON.parse(GdsApi::ContentStore.new(Plek.new.find("content-store")).content_item('/find-eu-exit-guidance-business'))
```

...makes a call to `http://content-store.dev.gov.uk/content/find-eu-exit-guidance-business`. Is that what we want? I'm worried I won't catch any issues encountered if I'm just stubbing the network request to return what I want 🤔

brucebolt: That would work.  If it were me, I'd use the publishing API, since you wouldn't then need to parse the JSON:
```
GdsApi::PublishingApiV2.new(Plek.new.find('publishing-api')).get_content('content_id_of_finder')
```
But I guess it doesn't really matter too much either way.

Even better, I'd probably define a `publishing_api` method, then just use that (it could even be defined in `./app/lib/services.rb`, so you can then access it in the tests using `Services.publishing_api`):
```
def publishing_api
   GdsApi::PublishingApiV2.new(Plek.new.find('publishing-api'))
end

....

publishing_api.get_content('content_id_of_finder')
```

I'd strongly suggest stubbing the request (hence the need for the fixture file) otherwise the tests will start failing if the live content item is changed elsewhere.  The test should be checking the method performs the desired functionality, which it can do with mock data and doesn't necessarily need real data.  This gives the advantage that the tests will still pass if you were to change the values in `facet_values_to_replace` at some point in the future but leave the code unchanged (my belief is this rake task should stick around until the finder is removed, as I'm sure the sectors will get changed again at some point 🤔).

ChrisBAshton: Oh I'm already stubbing, it's just I'm stubbing the https://www.gov.uk/api/content/find-eu-exit-guidance-business endpoint (which I can manually check and make sure it will return the right thing in production) rather than stubbing `http://content-store.dev.gov.uk/content/find-eu-exit-guidance-business` (which I can't manually check, so stubbing locally is fine but I have less confidence it will work on production).

Good tip about using publishing API to avoid having to parse the JSON. Also interesting you say the code will stick around, as I'd imagined us deleting it shortly after use, but agree it makes sense to keep for as long as the finder exists.

ChrisBAshton: Have now pushed up an update that should resolve this conversation 👍

spec/lib/tasks/fixtures/find-eu-exit-guidance-business.json
leenagupte: You can paste the json into https://jsonlint.com/ and it will prettify it for you.

I agree with cutting down the fixture to only what's needed. It will make the fixture, and therefore the tests easier to read.

lib/services.rb
```diff
@@ -3,6 +3,10 @@ def self.content_store
     @content_store ||= GdsApi::ContentStore.new(Plek.new.find("content-store"))
   end
 
+  def self.publishing_api
```
leenagupte: This is great, but can you fix up these changes to the first commit.

spec/lib/tasks/fixtures/find-eu-exit-guidance-business.json
brucebolt: I'd also recommend using mock data (e.g. replacing actual content_id values with things like `a-facet-value`) in both the fixture and specs.  It then removes any misunderstanding in the future about whether the tests have to be updated when the rake tasks are changed to cater for different sector merges.

spec/lib/tasks/eu_exit_business_finder_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+  def mock_facet_values_to_replace(replacements)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_values_to_replace).and_return(replacements)
+  end
+
+  def mock_facet_value_labels_to_replace(replacements)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_value_labels_to_replace).and_return(replacements)
+  end
+
+  def stub_find_eu_exit_guidance_business_json
+    abs_path_to_fixture_data = File.join(File.dirname(__FILE__), './fixtures/find-eu-exit-guidance-business.json')
```
brucebolt: Would it be possible to use `Rails.root` rather than `File.dirname(__FILE__)`?  I think it looks slightly nicer.

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,78 @@
...
+
+  # Iterate through the affected subscriber lists and if we encounter any
+  # facet values that need replacing with a newer one, we insert the new one
+  # to the array (inserted 'in place' to retain original order). Then remove
+  # the old facet values in bulk and write only the unique values that remain.
+  def update_facet_values(subscriber_lists)
+    subscriber_lists.each do |list|
+      links = list.links[:facet_values][:any]
+      updated_links = links.dup # copy the links to avoid corrupting the loop
+      links.each_with_index do |value, index|
```
brucebolt: I wonder if `map` could be utilised to remove the need to create `updated_links`, remove the `each` loop and remove the `insert`?  For example (I haven't tested it, but something like this maybe):
```
links.map { |link|
  EuExitFacetMigrationConfig::facet_values_to_replace.find { |replace_val| replace_val[:old_facet_value] == value }.fetch(:new_facet_value, nil) || link
}
```

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def self.facet_values_to_remove
+    [
+      "7536c0c4-fb41-43f4-a2c4-08f4fa9f5427",
+      "5faa1741-fc55-4110-b342-de92f6324118",
+      "14cf2a68-3297-44d3-ba01-a4426845b1b8",
+      "040649fc-4e2c-4028-b846-77fe3eebd1f7",
+      "94b3cfe2-af89-4744-b8d7-7fc79edcbc85"
+    ] + facet_values_to_replace.map { |f| f[:old_facet_value] }.uniq
```
brucebolt: If you use the `map` method that I've suggested on line 17, you probably won't need to add the `old_facet_value` values here, since they'll get replaced when doing the `map`.

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -21,9 +21,40 @@ namespace :eu_exit_business_finder do
...
+      chosen_facet_ids.include? content_id
+    })
+    chosen_facet_titles = chosen_facet_ids_and_titles.map { |_content_id, title| title }
+    comma_separated_facet_titles = chosen_facet_titles.map { |choice_name| "'#{choice_name}'" }.join(", ")
+    "EU Exit guidance for your business in the following #{'category'.pluralize(chosen_facet_titles.count)}: #{comma_separated_facet_titles}"
+  end
+
+  def flatten_facet_values_to_h(facets)
+    array_of_facet_value_hashes = facets.map do |facet|
+      facet['links']['facet_values'].map { |facet_value| [facet_value['content_id'], facet_value['title']] }.to_h
```
brucebolt: You could override the special values here, which means the `override_special_titles` method could be totally removed.
```
facet['links']['facet_values'].map { |facet_value| [facet_value['content_id'], facet_value_labels_to_replace.fetch(facet_value['content_id'], facet_value['title'])] }.to_h
```

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -21,9 +21,40 @@ namespace :eu_exit_business_finder do
...
     end
   end
+
+  # Title is constructed from finder content item, so we have to retrieve the
+  # facet titles from the JSON using their content IDs.
+  def title_from_facets(chosen_facet_ids)
+    business_finder_json = Services.publishing_api.get_content('42ce66de-04f3-4192-bf31-8394538e0734')
+    all_facets = business_finder_json['links']['facet_group'][0]['links']['facets']
+    all_facet_ids_and_titles = flatten_facet_values_to_h(all_facets)
+    chosen_facet_ids_and_titles = override_special_titles(all_facet_ids_and_titles.filter { |content_id, _title|
```
brucebolt: If you take my suggestion of doing the overrides in `flatten_facet_values_to_h` rather than having a separate method, a lot of this code could be removed.

`chosen_facet_ids_and_titles` wouldn't have to be created (i.e. remove lines 35 to 37).  You could go straight to `chosen_facet_titles`, e.g.
```
chosen_facet_titles = chosen_facet_ids.map { |chosen_facet_id| all_facet_ids_and_titles[chosen_facet_id] }
```

ChrisBAshton:
Thanks @brucebolt - I had something similar to what you suggested then refactored it (https://github.com/alphagov/email-alert-api/compare/3c237bc10bdbef2c9201fca23f3a076c5e548c1b..550560cef8b00b52e64298fc6b553ee71ee851ec) as it felt like there was too much going on. I'm happy to revisit though - still figuring out the balance of readability vs brevity!

lib/services.rb
thomasleese: Instead of defining `publishing_api` here ourselves, we can use the built in [`GdsApi.publishing_api_v2` instance provided by gds-api-adapters](https://github.com/alphagov/gds-api-adapters/blob/d30488d787c98c08702b047f99ce62d42b8a3b1f/lib/gds_api.rb#L142-L153).

It's something [other apps have been moving towards](https://github.com/alphagov/travel-advice-publisher/pull/602).

spec/lib/tasks/eu_exit_business_finder_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+    allow(EuExitFacetMigrationConfig).to receive(:facet_values_to_replace).and_return(replacements)
+  end
+
+  def mock_facet_value_labels_to_replace(replacements)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_value_labels_to_replace).and_return(replacements)
+  end
+
+  def stub_find_eu_exit_guidance_business_json
+    abs_path_to_fixture_data = Rails.root.join('spec/lib/tasks/fixtures/find-eu-exit-guidance-business.json')
+    stub_request(:get, "#{Plek.find('publishing-api')}/v2/content/42ce66de-04f3-4192-bf31-8394538e0734").
```
thomasleese: Here you could use [the test helpers provided by gds-api-adapters](https://github.com/alphagov/gds-api-adapters/blob/d30488d787c98c08702b047f99ce62d42b8a3b1f/lib/gds_api/test_helpers/publishing_api_v2.rb#L369-L378).

I imagine it would look something like:

```suggestion
    stub_publishing_api_has_item(JSON.parse(abs_path_to_fixture_data))
```

spec/lib/tasks/fixtures/find-eu-exit-guidance-business.json
```diff
@@ -0,0 +1,74 @@
+{
+	"base_path": "/find-eu-exit-guidance-business",
```
brucebolt: The indentation here is using tabs, could they be replaced with two spaces to maintain consistency with other code

ChrisBAshton: Good spot - I just took the default output from jsonlint.com

lib/services.rb
ChrisBAshton: Thanks, will do!

spec/lib/tasks/eu_exit_business_finder_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+  def mock_facet_values_to_replace(replacements)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_values_to_replace).and_return(replacements)
+  end
+
+  def mock_facet_value_labels_to_replace(replacements)
+    allow(EuExitFacetMigrationConfig).to receive(:facet_value_labels_to_replace).and_return(replacements)
+  end
+
+  def stub_find_eu_exit_guidance_business_json
+    abs_path_to_fixture_data = Rails.root.join('spec/lib/tasks/fixtures/find-eu-exit-guidance-business.json')
```
leenagupte: For info `Rails.root.join` is pretty cool it that it lets you get around operating system defaults by letting you only specify the directory names. So here, you could have used:

```
Rails.root.join('spec', 'lib', 'tasks', 'fixtures', 'find-eu-exit-guidance-business.json')
```

ChrisBAshton: Ah, so that it comes out as `spec\lib\tasks\...etc` on Windows? Not as readable but I see the benefit 👍 

We seem to use it lots of different ways in email-alert-api (and doubtless elsewhere), probably worth changing it to a consistent approach in a separate PR someday.

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,108 @@
...
+
+  def replace_facet_val(original_facet_val)
+    EuExitFacetMigrationConfig::facet_values_to_replace
+      .select { |replace_val| replace_val[:old_facet_value] == original_facet_val }
+      .map { |replace_val| replace_val[:new_facet_value] }[0]
+  end
+
+  # Title is constructed from finder content item, so we have to retrieve the
+  # facet titles from the JSON using their content IDs.
+  def title_from_facets(links)
```
leenagupte: I assume that this method is being passed a list of the content-ids of facet-values, and that the list is in correct order? i.e. sectors, then org activity etc.

If that's the case, then this method should be very simple. 

You what the subscription title prefix is, so it should just be a case of looping through the links array, getting the title of the content item of the facet value from publishing-api, and concatenating that to the "subscription title prefix"

Also, I think "links" should be renamed to "facet_values" so we're clear what they are.

e.g.
```
subscription_title_prefix = "EU Exit guidance for your business in the following "
# category or categories code goes here

titles = facet_values.map do |facet_value|
  if EuExitFacetMigrationConfig.facet_value_labels_to_replace[facet_value].present?
    return "'#{EuExitFacetMigrationConfig.facet_value_labels_to_replace[facet_value]}'"
 end
  content_item = GdsApi.publishing_api_v2.get_content(facet_value).to_h
  "'#{content_item['title']}'"
end
```
subscription_title_prefix + titles.join(', ')

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,108 @@
...
+    [
+      "7536c0c4-fb41-43f4-a2c4-08f4fa9f5427",
+      "5faa1741-fc55-4110-b342-de92f6324118",
+      "14cf2a68-3297-44d3-ba01-a4426845b1b8",
+      "040649fc-4e2c-4028-b846-77fe3eebd1f7",
+      "94b3cfe2-af89-4744-b8d7-7fc79edcbc85"
+    ]
+  end
+
+  def self.facet_value_labels_to_replace
```
leenagupte: I think this method title is unclear. I think a better name would be `facet_value_label_overrides`

lib/tasks/eu_exit_business_finder.rake
ChrisBAshton: Definitely more readable 👍think making one request to Publishing API (for business finder content item) is more efficient than lots of little ones, but not an issue at this scale.

Have refactored now.

brucebolt:
Whilst we're migrating facet values, would it be possible to add an additional migration from `7620da7a-0427-4b3c-9498-db9dc25209b` to `7620da7a-0427-4b3c-9498-db9dc25209b0` (since we were subscribing users to the incorrect facet value: https://github.com/alphagov/search-api/commit/8fce3db218821dfc7a2b50dbd1b03984b1ec41b1)?  This will save doing it separately.

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,102 @@
...
+      links = list.links[:facet_values][:any]
+      updated_links = links.map { |original_val| replace_facet_val(original_val) || original_val }
+      updated_links = (updated_links - EuExitFacetMigrationConfig::facet_values_to_remove).uniq
+      list.write_attribute('links', facet_values: { any: updated_links })
+      list.write_attribute('title', title_from_facets(updated_links))
+      list.save
+    end
+  end
+
+  def replace_facet_val(original_facet_val)
```
brucebolt: Could you rename `val` to `value` in `replace_facet_val` and `replace_val`, just so it's clear what this value is?

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,108 @@
...
+      # "94b3cfe2-af89-4744-b8d7-7fc79edcbc85" is a special case
+      # as it's been split into two facets. We should assume the
+      # user wants to remain subscribed to both.
+      {
+        "old_facet_value": "94b3cfe2-af89-4744-b8d7-7fc79edcbc85",
+        "new_facet_value": "9d54c591-f5ca-4d0c-a484-12d5591987cb"
+      },
+      {
+        "old_facet_value": "94b3cfe2-af89-4744-b8d7-7fc79edcbc85",
+        "new_facet_value": "afd45eef-743a-417d-9245-3eab8322116d"
```
leenagupte: I think this might raise an issue, as by the time this hash is processed, "94b3cfe2-af89-4744-b8d7-7fc79edcbc85" will have already been overwritten with 9d54c591-f5ca-4d0c-a484-12d5591987cb".

I think the code needs to be able to handle an array of "new_facet_value", and do the replace at the same time.

ChrisBAshton: Whoops - good spot!!

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,108 @@
...
+      list.write_attribute('links', facet_values: { any: updated_links })
+      list.write_attribute('title', title_from_facets(updated_links))
+      list.save
+    end
+  end
+
+  def replace_facet_value(original_facet_value)
+    EuExitFacetMigrationConfig::facet_values_to_replace
+      .select { |replace_value| replace_value[:old_facet_value] == original_facet_value }
+      .map { |replace_value| replace_value[:new_facet_value] }[0]
```
leenagupte: Is there an easier way to do this?

Would this work:

```
EuExitFacetMigrationConfig::facet_values_to_replace.map |facet_value_to_replace| do
   if facet_value_to_replace[:old_facet_value] != original_facet_value
       return facet_value_to_replace[:new_facet_value]
   end
   original_facet_value
end
```

lib/tasks/eu_exit_business_finder.rake
ChrisBAshton: Fixed. Added a test.

lib/tasks/eu_exit_business_finder.rake
brucebolt: If you make `facet_values_to_replace` into a hash (rather than an array of hashes) with the key being the `old_facet_value` and the value being the `new_facet_value` then this method won't be needed.

lib/tasks/eu_exit_business_finder.rake
```diff
@@ -0,0 +1,111 @@
...
+      {
+        "old_facet_value": "c1d8057c-76bf-431c-9ac8-6281a4b7b9ca",
+        "new_facet_value": "34a6edd0-46ea-4a76-ae80-8c96709d4f59"
+      },
+      # "94b3cfe2-af89-4744-b8d7-7fc79edcbc85" is a special case
+      # as it's been split into two facets. We should assume the
+      # user wants to remain subscribed to both.
+      {
+        "old_facet_value": "94b3cfe2-af89-4744-b8d7-7fc79edcbc85",
+        "new_facet_values": [
```
brucebolt: If you make all `new_facet_value` values into an array (even with just one item) then on line 31, you won't need the `||`.

ChrisBAshton: Good point 👍

spec/lib/tasks/eu_exit_business_finder_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+      stub_publishing_api_has_item(
+        "content_id": "444-content-id-of-eu-funding",
+        "title": "EU funding",
+      )
+      mock_behaviour!
+      create_list_and_execute_rake!
+      SubscriberList.last.title
+    }
+
+    context "referencing a facet value in the finder data" do
```
leenagupte: I think my previous comment about the test names was lost. "finder data" doesn't make sense in this context.

How about: "build title from single facet value" ?

spec/lib/tasks/eu_exit_business_finder_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+      let(:facet_values) { ["000-content-id-of-food-drink-tobacco-facet"] }
+
+      it {
+        is_expected.to eql "EU Exit guidance for your business in the following category: 'Food, drink and tobacco (retail and wholesale)'"
+      }
+    end
+
+    context "referencing multiple facet values in the finder data" do
```
leenagupte: The same point about the test name applies here too.

How about: "build title from multiple facet values" ?

spec/lib/tasks/eu_exit_business_finder_spec.rb
ChrisBAshton: It doesn't make sense now we're querying publishing API for individual facet values, agreed 👍 will change.

PR #895, opened 2019-06-26
Refactor email templates and implement text improvements

AgaDufrat:
**What**
Improve Email Builders to centralise email template terminology using i18n as well as avoid duplication and make it easier to change (more details in the commit message).

Update our email templates with the changes suggested by our content designers:
- Add headers for page description, change note and time updated at
- Change email subject 
- Add opening line to the body
- Reorder parts of the footer (to avoid inconsistent placement of manage and unsubscribe links)
- Remove unsubscribe link from immediate emails
- Update text for manage your subscriptions link
- Update permission reminder text and formatting
- Remove ‘do not reply’ and update feedback part formatting
- Remove line between content changes and footer

**Why**
These changes will make email-alert-api easier to maintain and reduce the amount of developer effort needed when future changes are required. It should also stop email for different frequencies getting out of sync with each other.

Our user research showed some points of confusion in the understanding of content within email.
These pattern-wide fixes we can make to improve email alerts across GOV.UK.

**Immediate update template before**
before-immediate

**NEW immediate update template**
new-immediate

---

**Digest update template before**
before-daily

**NEW digest update template**
new-weekly

[Trello card: Refactor templates in email-alert-api](https://trello.com/c/XfyyP9ry/159-refactor-templates-in-email-alert-api)
[Trello card: Make text improvements to GOV.UK email templates](https://trello.com/c/rVMddkis/164-make-text-improvements-to-govuk-email-templates)

### DIFF
--- a/app/builders/digest_email_builder.rb
+++ b/app/builders/digest_email_builder.rb
@@ -1,4 +1,5 @@
 class DigestEmailBuilder
+  include EmailBuilderHelper
   def initialize(subscriber:, digest_run:, subscription_content_change_results:)
     @subscriber = subscriber
     @digest_run = digest_run
@@ -25,8 +26,9 @@ private
   attr_reader :subscriber, :digest_run, :results
 
   def body
-    presented_results.concat("\n")
-      .concat(subscription_link).concat("\n")
+    presented_results.concat("\n\n&nbsp;\n\n---\n\n")
+      .concat(permission_reminder).concat("\n")
+      .concat(presented_manage_subscriptions_links).concat("\n\n&nbsp;\n\n")
       .concat(feedback_link)
   end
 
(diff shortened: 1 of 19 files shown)


app/builders/digest_email_builder.rb
```diff
@@ -1,4 +1,7 @@
+require 'email_builder_helper'
```
brucebolt: I see in the next PR you've moved EmailBuilderHelper so the require isn't needed here.  Would it be possible to have that in this commit too?  Then it reduces the diff between the commits.

AgaDufrat:
I had a chat with Leena (content designer) and amended commit 7e8293ab8566a58adf7c1207287ceaff27279191 to only remove the line and space before the footer from digest email. There is a new commit to add a line after opening line for immediate emails.

PR #909, opened 2019-07-23
Clean email lists

bilbof:
This adds `Clean::InvalidSubscriberLists` that will be called via the rake task `clean:migrate_subscriptions_to_valid_lists`.

This will enable us to find and fix situations where a user is subscribed to an email list which has invalid tags.

By default the tasks will be run in `dry_run` mode, so nothing will be modified unless you use `DRY_RUN=no.`

If it's not a dry run, then `clean:migrate_subscriptions_to_valid_lists` will find all the invalid subscriber lists and create a new valid subscriber list. It'll then create new subscriptions for the users on the invalid list.

Later on we can clean up the old data (invalid subscriber lists and their subscriptions) and add some validations against these invalid tags.

Trello: https://trello.com/c/rTXOLY1z/878.

### DIFF
--- a/lib/clean/invalid_subscriber_lists.rb
+++ b/lib/clean/invalid_subscriber_lists.rb
@@ -77,6 +77,20 @@ module Clean
       puts "#{dry} #{moved} of #{total_subscribers} subscribers from list #{from_list.slug} to #{to_list.slug}"
     end
 
+    def deactivate_invalid_subscriptions(dry_run: true)
+      dry_msg = dry_run ? "[DRY RUN] Would have deactivated" : "Deactivated"
+      lists.each do |list|
+        SubscriberList.transaction do
+          subscriptions = list.subscriptions.active
+          count = subscriptions.count
+          subscriptions.each do |subscription|
+            subscription.end(reason: :subscriber_list_changed) unless dry_run
+          end
+          puts "#{dry_msg} #{count} subscription(s) on subscriber list #{list.slug}"
+        end
+      end
+    end
+
   private
 
     def tags_to_links(tags)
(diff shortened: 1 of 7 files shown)


lib/services.rb
```diff
@@ -6,4 +6,8 @@ def self.content_store
   def self.rate_limiter
     @rate_limiter ||= Ratelimit.new("deliveries")
   end
+
+  def self.worldwide_api
```
thomasleese: Instead of needing to define this here you can use `GdsApi.worldwide` wherever you need to use the API:

https://github.com/alphagov/gds-api-adapters/blob/52d5d97e9b5822462bb533944666b02a1596bed1/lib/gds_api.rb#L206-L208

bilbof: Thanks @thomasleese, changed.

lib/clean/invalid_subscriber_lists.rb
```diff
@@ -0,0 +1,146 @@
...
+          # Check if they have already subscribed
+          subscribed_to_destination_subscriber_list = Subscription.find_by(
+            subscriber: subscriber,
+            subscriber_list: to_list
+          )
+
+          # Subscribed them if they haven't already been subscribed
+          if subscribed_to_destination_subscriber_list.nil?
+            unless dry_run
+              Subscription.create!(
```
hannako: Just a question about using the `!` ... Is it impossible for the validations to fail? Because there's no rescue if it does.

bilbof: The validations could fail, but that is OK. I suspect they won't, as the data is already in the DB, the task is just putting it back in a new list. I would prefer the task to fail if validations do fail, and we can always safely restart the task if that happens.

PR #910, opened 2019-07-23
Quicken SubscriptionContent healthcheck

benthorner:
https://trello.com/c/P9fomge6/646-%E2%9C%89%EF%B8%8F-email-alert-api-app-healthcheck-not-ok

This should reduce the multi-second timings we currently have warnings
for in production...

Previously running this healthcheck involved querying two large tables
twice: SubscriptionContent and Subscription. Since we expect the steps
taken to address the alert to be identical for both a warning and a
critical, this reduces the runtime of the check by just having one
(critical) severity level, but using the parameters for the (deleted)
warning severity level so that the alert will still be timely.

### DIFF
--- a/app/models/healthcheck/subscription_contents.rb
+++ b/app/models/healthcheck/subscription_contents.rb
@@ -7,8 +7,6 @@ module Healthcheck
     def status
       if critical_subscription_contents.positive?
         :critical
-      elsif warning_subscription_contents.positive?
-        :warning
       else
         :ok
       end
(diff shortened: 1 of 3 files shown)


benthorner:
It takes 5-6 seconds when it's going slow. We're not entirely sure why it's going slow. This seemed like the most pragmatic thing to do that would speed it up after thinking about what it's trying to check; what sorts of optimisation did you have in mind?

kevindew:
> It takes 5-6 seconds when it's going slow. We're not entirely sure why it's going slow. This seemed like the most pragmatic thing to do that would speed it up after thinking about what it's trying to check; what sorts of optimisation did you have in mind?

Yikes - that's a bit miserable. We could always have a compound index with email and created_at so that condition is entirely an index scan.

The thing seems to smell the most about the query though is the join, I wonder if we should just delete SubscriptionContent entries that'll never go out when we go to run the immediate_email_generator and then that part could be removed. I guess it depends on the query planner as to the impact that could have.

benthorner:
After a bit more though, I'm going to close this PR. The real problem here is that the check is sometimes fast and sometimes slow; we should figure out why this is before relieving the symptoms, as there may be a bigger problem underneath.

PR #912, opened 2019-07-24
Add rake task for removing sub lists with no subs

SamJamCul:
https://trello.com/c/TcWzthQL/880-clean-subscriber-lists-created-by-finder-frontend

### DIFF
--- a/lib/tasks/clean.rake
+++ b/lib/tasks/clean.rake
@@ -37,6 +37,13 @@ namespace :clean do
     cleaner.deactivate_invalid_subscriptions(dry_run: dry_run)
   end
 
+  desc "Remove SubscriberLists with no subscriptions"
+  task remove_empty_subscriberlists: :environment do
+    dry_run = is_dry_run?
+    cleaner = Clean::EmptySubscriberLists.new
+    cleaner.remove_empty_subscriberlists(dry_run: dry_run)
+  end
+
   def is_dry_run?
     dry = ENV["DRY_RUN"] != 'no'
     puts "Warning: Running in DRY_RUN mode. Use DRY_RUN=no to run live." if dry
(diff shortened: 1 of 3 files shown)


lib/tasks/clean.rake
```diff
@@ -37,6 +37,19 @@ namespace :clean do
...
   end
 
+  desc "Remove SubscriberLists with no subscriptions"
+  task remove_empty_subscriberlists: :environment do
+    dry_run = is_dry_run?
+    count = 0
+    SubscriberList.find_each do |l|
+      if l.subscriptions.count.zero?
+        count += 1
+        l.destroy unless dry_run
```
bilbof: I wonder if this could be added as a class in the `Clean` module, so we could add tests for it?

cbaines:
What's the motivation for this change? The commit message is a bit lacking.

Also, is there any data related to the subscriber lists in the Email Alert API database that's going to block deleting subscriber lists, or be deleted along with them?

bilbof:
What happens to the subscriptions that reference the deleted subscriber lists?

barrucadu:
@cbaines Creating the `SubscriberList` happens before the user actually enters their email address.  We also don't do any validation on facet values (there's a separate card for that), so as a result we've got a bunch of `SubscriberList`s for organisation with names like `"+response.write(9205521*9921377)+"` which don't have any subscribers.  So it's just to clean those up.

@bilbof It's only deleting `SubscriberList`s with no `Subscription`s, so I don't think there will be any?

bilbof:
@barrucadu 🤦‍♂ that's fine then 😄

spec/lib/clean/empty_subscriber_lists_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  context "when there are lists with subscribers" do
+    let!(:list) { create(:subscriber_list_with_subscribers) }
+
+    describe "#lists" do
+      it "returns zero subscriber lists" do
+        expect(subject.lists.count).to eq(0)
+        expect(subject.lists).to_not include(list)
+      end
+    end
```
bilbof: Thanks for adding specs. Looks great!

Could you add a test for `remove_empty_subscriberlists`?

barrucadu: Done

PR #925, opened 2019-08-12
Subscriber list validations

bilbof:
This adds a validator for tags, to ensure that only permitted tags can be used in subscriber lists.

The tags I've added to the list are those already in use, or that I found through an audit of the tags that could be used but aren't in use yet.

Trello: https://trello.com/c/3JqVLHZX/393

### DIFF
--- a/doc/fields.md
+++ b/doc/fields.md
@@ -1,0 +1,14 @@
+Email Alert API Field Reference
+===============================
+
+## Subscriber Lists
+
+Subscriber Lists in Email Alert API can have tags or links. When a document is
+is changed, the API does a search for the document's attributes in the API to
+find a matching subscriber list.
+
+### Tags
+
+The API has a whitelist of available tags that a user can subscribe to.
+
+You can find a list of permitted tags in [`lib/valid_tags.rb`](https://github.com/alphagov/email-alert-api/blob/3e0018510ea85f5d561e2865ad149832b94688a1/lib/valid_tags.rb).
--- a/lib/valid_tags.rb
+++ b/lib/valid_tags.rb
@@ -1,7 +1,79 @@
 class ValidTags
-  ALLOWED_TAGS = %i(vessel_type format case_type sector_business_area business_activity
-                    employ_eu_citizens personal_data intellectual_property public_sector_procurement
-                    location eu_uk_government_funding alert_type tribunal_decision_category policies
-                    topics content_store_document_type content_purpose_supergroup railway_type
-                    aircraft_category content_purpose_subgroup appear_in_find_eu_exit_guidance_business_finder).freeze
+  ALLOWED_TAGS = %i(
+    aircraft_category
+    aircraft_type
+    alert_type
+    appear_in_find_eu_exit_guidance_business_finder
+    assessment_date
+    business_activity
+    business_sizes
+    business_stages
+    case_state
+    case_type
+    certificate_status
+    closed_date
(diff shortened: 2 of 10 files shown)


bilbof:
Still in progress - I've found cases where tags such as `tribunal_decision_judges` should be permitted, but haven't yet been used so aren't in the allowed list.

As part of this task, I need to get the exhaustive list of tags one can use.

bilbof:
I audited the applications that can send tags, but it looks like there isn't a central place where all of the permitted tags are defined.

I wonder if this would be a good thing to add to the content schemas?

The tags in the list are those that I could find + those already used in the database.

doc/fields.md
```diff
@@ -0,0 +1,90 @@
...
+
+Subscriber Lists in Email Alert API can have tags or links. When a document is
+is changed, the API does a search for the document's attributes in the API to
+find a matching subscriber list.
+
+### Tags
+
+The API has a whitelist of available tags that a user can subscribe to.
+
+The following tags are permitted:
```
thomasleese: I wonder if it's better to link to the code here because I suspect it'll be easy for this list to go out of the date with the real list.

lib/valid_tags.rb
```diff
@@ -0,0 +1,79 @@
+class ValidTags
+  ALLOWED_TAGS = %i(
```
thomasleese: Do we need to update https://docs.publishing.service.gov.uk/apis/search/add-new-fields-or-document-types.html as well to mention updating this list?

doc/fields.md
bilbof: Added that, thanks. I like the approach Search API has (generating https://docs.publishing.service.gov.uk/apis/search/fields.html from the current list), so might be nice to do the same.

lib/valid_tags.rb
bilbof: It sounds like that doc concerns search fields, though I think there's a difference between fields that one can search with and fields accepted by email alert api.

I couldn't find a doc on adding email alerts for a finder, so it might be worth adding one here: https://github.com/alphagov/finder-frontend/tree/master/docs, and then I can link to this doc from there. What do you think?

thomasleese: Yeah, I reckon that's probably fine. I'm imaging a situation where someone adds a new field that should also be supported by email-alert-api and they have no idea about the whitelist and get confused wondering why it's not working. So as long as it's written down somewhere.

lib/valid_tags.rb
```diff
@@ -0,0 +1,79 @@
+class ValidTags
```
koetsier: Seeing how everyone _loves_ YAML, would it be useful to have this as a configuration file? It probably won't add much, but would allow people to configure the app without changing code.

koetsier:
What about links? Do we need validation for those as well?

Small possible suggestion but apart from that 👍

lib/valid_tags.rb
bilbof: I added documentation for this step to finder-frontend: https://github.com/alphagov/finder-frontend/commit/7ac871987abf36184b136b3b3513ebaa02d8e8a6#diff-b94f09a9b9baf14806b87c9db6ae7d0eR93

lib/valid_tags.rb
bilbof: Yeah, if having this in Yaml becomes a requirement we can add that later.

PR #927, opened 2019-08-19
Spike into allowing ad-hoc messages to subscriber lists

kevindew:
Trello: https://trello.com/c/b4rYHqUc/23-spike-into-working-out-if-email-alert-api-can-handle-bespoke-lists-and-messages

This adds functionality which allows sending a message to the members of one the GOV.UK subscription lists. It does this by providing an endpoint `/messages` which allows creating a message (fields are title, path and body) which can then be distributed to everyone matching the criteria for that message. 

This functionality is intended to be used as a bespoke means for communicating with people on mailing lists where we may want to provide more appropriate communication for a particular scenario or to alert them to things that outside normal content changes (this functionality would be an appropriate replacement for the [taxon unpublishing notifications](https://github.com/alphagov/email-alert-api/blob/f78035d1519328fc9bb47a8821eb81207bb64cf2/app/services/unpublish_handler_service.rb) and the [business readiness fake content-changes](https://github.com/alphagov/content-tagger/blob/a5f4273252639328e1b89d05f313860f1f1dd144/app/services/facets/tagging_update_publisher.rb#L27-L33)).

This spike was intended to work out if this sort of functionality was possible but in doing so I ended up doing a lot of the work towards it.

The follow up activities I expect, assuming we decide we want to proceed with this functionality is:

- [x] break out refactorings and renamings from this so they can be merged and deployed first
- [ ] test this for reals
- [x] add support for digests
- [x] determine the fields we actually need for a message
- [ ] update documentation
- [ ] update alerting


thomasleese:
All parts of this spike have now been merged in.

PR #928, opened 2019-08-20
Document rake task & make more robust

ChrisBAshton:
Documents the new subscriptions count rake task so that it can be searched for in the developer docs.

Also makes it more robust by addressing post-merge PR comments made in #887.

### DIFF
--- a/lib/tasks/count_subscribers.rake
+++ b/lib/tasks/count_subscribers.rake
@@ -2,15 +2,14 @@ namespace :query do
   desc "Query how many active subscribers there are to the given subscription slug"
   task :count_subscribers, %i[subscription_list_slug] => :environment do |_t, args|
     slug = args[:subscription_list_slug]
-    subscription_lists = SubscriberList.where(slug: slug)
-    raise "There is no SubscriberList called '#{slug}'" unless subscription_lists.count == 1
+    subscription_list = SubscriberList.find_by!(slug: slug)
+    active_subscriptions = subscription_list.subscriptions.active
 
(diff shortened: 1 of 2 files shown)


lib/tasks/count_subscribers.rake
```diff
@@ -2,15 +2,14 @@ namespace :query do
...
+    subscription_list = SubscriberList.find_by!(slug: slug)
+    active_subscriptions = subscription_list.subscriptions.active
 
-    active_subscriptions = subscription_lists.first.subscriptions.where(ended_at: nil)
     puts """
       The SubscriberList '#{slug}' has #{active_subscriptions.count} active subscriptions, of which:
-        - #{active_subscriptions.where(frequency: 0).count} are signed up for Immediate updates
-        - #{active_subscriptions.where(frequency: 1).count} are signed up for Daily updates
-        - #{active_subscriptions.where(frequency: 2).count} are signed up for Weekly updates
+        - #{active_subscriptions.where(frequency: Frequency::IMMEDIATELY).count} are signed up for Immediate updates
```
thomasleese: I'm not sure this will work, because `Frequency::IMMEDIATELY` will resolve to the string `"immediately"`, whereas [we want the enum value](https://github.com/alphagov/email-alert-api/blob/d57091a5d33e57d6b69a00a9164dd35279936b8a/app/models/subscription.rb#L7). I think you can just use `:immediately`.

ChrisBAshton:
Have manually tested this branch in Integration and the rake task still works 👍

PR #933, opened 2019-08-22
Create models related to messages

thomasleese:
This PR adds the various models necessary to implement the message feature.

The Message model is intended for ad-hoc messages to subscriber lists to send them a particular message. This could be used as a standard mailing list message or as a means to notify them to a change to a particular list.

The MatchedMessage model serves the same purpose as the MatchedContentChange model but for messages which acts as a many to many join table.

This code was cherry picked from #927 to make it easier to merge and test individually.

[Trello Card](https://trello.com/c/5rtAmAjc/54-finish-the-messages-endpoint-work-in-email-alert-api)


thomasleese:
One difference from https://github.com/alphagov/email-alert-api/pull/927 is that I've renamed `path` to `url`.

app/models/message.rb
```diff
@@ -0,0 +1,19 @@
+class Message < ApplicationRecord
+  include SymbolizeJSON
+
+  has_many :matched_messages
+  has_many :subscription_contents
+
+  validates_presence_of :title, :body
+  validates :url, absolute_path: true, allow_nil: true
```
kevindew: I wonder if we need to rename this absolute_path validator too? From a very non-authoritive looking website I've found the name seems to be root-relative, so we could maybe call the validator RootRelativeUrlValidator?

db/migrate/20190814163542_create_messages.rb
```diff
@@ -0,0 +1,22 @@
+class CreateMessages < ActiveRecord::Migration[5.2]
+  def change
+    create_table :messages, id: :uuid, default: "uuid_generate_v4()" do |t|
+      t.text :sender_message_id
+      t.text :title, null: false
+      t.text :path
```
kevindew: Looks like this needs to be changed to url

db/migrate/20190815191552_add_message_to_subscription_contents.rb
```diff
@@ -0,0 +1,9 @@
+class AddMessageToSubscriptionContents < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :subscription_contents,
+                  :message,
+                  type: :uuid,
+                  foreign_key: { on_delete: :restrict },
+                  index: false
```
kevindew: Just an FYI that when I wrote this before I wasn't quite sure whether adding a foreign key without an index was a table blocking operation. So just a heads up to be cautious with the deploy.

app/models/message.rb
thomasleese: I like `RootRelativeUrlValidator`... I'll do that 👍

db/migrate/20190814163542_create_messages.rb
thomasleese: Good point, I've amended that now.

db/migrate/20190815191552_add_message_to_subscription_contents.rb
thomasleese: It looks like it can cause problems, so I'm going to split this up into two migrations: https://github.com/ankane/strong_migrations#good-5

thomasleese: I've pushed up a change to split this up into two migrations if you'd be able to take another look.

PR #935, opened 2019-08-22
Add endpoint to email-alert-api to send single email

1pretz1:
Trello:
https://trello.com/c/68dWZW80/46-create-endpoint-on-email-alert-api-to-send-single-email

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,6 +9,8 @@ Rails.application.routes.draw do
     post "/unpublish-messages", to: "unpublish_messages#create"
 
     post "/notifications", to: "content_changes#create" # for backwards compaibility
+
+    resources :emails, only: %i[create]
     resources :content_changes, only: %i[create], path: "content-changes"
     resources :spam_reports, path: "spam-reports", only: %i[create]
     resources :status_updates, path: "status-updates", only: %i[create]
--- a/app/controllers/emails_controller.rb
+++ b/app/controllers/emails_controller.rb
@@ -1,0 +1,11 @@
+class EmailsController < ApplicationController
+  def create
+    email = Email.create!(
+      address: params.require(:address),
+      subject: params.require(:subject),
+      body: params.require(:body),
+    )
+
+    DeliveryRequestWorker.perform_async_in_queue(email.id, queue: :delivery_immediate)
+  end
+end
--- a/spec/features/emails_spec.rb
+++ b/spec/features/emails_spec.rb
@@ -1,0 +1,26 @@
+RSpec.describe "Sending an email", type: :request do
+  before do
+    stub_notify
+  end
+
+  scenario do
+    params = {
+      body: 'Description',
+      subject: 'Update from GOV.UK',
+      address: 'test@test.com'
+    }
+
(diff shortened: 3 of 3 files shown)


app/controllers/alert_controller.rb
```diff
@@ -0,0 +1,11 @@
+class AlertController < ApplicationController
```
thomasleese: I wonder if we should avoid the word `Alert` because it tends to be an overused term in the email product. It might be easier to just go with `EmailController` and `/emails` since that's what we're doing here.

app/controllers/alert_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  def create
+    email = Email.create!(
+      address: params.require(:recipient_address),
```
thomasleese:
```suggestion
      address: params.require(:address),
```

Small thing, I'm not sure if `recipient` adds anything here although I don't really mind.

config/routes.rb
```diff
@@ -9,6 +9,8 @@
     post "/unpublish-messages", to: "unpublish_messages#create"
 
     post "/notifications", to: "content_changes#create" # for backwards compaibility
+    post "/alert", to: "alert#create"
```
thomasleese:
```suggestion
    resources :alerts, only: %i[create]
```

I think this format of the route is better.

app/controllers/alert_controller.rb
1pretz1: As you were reviewing this I've changed the wording to use 'message' instead of 'alert', do you think this is better?

app/controllers/alert_controller.rb
1pretz1: Yeah fair point, i'll remove it.

app/controllers/alert_controller.rb
thomasleese: Unfortunately... that's already taken! (Blame @kevindew :smile:)

We decided to call the models added in https://github.com/alphagov/email-alert-api/pull/927 as messages. I think we can probably just change the wording to `email` here since that's the closest model to the controller.

config/routes.rb
1pretz1: Why do you think its better? I'm used to working on content-publisher where routes are defined as I've done above but am not too fussed either way

app/controllers/alert_controller.rb
1pretz1: Ah damn, okay dokay sure.

config/routes.rb
thomasleese: Maybe it's more my personal opinion but I think it's neater, especially when a controller is a typical create, read, update, delete controller for a model.

I don't really mind in this case, but it does fit the pattern we've already got with `content_changes`, `spam_reports`, etc.

1pretz1: Yeah true, but then further up there's the non-resource way of doing it. Maybe there just needs to be one pattern in the routes to settle it once and for all! & I've updated it to use the resource way. 😄

1pretz1:
@thomasleese ready for review no2

PR #936, opened 2019-08-22
Implement sending ad-hoc messages

thomasleese:
This adds functionality which allows sending a message to the members of one the GOV.UK subscription lists. It does this by providing an endpoint `/messages` which allows creating a message (fields are title, path and body) which can then be distributed to everyone matching the criteria for that message. 

This functionality is intended to be used as a bespoke means for communicating with people on mailing lists where we may want to provide more appropriate communication for a particular scenario or to alert them to things that outside normal content changes (this functionality would be an appropriate replacement for the [taxon unpublishing notifications](https://github.com/alphagov/email-alert-api/blob/f78035d1519328fc9bb47a8821eb81207bb64cf2/app/services/unpublish_handler_service.rb) and the [business readiness fake content-changes](https://github.com/alphagov/content-tagger/blob/a5f4273252639328e1b89d05f313860f1f1dd144/app/services/facets/tagging_update_publisher.rb#L27-L33)).

This is the final part of https://github.com/alphagov/email-alert-api/pull/927 and follows on from #933.

[Trello Card](https://trello.com/c/5rtAmAjc/54-finish-the-messages-endpoint-work-in-email-alert-api)


app/builders/message_email_builder.rb
```diff
@@ -0,0 +1,63 @@
...
+        #{permission_reminder(subscriptions)}
+
+        #{ManageSubscriptionsLinkPresenter.call(address)}
+      BODY
+    end
+
+    copy
+  end
+
+  def permission_reminder(subscriptions)
```
1pretz1: Maybe call this `subscription_reminder` as its more of a 'reminder' of your subscription than of your permission?

app/workers/immediate_email_generation_worker.rb
```diff
@@ -40,6 +42,19 @@ def update_content_change_cache(subscription_contents)
...
 
+  def messages
+    @messages ||= {}
+  end
+
+  def update_message_cache(subscription_contents)
+    message_ids = subscription_contents.flat_map { |_, sc| sc.map(&:message_id) }
+                                       .compact
+                                       .uniq
+    ids = message_ids - messages.keys
```
1pretz1: I don't really understand this line as here `messages` will always be an empty array at this point unless I'm misunderstanding something?

PR #938, opened 2019-08-23
Include links to subscriber lists in emails

thomasleese:
This adds a new `url` field to subscriber lists which allows emails to include a link back to the original thing that the user subscribed to.

This also allows us to implement #943 which includes a link back to the subscriber list in the subscription confirmation email.

[Trello Card](https://trello.com/c/pHWNpL5w/49-store-url-of-a-subscriber-list-and-allow-including-that-in-email)

### DIFF
--- a/app/builders/content_change_email_builder.rb
+++ b/app/builders/content_change_email_builder.rb
@@ -50,16 +50,23 @@ private
   def footer(subscriptions, address)
     return feedback_link if subscriptions.empty?
 
-    permission_reminder = I18n.t!("emails.content_change.permission_reminder",
-                                  topic: subscriptions.first.subscriber_list.title)
-
     <<~BODY
-      #{permission_reminder}
+      #{permission_reminder(subscriptions.first.subscriber_list)}
 
       #{ManageSubscriptionsLinkPresenter.call(address)}
 
       #{feedback_link}
     BODY
+  end
+
+  def permission_reminder(subscriber_list)
+    topic = if subscriber_list.url
+              "[#{subscriber_list.title}](#{Plek.new.website_root}#{subscriber_list.url})"
+            else
+              subscriber_list.title
+            end
+
+    I18n.t!("emails.content_change.permission_reminder", topic: topic)
   end
 
   def feedback_link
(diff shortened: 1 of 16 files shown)


app/builders/content_change_email_builder.rb
```diff
@@ -50,18 +50,25 @@ def body(content_change, subscriptions, address)
...
 
       #{ManageSubscriptionsLinkPresenter.call(address)}
 
       #{feedback_link}
     BODY
   end
 
+  def permission_reminder(subscriber_list)
+    topic = if subscriber_list.url
+              "[#{subscriber_list.title}](#{Plek.new.website_root}#{subscriber_list.url})"
```
1pretz1: Just a reminder here that if we're passing a relative url in to create the subscriber list then this is all good, if an absolute url is passed in then we don't need `#{Plek.new.website_root}#`.

1pretz1: Actually thinking about this, should the column name just be called `base_path` unless it's an absolute url?

thomasleese: The reason we didn't want to call it `base_path` is because it can include more than just a path (it could also include query parameters, for example `/results?c[]=fdsfss&c[]=fdsdfs). 

It shouldn't be possible to include an absolute URL here because we some validation that will prevent that: https://github.com/alphagov/email-alert-api/pull/938/files#diff-266065afad82bba6259d1fc97ce5612cR12

thomasleese: I agree the naming is a bit surprising, when we looked online it seemed like these were called "root relative URLs": https://mor10.com/html-basics-hyperlink-syntax-absolute-relative-and-root-relative/

1pretz1: Ah okay, url still makes me think of the full bodied thing but maybe that's just me. This line will have to change, substituting 'url' for 'path' in the current implementation for creating a subscriber list. https://github.com/alphagov/finder-frontend/pull/1350/commits/84db974cfa5e527cfe1922a8eed954b5456b3a5f#diff-4d79c5f76ab8940efcf34f0b531f0058R50

thomasleese: Good point about needing to change that from `url` to `path`.

1pretz1: 👍 I've given you a nice green tick

PR #951, opened 2019-08-29
Support subscriber lists for single pieces of content

thomasleese:
This PR adds the ability to create subscriber lists for receiving updates to single piece of content, represented by its `content_id`. It ignores `locale` as we currently don't support [sending emails for non-English content](https://github.com/alphagov/email-alert-service/blob/e7d06553ed0530249fab3f91cbd2b6b2ac8f7c0d/email_alert_service/models/major_change_message_processor.rb#L122-L127).

The need for this comes from being able to subscribe to travel advice updates and allowing travel-advice-publisher to avoid having to manually make a request to email-alert-api to send out the content change, and instead rely on the standard event queue based system which most other apps use.

Currently, we support this behaviour by travel-advice-publisher [creating a links hash which includes the `content_id` of the current page](https://github.com/alphagov/travel-advice-publisher/blob/1b67291be783370f67c2e6f1b444e4c2b3a23fa7/app/presenters/email_alert_presenter.rb#L49-L51) which isn't the links hash of the content itself in the publishing-api. Because of this, if we were to switch over to sending these emails through email-alert-service right now, the content wouldn't match with any travel advice subscriber lists. Rather than making sure that each travel advice document has a link to itself (which doesn't make much sense), we can support a new field on subscriber lists called `content_id` (similar to `document_type` at the moment).

To fully support travel-advice-publisher not talking directly to email-alert-api we will need to [update the sign up page to include the `content_id`](https://github.com/alphagov/travel-advice-publisher/blob/1b67291be783370f67c2e6f1b444e4c2b3a23fa7/app/presenters/email_alert_signup/edition_presenter.rb#L64-L66) and then update the existing subscriber lists.

### DIFF
--- a/spec/queries/find_exact_query_spec.rb
+++ b/spec/queries/find_exact_query_spec.rb
@@ -46,6 +46,12 @@ RSpec.describe FindExactQuery do
       _subscriber_list = create_subscriber_list(document_type: 'travel_advice')
       expect(query.exact_match).to be_nil
     end
+
+    it "not matched on content_id - even if they match" do
+      query = build_query(links: { policies: { any: %w[aa-11] } }, document_type: 'cd4179c9-6fe4-49bc-8790-da8a1406c8c5')
+      _subscriber_list = create_subscriber_list(document_type: 'cd4179c9-6fe4-49bc-8790-da8a1406c8c5')
+      expect(query.exact_match).to be_nil
+    end
   end
 
   context "when tags are in the query" do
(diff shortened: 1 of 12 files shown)


spec/queries/find_exact_query_spec.rb
```diff
@@ -46,6 +46,12 @@
       _subscriber_list = create_subscriber_list(document_type: 'travel_advice')
       expect(query.exact_match).to be_nil
     end
+
+    it "not matched on content_id - even if they match" do
+      query = build_query(links: { policies: { any: %w[aa-11] } }, document_type: 'cd4179c9-6fe4-49bc-8790-da8a1406c8c5')
```
bevanloon: not sure I fully understand this test - should `document_type` be `content_id`? (similar question for the other tests you've added in this commit)

thomasleese: Hmm you're right this a bit unclear. I think I just wanted a random string for the document type so I generated a UUID which then makes it look like a content ID, and I also forgot to rename the tests.

thomasleese:
I'm going to close this for now because we've got a new way of matching against fields (`criteria_rules`) and it seems better to support this out of the box, so I'll wait for that to be implemented further.

PR #952, opened 2019-08-29
Add tracking to the subscription confirmation title link

thomasleese:
This allows us to know if a user clicks on the title in an email subscription confirmation.

[Trello Card](https://trello.com/c/OwwRCH7s/62-add-utm-analytics-to-the-subscription-confirmation-message)

### DIFF
--- a/app/builders/subscription_confirmation_email_builder.rb
+++ b/app/builders/subscription_confirmation_email_builder.rb
@@ -35,12 +35,6 @@ private
   end
 
   def body
-    title = if subscriber_list.url
-              "[#{subscriber_list.title}](#{Plek.new.website_root}#{subscriber_list.url})"
-            else
-              subscriber_list.title
-            end
-
     <<~BODY
       You’ll get an email each time there are changes to #{title}.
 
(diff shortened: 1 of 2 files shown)


app/builders/subscription_confirmation_email_builder.rb
```diff
@@ -51,4 +45,23 @@ def body
...
+
+  def title_url
+    query = {
+      utm_source: subscriber_list.slug,
+      utm_medium: "email",
+      utm_campaign: "govuk-notifications-subscription-confirmation",
+    }.to_query
+
+    url = subscriber_list.url
+    tracked_url = url + (url.include?("?") ? "&" : "?") + query
```
1pretz1: I don't think you can actually subscribe to a blank results page so there always should be query params so this should be okay - `tracked_url = url + "&" + query`?

thomasleese: You're right, but I think I'll leave it in just to be sure, and in the future we might have subscriber lists without parameters.

PR #956, opened 2019-09-02
Optimise find exact query

bilbof:
The `FindExactMatch` query is used to lookup a subscriber list that a user can subscribe to. This change will speed up that query.

Currently we find lists that have matching keys in the JSON, and then loop through them to find those with matching values. This change would store a digest of the links or tags in a new column with an index, and lookup lists using that digest, which'd be much faster.

Should be merged after https://github.com/alphagov/email-alert-api/pull/965 and https://github.com/alphagov/email-alert-api/pull/963

### Benchmark

I ran a benchmark to demonstrate the performance improvement.

99% of requests were completed with 3161ms on master, compared to 66ms on this branch, or almost 50x faster.

I created 25k subscriber lists locally, and ran both master and this
branch, and fired requests at it using `ab` (Apache Benchmarking tool).

ab -t 10 -r -c 5 http://email-alert-api.dev.gov.uk?/subscriber-lists?tags...

The test was quite biased towards demonstrating the best case for this test.
I created 25k subscriber lists with the same tag keys, but different values.
We likely have a situation similar to this in production, as we have many
more subscriber lists, with subscriptions mostly on the same sets of tags.

Results of benchmarking:

Master:

18 requests completed.

Percentage of the requests served within a certain time (ms)
  50%   2518
  66%   2549
  75%   2744
  80%   2873
  90%   3138
  95%   3161
  98%   3161
  99%   3161
 100%   3161 (longest request)

This branch:

1167 requests completed.

 Percentage of the requests served within a certain time (ms)
   50%     41
   66%     45
   75%     47
   80%     49
   90%     53
   95%     57
   98%     63
   99%     66
  100%    598 (longest request)

### Alternative approaches

Another option would be return early from https://github.com/alphagov/email-alert-api/blob/master/app/queries/find_exact_match.rb#L17 when the first match is found, since we just call `.first` on the result of the query. But I think this would be less performant than this solution.

Another option would be to do a fancier JSON query on the keys and values, and add indexes on those values? I think we're using an older version of postgres without jsonb which might make this trickier but not impossible?

https://trello.com/c/TUITRnQv/993-improve-performance-of-subscriber-list-lookup-query


app/models/subscriber_list.rb
```diff
@@ -16,6 +16,12 @@ class SubscriberList < ApplicationRecord
   has_many :matched_content_changes
   has_many :matched_messages
 
+  before_validation do
```
kevindew: I'd suggest using a before_save block here rather than validation since we don't use this data in validation

app/models/subscriber_list.rb
```diff
@@ -16,6 +16,12 @@ class SubscriberList < ApplicationRecord
...
+    tags_or_links = links.any? ? links : tags
+    data_type = links.any? ? :links : :tags
+    self.tags_links_digest = HashDigest.generate(tags_or_links, data_type)
```
kevindew: As there isn't validation that ensures you can only have links or tags I wouldn't suggest that we have this based on or conditions.

I think it'd be make more sense to hash both links and tags as columns in the database, which seems to be a less domain specific approach to the optimisation.

app/queries/find_exact_query.rb
```diff
@@ -38,6 +38,6 @@ def base_scope
   end
 
   def find_exact(query_field, query)
-    FindExactMatch.new(query_field: query_field, scope: base_scope).call(query).first
+    FindExactMatch.new(query_field: query_field, scope: base_scope).call(query)
```
kevindew: Does FindExactMatch still make sense as an individual class once changed?

db/migrate/20190902133711_add_tags_links_digest_to_existing_subscriber_lists.rb
```diff
@@ -0,0 +1,10 @@
+class AddTagsLinksDigestToExistingSubscriberLists < ActiveRecord::Migration[5.2]
+  def change
+    SubscriberList.all.each do |list|
```
kevindew: You'd probably want to do a find_each here so you don't load the entire query in memory.

db/migrate/20190902133711_add_tags_links_digest_to_existing_subscriber_lists.rb
```diff
@@ -0,0 +1,10 @@
...
+      tags_or_links = list.links.any? ? list.links : list.tags
+      data_type = list.links.any? ? :links : :tags
+      list.tags_links_digest = HashDigest.generate(tags_or_links, data_type)
+      list.save!
+    end
```
kevindew: For this migration it'd probably make sense to disable the transaction as this could gradually lock the whole table. 

If it takes a long time to run you could set it so that it looks up where digests are nil in the table, which would then mean this migration could pick up again if it failed half way through.

db/schema.rb
```diff
@@ -146,10 +146,12 @@
     t.string "slug", limit: 10000, null: false
     t.string "url"
     t.string "description", default: "", null: false
+    t.string "tags_links_digest"
```
kevindew: Once this is populated with data should it be null: false?

lib/hash_digest.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class HashDigest
+  def self.generate(hash = {}, query_field = :tags)
+    raise ArgumentError.new("query_field must be `:tags` or `:links`") unless %i{tags links}.include?(query_field)
+
+    return nil if hash.empty?
+
+    wrapped = {}
+    wrapped[query_field] = sort_hash(hash)
+    Digest::MD5.new.update(wrapped.to_s).hexdigest
```
kevindew: I'd imagine it'd be a bit safer to use a SHA hash to just reduce the fatal risk of collisions further.

app/queries/find_exact_match.rb
```diff
@@ -6,37 +6,14 @@ def initialize(query_field: :tags, scope: SubscriberList)
...
-      subscriber_list_tags_or_links.all? do |key, subscriber_list_tag_or_link|
-        query_any = query_hash[key].fetch(:any, [])
-        query_all = query_hash[key].fetch(:all, [])
-        links_or_tags_any = subscriber_list_tag_or_link.fetch(:any, [])
-        links_or_tags_all = subscriber_list_tag_or_link.fetch(:all, [])
-        query_any.sort == links_or_tags_any.sort && query_all.sort == links_or_tags_all.sort
-      end
-    end
+    digest = HashDigest.generate(tags_links, query_field)
+    scope.where(tags_links_digest: digest).first
```
kevindew: Given there is a risk of collisions (albeit minute) should this pull out all results that match digest and then do a final sanity check on which is equal? or is that overkill?

lib/hash_digest.rb
```diff
@@ -0,0 +1,29 @@
+require 'digest'
+
+class HashDigest
```
kevindew: Minor, but I'd suggest putting this as a model concern or under the namespace of SubscriberList to associate it more with the domain it's within. Being in `lib` makes it seem generic but it's rather specific to SubscriberList.

bilbof:
Thank you for a great review, @kevindew! I agree with the changes you've requested. Before I start on them I had a look at using JSONB instead:

Using JSONB would have advantages:

- we can add indexes on known keys, e.g. (but don't know if this one would work) `CREATE INDEX idx_topics_any ON subscriber_lists (tags->'topics'->'any');`. Caveat is that we don't know the permitted keys for links to do this, but we have an allow list for tag keys https://github.com/alphagov/email-alert-api/blob/master/lib/valid_tags.rb#L1
- simpler queries, e.g. `SELECT * FROM subscriber_lists WHERE tags->'topics'->'any' @> '[my-topic]'::jsonb) LIMIT 1;` with indexes on tags->topic->any I imagine it would have decent performance
- containment queries e.g. `SELECT '[1, 2, 3]'::jsonb @> '[1, 3]'::jsonb;`, to avoid query work in ruby: https://www.postgresql.org/docs/current/datatype-json.html#JSON-CONTAINMENT

A bigger benefit to using JSONB would be that it looks useful for the matched for notification query. My digest solution is only applicable for the exact match query.

My solution also feels like a hack (e.g. the digests rely on consistent ordering of hash keys, the table requires extra columns). However if we don't want to upgrade to postgres 9.4 right now this would solve the immediate issue that requests get slower as the lists table gets bigger with the same keys being used for the checker. We could also remove this once we can query JSONB.

Do you think we'll be upgrading to postgres 9.4 soon?

kevindew:
> Thank you for a great review, @kevindew! I agree with the changes you've requested. Before I start on them I had a look at using JSONB instead:
> 
> Using JSONB would have advantages:
> 
> * we can add indexes on known keys, e.g. (but don't know if this one would work) `CREATE INDEX idx_topics_any ON subscriber_lists (tags->'topics'->'any');`. Caveat is that we don't know the permitted keys for links to do this, but we have an allow list for tag keys https://github.com/alphagov/email-alert-api/blob/master/lib/valid_tags.rb#L1

Yeah I'm not super keen on that approach as it feels a bit of a strange coupling between the db and a validation rule, and odd to be imbalanced with links of course.

> * simpler queries, e.g. `SELECT * FROM subscriber_lists WHERE tags->'topics'->'any' @> '[my-topic]'::jsonb) LIMIT 1;` with indexes on tags->topic->any I imagine it would have decent performance
> * containment queries e.g. `SELECT '[1, 2, 3]'::jsonb @> '[1, 3]'::jsonb;`, to avoid query work in ruby: https://www.postgresql.org/docs/current/datatype-json.html#JSON-CONTAINMENT
> 
> A bigger benefit to using JSONB would be that it looks useful for the matched for notification query. My digest solution is only applicable for the exact match query.

Yeah I think it'd be a lot nicer to use JSONB 👍 - it makes me sad each time I find I can't use a function as it's 9.4 or JSONB only.

> 
> My solution also feels like a hack (e.g. the digests rely on consistent ordering of hash keys, the table requires extra columns). However if we don't want to upgrade to postgres 9.4 right now this would solve the immediate issue that requests get slower as the lists table gets bigger with the same keys being used for the checker. We could also remove this once we can query JSONB.

Sure do you think without specific indexes we'd get decent enough performance with JSONB to not need the optimisation?

> Do you think we'll be upgrading to postgres 9.4 soon?

"Soon" is hard to answer. My hunch would be: unlikely in next 3 months, maybe in next 6 months and would be disappointed if we hadn't within a year. We have to upgrade to 9.6 to move Email Alert API to AWS (it's already running on 9.6 in integration) so it should just happen as part of that. We do have prior art of a 9.6 db in Carrenza (Content Data API) but I'd doubt we'd want to move this to a new db server in Carrenza for that upgrade - given this a sensitive part of the architecture for downtime and the DB is pretty huge.

---

If Email Alert API goes down the future strategy approach discussed in the doc linked above then we'd likely get rid of links and tags fields and potentially the various document type ones for a single JSON blob that describes rules.

I'd imagine if we did this it'd be useful to have this digest approach as that JSON blob would not be parsable at a query level. 

To achieve matching I anticipated we'd need to denormalise aspects of that JSON blob onto subscriber list so it could be queryable. For example if we had criteria of say:

```
[
  { type: tag, key: "topic", value: "value" },
  { type: link, key: "organisation", value: "12345-1234-1234-123456" }
]
```

then that'd probably need to be queryable in the form of: 

```
{
  tags: { topic: ["value"] },
  link: { organisation: ["12345-1234-1234-123456"] },
}
```

and then have final check done in Ruby.

bilbof:
> Sure do you think without specific indexes we'd get decent enough performance with JSONB to not need the optimisation?

Yeah we might be able to get decent performance from [a GIN index](https://www.postgresql.org/docs/9.4/textsearch-indexes.html), to get faster containment queries on tags. This wouldn't require us to specify the valid keys that tags can have, so something like this might work

```
CREATE INDEX idx_topics ON subscriber_lists USING GIN (tags jsonb_path_ops);
SELECT * FROM subscriber_lists WHERE tags->topics->any @> '["my-tag"]'::jsonb;
```

The benchmarks [here](https://coussej.github.io/2016/01/14/Replacing-EAV-with-JSONB-in-PostgreSQL/) on a containment query suggest a gin index makes queries pretty quick (0.153ms for their containment queries):



We might get different results, but it sounds decent. Maybe we could do a benchmark on JSONB once we've upgraded.

> I'd imagine if we did this it'd be useful to have this digest approach as that JSON blob would not be parsable at a query level.

In that case, if we don't plan on an upgrade for a few months at least, I'd like to implement this optimisation in the meantime. What do you think?

kevindew:
Great, sounds good that JSONB offers us a different route forward.

Yeah I think go for it 👍

db/schema.rb
bilbof: I'm not sure actually. If we have a not null constraint, we'd have to put in the hash of '{}' for every row of the table, but would we gain much from that?

app/queries/find_exact_query.rb
bilbof: I think so, though they're very tightly coupled, this change didn't change the relationship between the classes

bilbof:
@kevindew awesome, I made the requested changes.

I re-ran the benchmark and verified there's no change in performance. Perhaps having a larger table and longer hash digests might have a minor effect, but I doubt we'll notice.

```
Percentage of the requests served within a certain time (ms)
  50%     39
  66%     42
  75%     44
  80%     45
  90%     50
  95%     55
  98%     61
  99%     67
 100%    338 (longest request)
```

PR #958, opened 2019-09-03
Add a spec to document /messages URL validation

benthorner:
Previously the URL validation for messages wasn't clear for someone testing the API. This makes the validation explicit and may help future developers.


benthorner:
Oops - finally added that to my global ignore file!

PR #962, opened 2019-09-05
Switch to using criteria for messages instead of links/tags/document-types

kevindew:
Trello: https://trello.com/c/9pic1BWS/78-create-criteria-rules-for-messages-model

This PR switches the looking up of SubscriberLists for a Message to be based upon CriteriaRules rather than by using SubscriberListQuery.

Following this this then removes the unused fields on a message (links, tags, document types) and then validates that sender_message_id should be a UUID.

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -4,7 +4,7 @@ class Message < ApplicationRecord
   has_many :matched_messages
   has_many :subscription_contents
 
-  validates_presence_of :title, :body, :criteria_rules
+  validates_presence_of :title, :body, :criteria_rules, :govuk_request_id
   validates :url, root_relative_url: true, allow_nil: true
   validates :criteria_rules, criteria_schema: true
   validates :sender_message_id,
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -4,9 +4,16 @@ class Message < ApplicationRecord
   has_many :matched_messages
   has_many :subscription_contents
 
-  validates_presence_of :title, :body
+  validates_presence_of :title, :body, :criteria_rules
   validates :url, root_relative_url: true, allow_nil: true
-  validates :criteria_rules, criteria_schema: true, allow_nil: true
+  validates :criteria_rules, criteria_schema: true
+  validates :sender_message_id,
+            format: {
+              with: /\A[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\Z/i,
+              message: "is not a UUID"
+            },
+            uniqueness: true,
+            allow_nil: true
 
   enum priority: { normal: 0, high: 1 }
 
--- a/spec/models/message_spec.rb
+++ b/spec/models/message_spec.rb
@@ -1,5 +1,5 @@
 RSpec.describe Message do
-  describe "url" do
+  describe "url validation" do
     it "is valid when url is nil" do
       expect(build(:message, url: nil)).to be_valid
     end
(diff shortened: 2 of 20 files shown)


spec/models/message_spec.rb
```diff
@@ -13,6 +13,25 @@
     end
   end
 
+  describe "criteria_rules validation" do
+    it "is valid with criteria_rules are valid" do
```
1pretz1: Feels like a slight overlap of concern for the tests below compared with tests in CriteriaSchemaValidator but I suppose that's okay as this is explicitly testing the validity of the record instead of the attribute.

spec/models/message_spec.rb
```diff
@@ -13,6 +13,25 @@
...
+      message = build(:message, criteria_rules: [
```
1pretz1: Feels like a slight overlap of concern for the tests below compared with tests in CriteriaSchemaValidator but I suppose that's okay as this is explicitly testing the validity of the record instead of the attribute.

app/models/message.rb
```diff
@@ -4,7 +4,7 @@ class Message < ApplicationRecord
   has_many :matched_messages
   has_many :subscription_contents
 
-  validates_presence_of :title, :body
+  validates_presence_of :title, :body, :criteria_rules
```
1pretz1: If we are validating the presence of criteria_rules here we should probably remove `allow_nil: true` from line 9 below - `validates :criteria_rules, criteria_schema: true, allow_nil: true`?

spec/queries/email_archive_query_spec.rb
```diff
@@ -48,7 +48,39 @@
       end
     end
 
-    context "when an email is not associated with content changes" do
+    context "when an email is associated with messages" do
+      let!(:subscriber) { create(:subscriber) }
+      let!(:email) { create(:archivable_email, subscriber_id: subscriber.id) }
+      let!(:subscription_contents) do
```
1pretz1: Minor but I don't think the contents of this let really make sense due to both subscription contents being for the same subscriber associated with the same email which wouldn't happen in reality unless I'm misunderstanding something?

spec/models/message_spec.rb
kevindew: Yup it is, we should test that criteria rules are validated so either we test with an example thing that overlaps or use a test that checks whether the validator is called. I prefer the former option since it's a real test but this is an area where there are differing opinions.

app/models/message.rb
kevindew: It's ok to still have the allow_nil here as that just means that the criteria_schema validation isn't run if the value is nil. 

Though it looks like the validation won't error if the input is nil so I'll remove it.

spec/queries/email_archive_query_spec.rb
kevindew: I think the scenario this can occur in is a digest, where you have multiple messages associated with the same email and same subscriber.

kevindew:
@1pretz1 comments are addressed.

app/models/message.rb
1pretz1: Oh I see, I thought the `allow_nil: true` meant that nil was an allowed value to be saved (hence my comment). But thinking about it, nil is allowed anyway unless there is a db constraint so ignore me.

spec/models/message_spec.rb
1pretz1: Nice, good to know! 👍

spec/queries/email_archive_query_spec.rb
1pretz1: Oh okay, I always forget about the digest stream. 🤦‍♂

PR #963, opened 2019-09-05
Add tags & links digest columns to subscriber list

bilbof:
The tags_digest and links_digest columns will be used to find exact matches using provided tags or links.

A later commit will use these in queries.

https://trello.com/c/TUITRnQv/993-improve-performance-of-subscriber-list-lookup-query

### DIFF
--- a/spec/integration/create_subscriber_list_spec.rb
+++ b/spec/integration/create_subscriber_list_spec.rb
@@ -46,6 +46,8 @@ RSpec.describe "Creating a subscriber list", type: :request do
           email_document_supertype
           government_document_supertype
           active_subscriptions_count
+          tags_digest
+          links_digest
         }.to_set.sort
       )
 
@@ -66,6 +68,8 @@ RSpec.describe "Creating a subscriber list", type: :request do
       )
 
       expect(subscriber_list["slug"]).to eq("oil-and-gas-licensing")
+      expect(subscriber_list["links_digest"]).to eq(digested(subscriber_list['links']))
+      expect(subscriber_list["tags_digest"]).to eq(digested(subscriber_list['tags']))
     end
 
     it "can create a subscriber_list with a url" do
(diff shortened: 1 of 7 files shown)


lib/subscriber_list/hash_digest.rb
```diff
@@ -0,0 +1,27 @@
+require 'digest'
+
+class SubscriberList
```
kevindew: Since this class exists in `app/models` you would anticipate this class would live in app/models/subscriber_list`. However since this class is now free from any coupling to SubscriberList (since https://github.com/alphagov/email-alert-api/commit/331b76bcdf624cb618bc912dc1ddfd2aeb8cea95#diff-8c94b9762f86e14636fd40a443d65fbcR5) we probably don't need it coupled within namespacing.

Maybe it does make sense just in `lib/` again?

lib/subscriber_list/hash_digest.rb
```diff
@@ -0,0 +1,27 @@
...
+  class HashDigest
+    def self.generate(hash = {})
```
kevindew: This class has ended up with a somewhat unusual interface with this and sort_hash quasi private method. I'd suggest you actually initialize an object to perform this conversion and then have a self.generate method that is like `new(hash).generate`

Also minor but I assume we don't really need the default argument of `{}` since we'd not expect someone to call this without an argument.

spec/integration/create_subscriber_list_spec.rb
```diff
@@ -93,6 +97,19 @@
...
+      create_subscriber_list(tags: { topics: { any: ["oil-and-gas/licensing"] },
+                                     location: { all: %w[france germany] } },
+                             links: { topics: { any: ["oil-and-gas/licensing"] },
+                                     location: { all: %w[france germany] } })
+      expect(SubscriberList.last.tags_digest).to eq("0cac2ccdb5607c66f8c0ad62cc29e9bdcc09a0b349191406f9db743dac871d4c")
+      expect(SubscriberList.last.links_digest).to eq('0cac2ccdb5607c66f8c0ad62cc29e9bdcc09a0b349191406f9db743dac871d4c')
+
+      create_subscriber_list(tags: { topics: { any: ["oil-and-gas/licensing"] } })
+      expect(SubscriberList.last.tags_digest).to eq("06982c6eaa16c38056dcf81da2bc7ccb7c58e6ef68349db9012534d025a7dc5a")
+      expect(SubscriberList.last.links_digest).to be_nil
```
kevindew: I'd avoid hard coding these digests in here since they make this test rather brittle to change. I think it'd be sufficient that these expectations were just that they matched the output of a call to the digest generator call.

It feels more the role of the digest generator to test that it's returning the right digests.

spec/lib/subscriber_list/hash_digest_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+          more_tags: %w(more more and),
+          number: 2,
+          deep_nest: { foo: 'bar' },
+        },
+        prior: 'one'
+      }
+      expected_digest = "1528bd4ce9c8d03c240c87ddb7d44f5892e23136ffe692cd150ef4a75184aa77"
+
+      expect(described_class.generate(first_input)).to eq(expected_digest)
+      expect(described_class.generate(second_input)).to eq(expected_digest)
```
kevindew: I think it'd be reasonable here to just check these return matching digests rather than they both explicitly match a hardcoded digest.

bilbof:
Thanks @kevindew I made those changes and rebased https://github.com/alphagov/email-alert-api/pull/965 and https://github.com/alphagov/email-alert-api/pull/956. Please could I have another review?

PR #967, opened 2019-09-10
Group field

eamonn-mcelroy-gds:
Added list_group_id to subscriber list

1. Kevin performed migration.
2. Added value to permitted params in controller.
3. Added validation to model, using bespoke UUID validation in validators.
4. Added testing where necessary.
5. Kevin updated documentation to reflect change.

### DIFF
--- a/app/validators/uuid_validator.rb
+++ b/app/validators/uuid_validator.rb
@@ -1,0 +1,9 @@
+class UuidValidator < ActiveModel::EachValidator
+  UUID_REGEX = /\A[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\Z/i.freeze
+
+  def validate_each(record, attribute, value)
+    unless UUID_REGEX =~ value
+      record.errors.add(attribute, "is not a valid UUID")
+    end
+  end
+end
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -36,6 +36,7 @@ private
       slug: slug,
       url: params[:url],
       description: (params[:description] || ""),
+      list_group_id: params[:list_group_id],
       signon_user_uid: current_user.uid,
     )
   end
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -6,13 +6,7 @@ class Message < ApplicationRecord
 
   validates_presence_of :title, :body, :criteria_rules, :govuk_request_id
   validates :criteria_rules, criteria_schema: true, allow_blank: true
-  validates :sender_message_id,
-            format: {
-              with: /\A[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\Z/i,
-              message: "is not a UUID"
-            },
-            uniqueness: true,
-            allow_nil: true
+  validates :sender_message_id, uuid: true, uniqueness: true, allow_nil: true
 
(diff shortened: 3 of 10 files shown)


app/validators/uuid_validator.rb
```diff
@@ -0,0 +1,13 @@
...
+  def validate_each(record, attribute, value)
+    unless valid_uuid?(value)
+      record.errors.add(attribute, "is not a valid UUID")
+    end
+  end
+
+private
+
+  def valid_uuid?(uuid)
+    uuid =~ /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
```
kevindew: I'd suggest setting this regex as a constant and then you can avoid the need for this particular method and do the check in the `validate_each` method. I'd also suggest using the regex from https://github.com/alphagov/email-alert-api/blob/b6428880aa730e316803d7129db3ec47304e933b/app/models/message.rb#L11 as it uses `\A` and `\Z` start and end of string (see https://stackoverflow.com/questions/577653/difference-between-a-z-and-in-ruby-regular-expressions for more).

eamonn-mcelroy-gds: Updated as requested.

app/validators/uuid_validator.rb
```diff
@@ -0,0 +1,10 @@
+class UuidValidator < ActiveModel::EachValidator
+
+  UUID_REGEX = /\A[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\Z/i
```
kevindew: This is currently failing linting as it needs to be frozen:

```
app/validators/uuid_validator.rb:2:1: C: Layout/EmptyLinesAroundClassBody: Extra empty line detected at class body beginning. (https://rubystyle.guide#empty-lines-around-bodies)
app/validators/uuid_validator.rb:3:16: C: Style/MutableConstant: Freeze mutable objects assigned to constants.
  UUID_REGEX = /\A[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\Z/i
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

If you run `bundle exec rake lint` you can run the linting, you can do `bundle exec rake` to run both tests and linting sequentially.

spec/integration/subscriber_lists_controller_spec.rb
```diff
@@ -55,6 +55,22 @@
         end
       end
 
+      context "creating subscriber list with a list_group_id" do
+        let!(:list_group_id) { SecureRandom.uuid }
```
kevindew: this doesn't need to have a bang: https://relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let

Besides that I don't think this even should be a let since we just use it in the one test, it'd be easier to follow if it was specified in the test block itself.

app/validators/uuid_validator.rb
eamonn-mcelroy-gds: Took the linter's advice and applied /\A[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\Z/i.freeze.

spec/integration/subscriber_lists_controller_spec.rb
eamonn-mcelroy-gds: Remove the let and just declared as a standard variable.

PR #968, opened 2019-09-10
Truncate and randomise slug generation

kevindew:
Trello: https://trello.com/c/XBuIOfXj/86-reduce-slug-lengths-used-by-brexit-checklist

This truncates slugs for subscriber lists so they don't exceed 255
characters and use random data to make them unique.

The 255 characters is a somewhat arbitrary number that is picked since
a) it's a common length being 256 bytes long and b) is long enough to
express a significant amount of meaning.

If a slug already exists random data is used instead of a slug so as to
save potentially thousands of query on a popular list and to also not
reveal information on many versions of the list exist. We use 10 bytes
of data to randomise the slug, this will mean there is a 1 in 1 trillion
chance of a collision, so we'd have to have a crazy amount of items with
the same title before a collision may occur.

The rails truncate function is used on the string since that'll truncate
the string only at the point of a word and not leave a rogue trailing
slash.

Also, believe it or not, the title used in the test is a real list
title.


kevindew:
Thanks for the review @thomasleese , I just made a small edit to use 5 bytes of hex rather than 4 to take the collision risk from 1 in ~400 million to 1 in ~1 trillion.

> Is there a plan to go back and truncate old ones?

Yup, I plan to get this deployed first then go back and disable setting a custom slug and truncate old ones

PR #973, opened 2019-09-12
group_id field for associating similar subscriber lists

kevindew:
Trello: https://trello.com/c/sWBqpBeQ/84-create-an-id-to-represent-a-family-of-subscriberlists

This picks up the work from https://github.com/alphagov/email-alert-api/pull/971 and https://github.com/alphagov/email-alert-api/pull/967

This introduces a UUID that can be set on a subscriber_list as a means to group together variants of the same list. The intention is that an application calling Email Alert API can set an id each time they create variants of a list and this id can then be used as a means for calculating statistics for all variants of a list.

The anticipated usage of this is for the Get ready for Brexit checker where there are currently over 8000 lists, this would likely have been useful for the Business Readiness finder that has over 4000 distinct lists. 

In many ways this is actually just a plaster over a gap where the system should probably have a concept of variant of SubscriberList so we can store options on a core list and then have means to model variances separately. However this isn't something that is realistic to do in the short term and represents a bigger architectural change.

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -35,6 +35,7 @@ private
       slug: slugify(title),
       url: params[:url],
       description: (params[:description] || ""),
+      group_id: params[:group_id],
       signon_user_uid: current_user.uid,
     )
   end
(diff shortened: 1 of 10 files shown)


app/controllers/subscriber_lists_controller.rb
```diff
@@ -35,6 +35,7 @@ def subscriber_list_params
       slug: slugify(title),
       url: params[:url],
       description: (params[:description] || ""),
+      list_group_id: params[:list_group_id],
```
thomasleese: Do we need to prefix this with `list_` given that it's on the subscriber list model? It feels like `group_id` would be enough.

kevindew: I've made this change now.

PR #977, opened 2019-09-16
Order subscriptions by when they were created

1pretz1:
This is needed as we are adding when subscriptions have been created on
the subscription management page in email-alert-frontend and it makes
sense for the order to default to created_at descending.

The api has also been updated to allow an order parameter to be passed
for either the `created_at` field for a subscription or the `title` for
a subscriber list. The syntax for descending or ascending values is as
follows `-title` for descending and just `title` for ascending. This
makes the endpoint less coupled to email-alert-frontends code and more
flexible if this endpoint was to be used by something else.

This specific action seemes to be only used for the subscription
management flow, from a git search and a local grep.

Trello:
https://trello.com/c/FGYU4NlN/92-differentiate-between-your-get-ready-for-brexit-results-in-subscription-management

### DIFF
--- a/app/controllers/subscribers_controller.rb
+++ b/app/controllers/subscribers_controller.rb
@@ -1,12 +1,13 @@
 class SubscribersController < ApplicationController
-  def subscriptions
-    subscriptions = Subscription.active.
-      includes(:subscriber_list).
-      where(subscriber: subscriber).
-      order('subscriber_lists.title').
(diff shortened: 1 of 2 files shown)


1pretz1:
@kevindew good point and done, the api is now super flexible in how you can order things, maybe even too flexible! 👀

The periods are also now in the correct place.

PR #998, opened 2019-10-10
Increase the number of workers

oscarwyatt:
A backlog built up yesterday which caused an incident.
Increasing the number of workers should help prevent that
happening again

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,6 +1,6 @@
 ---
 :verbose: true
-:concurrency: 30
+:concurrency: 60
 :logfile: ./log/sidekiq.json.log
 :timeout: 4
 :queues:


thomasleese:
If we want to make this change we'll probably need to think about [increasing the maximum number of database connections as well](https://github.com/alphagov/govuk-puppet/blob/2fb121ba8b388a3dd9f43f9eddd7fca15027f620/hieradata/class/production/postgresql_primary.yaml#L1).

That being said, this is only likely speed up the delivery of emails once they've been created, and I don't think that's where we're seeing the bottleneck.

I think we're seeing the bottleneck in the [`ImmediateEmailGenerationWorker`](https://github.com/alphagov/email-alert-api/blob/master/app/workers/immediate_email_generation_worker.rb) which is the stage which creates the emails and queues them up to be delivered.

I think the first thing we might want to look into is if there are any particularly slow database queries, which you can find by looking at the PostgreSQL logs (`/var/log/postgresql`) and grepping for the word `duration`.

AlanGabbianelli:
@thomasleese https://github.com/alphagov/govuk-puppet/pull/9670

bilbof:
It looks like database lock prevent the jobs from being done concurrently. If one job has a lock for an hour, no new jobs can be completed during that hour. An increased number of workers would help speed up processing of those jobs that have backed up in the queue. But it looks like this happens quite quickly already (2 mins to process the queue after an hour delay).

PR #1000, opened 2019-10-10
Allow rake task to run with multiple ids

AlanGabbianelli:
This will allow us to run either
`rake task report:content_change_email_status_count[id_1]`
or
`report:content_change_email_status_count[id_1,id_2,id_n]`

### DIFF
--- a/lib/reports/content_change_email_status_count.rb
+++ b/lib/reports/content_change_email_status_count.rb
@@ -1,8 +1,7 @@
 module Reports
   class ContentChangeEmailStatusCount
-    def initialize(content_change)
-      @content_change = content_change
-      @email_status_count = email_status_count
+    def initialize(content_changes)
+      @content_changes = content_changes
     end
 
     def self.call(*args)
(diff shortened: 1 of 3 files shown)


lib/tasks/content_change_email.rake
```diff
@@ -1,8 +1,18 @@
...
-  desc "Produce a report on sent, pending and failed emails for a given content change"
-  task :content_change_email_status_count, [:id] => :environment do |_t, args|
-    content_change = ContentChange.find(args[:id])
-    Reports::ContentChangeEmailStatusCount.call(content_change)
+  desc "Produce a report on sent, pending and failed emails for given content change id or ids"
+  task content_change_email_status_count: :environment do |_t, args|
+    ids = args.extras
+    if ids.present?
+      ids.each do |id|
+        content_change = ContentChange.find(id)
```
thomasleese: Perhaps we should do `ContentChange.where(id: id)` and then iterate over the `ContentChanges` to avoid unnecessary queries.

lib/tasks/content_change_email.rake
```diff
@@ -1,8 +1,18 @@
...
+        Reports::ContentChangeEmailStatusCount.call(content_change)
```
thomasleese: Is there a chance we can upgrade this query to run once against each `ContentChange` (rather than having to run it each time for each content change)? It might make it faster to run with multiple IDs.

lib/tasks/content_change_email.rake
```diff
@@ -1,8 +1,18 @@
...
+      end
+    else
+      puts "At least one ContentChange id must be given."
```
thomasleese: Perhaps it's better to put this in the `desc` above so anyone looking at the documentation for this task would get this information without having to run it.

AlanGabbianelli: I'll write a longer description but I want it here too so if someone happens to not pass an id, at least has a helpful message.

lib/reports/content_change_email_status_count.rb
```diff
@@ -1,34 +1,32 @@
...
-      TEXT
+      @content_changes.each do |content_change|
+        puts <<~TEXT
+
+          ---------------------------------------------------------------------------
+          Email status counts for Content Change #{content_change.id}
+          ---------------------------------------------------------------------------
+          Sent emails: #{email_status_count(content_change)['sent']}
+          Pending emails: #{email_status_count(content_change)['pending']}
+          Failed emails: #{email_status_count(content_change)['failed']}
```
thomasleese: We should save the results of this call because right now it means we'll be running the query three times for each `content_change` to get the `sent`, `pending` and `failed` values.

PR #1007, opened 2019-10-17
Multiple ImmediateEmailGenerationWorkers

oscarwyatt:
One per content change or message
Additionally, has different queues for high
and lower priority ImmediateEmailGenerationWorkers

In response to an incident in which a
Travel Advice email was stuck in a backlog
of ~1 million unprocessed SubscriptionContents.
The bottleneck was that there was only
one ImmediateEmailGenerationWorkers and it
could not prioritise those which were more
important or process them fast enough to
prevent a buildup and a delay of around
4 hours.

In the last year, the most content changes
created in a single day was 1581 and the mean
per day was 277 so this change won't flood Sidekiq

### DIFF
--- a/app/queries/subscribers_for_immediate_email_query.rb
+++ b/app/queries/subscribers_for_immediate_email_query.rb
@@ -1,9 +1,11 @@
 class SubscribersForImmediateEmailQuery
-  def self.call
+  def self.call(content_change_id: nil, message_id: nil)
+    raise ArgumentError("Must specify either content_change_id or message_id") unless content_change_id || message_id
+
     unprocessed_subscription_content_exists =
       SubscriptionContent
         .joins(:subscription)
-        .where(email_id: nil)
+        .where(email_id: nil, content_change_id: content_change_id, message_id: message_id)
         .where("subscriptions.subscriber_id = subscribers.id")
         .arel
         .exists
--- a/app/services/immediate_email_generator_service.rb
+++ b/app/services/immediate_email_generator_service.rb
@@ -1,27 +1,20 @@
-class ImmediateEmailGenerationWorker
-  include Sidekiq::Worker
+module ImmediateEmailGeneratorService
+private
 
-  sidekiq_options queue: :email_generation_immediate
-
-  LOCK_NAME = "immediate_email_generation_worker".freeze
-
-  def perform
-    ensure_only_running_once do
-      SubscribersForImmediateEmailQuery.call.find_in_batches(batch_size: 5000) do |group|
-        subscription_contents = grouped_subscription_contents(group.pluck(:id))
-        update_content_change_cache(subscription_contents)
-        update_message_cache(subscription_contents)
-        create_content_change_emails(group, subscription_contents)
-        create_message_emails(group, subscription_contents)
-      end
+  def ensure_only_running_once(type, id)
+    lock_name = "immediate_email_generation_worker_#{type}_#{id}".freeze
+    Subscriber.with_advisory_lock(lock_name, timeout_seconds: 0) do
+      yield
     end
   end
 
-private
-
-  def ensure_only_running_once
-    Subscriber.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
-      yield
+  def generate_emails(content_change_id: nil, message_id: nil)
+    SubscribersForImmediateEmailQuery.call(content_change_id: content_change_id, message_id: message_id).find_in_batches(batch_size: 5000) do |group|
+      subscription_contents = grouped_subscription_contents(group.pluck(:id))
+      update_content_change_cache(subscription_contents)
+      update_message_cache(subscription_contents)
+      create_content_change_emails(group, subscription_contents)
+      create_message_emails(group, subscription_contents)
     end
   end
 
--- a/app/services/queue_courtesy_email_service.rb
+++ b/app/services/queue_courtesy_email_service.rb
@@ -1,0 +1,21 @@
+class QueueCourtesyEmailService
+  def self.call(email_builder, content_change: nil, message: nil)
+    raise ArgumentError.new("Filter must be either :for_content_change or :for_message") unless content_change || message
+
+    subscriber = Subscriber.find_by(address: Email::COURTESY_EMAIL)
+    return unless subscriber
+
+    additional_parameter = content_change ? { content_change: content_change } : { message: message }
+    email_id = email_builder.call([
+                                      {
+                                          address: subscriber.address,
+                                          subscriptions: [],
+                                          subscriber_id: subscriber.id,
+                                      }.merge(additional_parameter),
+                                  ]).ids.first
+
+    DeliveryRequestWorker.perform_async_in_queue(
+      email_id, queue: :delivery_immediate
+    )
+  end
+end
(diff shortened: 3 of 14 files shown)


app/queries/subscribers_for_immediate_email_query.rb
```diff
@@ -1,9 +1,9 @@
 class SubscribersForImmediateEmailQuery
-  def self.call
+  def self.call(filter_hash)
```
thomasleese: I can see the benefit of having a generic argument here, but since this method is already highly specialiased I wonder if it's better to define it like this:

```suggestion
  def self.call(content_change_id: nil, message_id: nil)
    raise "Must specify either content_change_id or message_id" unless content_change_id || message_id
```

The advantage being that we can catch any unusual uses of this class which might not be what users expect and it helps to document this method better (the first thing I did when I saw this was look down to see what arguments are being passed to it).

app/workers/concerns/immediate_email_generator.rb
```diff
@@ -1,27 +1,20 @@
...
-  def perform
-    ensure_only_running_once do
-      SubscribersForImmediateEmailQuery.call.find_in_batches(batch_size: 5000) do |group|
-        subscription_contents = grouped_subscription_contents(group.pluck(:id))
-        update_content_change_cache(subscription_contents)
-        update_message_cache(subscription_contents)
-        create_content_change_emails(group, subscription_contents)
-        create_message_emails(group, subscription_contents)
-      end
+  def ensure_only_running_once(type, id)
```
thomasleese: Do we need this anymore now that we're having a single worker for each content change/message? Sidekiq should make sure that only one worker picks up the job, so I don't think we would be in a situation where the job could run at the same time by multiple workers.

app/workers/concerns/processor.rb
```diff
@@ -0,0 +1,32 @@
+module Processor
```
thomasleese: I'm not sure what this class is from the name, it seems like a place to put miscellaneous methods used by workers which could end up becoming untidy.

The idea of "concerns" is also a new one for email-alert-api, generally we've been using the concept of "services" instead which might be a better fit here. For example, I can imagine the courtesy email part being refactored into a service: `CourtesyEmailService.queue(ContentChangeEmailBuilder, content_change: content_change)`

As I understand it, while this is a very nice refactor, it's not specifically part of the code to have multiple immediate email generation workers, so I think it's worth pulling it out into a separate commit (or even a separate PR which could get merged in sooner).

config/sidekiq.yml
```diff
@@ -5,16 +5,14 @@
...
+  - [email_generation_immediate_high, 5]
+  - [email_generation_immediate, 3]
   - [default, 3]
   - [delivery_digest, 2]
   - [email_generation_digest, 1]
   - cleanup
 :schedule:
-  immediate_email_generation:
-    every: '5s'
-    class: ImmediateEmailGenerationWorker
```
thomasleese: :tada: :fire:

app/queries/subscribers_for_immediate_email_query.rb
oscarwyatt: Excellent suggestion! 👍

app/workers/concerns/immediate_email_generator.rb
oscarwyatt: I've done some thinking/digging and as far as I know/can tell Sidekiq will ensure at least one worker will pick up the job, but won't guarantee that only one will. It's theoretically possible that an instance of ProcessContentChangeWorker will queue up a job, fail, get retried and queue up _another_ job. My understanding is that the second job could then process SubscriptionContent which the first job is working on creating emails for but don't yet have one. I _think_ therefore, in theory, that removing it _could_ result in duplicate emails being sen

app/workers/concerns/processor.rb
oscarwyatt: It was a bit unclear, I've split those into their own classes and hopefully that'll be clearer

I've also split it into two commits

app/services/queue_courtesy_email_service.rb
```diff
@@ -0,0 +1,21 @@
+class QueueCourtesyEmailService
+  def self.call(email_builder, content_change: nil, message: nil)
```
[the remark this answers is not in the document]
oscarwyatt: 👍

app/services/queue_courtesy_email_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    raise ArgumentError.new("Filter must be either :for_content_change or :for_message") unless content_change || message
+
+    subscriber = Subscriber.find_by(address: Email::COURTESY_EMAIL)
```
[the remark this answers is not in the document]
oscarwyatt: 👍

oscarwyatt:
@thomasleese I've tried it on integration and it works just fine, it was able to chew through a decent number of subscribers for travel advice in very little time (although, it's not under proper load it's not a real test) and emails to the courtesy group got through so I'm happy with it

app/services/immediate_email_generator_service.rb
```diff
@@ -1,27 +1,20 @@
-class ImmediateEmailGenerationWorker
-  include Sidekiq::Worker
-
-  sidekiq_options queue: :email_generation_immediate
-
-  LOCK_NAME = "immediate_email_generation_worker".freeze
+module ImmediateEmailGeneratorService
```
kevindew: This mixin used for inheritance doesn't seem an ideal approach to me here, at first I wondered should you just use standard inheritance as that seems less surprising than the module choice for it, though after having a think I don't think you need it at all.

I don't think you gain much by having the lock in a separate method and it might be better to ditch `ensure_only_running_once` to just use the actual lock in both things. Like Tom I do wonder if we can advisory the need for an advisory lock. I'd agree with your comment that we need some form of protection but I'd hope we could do it with db transactions and constraints rather than explicit locks.

With the generate_emails method I think this should have distinct variations for content change and messages. At the moment it does 5 tasks where only 3 are relevant for ContentChanges and Messages respectively so there's a bunch of shared methods that are only relevant on the subclasses.

There's also likely quite a lot of redundant stuff going on. As we only have one Content Change or one Message we shouldn't need a cache of them and it may be that we want to combine looking up by subscribers and their subscriptions given they'd be for the same item.

app/services/queue_courtesy_email_service.rb
```diff
@@ -0,0 +1,51 @@
+class QueueCourtesyEmailService
+  def initialize(content_change_or_message)
+    raise ArgumentError.new("Must be a Message or a Content Change") unless
+        content_change_or_message.is_a?(Message) ||
+          content_change_or_message.is_a?(ContentChange)
```
kevindew: Minor but this guard clause seems rather unusually presented. Might be an easier read as:

if !content_change_or_message.is_a?(Message) && content_change_or_message.is_a?(ContentChange)
  raise ArgumentError.new("Must be a Message or a Content Change")
end

app/services/queue_courtesy_email_service.rb
```diff
@@ -0,0 +1,51 @@
...
+                            address: subscriber.address,
+                            subscriptions: [],
+                            subscriber_id: subscriber.id,
+                          }.merge(email_builder_parameter),
+                      ]).ids.first
+  end
+
+  def is_a_content_change?
+    content_change_or_message.is_a?(ContentChange)
+  end
```
kevindew: This class seems to have ended up as quite a lot to understand for the 4 statements it replaced. I wonder if actually adds value overall?

oscarwyatt:
Closed in favour of #1014

PR #1014, opened 2019-10-28
Generate emails in ProcessWorkers

oscarwyatt:
In response to an incident in which a
Travel Advice email was stuck in a backlog
of ~1 million unprocessed SubscriptionContents.
The bottleneck was that there was only
one ImmediateEmailGenerationWorker and it
could not process SubscriptionContents
concurrently.

By moving the generation of emails to the
ProcessContentChangeAndGenerateEmailsWorker and
ProcessMessageAndGenerateEmailsWorker, it will now be able
to create emails concurrently, thus
removing the bottleneck. It will also make
the system conceptually simpler and make it easier
to see where emails are in the system as the
length of job time or number in the queue
will be more visible.

In the last year, the most content changes
created in a single day was 1581 and the mean
per day was 277 so although this will increase
the time these new workers will take to run,
it shouldn't cause huge backlogs to build up.

The existing ProcessContentChangeWorker
and ProcessMessageWorker and
ImmediateEmailGenerationWorkers are left in.
This is to make the deploying process simpler
and so existing jobs can continue to run and be
restarted if necessary. A few months after
this is merged, we can remove the old workers

Made with @AlanGabbianelli 

Trello: https://trello.com/c/aYHlLIJt/1491-continue-working-on-existing-multiple-immediateemailgenerationworkers-pr

### DIFF
--- a/app/workers/immediate_email_generation_worker.rb
+++ b/app/workers/immediate_email_generation_worker.rb
@@ -1,19 +1,28 @@
-class ImmediateEmailGenerationWorker
+class ImmediateEmailGenerationWorker < ProcessAndGenerateEmailsWorker
   include Sidekiq::Worker
 
   sidekiq_options queue: :email_generation_immediate
 
   LOCK_NAME = "immediate_email_generation_worker".freeze
-  BATCH_SIZE = 5000
 
   def perform
     ensure_only_running_once do
       SubscribersForImmediateEmailQuery.call.find_in_batches(batch_size: BATCH_SIZE) do |group|
-        subscription_contents = grouped_subscription_contents(group.pluck(:id))
-        update_content_change_cache(subscription_contents)
-        update_message_cache(subscription_contents)
-        create_content_change_emails(group, subscription_contents)
-        create_message_emails(group, subscription_contents)
+        content_change_email_data = []
+        content_change_email_ids = {}
+        message_email_data = []
+        message_email_ids = {}
+        ActiveRecord::Base.transaction do
+          subscription_contents = UnprocessedSubscriptionContentsBySubscriberQuery.call(group.pluck(:id))
+          if subscription_contents.any?
+            update_content_change_cache(subscription_contents)
+            update_message_cache(subscription_contents)
+            content_change_email_data, content_change_email_ids = create_content_change_emails(group, subscription_contents)
+            message_email_data, message_email_ids = create_message_emails(group, subscription_contents)
+          end
+        end
+        queue_for_delivery(content_change_email_data, content_change_email_ids)
+        queue_for_delivery(message_email_data, message_email_ids)
       end
     end
   end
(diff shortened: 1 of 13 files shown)


kevindew:
Hey Oscar - thanks for putting this together and the meeting we just had about it.

Here are some notes of the areas we covered that we may want to work on before getting this merged and out.

## Concurrency

We have 90 workers so in theory 90 content changes processed concurrently could block all emails going out. I personally don't think it's too likely that will be the case but to make headroom we may need to run separate processes for separate queues.

## Idempotency

The previous ProcessContentChangeWorker was quite resilient to being multiple times and concurrently by the import task [being resilient to duplicates](https://github.com/alphagov/email-alert-api/blob/master/app/workers/process_content_change_worker.rb#L16-L23). ImmediateEmailGeneratorWorker was resilient due to db lock.

I wonder how hard it could be to make this also resilient to unfriendly side effects if problems occur mid running and it needs to retry or we end up with duplicate processes. The most likely of these seems to be that if any goes wrong creating emails before subscription_contents is updated we can end up with dupe emails. We may be able to resolve this with a transactional process like:

```
Subscribers.in_batches(of: 1000) do |relation|
     subscriber_ids = pluck(:id)
     lock_subscription_contents
     create_emails
     update_subscription_contents
  end
end
```

I don't know effects transaction will have on performance though 🤷‍♂ 

## Recoverability

One thing that we may want to be aware of is that if the sidekiq crashes mid way through a process content change then there'll be nothing to create those emails without human intervention. This was the case before for processing content changes and has never been a problem, but since they will now run for much longer this risk increases.

## ImmediateEmailGenerationWorker

With the approach to launch this with ImmediateEmailGenerationWorker still running we'll probably need ways to stop this and ProcessContentChangeAndGenerateEmailsWorker from running at the same time and both creating the same emails (creating orphaned emails that are distinct from a SubscriptionContent record). The approach suggested above about transactions and locking subscription_contents might allow dealing with this.

## Monitoring

Are we concerned this will blow any healthchecks?
Will this adversely affect [the dashboard](https://grafana.publishing.service.gov.uk/dashboard/file/email_alert_api.json?refresh=10s&orgId=1) at all?

## Measuring

As Bevan raised we probably need to do something on how we can measure change, whether success or failure.

Some of the suggested things to look at are:

- Time from content change created to first email attempt?
- Time from content change created to email created?
- Sidekiq queues

## Approach

Should we be enqueuing with priority? 

There's a bunch of potential improvements/optimisations we may be able to make to the code for processing a content change as some of the bits from ImmediateEmailGeneratorWorker aren't needed or involve looking up information we already have.

For example:
- we shouldn't need to cache or lookup content changes as there will only be one and we already have it loaded
- we don't really need to look up [subscriber and subscription ids](https://github.com/alphagov/email-alert-api/pull/1014/files#diff-96a32650033446c4be893ee0d5988f18R11-R12) as part of the email generation process as we already load [them beforehand](https://github.com/alphagov/email-alert-api/pull/1014/files#diff-96a32650033446c4be893ee0d5988f18R29-R37)

I also think it'd make sense to send the courtesy email after we have generated the emails rather than before so that the presence of that is more of a sign that we've generated our emails - which may help some incidents where seeing that has made us think emails are working when actually they aren't being generated.

app/workers/immediate_email_generation_worker.rb
```diff
@@ -7,12 +7,23 @@ class ImmediateEmailGenerationWorker
...
+              create_content_change_emails(group, subscription_contents)
+              create_message_emails(group, subscription_contents)
+            end
+          end
+          queue_for_delivery
+          clear_emails_to_enqueue
+        end
+      rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid, ActiveRecord::StatementInvalid, ActiveRecord::InvalidForeignKey
+        GovukStatsd.increment("immediate_email_generation_worker.rescheduled")
+        ImmediateEmailGenerationWorker.perform_in(5.seconds)
```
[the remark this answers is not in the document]
thomasleese: And there are a few other places in this PR where we would want to do this I think.

app/workers/process_content_change_and_generate_emails_worker.rb
```diff
@@ -0,0 +1,74 @@
+class ProcessContentChangeAndGenerateEmailsWorker < ProcessAndGenerateEmailsWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :process_and_generate_emails
+
+  def perform(content_change_id)
+    content_change = ContentChange.find(content_change_id)
```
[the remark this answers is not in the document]
kevindew: @thomasleese in a Rails bang quirk `#find` raises an error if it fails to find things in the same way as `#find_by!(id: id)` does.

app/workers/process_content_change_and_generate_emails_worker.rb
```diff
@@ -0,0 +1,74 @@
...
+            update_content_change_cache(subscription_contents)
+            create_content_change_emails(group, subscription_contents)
+          end
+        end
+      end
+      queue_for_delivery
+      content_change.mark_processed!
+      queue_courtesy_email(content_change)
+    rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid, ActiveRecord::StatementInvalid, ActiveRecord::InvalidForeignKey
+      GovukStatsd.increment("process_content_change_and_generate_emails_worker.rescheduled")
```
kevindew: I'm not sure we get a lot from this statsd value, as if an exception outside of those listed raises an error the job would get rescheduled anyway so the stat would be inaccurate.

app/workers/process_content_change_and_generate_emails_worker.rb
```diff
@@ -0,0 +1,74 @@
...
+          subscription_contents = UnprocessedSubscriptionContentsBySubscriberQuery.call(group.pluck(:id))
+          if subscription_contents.any?
+            update_content_change_cache(subscription_contents)
+            create_content_change_emails(group, subscription_contents)
+          end
+        end
+      end
+      queue_for_delivery
+      content_change.mark_processed!
+      queue_courtesy_email(content_change)
```
kevindew: 👍

app/workers/process_content_change_and_generate_emails_worker.rb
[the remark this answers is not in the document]
thomasleese: 🤦‍♂

app/workers/process_content_change_and_generate_emails_worker.rb
oscarwyatt: True. From writing the tests I found that the errors listed occur when there is contention for records - so if you think there is value in recording errors arising from that we could keep it but otherwise you're right it's not especially useful

oscarwyatt: Removed now 😄

PR #1015, opened 2019-10-29
Add subscriber information report

bevanloon:
Add a report that will take one or more email addresses and return a
JSON payload of subscription information. This will include information
about the subscriber, their subscriptions and the associated
subscriber lists.

The report simply pretty prints to the console. I considered saving to a
file, but the reports are usually fairly one-off and the extra hurdle of
getting onto a machine and grabbing the file seemed unnecessary when we
can just run the task in Jenkins and see the results.

### DIFF
--- a/lib/reports/subscriber_subscriptions.rb
+++ b/lib/reports/subscriber_subscriptions.rb
@@ -1,0 +1,44 @@
+module Reports
+  class SubscriberSubscriptions
+    def initialize(email_addresses)
+      @email_addresses = email_addresses
+    end
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def call
+      pp report
+    end
+
+  private
+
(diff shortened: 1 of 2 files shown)


lib/reports/subscriber_subscriptions.rb
```diff
@@ -0,0 +1,44 @@
+module Reports
+  class SubscriberSubscriptions
+    def initialize(email_addresses)
+      @email_addresses = email_addresses
+    end
+
+    def self.call(email_addresses:)
+      new(email_addresses).call
```
thomasleese: It might be better to do `self.call(*args)` here so if the arguments to `initialize` changes, we won't need to update it in two places.

bevanloon: 👍 updated in bde568a

PR #1020, opened 2019-11-01
Add migration to rename brexit related subscriber list titles

koetsier:
Adds a migration to rename Brexit related subscriber list titles

Trello card: https://trello.com/c/dhHPa8eK/201-pre-election-changes-to-brexit-checker-notification-emails


koetsier:
@benthorner We don't really no. I don't mind removing it later (probably makes sense to). It was just for my own peace of mind

PR #1034, opened 2019-11-08
Reduce post-archival retention of emails from 9 to 8 days.

sengi:
See #1028 for context.


sengi:
Won't submit this today; just getting it ready for Monday :)

PR #1039, opened 2019-11-11
Replace `govuk-lint`

issyl0:
- The GOV.UK Lint gem is deprecated in favour of using Rubocop directly
  with a set of shared configs.
- See https://github.com/alphagov/govuk-rfcs/pull/ 100 for more context.

https://trello.com/c/CdBFg6yw/1521-replace-govuk-lint-with-rubocop-govuk


Rakefile
```diff
@@ -5,4 +5,4 @@ require File.expand_path("config/application", __dir__)
```
theseanything: I think most projects lint as part of their default rake task - you could add:

```suggestion
require 'rubocop/rake_task'

Rails.application.load_tasks

RuboCop::RakeTask.new
task default %i[spec rubocop]
```

PR #1045, opened 2019-11-20
Fix prod rake tasks due to missing rubocop

theseanything:
As RuboCop is only a test dependency, it is not available during production. This causes rake in production mode to fail as the Rakefile requires rubocop to define the tasks.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,9 +2,7 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
 require File.expand_path("config/application", __dir__)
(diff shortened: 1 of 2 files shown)


issyl0:
I think the same will happen in [Travel Advice Publisher])(https://github.com/alphagov/travel-advice-publisher/blob/master/Rakefile)?

theseanything:
> I think the same will happen in [Travel Advice Publisher])(https://github.com/alphagov/travel-advice-publisher/blob/master/Rakefile)?

https://github.com/alphagov/travel-advice-publisher/pull/744

PR #1048, opened 2019-11-22
Fixes the current create_subscription api call

koetsier:
Previously, when a user attempted to create a subscription multiple times the code would end the existing subscription, recreate the subscription and send an email.

This PR fixes the behaviour of the API so that it returns a no-op success when the specified subscription already exists. The API still replaces an existing subscription if the one specified has a different frequency.

Trello: https://trello.com/c/L9Oxr2Qx/279-double-opt-in-ignore-multiple-attempts-to-subscribe

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -2,31 +2,37 @@ class SubscriptionsController < ApplicationController
   def create
     return render json: { id: 0 }, status: :created if smoke_test_address?
 
-    existing_subscription = nil
-    subscription = nil
+    subscription = send_email = status = nil
 
     Subscription.transaction do
+      if (subscriber_was_deactivated = subscriber.deactivated?)
+        subscriber.activate!
+      end
+
       existing_subscription = Subscription.active.lock.find_by(
         subscriber: subscriber,
         subscriber_list: subscriber_list,
       )
 
-      existing_subscription.end(reason: :frequency_changed) if existing_subscription
+      create_new_subscription = existing_subscription.nil? ||
+        (frequency.to_s != existing_subscription.frequency.to_s)
 
-      subscriber.activate! if subscriber.deactivated?
+      if create_new_subscription
+        existing_subscription&.end(reason: :frequency_changed)
+        new_subscription = Subscription.create!(
+          subscriber: subscriber,
+          subscriber_list: subscriber_list,
+          frequency: frequency,
+          signon_user_uid: current_user.uid,
+          source: existing_subscription ? :frequency_changed : :user_signed_up,
+          )
+      end
 
-      subscription = Subscription.create!(
-        subscriber: subscriber,
-        subscriber_list: subscriber_list,
-        frequency: frequency,
-        signon_user_uid: current_user.uid,
-        source: existing_subscription ? :frequency_changed : :user_signed_up,
-      )
+      subscription = new_subscription || existing_subscription
+      send_email = create_new_subscription || subscriber_was_deactivated
+      status = existing_subscription.nil? ? :created : :ok
     end
-
-    send_subscription_confirmation_email(subscription)
-
-    status = existing_subscription ? :ok : :created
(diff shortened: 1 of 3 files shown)


app/controllers/subscriptions_controller.rb
```diff
@@ -2,31 +2,36 @@ class SubscriptionsController < ApplicationController
...
-      subscription = Subscription.create!(
-        subscriber: subscriber,
-        subscriber_list: subscriber_list,
-        frequency: frequency,
-        signon_user_uid: current_user.uid,
-        source: existing_subscription ? :frequency_changed : :user_signed_up,
-      )
+      subscription = new_subscription || existing_subscription
+      send_email = create_new_subscription || deactivated_subscriber
+      status = existing_subscription.nil? ? :created : :ok
```
benthorner: Should this be `created` if `create_new_subscription`?

doc/api.md
```diff
@@ -178,13 +178,18 @@ new email address.
...
 ```
 
 and it will create a new subscription between the email address and the
 subscriber list. It will respond with a `201 Created` if it's a new
-subscription or a `200 OK` if the subscription already exists.
+subscription or a `200 OK` if the subscription already exists. If a 
+subscription already exists but the frequency is different, the
+current subscription is ended and a new one with the updated frequency
+is created. A confirmation email will be sent if a new subscription is 
+created.
```
benthorner: Minor: could we add a quick note that this is also if the subscriber is re-activated?

app/controllers/subscriptions_controller.rb
```diff
@@ -2,31 +2,36 @@ class SubscriptionsController < ApplicationController
...
 
-      subscription = Subscription.create!(
-        subscriber: subscriber,
-        subscriber_list: subscriber_list,
-        frequency: frequency,
-        signon_user_uid: current_user.uid,
-        source: existing_subscription ? :frequency_changed : :user_signed_up,
-      )
+      subscription = new_subscription || existing_subscription
+      send_email = create_new_subscription || deactivated_subscriber
```
benthorner: Minor, but `deactivated_subscriber` is strange to read down here. How about changing the assignment to be

```
if subscriber.deactivated?
   subscriber_was_deactivated = true
   subscriber.activate!
end
```

It's not great, but it does mean the subsequent code is easier to understand.

app/controllers/subscriptions_controller.rb
```diff
@@ -2,31 +2,36 @@ class SubscriptionsController < ApplicationController
...
         subscriber_list: subscriber_list,
       )
 
-      existing_subscription.end(reason: :frequency_changed) if existing_subscription
+      create_new_subscription = existing_subscription.nil? ||
+        (frequency.to_s != existing_subscription.frequency.to_s)
 
-      subscriber.activate! if subscriber.deactivated?
+      if create_new_subscription
+        existing_subscription&.end(reason: :frequency_changed)
```
benthorner: Minor: I still find the `frequency_changed` stuff hard to follow, since you have to know about how `create_new_subscription` can be true. I think it might be easier to read this if we had another variable of `frequency_changed` and did the 'ending stuff' in a separate `if` block. How does that sound?

doc/api.md
koetsier: 👍

app/controllers/subscriptions_controller.rb
koetsier: That was not the behaviour of the previous code. I suppose it depends on how you think about it - if you end the previous subscription and create a new one with a different frequency you could see it as modifying the the existing subscription

app/controllers/subscriptions_controller.rb
koetsier: I tried a few permutations but still can't come up with something better looking than this.

app/controllers/subscriptions_controller.rb
koetsier: updated it. let me know what you think!

PR #1050, opened 2019-11-27
Replace #permit! call

koetsier:
Newer versions of Brakeman correctly report #permit! as a security
issue. This PR replaces the call with a more secure version

https://github.com/presidentbeef/brakeman/issues/1426

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -47,14 +47,13 @@ private
   end
 
   def find_exact_query_params
-    permitted_params = params.permit!.to_h
     {
-      tags: convert_legacy_params(permitted_params.fetch(:tags, {})),
-      links: convert_legacy_params(permitted_params.fetch(:links, {})),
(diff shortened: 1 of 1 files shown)


app/controllers/subscriber_lists_controller.rb
```diff
@@ -47,14 +47,14 @@ def convert_legacy_params(link_or_tags)
   end
 
   def find_exact_query_params
-    permitted_params = params.permit!.to_h
+    permitted_params = params.permit(:document_type, :email_document_supertype, :government_document_supertype, :gov_delivery_id)
```
benthorner: This seems like a strange pattern to me, since it involves listing keys like `document_type` twice: one to 'permit' them, and once to actually use them. Surely the usage of them is enough.

Can we just do `params.fetch` everywhere, so we don't need `permitted_params` in the first place?

koetsier: fixed

PR #1051, opened 2019-11-27
Send a confirmation email as part of double opt in

koetsier:
This PR creates a new internal API in email-alert-api that will generate a confirmation email to a specified address. The API does not rely upon or create a subscriber; this will be done as part of confirming the subscription.

The email contains a link back to the signup flow, where the user completes their subscription. The email contains a 'token' query param that will be used to authenticate the address being subscribed.

Trello: https://trello.com/c/Dte8k1Aa/280-double-opt-in-create-an-internal-api-to-send-double-opt-in-emails

### DIFF
--- a/app/builders/subscription_auth_email_builder.rb
+++ b/app/builders/subscription_auth_email_builder.rb
@@ -1,0 +1,55 @@
+class SubscriptionAuthEmailBuilder
+  def initialize(address:, token:, topic_id:, frequency:)
+    @address = address
+    @token = token
+    @topic_id = topic_id
+    @frequency = frequency
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    Email.create!(
+      subject: subject,
+      body: body,
+      address: address,
+      )
+  end
+
+  private_class_method :new
+
+private
+
+  attr_reader :address, :token, :frequency, :topic_id
+
+  def subject
+    "Confirm your subscription"
+  end
+
+  def body
+    <<~BODY
+      # Click the link to confirm your subscription
+
+      ^[Confirm your subscription](#{link})^
+
+      This link will stop working in 7 days.
+
+      **Didn’t request this email?**
+
+      Ignore or delete this email if you didn’t request it.
+
+      [Read our privacy policy (opens in a new tab)](https://www.gov.uk/help/privacy-notice) to find out how we use and protect your data.
+
+      [Contact GOV.UK](https://www.gov.uk/contact/govuk) if you have any problems with your email subscriptions.
+    BODY
+  end
+
+  def link
+    Plek.new.website_uri.tap do |uri|
+      uri.path = "/email/subscriptions/authenticate"
+      uri.query = "token=#{token}&topic_id=#{topic_id}&frequency=#{frequency}"
+    end
+  end
+end
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -31,7 +31,7 @@ private
   end
 
   def build_email(subscriber, token)
-    AuthEmailBuilder.call(
+    SubscriberAuthEmailBuilder.call(
       subscriber: subscriber,
       destination: expected_params.require(:destination),
       token: token,
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -22,8 +22,10 @@ private
 
   def generate_token(subscriber)
     AuthTokenGeneratorService.call(
-      subscriber,
-      redirect: expected_params[:redirect],
+      {
+        "subscriber_id" => subscriber.id,
+        "redirect" => expected_params[:redirect],
+      },
       expiry: 1.week.from_now,
     )
   end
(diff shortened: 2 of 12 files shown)


app/controllers/subscriptions_auth_token_controller.rb
```diff
@@ -1,5 +1,25 @@
...
+
+  def expected_params
+    params.permit(:address, :topic_id, :frequency)
+  end
+
+  class ParamsValidator < OpenStruct
+    include ActiveModel::Validations
+
+    validates :address, presence: true
+    validates :address, email_address: true, allow_blank: true
```
benthorner: Could we combine these two lines? I'm wondering if we need `allow_blank` too.

app/controllers/subscribers_auth_token_controller.rb
```diff
@@ -21,9 +21,12 @@ def find_subscriber(address)
   end
 
   def generate_token(subscriber)
+    data = {
```
benthorner: Minor: could we inline this in the method call, if we're not using it again?

spec/services/auth_token_generator_service_spec.rb
```diff
@@ -1,11 +1,16 @@
 RSpec.describe AuthTokenGeneratorService do
   describe ".call" do
-    let(:subscriber) { create(:subscriber) }
     let(:secret) { Rails.application.secrets.email_alert_auth_token }
     let(:algorithim) { "HS256" }
     let(:expiry) { 1.day.from_now }
+    let(:data) do
+      {
+        token: "token",
```
benthorner: Minor: it's a bit confusing for the token data to be a `token`. Could we use a string like `"data"`?

app/builders/subscription_auth_email_builder.rb
```diff
@@ -0,0 +1,59 @@
+class SubscriberAuthEmailBuilder
```
benthorner: Should this be in the next commit?

app/builders/subscription_auth_email_builder.rb
```diff
@@ -13,47 +14,42 @@ def call
...
-      You need to do this to manage your GOV.UK email subscriptions. The link will stop working in 7 days.
+      This link will stop working in 7 days.
 
-      # Didn’t request this email?
+      **Didn’t request this email?**
 
-      Ignore or delete this email if you didn’t request it. Your subscriptions won’t be changed.
+      Ignore or delete this email if you didn’t request it.
+
+      [Read our privacy policy (opens in a new tab)](https://www.gov.uk/help/privacy-notice) to find out how we use and protect your data.
```
benthorner: It might just be me, but the `opens in a new tab` copy is a bit weird. Are we sure that's right?

app/controllers/subscriptions_auth_token_controller.rb
```diff
@@ -2,13 +2,39 @@ class SubscriptionsAuthTokenController < ApplicationController
...
 
+  def build_email(address:, token:, topic_id:, frequency:)
+    SubscriptionAuthEmailBuilder.call(address: address, token: token, topic_id: topic_id, frequency: frequency)
+  end
+
+  def build_token(address:, topic_id:, frequency:)
+    token_hash = {
+      address: address,
+      topic_id: topic_id,
+      frequency: frequency,
```
benthorner: We don't actually need the frequency in the token, but it doesn't do any harm to have it in there.

app/controllers/subscriptions_auth_token_controller.rb
```diff
@@ -2,13 +2,39 @@ class SubscriptionsAuthTokenController < ApplicationController
   before_action :validate_params
 
   def auth_token
+    address = params.fetch(:address)
+    topic_id = params.fetch(:topic_id)
+    frequency = params.fetch(:frequency)
+
+    token = build_token(address: address, topic_id: topic_id, frequency: frequency)
```
benthorner: Minor: the amount of parameter passing means it takes some effort to read this code. Since this is the only responsibility of the controller, could we do all or most of the work in the action method?

app/controllers/subscriptions_auth_token_controller.rb
```diff
@@ -1,5 +1,25 @@
 class SubscriptionsAuthTokenController < ApplicationController
```
benthorner: Minor: commit messages normally start with a capital 😸

spec/integration/subscriptions_auth_token_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+    context "when we're provided with a badly formatted email address" do
+      let(:address) { "wrong.bad" }
+
+      it "returns a 422" do
+        post path, params: params
+        expect(response.status).to eq(422)
+      end
+    end
+    context "when we're provided with no topic_id" do
```
benthorner: Minor: we seem to be missing some newlines here, compared to other specs.

app/builders/subscription_auth_email_builder.rb
koetsier: Yes it is weird and I already flagged it. It is considered part of MVP and is going to be iterater

app/controllers/subscriptions_auth_token_controller.rb
koetsier: 👍

app/controllers/subscribers_auth_token_controller.rb
koetsier: done

app/builders/subscription_auth_email_builder.rb
koetsier: good point!

PR #1065, opened 2019-12-05
Update anonymisation SQL script for govuk_env_sync.

sengi:
Change the comments to be valid SQL (`--` instead of `#`) and update the comment to reflect that the script will be used in govuk_env_sync in the govuk-puppet repo instead of the old env-sync-and-backup repo.

https://github.com/alphagov/govuk-puppet/pull/9922 is the corresponding PR in govuk-puppet.

### DIFF
--- a/lib/data_hygiene/anonymise_email_addresses.sql
+++ b/lib/data_hygiene/anonymise_email_addresses.sql
@@ -1,38 +1,38 @@
-# This SQL file anonymises email addresses in the email-alert-api database.
-#
-# It tries to preserve structure so that multiple occurences of the same address
-# are mapped to the same anonymous address.
-#
-# It runs when we copy data to integration:
-# https://deploy.publishing.service.gov.uk/job/Copy_Data_to_Integration/
-#
-# Please make changes to this script in email-alert-api where it is tested.
-# Then copy it to the env-sync-and-backup repository.
+-- This SQL file anonymises email addresses in the email-alert-api database.
+--
+-- It tries to preserve structure so that multiple occurences of the same address
+-- are mapped to the same anonymous address.
+--
+-- It runs when we copy data to integration:
+-- https://deploy.publishing.service.gov.uk/job/Copy_Data_to_Integration/
+--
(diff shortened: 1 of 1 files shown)


lib/data_hygiene/anonymise_email_addresses.sql
```diff
@@ -1,45 +1,46 @@
...
+--
+-- It tries to preserve structure so that multiple occurences of the same address
+-- are mapped to the same anonymous address.
+--
+-- It runs when we copy data to Integration via govuk_env_sync:
+-- https://docs.publishing.service.gov.uk/manual/govuk-env-sync.html
+--
+-- Please make changes to this script in
+-- https://github.com/alphagov/email-alert-api/blob/master/lib/data_hygiene/anonymise_email_addresses.sql
+-- where it is tested. Then copy it to the govuk-puppet repository.
```
issyl0: Is there a better way? It feels like we might forget to do this? But how often do we expect edits to happen?

sengi: Yeah this feels icky, but it's what was there before and we have limited time to get a working solution in place.

I thought about whether we could just have Puppet download the file directly from GitHub, but Puppet 3 doesn't support that and in any case it would probably involve an undesirable dependency.

I'm open to exploring other ideas for improving this but I'd prefer that this issue not block updating the script once the govuk_env_sync solution is up and running.

sengi: For what it's worth, I plan to delete the other copy that's in the env-sync-and-backup repo as soon as the govuk_env_sync thing is in place.

issyl0: I thought about the "download directly from GitHub" solution too. Let's go with this for now but note it somewhere for future improvements?

sengi: Sounds good to me. Thanks!

Maybe we could make the one in the Puppet repo canonical and have the tests in email-alert-api pull it from there? Anyway hopefully we can find some solution.

PR #1073, opened 2019-12-16
Add status updates sidekiq job

deborahchua:
We are moving away from using a healthcheck to monitor the status
of delivery attempts created within the past hour.

This sidekiq job will run every 1 min and send the total number
of delivery attempts as well as the number of pending to Graphite.
We can use StatD's `gauge` feature to maintain the current values
until they are next updated.

Trello card: https://trello.com/c/hPiOuVko/1581-5-extract-the-statusupdates-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check

### DIFF
--- a/app/services/global_metrics_service.rb
+++ b/app/services/global_metrics_service.rb
@@ -1,0 +1,25 @@
+class GlobalMetricsService
+  class << self
+    def delivery_attempt_pending_status_total(total)
+      gauge("delivery_attempt.pending_status_total", total)
+    end
+
+    def delivery_attempt_total(total)
+      gauge("delivery_attempt.total", total)
+    end
+
+  private
+
+    def statsd
+      @statsd ||= begin
+        statsd = Statsd.new
+        statsd.namespace = "govuk.email-alert-api"
+        statsd
+      end
+    end
+
+    def gauge(stat, metric)
+      statsd.gauge(stat, metric)
+    end
+  end
+end
--- a/app/services/metrics_service.rb
+++ b/app/services/metrics_service.rb
@@ -64,14 +64,6 @@ class MetricsService
       increment("delivery_attempt.status.#{status}")
     end
 
-    def delivery_attempt_pending_status_total(total)
-      gauge("delivery_attempt.pending_status_total", total)
-    end
-
-    def delivery_attempt_total(total)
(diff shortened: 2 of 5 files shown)


app/services/metrics_service.rb
```diff
@@ -77,5 +77,17 @@ def time(metric, &block)
     def timing(namespace, difference)
       GovukStatsd.timing(namespace, difference)
     end
+
+    def gauge(stat, metric)
+      statsd.gauge(stat, metric)
+    end
+
+    def statsd
```
cbaines: It would be good to name this in some way to identify why it should be used compared to the `GovukStatsd` client.

deborahchua: Thanks @cbaines , I have updated this in https://github.com/alphagov/email-alert-api/pull/1073/commits/6c284698227eb0b3f380f036e4839051beee0382

app/services/metrics_service.rb
```diff
@@ -64,6 +66,14 @@ def delivery_attempt_status_changed(status)
       increment("delivery_attempt.status.#{status}")
     end
 
+    def delivery_attempt_pending_status_total(total)
+      OverallMetricsService.gauge("delivery_attempt.pending_status_total", total)
+    end
+
+    def delivery_attempt_total(total)
+      OverallMetricsService.gauge("delivery_attempt.total", total)
+    end
```
thomasleese: I wonder if it's better to make these methods of `OverallMetricsService` to make the distinction explicit.

app/services/overall_metrics_service.rb
```diff
@@ -0,0 +1,13 @@
+module OverallMetricsService
```
thomasleese:
```suggestion
module GlobalMetricsService
```

I wonder if this is a clearer name for this service.

app/services/global_metrics_service.rb
```diff
@@ -0,0 +1,25 @@
+class GlobalMetricsService
+  class << self
+    def statsd
```
thomasleese: Can this also be private?

deborahchua: Thanks Tom! I have updated this now

PR #1083, opened 2019-12-20
Encrypt token data (as well as signing it)

benthorner:
https://trello.com/c/FwB5bALg/336-double-opt-in-fix-data-leak

Previously we used JWT tokens as part of an authentication mechanism to
create or manage subscriptions. Only GOV.UK with the shared secret could
decode a token, thus confirming the origin of the data within. However,
only the signature of the token is protected by the shared secret; the
rest of the payload can be decoded to plaintext without.

This replaces the use of JWT with Rails' MessageEncryptor, which signs
and encrypts the payload, thus protecting all of it from exposure in
logs and analytics. Since the exact characters in the token vary with
the same payload, the testing approach now involves decoding the token
in order to verify its content.

Since the token is Base64 encoded, we need to escape the token string,
to avoid the '+', '=' and '/' characters, which are reserved as part of
URLs but can be present in the initial token.

https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/activesupport/lib/active_support/message_encryptor.rb#L179
https://en.wikipedia.org/wiki/Base64

### DIFF
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -22,11 +22,8 @@ private
 
   def generate_token(subscriber)
     AuthTokenGeneratorService.call(
-      {
-        "subscriber_id" => subscriber.id,
-        "redirect" => expected_params[:redirect],
-      },
-      expiry: 1.week.from_now,
+      "subscriber_id" => subscriber.id,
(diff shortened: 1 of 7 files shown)


benthorner:
I've tested this working with email-alert-frontend locally 🆗

PR #1091, opened 2019-12-31
Add validation to SubsciberList tags

MahmudH:
The tag is allows characters that cannot be processed in email alert api.
Therefore, we are adding validations to tags, which will only allow alphanumeric, dashes, quotes, brackets and underscores.

Trello: https://trello.com/c/so20SdvH/808

### DIFF
--- a/app/validators/links_validator.rb
+++ b/app/validators/links_validator.rb
@@ -1,0 +1,19 @@
+class LinksValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, links)
+    invalidly_formatted = invalid_formatted_links(links)
+    if invalidly_formatted.any?
+      record.errors.add(attribute, "#{invalidly_formatted.to_sentence} has a value with an invalid format.")
+    end
+  end
+
+private
+
+  def invalid_formatted_links(links)
+    invalid = links.select do |_key, link_values|
+      link_values.flat_map(&:last)
+        .any? { |link| !link.to_s.match?(/\A[a-zA-Z0-9\-_]*\z/) }
+    end
+
+    invalid.keys
+  end
+end
--- a/app/validators/tags_validator.rb
+++ b/app/validators/tags_validator.rb
@@ -7,6 +7,11 @@ class TagsValidator < ActiveModel::EachValidator
     if invalid_tags(tags).any?
       record.errors.add(attribute, "#{invalid_tags(tags).to_sentence} are not valid tags.")
     end
+
+    invalidly_formatted = invalid_formatted_tags(tags)
+    if invalidly_formatted.any?
+      record.errors.add(attribute, "#{invalidly_formatted.to_sentence} has a value with an invalid format.")
+    end
   end
 
 private
@@ -21,5 +26,14 @@ private
         %i[all any].include?(operator) && values.is_a?(Array)
       end
     end
+  end
+
+  def invalid_formatted_tags(tags)
+    invalid = tags.select do |_key, tag_values|
+      tag_values.flat_map(&:last)
+        .any? { |tag| !tag.to_s.match?(/\A[a-zA-Z0-9\-_\/]*\z/) }
+    end
+
+    invalid.keys
   end
 end
--- a/spec/validators/tags_validator_spec.rb
+++ b/spec/validators/tags_validator_spec.rb
@@ -10,7 +10,14 @@ RSpec.describe TagsValidator do
   subject(:model) { TagsValidatable.new }
 
   context "when valid tags are provided" do
-    before { model.tags = { topics: { any: %w(dogs cats), all: %w(horses) } } }
+    before {
+      model.tags = {
+        topics: { any: %w(dogs cats), all: %w(horses) },
+        policies: { any: %w(wWelcome1098-_/), all: %w(news_story) },
+        commodity_type: { any: %w(f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a), all: %w(123-Abc) },
+      }
+    }
+
     it { is_expected.to be_valid }
   end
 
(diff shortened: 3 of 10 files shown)


app/validators/tags_validator.rb
```diff
@@ -22,4 +27,13 @@ def tag_values_are_arrays(tags)
       end
     end
   end
+
+  def invalid_formatted_tags(tags)
+    invalid = tags.select do |_key, tag_values|
+      tag_values.flat_map(&:last)
+        .any? { |tag| !tag.to_s.match?(/\A[a-zA-Z0-9-_ \/"\'\/{\}\]\[]*\z/) }
```
bilbof: It looks like this regex permits quotes in strings. Is that something we want to permit? If so I think it'd be worth adding a test for this case.

If we don't want quotes, would this work? `(/\A[a-zA-Z0-9-_{}\]\[]*\z/)`

spec/validators/tags_validator_spec.rb
```diff
@@ -19,8 +25,8 @@ class TagsValidatable
       model.tags = {
         organisations: { any: %w(dogs cats) },
         topics: { any: %w(dogs cats) },
-        foo: { any: %w(dogs cats) },
-        people: { any: %w(dogs cats) },
+        foo: { any: %w([dogs] !cats) },
+        people: { any: %w(\u0000) },
         world_locations: { any: "dogs" },
```
bilbof: 💯 Nice work adding these. I actually think a couple more examples of valid and invalid tags would be helpful, since there are a lot of cases that we want to ensure we allow/disallow. E.g. a valid uuid like `f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a` and an invalid string like `"><script>alert(1);</script>"` or something like that.

app/validators/tags_validator.rb
MahmudH: There is a test in spec/queries/subscriber_lists_by_criteria_query_spec.rb line 65, which includes quotes. That's the reason I permitted quotes. I'm not sure if this is something we want to change?

bilbof: Ah good spot! In that case I think you've got it right 👍 

We allow some pretty funky stuff in tags, this'd be a valid tag: `[]{}/" '123-Abc`.

kevindew: I don't know if that test means we need to support those things. I wrote that test originally and I expect I was just unit testing places where I thought this was liable to break. 

If we don't have use of those type of tags in the wild we can kill off that test.

I just did a quick experiment and it looks like we don't have any outside alpha numeric characters, numbers, dashes, underscores and forward slash:

```
irb(main):042:0> SubscriberList.where("tags::text != '{}'::text").pluck(:tags).inject([]) { |memo, tags| memo + tags.values.last.flat_map(&:last) }.uniq.all? { |a| a.match?(/\A[a-z0-9\-_\/]*\Z/) }
=> true
```
(Ran on staging)

app/validators/links_validator.rb
```diff
@@ -0,0 +1,19 @@
...
+      record.errors.add(attribute, "#{invalidly_formatted.to_sentence} has a value with an invalid format.")
+    end
+  end
+
+private
+
+  def invalid_formatted_links(links)
+    invalid = links.select do |_key, link_values|
+      link_values.flat_map(&:last)
+        .any? { |link| !link.to_s.match?(/\A[a-zA-Z0-9\-_\/]*\z/) }
```
kevindew: Links should be UUID values so we can probably use the regex from UuidValidator: https://github.com/alphagov/email-alert-api/blob/master/app/validators/uuid_validator.rb#L2

app/validators/tags_validator.rb
AlanGabbianelli: If we don't currently use them, it probably means we don't have a need for them. I'd say it's safe to make the regex simpler and remove the chars that are not needed.

bilbof: Thanks for looking those up Kevin! That sounds good to me. @MahmudH do you want to try using `(/\A[a-zA-Z0-9\-_{}\]\[]*\z/)` and updating/removing those other tests?

kevindew: I'd suggest you can reduce this further to not have `{}` and `[]` since we don't have known usage of them or plans to use them.

spec/factories.rb
```diff
@@ -156,7 +156,7 @@
   factory :subscriber_list do
     sequence(:title) { |n| "title #{n}" }
     sequence(:slug) { |n| "title-#{n}" }
-    tags { { topics: { any: ["motoring/road_rage"] } } }
+    tags { { topics: { any: %w[motoring-road_rage] } } }
```
bilbof: This change isn't necessary now that you've permitted `/` in the regex.

spec/workers/process_message_worker_spec.rb
```diff
@@ -2,15 +2,15 @@
...
+             { type: "tag", key: "topics", value: "oil-and-gas-licensing" },
            ])
   end
 
   let(:email) { create(:email) }
 
   context "with a subscription" do
     let(:subscriber) { create(:subscriber) }
-    let(:subscriber_list) { create(:subscriber_list, tags: { topics: { any: ["oil-and-gas/licensing"] } }) }
+    let(:subscriber_list) { create(:subscriber_list, tags: { topics: { any: %w[oil-and-gas-licensing] } }) }
```
bilbof: Looks like these changes are not necessary.

kevindew:
@MahmudH nice one getting this sorted 👍 

It looks like we lost the validation of using the UUID regex that we discussed here: https://github.com/alphagov/email-alert-api/pull/1091#pullrequestreview-337766258 - Shall I raise a PR to amend that?

bilbof:
@kevindew we found that there were also `content_purpose_supergroup` and `content_purpose_subgroup` present in links which were added for subscriber lists to match on both links and tags (e.g. filter by organisations and document type). These have values like `news_and_communications`. That means UUIDs aren't the only value present in links, so the regex is a little more permissive about structure and values (`-`, `_`, and `/` permitted).

kevindew:
> @kevindew we found that there were also `content_purpose_supergroup` and `content_purpose_subgroup` present in links which were added for subscriber lists to match on both links and tags (e.g. filter by organisations and document type). These have values like `news_and_communications`. That means UUIDs aren't the only value present in links, so the regex is a little more permissive about structure and values (`-`, `_`, and `/` permitted).

Oh no really? that seems like it's wrong :( but fair enough can understand why you couldn't do it. Thanks for explaining.

bilbof:
Yeah I agree. Would be good to find a solution to it.

PR #1099, opened 2020-01-06
Record stats on content changes

edwardkerry:
We are moving away from healthcheck monitoring for individual
metrics as this places additional DB load on each machine and
creates duplicated Icinga alerts.

This PR adds a sidekiq job which to send `critical` and `warning`
content change metrics to Graphite. This is in advance of a new
Icinga alert being configured in Puppet to monitor these values
and report.

[Trello](https://trello.com/c/c6mWg9jP/1627-3-extract-the-contentchanges-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check)

### DIFF
--- a/spec/workers/content_changes_worker_spec.rb
+++ b/spec/workers/content_changes_worker_spec.rb
@@ -1,0 +1,46 @@
+RSpec.describe ContentChangesWorker do
+  describe ".perform" do
+    shared_examples "tests for critical and warning states" do
+      context "find content changes from last 15 minutes" do
+        let(:statsd) { double }
+
+        before do
+          create(:content_change, created_at: 11.minutes_ago)
+          create(:content_change, created_at: 6.minutes_ago)
+          allow(GlobalMetricsService.send(:statsd)).to receive(:gauge)
+        end
+
+        it "records number of unprocessed content changes over 10 minutes old (critical)" do
+          expect(GlobalMetricsService).to receive(:critical_content_changes_total).with(1)
+          described_class.new.perform
+        end
+
+        it "records number of unprocessed content changes over 5 minutes old (warning)" do
+          expect(GlobalMetricsService).to receive(:warning_content_changes_total).with(2)
+          described_class.new.perform
+        end
+
+        it "sends the correct values to statsd" do
+          expect(GlobalMetricsService.send(:statsd)).to receive(:gauge)
+            .with("content_changes.critical_total", 1)
+          expect(GlobalMetricsService.send(:statsd)).to receive(:gauge)
+            .with("content_changes.warning_total", 2)
+
+          described_class.new.perform
+        end
+      end
+    end
+  end
+
+  describe ".perform_async" do
+    before do
+      Sidekiq::Testing.fake! do
+        described_class.perform_async
+      end
+    end
+
+    it "gets put on the low priority 'cleanup' queue" do
+      expect(Sidekiq::Queues["cleanup"].size).to eq(1)
+    end
+  end
+end
--- a/app/services/global_metrics_service.rb
+++ b/app/services/global_metrics_service.rb
@@ -16,6 +16,14 @@ class GlobalMetricsService
       gauge("subscription_contents.warning_total", total)
     end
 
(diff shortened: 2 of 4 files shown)


spec/workers/content_changes_worker_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+          create(:content_change, created_at: 6.minutes_ago)
+          allow(GlobalMetricsService.send(:statsd)).to receive(:gauge)
+        end
+
+        it "records number of cricial content changes created over 10 minutes ago" do
+          expect(GlobalMetricsService).to receive(:critical_content_changes_total).with(1)
+          described_class.new.perform
+        end
+
+        it "ecords number of warning content changes created over 5 minutes ago" do
```
AlanGabbianelli:
```suggestion
        it "records number of warning content changes created over 5 minutes ago" do
```

spec/workers/content_changes_worker_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        end
+
+        it "records number of unprocessed content changes over 5 minutes old (warning)" do
+          expect(GlobalMetricsService).to receive(:warning_content_changes_total).with(2)
+          described_class.new.perform
+        end
+
+        it "sends the correct values to statsd" do
+          expect(GlobalMetricsService.send(:statsd)).to receive(:gauge)
+          .with("content_changes.critical_total", 1)
```
thomasleese:
```suggestion
            .with("content_changes.critical_total", 1)
```

spec/workers/content_changes_worker_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+          expect(GlobalMetricsService.send(:statsd)).to receive(:gauge)
+          .with("content_changes.warning_total", 2)
```
thomasleese:
```suggestion
            .with("content_changes.warning_total", 2)
```

spec/workers/content_changes_worker_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+        before do
+          create(:content_change, created_at: 11.minutes_ago)
+          create(:content_change, created_at: 6.minutes_ago)
+          allow(GlobalMetricsService.send(:statsd)).to receive(:gauge)
+        end
+
+        it "records number of unprocessed content changes over 10 minutes old (critical)" do
+          expect(GlobalMetricsService).to receive(:critical_content_changes_total).with(1)
+          described_class.new.perform
```
thomasleese:
```suggestion
          subject.perform
```

I think by default `subject` is defined as `described_class.new`.

spec/workers/content_changes_worker_spec.rb
```diff
@@ -0,0 +1,34 @@
+RSpec.describe ContentChangesWorker do
+  describe ".perform" do
```
thomasleese: I wonder if we should add a test on `perform_async` to check it gets added to the right queues similar to https://github.com/alphagov/email-alert-api/blob/63799a71b45c869ffa009543ff9f8ba38ce1554b/spec/workers/subscription_contents_worker_spec.rb#L89-L99.

spec/workers/content_changes_worker_spec.rb
edwardkerry: 👍

spec/workers/content_changes_worker_spec.rb
edwardkerry: 👍

spec/workers/content_changes_worker_spec.rb
edwardkerry: Thanks, I've added this test

spec/workers/content_changes_worker_spec.rb
edwardkerry: This makes sense - however as `perform_async` is a class method, that test below uses `described_class`. To avoid mixing `subject` and `described_class` I might suggest leaving this in the more verbose form.

It also matches the style for the other `...Worker` specs which tend to use `described_class`

PR #1105, opened 2020-01-08
Add digest runs sidekiq job

edwardkerry:
We are moving away from healthcheck monitoring for individual
metrics as this places additional DB load on each machine and
creates duplicated Icinga alerts.

This PR adds a sidekiq job which to send critical and warning
digest run metrics to Graphite. This is in advance of a new
Icinga alert being configured in Puppet to monitor these values
and report.

[Trello](https://trello.com/c/vmcbj23s/1662-5-extract-the-digestruns-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check)

### DIFF
--- a/spec/workers/digest_run_worker_spec.rb
+++ b/spec/workers/digest_run_worker_spec.rb
@@ -1,0 +1,46 @@
+RSpec.describe DigestRunWorker do
+  describe ".perform" do
+    shared_examples "tests for critical and warning states" do
+      context "find incomplete digest runs" do
+        let(:statsd) { double }
+
+        before do
+          create(:digest_run, created_at: 61.minutes_ago)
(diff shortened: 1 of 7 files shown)


spec/workers/digest_run_worker_spec.rb
```diff
@@ -0,0 +1,46 @@
+RSpec.describe DigestRunWorker do
+  describe ".perform" do
+    shared_examples "tests for critical and warning states" do
+      context "find digest runs from last  minutes" do
```
deborahchua: Just a minor thing but there's an extra space here, or did you mean to say `last x minutes`?

edwardkerry: Thanks @deborahchua, have reworded

PR #1117, opened 2020-01-20
Add clean:migrate_specialist_subscribers rake task

bilbof:
This rake task will migrate subscriptions to valid subscriber lists.

Practice run on integration: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/114878/console

https://trello.com/c/yLsG8ecR/1292-fix-specialist-publisher-email-signups

### DIFF
--- a/lib/clean/migrate_specialist_subscriber_lists.rb
+++ b/lib/clean/migrate_specialist_subscriber_lists.rb
@@ -1,0 +1,88 @@
+module Clean
+  # A document_type filter was incorrectly added to some subscriber lists
+  # via a change to specialist publisher.
+  class MigrateSpecialistSubscriberLists
+    def migrate_subscribers_to_working_lists(dry_run: true)
+      lists.each { |from_list|
+        to_list = find_to_list(from_list)
+        migrate_list(from_list, to_list, dry_run: dry_run)
+      }
+
+      puts "Found #{dry_run ? '' : 'and migrated '}#{lists.count} #{'lists'.pluralize(lists.count)}"
+    end
+
+    def lists
+      @lists ||= begin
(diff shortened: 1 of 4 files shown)


lib/clean/migrate_specialist_subscriber_lists.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  private
+
+    def find_to_list(from_list)
+      new_tags = from_list.tags
+      new_tags.delete :document_type
+      query = FindExactQuery.new(
+        tags: new_tags,
+        links: from_list.links,
+        document_type: from_list.document_type,
```
barrucadu: Should this be searching on `format`?

bilbof: In this case no, since `document_type` is an attribute of a SubscriberList, rather than part of the `tags`. The bug that occurred is that `document_type` was added to the SubscriberList tags.

lib/clean/migrate_specialist_subscriber_lists.rb
```diff
@@ -0,0 +1,88 @@
...
+              source: :subscriber_list_changed,
+            )
+          end
+        end
+      end
+    end
+
+    def create_list!(from_list, new_tags)
+      SubscriberList.create!(
+        slug: from_list.slug += "-#{SecureRandom.hex(5)}",
```
barrucadu: Do slugs often contain random values?

bilbof: Yes, this is a less robust copy of the current implementation at https://github.com/alphagov/email-alert-api/blob/master/app/controllers/subscriber_lists_controller.rb#L63-L66

barrucadu:
Whoops, merged this on autopilot

PR #1123, opened 2020-01-28
Process content change in transaction

oscarwyatt:
If any errors are encountered during this process, the content change should
be rolled back and the caller should re-attempt.
This is in response to a problem where the content change
was created but no MatchedContentChanges were created.
When this happens any further attempts to create the content
change will fail (as one already exists) so we should
rollback creating the content change so that further
attempt(s) can be made.

### DIFF
--- a/app/services/content_change_handler_service.rb
+++ b/app/services/content_change_handler_service.rb
@@ -10,9 +10,12 @@ class ContentChangeHandlerService
   end
 
   def call
-    content_change = ContentChange.create!(content_change_params)
-    MetricsService.content_change_created
-    MatchedContentChangeGenerationService.call(content_change: content_change)
+    content_change = nil
+    ActiveRecord::Base.transaction do
+      content_change = ContentChange.create!(content_change_params)
+      MetricsService.content_change_created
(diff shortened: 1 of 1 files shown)


app/services/content_change_handler_service.rb
```diff
@@ -10,10 +10,12 @@ def self.call(*args)
...
   def call
-    content_change = ContentChange.create!(content_change_params)
-    MetricsService.content_change_created
-    MatchedContentChangeGenerationService.call(content_change: content_change)
-    ProcessContentChangeAndGenerateEmailsWorker.perform_async(content_change.id)
+    ActiveRecord::Base.transaction do
+      content_change = ContentChange.create!(content_change_params)
+      MetricsService.content_change_created
+      MatchedContentChangeGenerationService.call(content_change: content_change)
+      ProcessContentChangeAndGenerateEmailsWorker.perform_async(content_change.id)
```
cbaines: This should happen after the transaction, to ensure that it has committed prior to the job starting.

oscarwyatt: Good catch, I'd changed my mind about it, I was right the first time!

PR #1129, opened 2020-02-05
Don't send invalid email address errors to Sentry

callumknights:
https://trello.com/c/9s6vdqQP/1712-3-prevent-email-alert-api-from-sending-invalid-email-domain-errors-to-sentry

This creates unneccesary noise in Sentry

### DIFF
--- a/app/providers/notify_provider.rb
+++ b/app/providers/notify_provider.rb
@@ -23,7 +23,9 @@ class NotifyProvider
     :sending
   rescue Notifications::Client::RequestError => e
     MetricsService.failed_to_send_to_notify
-    GovukError.notify(e, tags: { provider: "notify" })
+    unless e.message.end_with?("Not a valid email address")
+      GovukError.notify(e, tags: { provider: "notify" })
+    end
     :technical_failure
   end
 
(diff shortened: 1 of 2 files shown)


app/providers/notify_provider.rb
```diff
@@ -23,7 +23,9 @@ def call(address:, subject:, body:, reference:)
     :sending
   rescue Notifications::Client::RequestError => e
     MetricsService.failed_to_send_to_notify
-    GovukError.notify(e, tags: { provider: "notify" })
+    unless e.message.end_with?("Not a valid email address")
```
thomasleese:
```suggestion
    unless e.message.include?("valid email address")
```

I wonder if this allows us to be more open to the format of this message changing, but ultimately it's unfortunate that we have to rely on the human-readable string for this check, so we're already at risk of that.

spec/providers/notify_provider_spec.rb
```diff
@@ -53,6 +53,19 @@
         expect(GovukError).to receive(:notify)
         described_class.call(arguments)
       end
+
+      context "and it's an invalid email address error" do
+        before do
+          error_response = double(code: 404, body: '{"errors": [{"error": "ValidationError", "message": "email_address Not a valid email address"}]}')
+          allow_any_instance_of(Notifications::Client).to receive(:send_email)
+            .and_raise(Notifications::Client::RequestError.new(error_response))
```
AlanGabbianelli:
```suggestion
            .and_raise(Notifications::Client::BadRequestError.new(error_response))
```

PR #1140, opened 2020-02-20
Add a service to unsubscribe all Brexit Readiness Finder emails

koetsier:
and deactiveate all subscriptions.

trello: 

https://trello.com/c/ChmNnVRZ/523-remove-all-business-readiness-finder-subscriptions
https://trello.com/c/EgTYD0qt/497-send-a-message-to-subscribers-to-the-business-readiness-finder

### DIFF
--- a/app/services/unsubscribe_brf_service.rb
+++ b/app/services/unsubscribe_brf_service.rb
@@ -1,0 +1,79 @@
+class UnsubscribeBRFService
+  MESSAGE = <<~BODY.freeze
+    goes here!
+  BODY
+
+  SUBJECT = "subject goes here".freeze
+
+  def self.call
+    UnsubscribeBRFService.new.call
+  end
+
+  def call
+    subscriber_lists = SubscriberList.where("title like ?", "%Brexit guidance for your business%")
+    puts "Unsubscribing #{subscriber_lists.count} subscriptions"
+    subscriber_lists.inject(false) do |result, subscriber_list|
+      puts "Unsubscribe #{subscriber_list.title}"
+      result | process_subscriber_list(subscriber_list)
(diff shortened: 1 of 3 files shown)


app/services/unsubscribe_brf_service.rb
```diff
@@ -0,0 +1,79 @@
+class UnsubscribeBRFService
+  MESSAGE = <<~BODY.freeze
+    goes here!
```
huwd: Needs a body

app/services/unsubscribe_brf_service.rb
```diff
@@ -0,0 +1,79 @@
...
+  BODY
+
+  SUBJECT = "subject goes here".freeze
```
huwd: needs a subject

app/services/unsubscribe_brf_service.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def self.call
+    UnsubscribeBRFService.new.call
+  end
+
+  def call
+    subscriber_lists = SubscriberList.where("title like ?", "%Brexit guidance for your business%")
+    puts "Unsubscribing #{subscriber_lists.count} subscriptions"
+    subscriber_lists.inject(false) do |result, subscriber_list|
```
huwd: This is cool,
So do i understand this correnctly `.inject(false)` takes the enumberable `subscriber_lists` and sets a default value of false for it... then overwrites that only if the block returns true?

koetsier: it's a way of or-ing all the results so we know if a message has been sent at all

koetsier:
- Need to unique the subscribers
- Update the content

PR #1167, opened 2020-03-17
Add delivery_interaction queue

theseanything:
This adds a separate Sidekiq queue for emails that are in the critical path of user interaction. e.g. subscription confirms. Having a separate queue to lower the delay it takes to send these emails.

### DIFF
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -7,7 +7,7 @@ class SubscribersAuthTokenController < ApplicationController
     email = build_email(subscriber, token)
 
     DeliveryRequestWorker
-      .perform_async_in_queue(email.id, queue: :delivery_immediate_high)
+      .perform_async_in_queue(email.id, queue: :delivery_transactional)
 
     render json: { subscriber: subscriber }, status: :created
   end
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -4,6 +4,7 @@
 :logfile: ./log/sidekiq.json.log
 :timeout: 4
 :queues:
+  - [delivery_transactional, 7]
   - [delivery_immediate_high, 6]
   - [delivery_immediate, 5]
   - [email_generation_immediate, 4]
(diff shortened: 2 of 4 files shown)


app/controllers/subscribers_auth_token_controller.rb
```diff
@@ -7,7 +7,7 @@ def auth_token
     email = build_email(subscriber, token)
 
     DeliveryRequestWorker
-      .perform_async_in_queue(email.id, queue: :delivery_immediate_high)
+      .perform_async_in_queue(email.id, queue: :delivery_interaction)
```
huwd: in terms of the queue name, want we going to use this for?
The mention of naming this stuff "transactional emails" seems to fit with the usecase?

Any chance `:delivery_transactional` is more descriptive?

theseanything: Yup - wasn't really sure what to call it. Didn't think it was useful relating it to priority e.g. delivery_immediate_highest, delivery_immediate_utmost_importance

huwd: lol, sounds like an MS word versioning system `:delivery_immediate_001_FINAL_FINAL_REALLY_FINAL`

config/sidekiq.yml
```diff
@@ -4,6 +4,7 @@
 :logfile: ./log/sidekiq.json.log
 :timeout: 4
 :queues:
+  - [delivery_transactional, 7]
```
thomasleese:
```suggestion
  - delivery_transactional
```

I'm not sure whether this is a good idea, but if we did this, it would mean that Sidekiq _only_ takes jobs from the `delivery_transactional` queue until it's empty. I wonder if this is better for getting the emails out straight away, but it does mean if the queue gets long, other emails are delayed... not sure...

theseanything: Yeah, I was thinking about this - I think its worth trying the weighted method first. I'd expect to see a good improvement anyway because the emails will be at the start of a queue, instead of behind potentially 100,000s of other emails. If this didn't work, I'd be tempted to try a dedicated Sidekiq process for that queue instead make it a blocking queue.

PR #1173, opened 2020-03-19
Change wording on manage subscriptions email

brucebolt:
Users have been getting confused by the wording in the email that is sent out when they manage their subscriptions, leading to a large number of Zendesk tickets.

Therefore changing "Confirm your email address" to "Manage your subscriptions".


huwd:
Ah bruce hadn't spotted this, we've picked this up in the covid team here:
https://github.com/alphagov/email-alert-api/pull/1175

Karl gave us some example copy too

brucebolt:
No problem!  Raised this mostly so I could see how the changes looked visually.

PR #1175, opened 2020-03-19
Add new email content if user is authenticating to manage their subscriptions

huwd:
Trello: https://trello.com/c/QV5aaDq1/37-update-content-to-existing-manage-your-subscription-email

# What

If the destination is /manage/authenticate, we can use a different subject line and body

# Why

We're getting a lot of user confusion about the unsubscription flow. This is a first attempt at clarifying the content to see if it helps

### DIFF
--- a/app/builders/subscriber_auth_email_builder.rb
+++ b/app/builders/subscriber_auth_email_builder.rb
@@ -25,16 +25,20 @@ private
   attr_reader :subscriber, :destination, :token
 
   def subject
-    "Confirm your email address"
+    "Manage your GOV.UK email subscriptions"
   end
 
   def body
     <<~BODY
-      # Click the link to confirm your email address
+      # Manage your GOV.UK email subscriptions
 
-      ^ [Confirm your email address](#{link})
+      You can unsubscribe from emails or change your subscription at:
 
-      You need to do this to manage your GOV.UK email subscriptions. The link will stop working in 7 days.
+      #{link}
+
+      You’ll need to confirm your email address before you can make any changes.
+
+      The link will stop working in 7 days.
 
       # Didn’t request this email?
 
--- a/spec/builders/subscriber_auth_email_builder_spec.rb
+++ b/spec/builders/subscriber_auth_email_builder_spec.rb
@@ -18,7 +18,13 @@ RSpec.describe SubscriberAuthEmailBuilder do
       expect { call }.to change(Email, :count).by(1)
     end
 
-    it "has a link to authenticate" do
+    it "has a subject line prompting the user to manage their subscriptions" do
+      subject = "Manage your GOV.UK email subscriptions"
+      email = call
+      expect(email.subject).to include(subject)
+    end
+
+    it "has body content has a link allowing users to authenticate and manage their subscriptions" do
       link = "http://www.dev.gov.uk/destination?token=secret"
       email = call
       expect(email.body).to include(link)
--- a/spec/features/create_an_auth_token_spec.rb
+++ b/spec/features/create_an_auth_token_spec.rb
@@ -39,13 +39,13 @@ RSpec.describe "Create an auth token", type: :request do
         "body" => hash_including(
           "email_address" => subscriber.address,
           "personalisation" => hash_including(
-            "subject" => "Confirm your email address",
+            "subject" => "Manage your GOV.UK email subscriptions",
             "body" => include("http://www.dev.gov.uk#{destination}?token="),
           ),
         ),
       )
       .with { |request|
-        token = request.body.match(/token=([^&)]+)/)[1]
+        token = request.body.match(/token=([^&\\]+)/)[1]
 
         expect(decrypt_and_verify_token(token)).to eq(
           "subscriber_id" => subscriber.id,
--- a/spec/integration/subscribers_auth_token_spec.rb
+++ b/spec/integration/subscribers_auth_token_spec.rb
@@ -35,7 +35,7 @@ RSpec.describe "Subscribers auth token", type: :request do
     it "sends an email with the correct token" do
       post path, params: params
       expect(Email.count).to be 1
-      token = Email.last.body.match(/token=([^&)]+)/)[1]
+      token = Email.last.body.match(/token=([^&\n]+)/)[1]
       expect(decrypt_and_verify_token(token)).to eq(
         "subscriber_id" => subscriber.id,
         "redirect" => redirect,


app/builders/subscriber_auth_email_builder.rb
```diff
@@ -25,10 +25,39 @@ def call
   attr_reader :subscriber, :destination, :token
 
   def subject
-    "Confirm your email address"
+    return manage_your_subscriptions_subject if destination == "/manage/authenticate"
```
theseanything: I'm not sure this the destination https://github.com/alphagov/email-alert-frontend/blob/19223e768f85627fa917f456033fbcfc1b44b5fa/app/controllers/subscriber_authentication_controller.rb#L17

app/builders/subscriber_auth_email_builder.rb
```diff
@@ -44,16 +73,23 @@ def body
...
-                  else
-                    "token=#{token}"
-                  end
-      uri.fragment = destination_uri.fragment
-    end
+  def manage_your_subscriptions_body
+    <<~BODY
+      # Manage your GOV.UK email subscriptions
+
+      You can manage which GOV.UK emails you receive at:
```
karlchillmaid:
```suggestion
      You can unsubscribe from emails or change your subscription at:
```

spec/features/create_an_auth_token_spec.rb
```diff
@@ -39,13 +39,13 @@ def notify_email(subscriber, destination, redirect)
...
           "personalisation" => hash_including(
-            "subject" => "Confirm your email address",
+            "subject" => "Manage your GOV.UK email subscriptions",
             "body" => include("http://www.dev.gov.uk#{destination}?token="),
           ),
         ),
       )
       .with { |request|
-        token = request.body.match(/token=([^&)]+)/)[1]
+        token = request.body.match(/token=([^&\\]+)/)[1]
```
huwd: I'm not really clear why this regex isn't matching [this line](https://github.com/alphagov/email-alert-api/pull/1175/files#diff-fc85cafa53fc05719fa7019655ddc05fR38). However it's aimed at stripping the token off of the URL. We've changed it from some markdown to something that fits onto a newline.

By dropping a debugger here and taking the `request.body` into rubular. I got the following to work.

spec/integration/subscribers_auth_token_spec.rb
```diff
@@ -35,7 +35,7 @@
     it "sends an email with the correct token" do
       post path, params: params
       expect(Email.count).to be 1
-      token = Email.last.body.match(/token=([^&)]+)/)[1]
+      token = Email.last.body.match(/token=([^&\n]+)/)[1]
```
huwd: I'm not really clear why this regex isn't matching [this line](https://github.com/alphagov/email-alert-api/pull/1175/files#diff-24f45f86c97784ff32a8a8f2ff9e604eR48). However it's aimed at stripping the token off of the URL. We've changed it from some markdown to something that fits onto a newline.

By dropping a debugger here and taking the `request.body` into rubular. I got the following to work.

app/builders/subscriber_auth_email_builder.rb
```diff
@@ -25,23 +25,11 @@ def call
...
-
-      You need to do this to manage your GOV.UK email subscriptions. The link will stop working in 7 days.
-
-      # Didn’t request this email?
-
-      Ignore or delete this email if you didn’t request it. Your subscriptions won’t be changed.
-
-      [Contact GOV.UK](https://www.gov.uk/contact/govuk) if you have any problems with your email subscriptions.
-    BODY
+    manage_your_subscriptions_body
```
theseanything: Can we put the body text directly into this method? Do we need the extra method call?

app/builders/subscriber_auth_email_builder.rb
```diff
@@ -25,23 +25,11 @@ def call
   attr_reader :subscriber, :destination, :token
 
   def subject
-    "Confirm your email address"
+    manage_your_subscriptions_subject
```
theseanything: Can we put the subject text directly into this method? Do we need the extra method call?

spec/builders/subscriber_auth_email_builder_spec.rb
```diff
@@ -18,10 +18,30 @@
...
+      line_eight = "[Contact GOV.UK](https://www.gov.uk/contact/govuk) if you have any problems with your email subscriptions."
+      email = call
+      expect(email.body).to include(line_one)
+      expect(email.body).to include(line_two)
+      expect(email.body).to include(line_three)
+      expect(email.body).to include(line_four)
+      expect(email.body).to include(line_five)
+      expect(email.body).to include(line_six)
+      expect(email.body).to include(line_seven)
+      expect(email.body).to include(line_eight)
```
theseanything: How come this is multiple expectations?

spec/integration/subscribers_auth_token_spec.rb
theseanything: Hmmm weird - wouldn't expect the token to have a `\n` char in it

app/builders/subscriber_auth_email_builder.rb
huwd: yeah can move it back, thought i'd leave it in case we followed up with further changes?

spec/integration/subscribers_auth_token_spec.rb
huwd: it's not the token, it gets the entire body out of the Email.last call then extracts the token from the link we send the user.

I think the test then expects we should be able to encrypt/decrypt that token to prove it's valid

spec/builders/subscriber_auth_email_builder_spec.rb
huwd: yeah... i know it's a bit ugly. Had some difficulty with newlines and whitespace matching if I put the fixture in as one string the right level of indentation.

huwd: is there a better way to check a block of text with line breaks?

theseanything: hmmm does this work?
```ruby
it "has body content describing how to unsubscribe or manage subscriptions" do
    <<~BODY
       # Click the link to confirm your email address
       # Manage your GOV.UK email subscriptions

       ^ [Confirm your email address](#{link})
       You can unsubscribe from emails or change your subscription at:

       You need to do this to manage your GOV.UK email subscriptions. The link will stop working in 7 days.
       #{link}

       You’ll need to confirm your email address before you can make any changes.

       The link will stop working in 7 days.

       # Didn’t request this email?

      Ignore or delete this email if you didn’t request it. Your subscriptions won’t be changed.
      [Contact GOV.UK](https://www.gov.uk/contact/govuk) if you have any problems with your email subscriptions.
    BODY
    email.call
    expect(email.body).to eq(BODY)
end
```

theseanything: Although that feel like a pointless test

theseanything: I don't mind either way - maybe just checking for the link would suffice?

vanitabarrett: @huwd I agree with @theseanything , I'm not sure it's worth checking every single line. I think just checking for the link is ok.

huwd: @vanitabarrett cool cool, I think initially there was more of a switch we had two types of email copy i wanted to test. But yeah, now it's simpler, i should have simplified this too!

huwd: that's amended

PR #1178, opened 2020-03-20
Count failures / successes per subscriber

koetsier:
We need to know which subscribers are invalid and cause repeated failures so we can remove them.

This PR adds success/failure counts for each email sent for a subscriber. This is being updated each time the emails are archived.

Trello: https://trello.com/c/85zUTpds/24-investigate-removing-invalid-email-addresses-so-we-reduce-the-number-of-email-retries

### DIFF
--- a/app/workers/email_archive_worker.rb
+++ b/app/workers/email_archive_worker.rb
@@ -9,6 +9,9 @@ class EmailArchiveWorker
   def perform
     return unless ENV.include?("EMAIL_ARCHIVE_S3_ENABLED")
 
+    update_counts(type: :successes)
+    update_counts(type: :failures)
+
     Email.with_advisory_lock(LOCK_NAME, timeout_seconds: 0) do
       start_time = Time.zone.now
       archived_count = 0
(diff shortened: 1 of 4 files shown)


thomasleese:
How much of an effect does this have on the time this worker takes? It seems like these queries might be quite expensive, querying against the email table and that might put a lot of load on the system.

I also wonder whether we might need to be a bit clearer with the column names because, reading through the code, their values would only be accurate for the emails currently in the database, rather than of all time.

Is there a way this information could be gathered by running queries against athena?

koetsier:
I will try to figure out how much strain it puts on the system - it is something I am worried about as well.

I had a look and I can query the Athena db with
'''
select subscriber_id, COUNT(id) from email_archive where sent=false and subscriber_id is not null group by subscriber_id
'''
which gives me some info but the Athena email_archive table is itself gzipped and archived several times a day it seems and doesn't really allow easy querying (that I know of).  

Any thoughts or comments would be most welcome

huwd:
@thomasleese i'm always a bit fuzzy on performance optimizations. If we have a need to record this somewhere, what should we be optimizing for on email-alert-api?

Is this something where we're worried about disk space/table size or is it query speed/processing power?

thomasleese:
> @thomasleese i'm always a bit fuzzy on performance optimizations. If we have a need to record this somewhere, what should we be optimizing for on email-alert-api?
>
> Is this something where we're worried about disk space/table size or is it query speed/processing power?

I think for email-alert-api (especially at the moment) we should be optimising for writes on the database (emails, subscription contents, etc) and so we should avoid any queries which might affect the speed of write operations.

koetsier:
The write queries are done in batches of 1000 so that is a maximum of 1927 queries

huwd:
Humm... so this PR both reads and updates right?

Could we do something really stupid like just read from the DB and then use Logger to pipe it out to icinga or something?

Then perhaps we could build a report from the logs?

koetsier:
Closing this now. Needs a bit more thought

PR #1182, opened 2020-03-20
Increase concurrency to 50

bevanloon:
Notify figure they can handle our concurrency at 50.
Let's give it a go


bevanloon:
thank you Issy! ⭐️

PR #1201, opened 2020-04-09
Travel advice subscriber lists report

laurentqro:
Creates a rake task that exports the number of subscriptions to travel advice lists as of a given date (format: 'yyyy-mm-dd'), to std out. Output can then easily be pasted into a spreadsheet.

Usage: `rake export:csv_from_travel_advice_at[date]`

This is pretty barebones for now, but the plan is to make email signup data accessible to publishers autonomously.

https://trello.com/c/SJV50IIQ/70-create-method-for-subscriber-list-information-requested-by-fco-and-run-first-report


lib/data_exporter.rb
```diff
@@ -61,4 +65,33 @@ def export_csv(subscriber_lists, at: nil)
...
+      .group(:frequency)
+      .count
+  end
+
+  def travel_advice_subscriber_lists
+    SubscriberList.where(slug: travel_advice_slugs)
+  end
+
+  def travel_advice_slugs
+    YAML.safe_load(File.open("lib/countries.yml")).map { |country| "#{country}-travel-advice" }
```
vanitabarrett, line 95: I'm not sure if we can infer the slug from the country name - I've stumbled on this before. For example: The Gambia has a content slug of `the-gambia` but an email slug of `gambia-travel-advice`

lib/data_exporter.rb
```diff
@@ -61,4 +65,33 @@ def export_csv(subscriber_lists, at: nil)
...
+    Subscription
+      .where(subscriber_list: subscriber_list)
+      .where("created_at < ?", at)
+      .where("ended_at IS NULL OR ended_at >= ?", at)
+      .group(:frequency)
+      .count
+  end
+
+  def travel_advice_subscriber_lists
+    SubscriberList.where(slug: travel_advice_slugs)
```
thomasleese, line 91: It would probably be better to query against the subscriber lists which are linked to a `country`. For example:

```rb
SubscriberList.where("links->'countries' IS NOT NULL")
```

That way we don't need to keep the countries list up to date.

laurentqro: Great idea, thanks @thomasleese !

PR #1202, opened 2020-04-14
Add subscribes and unsubscribes count to FCO subscriber list report

laurentqro:
https://trello.com/c/d1y37f2t/87-add-subs-and-unsubs-to-travel-advice-subscriber-list-report-for-the-fco

### DIFF
--- a/lib/data_exporter.rb
+++ b/lib/data_exporter.rb
@@ -67,7 +67,7 @@ private
   end
 
   def travel_advice_csv(subscriber_lists, at: nil)
-    CSV($stdout, headers: %i(title immediately daily weekly), write_headers: true) do |csv|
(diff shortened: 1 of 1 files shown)


lib/data_exporter.rb
```diff
@@ -76,10 +76,25 @@ def travel_advice_csv(subscriber_lists, at: nil)
 
   def present_travel_advice_report(subscriber_list, at: nil)
     {
-      "title" => cleanup(subscriber_list.title),
+      title:        cleanup(subscriber_list.title),
```
vanitabarrett, line 79: Code formatting question: do the values need to all be in line with each other?

laurentqro: they don't have to, no, it's just something I like doing, but also happy not to :)

vanitabarrett: Not sure if we have a preferred option on GOVUK - lets go with what's used most commonly elsewhere 🙂

PR #1215, opened 2020-04-30
Mark emails that haven't sent because of a technical failure as "failed"

ChrisBAshton:
Previously, if a delivery attempt was marked as "technical_failure",
the email would stay marked as "pending".

This meant we weren't able to use [this rake task][rake-task] to resend
failed emails during a recent incident (after the cause of the failure
had been fixed), so we had to manually query for them.

Trello: https://trello.com/c/M8dulgoy/1938-3-mark-emails-that-havent-sent-because-of-a-technical-failure-as-failed

[rake-task]: https://github.com/alphagov/email-alert-api/blob/c7c03a0f292df40d4d7c2957014b67d8ac11b2dc/lib/tasks/deliver.rake#L24-L32

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -11,7 +11,7 @@ class Email < ApplicationRecord
   }
 
   enum status: { pending: 0, sent: 1, failed: 2 }
-  enum failure_reason: { permanent_failure: 0, retries_exhausted_failure: 1 }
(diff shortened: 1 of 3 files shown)


thomasleese:
It looks like this introduced a bug with archiving. I've written up a card to fix it: https://trello.com/c/nbNAVLwS/1699-fix-archiving-of-technical-failure-emails

PR #1216, opened 2020-04-30
Add rake task to migrate COVID-19 topic to taxon

chao-xian:
Subscribers are currently on a Topical Event as opposed to a Taxonomy
based SubscriberList. This rake task moves them across, and emails
users to let them know of the upcoming change.

https://trello.com/c/Azwahpgq

### DIFF
--- a/lib/subscriber_list_mover.rb
+++ b/lib/subscriber_list_mover.rb
@@ -1,13 +1,15 @@
 class SubscriberListMover
-  attr_reader :from_slug, :to_slug
+  attr_reader :from_slug, :to_slug, :send_email
 
-  def initialize(from_slug:, to_slug:)
+  def initialize(from_slug:, to_slug:, send_email: false)
+    @send_email = send_email
     @from_slug = from_slug
     @to_slug = to_slug
   end
 
   def call
     source_subscriber_list = SubscriberList.find_by(slug: from_slug)
+    sub_count = source_subscriber_list.subscriptions.active.count
     raise "Source subscriber list #{from_slug} does not exist" if source_subscriber_list.nil?
 
     source_subscriptions = Subscription.active.find_by(subscriber_list_id: source_subscriber_list.id)
@@ -17,7 +19,11 @@ class SubscriberListMover
     raise "Destination subscriber list #{to_slug} does not exist" if destination_subscriber_list.nil?
 
     subscribers = source_subscriber_list.subscribers.activated
-    puts "#{subscribers.count} active subscribers moving from #{from_slug} to #{to_slug}"
+    puts "#{sub_count} active subscribers moving from #{from_slug} to #{to_slug}"
+
+    if send_email
+      email_change_to_subscribers(source_subscriber_list)
+    end
 
     subscribers.each do |subscriber|
       Subscription.transaction do
@@ -36,6 +42,8 @@ class SubscriberListMover
         )
 
         if subscribed_to_destination_subscriber_list.nil?
+          puts "Moving #{subscriber.address} with ID #{subscriber.id} to #{destination_subscriber_list.title} list"
+
           Subscription.create!(
             subscriber: subscriber,
             subscriber_list: destination_subscriber_list,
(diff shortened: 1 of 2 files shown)


lib/subscriber_list_mover.rb
```diff
@@ -48,4 +53,30 @@ def call
 
     puts "#{subscribers.count} active subscribers moved from #{from_slug} to #{to_slug}"
   end
+
+  def email_change_to_subscribers(source_subscriber_list)
+    email_subject = "Changes to GOV.UK email alerts"
+    email_utm_parameters = {
+        utm_source: email_subject,
```
benthorner, line 60: I think we should try to use a more unique identifier here, as for [content changes](https://github.com/alphagov/email-alert-api/blob/cb0575b58b7cfc764c99fdaac0ade6c9e305eb74/app/presenters/content_change_presenter.rb#L33). This means we could filter for this specific change in GA.

How about the `from_slug`?

lib/subscriber_list_mover.rb
```diff
@@ -48,4 +53,30 @@ def call
...
+        utm_medium: "email",
+        utm_campaign: "govuk-subscription-ended",
+    }
+    email_redirect = EmailTemplateContext.new.add_utm("https://gov.uk/email/manage", email_utm_parameters)
+    list_title = source_subscriber_list.title.freeze
```
benthorner, line 65: Could you help me understand what the `freeze` is for?

lib/subscriber_list_mover.rb
```diff
@@ -48,4 +53,30 @@ def call
...
+
+    bulk_move_template = <<~BODY.freeze
+      Hello,
+      You've subscribed to get email alerts about #{list_title} information.
```
benthorner, line 69: Minor: I'm guessing this will come up in a content review, but we probably want some new lines here.

lib/subscriber_list_mover.rb
```diff
@@ -19,6 +20,10 @@ def call
     subscribers = source_subscriber_list.subscribers.activated
     puts "#{subscribers.count} active subscribers moving from #{from_slug} to #{to_slug}"
 
+    if send_message
```
benthorner, line 23: We should be careful here: if the actual move fails, we'll need to re-run this task, but of course we don't want to send the emails twice.

This should go after the actual move. However, sending the emails could still fail, and if that happens, we'll have lost the list of subscribers that have been moved.

We should add another layer of protection: print out the list of subscriber IDs that have been moved, so that we can manually go through and send emails if we need to.

lib/subscriber_list_mover.rb
chao-xian: [Done](https://github.com/alphagov/email-alert-api/commit/597fc4aeaf9d70896b7c826d0b2716d3c560f045)

lib/subscriber_list_mover.rb
chao-xian: Copy paste error

lib/subscriber_list_mover.rb
chao-xian: It was a mistake on my part

lib/subscriber_list_mover.rb
chao-xian: [Have done as you suggested](https://github.com/alphagov/email-alert-api/commit/597fc4aeaf9d70896b7c826d0b2716d3c560f045)

chao-xian:
@benthorner agreed - have deleted the additional rake task and added support for reading in an ENV var.

lib/tasks/manage.rake
```diff
@@ -39,7 +39,9 @@ namespace :manage do
 
   desc "Move all subscribers from one subscriber list to another"
   task :move_all_subscribers, %i[from_slug to_slug] => :environment do |_t, args|
-    SubscriberListMover.new(**args).call
+    send_email = ENV["SEND_EMAIL"] || false
```
benthorner, line 42: Minor: `|| false` is redundant, because `||` only happens when the left-hand side is `false`y.

lib/subscriber_list_mover.rb
```diff
@@ -47,5 +50,37 @@ def call
...
+        utm_medium: "email",
+        utm_campaign: "govuk-subscription-ended",
+    }
+    email_redirect = EmailTemplateContext.new.add_utm("https://gov.uk/email/manage", email_utm_parameters)
+    list_title = source_subscriber_list.title
+
+    bulk_move_template = <<~BODY.freeze
+      Hello,
+
+      You've subscribed to get email alerts about #{list_title} information.
```
benthorner, line 72: Minor: should we have some newlines here?

lib/subscriber_list_mover.rb
```diff
@@ -47,5 +50,38 @@ def call
     end
 
     puts "#{subscribers.count} active subscribers moved from #{from_slug} to #{to_slug}"
+
+    if send_email
```
vanitabarrett, line 54: This won't work at the moment as it's using the subscriber list for the old slug (which at this point has no subscribers as we've moved them all). I guess our options are:

1. Send the email first (risky if we send the email and then the move fails)
2. Keep track of each successful move and pass that data to the `email_change_to_subscribers` method (although I'm not sure exactly what data it's expecting).

chao-xian: Thanks @vanitabarrett - I'll rework this for option 2 I think, if I can.

chao-xian: I've gone with option 1 as the BulkEmailService/BulkEmailBuilder only supports sending emails to a SubscriberList and includes a SQL query. I don't think I could build a transient SubscriberList to keep track and then email the users of. See https://github.com/alphagov/email-alert-api/pull/1216/commits/18b301b04ddd5adc4ca8360e833c91754117356c

lib/subscriber_list_mover.rb
chao-xian: Have undone this. See https://github.com/alphagov/email-alert-api/pull/1216#discussion_r421647412

chao-xian: Have undone this. See https://github.com/alphagov/email-alert-api/pull/1216#discussion_r421647412

vanitabarrett:
@chao-xian Looks good to me, just need to tidy up the commits 👍

PR #1221, opened 2020-05-06
Support archiving unsent emails

benthorner:
Fixes: https://sentry.io/organizations/govuk/issues/1639671311/?project=202220&query=is%3Aunresolved

Previously this would raise a NoMethodError on nil in the case we
try to archive an email that wasn't sent. This fixes the presenter
to cope with that scenario, noting that it's still useful to store
history about emails, whether or not they successfully sent.

### DIFF
--- a/app/presenters/email_archive_presenter.rb
+++ b/app/presenters/email_archive_presenter.rb
@@ -13,7 +13,7 @@ class EmailArchivePresenter
       content_change: build_content_change(record),
       message: build_message(record),
       created_at_utc: record.fetch("created_at").utc.strftime(S3_DATETIME_FORMAT),
-      finished_sending_at_utc: record.fetch("finished_sending_at").utc.strftime(S3_DATETIME_FORMAT),
+      finished_sending_at_utc: record.fetch("finished_sending_at")&.utc&.strftime(S3_DATETIME_FORMAT),
       id: record.fetch("id"),
       marked_as_spam: record.fetch("marked_as_spam"),
       sent: record.fetch("sent"),
--- a/spec/presenters/email_archive_presenter_spec.rb
+++ b/spec/presenters/email_archive_presenter_spec.rb
@@ -79,5 +79,18 @@ RSpec.describe EmailArchivePresenter do
           .to raise_error(KeyError)
       end
     end
+
+    context "when the email was never sent" do
+      it "still presents the data" do
+        record["finished_sending_at"] = nil
+
+        expect(described_class.for_s3(record, archived_at)).to match(
+          a_hash_including(
+            finished_sending_at_utc: nil,
+            id: record["id"],
+          ),
+        )
+      end
+    end
   end
 end


kevindew:
Code looks good but logically I don't think it makes sense to fall into this state. How are we ending up with archivable things that have not a time for finished sending? They should still have a status of pending which would not make them be archivable.

One of the pain points I've known of (and not sure if anyone has taken on recently) is that emails can fall into an eternally pending state and I wonder how we can mark them as so.

Sorry to be a pain

thomasleese:
I wonder if this has started happening because of this change: https://github.com/alphagov/email-alert-api/pull/1215

kevindew:
> I wonder if this has started happening because of this change: #1215
> 
> We were motivated to make that change because there exists a Rake task to re-send failed emails which couldn't be used on emails that had a technical failure:
> 
> https://github.com/alphagov/email-alert-api/blob/c7c03a0f292df40d4d7c2957014b67d8ac11b2dc/lib/tasks/deliver.rake#L24-L32

Oh so can we now end up in a situation where we've not finished sending a technical failure? hmm that sounds logically challenging as we want states to either indicate they are in flight or complete and being both sounds tricky.

benthorner:
Thanks for flagging @kevindew, @thomasleese. I didn't dig into if there was an underlying issue, and it sounds like there is. We seem to have a tangential concern over whether the correct status is being set for technical failures. Putting that to one side, does this suggest we're getting an empty param [here](https://github.com/alphagov/email-alert-api/blob/master/app/controllers/status_updates_controller.rb#L15) in the case of a technical failure?

(Caveat: I also haven't checked which particular emails are having an archive issue.)

thomasleese:
> Oh so can we now end up in a situation where we've not finished sending a technical failure? hmm that sounds logically challenging as we want states to either indicate they are in flight or complete and being both sounds tricky.

Yes, I wasn't sure which way to go with this. I managed to convince myself it was okay for technical failure to be a final state because unless we manually intervene (i.e. by running those Rake tasks) then the email has "finished" sending, and its final state was technical failure. The system isn't going to anything with those to retry them, so leaving them in a pending state was confusing as developers thought that meant the system was still trying to send them.

thomasleese:
> Thanks for flagging @kevindew, @thomasleese. I didn't dig into if there was an underlying issue, and it sounds like there is. We seem to have a tangential concern over whether the correct status is being set for technical failures. Putting that to one side, does this suggest we're getting an empty param [here](https://github.com/alphagov/email-alert-api/blob/master/app/controllers/status_updates_controller.rb#L15) in the case of a technical failure?

We won't receive a technical failure from GOV.UK Notify's status API (where they make calls back to us to tell us about the state of emails), we'll only get it [when the API call to Notify itself fails](https://github.com/alphagov/email-alert-api/blob/51090366ae82b82af41ee52c27eaf9859f0353a9/app/providers/notify_provider.rb#L29) so this shouldn't be a problem with the empty parameter.

kevindew:
Ok hmm so can things only be given a technical failure once they've exhausted pending?

Logically things should only be archived once they're not going to change anymore (otherwise the archived data is incorrect).

If we are to retry technical failures by a rake task (and I don't know the process used for retrying them so I may be way off) could we create new email objects based off the technical failures rather than resurrecting things in a final state?

benthorner:
I've followed the code, and see what you mean: we're setting the status, but nothing else. However, I believe that should still exclude the email from archiving [here](https://github.com/alphagov/email-alert-api/blob/51090366ae82b82af41ee52c27eaf9859f0353a9/app/queries/email_archive_query.rb#L67).

Am I missing something (else) subtle?

benthorner:
I did a quick check of this in production. Does look like it's `technical_failure` that's causing the issue.

```
irb(main):003:0> rs = Email.failed.where(finished_sending_at: nil).pluck(:failure_reason); nil
=> nil
irb(main):004:0> rs.uniq
=> ["technical_failure"]
```

kevindew:
I'd plan to see if we can replace this with resolving whatever is allowing technical failures to continue sending.

benthorner:
Closing, as we've agreed to address the underlying reason why `technical_failure` Emails are not suitable for archiving.

PR #1226, opened 2020-05-12
Adds rake tasks for querying and unsubscribing subscriptions, and pulls in some docs

ChrisBAshton:
We're sometimes asked to unsubscribe a user from all subscriptions, which we have an [existing rake task](https://docs.publishing.service.gov.uk/manual/manage-email-subscribers.html#unsubscribe-a-subscriber-from-all-emails) for. But if a user wants to only unsubscribe from some lists and leave the rest alone, we currently have to SSH into email-alert-api machines and manually process that, which is a laborious and slightly risky exercise.

The [code to unsubscribe a specific subscription already exists](https://github.com/alphagov/email-alert-api/blob/70797224f3ba09cdfc02ee9b517d43570df2f363/app/services/unsubscribe_service.rb#L7-L9), but just isn't exposed in a rake task yet. So...

- In the first commit, I add a rake task that lets us unsubscribe a user from a given subscription. All that is needed is their email address and the slug of the subscriber list.
- There was currently no easy way of getting the slug of the subscriber list, so this would still involve some manual SSHing and querying. So...
- In the second commit, I add a rake task that lets us query which subscriptions a user has. We can use this to find out the slug of the subscription we wish to remove, as well as to perform due diligence in seeing which subscriptions a user has and double-checking that an unsubscribe task was successful.

This should hopefully streamline our 2nd line workflow a fair bit, and remove the need to SSH into email-alert-api machines for routine work.

EDIT: while I'm here, I've also pulled in the rake task docs from govuk-developer-docs. Now that email-alert-api docs are indexed, it makes sense to keep the docs in the same repo as the code. The sister PR to remove the docs elsewhere is https://github.com/alphagov/govuk-developer-docs/pull/2540. This made it very easy to then rename a badly named rake task, which no longer needs updating in tandem in a separate repository! 🎉

### DIFF
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -1,6 +1,33 @@
 require "csv"
 
 namespace :manage do
+  desc "View all subscriptions for a subscriber"
+  task :view_subscriptions, [:email_address] => :environment do |_t, args|
+    email_address = args[:email_address]
+    subscriber = Subscriber.find_by_address(email_address)
+    abort("Cannot find any subscriber with email address #{email_address}.") if subscriber.nil?
+
+    results = subscriber.subscriptions.map do |subscription|
+      subscriber_list = SubscriberList.find(subscription.subscriber_list_id)
+      {
+        status: subscription.ended_at.present? ? "Inactive (#{subscription.ended_reason})" : "Active",
+        subscriber_list: "#{subscriber_list.title} (slug: #{subscriber_list.slug})",
+        frequency: subscription.frequency.to_s,
+        timeline: "Subscribed #{subscription.created_at}#{subscription.ended_at.present? ? ", Ended #{subscription.ended_at}" : ''}",
+      }
+    end
+    columns = results.first.each_with_object({}) do |(col, _), h|
+      heading = col.to_s.humanize
+      h[col] = { label: heading, width: [results.map { |g| g[col].size }.max, label.size].max }
+    end
+
+    puts "| #{columns.map { |_, g| g[:label].ljust(g[:width]) }.join(' | ')} |"
+    results.each do |h|
+      str = h.keys.map { |k| h[k].ljust(columns[k][:width]) }.join(" | ")
+      puts "| #{str} |"
+    end
+  end
+
   desc "Change the email address of a subscriber"
   task :change_email_address, %i[old_email_address new_email_address] => :environment do |_t, args|
     old_email_address = args[:old_email_address]
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -17,7 +17,29 @@ namespace :manage do
     end
   end
 
-  desc "Unsubscribe a single subscriber"
+  desc "Unsubscribe a subscriber from a single subscription"
+  task :unsubscribe_single_subscription, %i[email_address subscriber_list_slug] => :environment do |_t, args|
+    email_address = args[:email_address]
+    subscriber = Subscriber.find_by_address(email_address)
+    subscriber_list = SubscriberList.find_by(slug: subscriber_list_slug)
+    if subscriber.nil?
+      puts "Subscriber #{email_address} not found"
+    elsif subscriber_list.nil?
+      puts "Subscriber list #{subscriber_list_slug} not found"
+    elsif !(subscriber.subscriptions.pluck(:subscriber_list_id).include? subscriber_list.id)
+      puts "Subscriber #{email_address} does not appear to be signed up for #{subscriber_list_slug}"
+    else
+      active_subscriptions = Subscription.active.where(subscriber_list: subscriber_list, subscriber: subscriber)
+      if active_subscriptions.empty?
+        puts "Subscriber #{email_address} already unsubscribed from #{subscriber_list_slug}"
+      else
+        UnsubscribeService.subscription!(active_subscriptions.last, :unsubscribed)
+        puts "Unsubscribing from #{email_address} from #{subscriber_list_slug}"
+      end
+    end
+  end
+
+  desc "Unsubscribe a subscriber from all subscriptions"
   task :unsubscribe_single, [:email_address] => :environment do |_t, args|
     email_address = args[:email_address]
     subscriber = Subscriber.find_by_address(email_address)
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -71,7 +71,7 @@ namespace :manage do
   end
 
   desc "Unsubscribe a subscriber from all subscriptions"
-  task :unsubscribe_single, [:email_address] => :environment do |_t, args|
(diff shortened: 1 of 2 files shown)


lib/tasks/manage.rake
```diff
@@ -17,7 +48,33 @@ namespace :manage do
...
+             "#{subscriber_list_slug} at #{latest_subscription.ended_at} (#{latest_subscription.ended_reason})"
+      else
+        puts "Unsubscribing from #{email_address} from #{subscriber_list_slug}"
+        UnsubscribeService.subscription!(latest_subscription, :unsubscribed)
+      end
+    end
+  end
+
+  desc "Unsubscribe a subscriber from all subscriptions"
   task :unsubscribe_single, [:email_address] => :environment do |_t, args|
```
Rosa-Fox, line 78: Look good! The name of this pre-exisiting task could also be changed to `unsubscribe_all` to avoid confusion between the two.

lib/tasks/manage.rake
```diff
@@ -17,7 +17,33 @@ namespace :manage do
...
+    email_address = args[:email_address]
+    subscriber = Subscriber.find_by_address(email_address)
+    subscriber_list = SubscriberList.find_by(slug: subscriber_list_slug)
+    if subscriber.nil?
+      puts "Subscriber #{email_address} not found"
+    elsif subscriber_list.nil?
+      puts "Subscriber list #{subscriber_list_slug} not found"
+    elsif !(subscriber.subscriptions.map(&:subscriber_list_id).include? subscriber_list.id)
+      puts "Subscriber #{email_address} does not appear to be signed up for #{subscriber_list_slug}"
+    else
```
1pretz1, line 62: We could probably slim down this else block to whats below and do most of the work in a query:

```
active_subscriptions = Subscription.active.where(subscriber_list: subscriber_list, subscriber: subscriber)

puts "Unsubscribing from #{email_address} from #{subscriber_list_slug}"
active_subscriptions.each { |s| UnsubscribeService.subscription!(s, :unsubscribed) }
```

What do you think?

lib/tasks/manage.rake
```diff
@@ -1,6 +1,33 @@
...
+  task :view_subscriptions, [:email_address] => :environment do |_t, args|
+    email_address = args[:email_address]
+    subscriber = Subscriber.find_by_address(email_address)
+    abort("Cannot find any subscriber with email address #{email_address}.") if subscriber.nil?
+
+    results = subscriber.subscriptions.map do |subscription|
+      subscriber_list = SubscriberList.find(subscription.subscriber_list_id)
+      {
+        status: subscription.ended_at.present? ? "Inactive (#{subscription.ended_reason})" : "Active",
+        subscriber_list: "#{subscriber_list.slug} (id: #{subscriber_list.id})",
```
1pretz1, line 14: Maybe also include the title of the subscriber list. This would be helpful as it would allow you to feed back to the user which lists they're subscribed to as each email they receive will have the title mentioned at the top of the email.

lib/tasks/manage.rake
```diff
@@ -1,6 +1,33 @@
...
+        frequency: subscription.frequency.to_s,
+        timeline: "Subscribed #{subscription.created_at}#{subscription.ended_at.present? ? ", Ended #{subscription.ended_at}" : ''}",
+      }
+    end
+    columns = results.first.each_with_object({}) do |(col, _), h|
+      heading = col.to_s.humanize
```
1pretz1, line 20: Could you explain why we're doing this below, my brain is struggling to parse this bit of code? 😄

lib/tasks/manage.rake
```diff
@@ -21,6 +21,10 @@ namespace :manage do
       h[col] = { label: heading, width: [results.map { |g| g[col].size }.max, label.size].max }
     end
```
1pretz1, line 23: Nice!

lib/tasks/manage.rake
```diff
@@ -21,6 +21,10 @@ namespace :manage do
...
 
+    # Example output:
+    # | Status                  | SubscriberList                  | Frequency | Timeline                                                       |
+    # | Inactive (unsubscribed) | test-foo (id: 58507)            | daily     | Subscribed 2020-04-18 15:21:20 +0100, Ended 2020-05-12 11:41:04 +0100 |
+    # | Active                  | recruiting-foo-staff (id: 9182) | weekly    | Subscribed 2019-06-10 13:48:43 +0100                           |
```
1pretz1, line 27: Nice example!

lib/tasks/manage.rake
ChrisBAshton: I did originally do that but it made the table very wide. Happy to add it back in!

lib/tasks/manage.rake
ChrisBAshton: Converts the key (e.g. `subscriber_list`) to a heading we can display (e.g. `SubscriberList`).
The StackOverflow I took inspiration from had a hardcoded mapping instead (e.g. `subscriber_list: "SubscriberList"` but this added bloat and felt unnecessary.

lib/tasks/manage.rake
ChrisBAshton: Nice 👍 I like it! I'll just do a `.last` to make it `active_subscription` (singular), as there should never be more than 1 active subscription to the same subscriber list.

ChrisBAshton: This gets us from 12 lines to 9, although we lose the end date & reason:

```ruby
    else
      active_subscriptions = Subscription.active.where(subscriber_list: subscriber_list, subscriber: subscriber)
      else
        UnsubscribeService.subscription!(active_subscriptions.last, :unsubscribed)
        puts "Unsubscribing from #{email_address} from #{subscriber_list_slug}"
      end
    end
```

What do you think? Worth doing?

lib/tasks/manage.rake
ChrisBAshton: I agree with you, although it does mean updating the [dev doc instructions](https://docs.publishing.service.gov.uk/manual/manage-email-subscribers.html#unsubscribe-a-subscriber-from-all-emails) too.

I'd better raise a PR there, ready, so we can merge them both at the same time..

ChrisBAshton: Actually I've gone one better and moved those docs into this repo!

lib/tasks/manage.rake
1pretz1: Getting the 'last' active subscription sounds fine as long as it's not possible to have multiple otherwise we could risk not unsubscribing someone.

1pretz1: Yup it looks an improvement to me 👍I'm not sure the reason and date are important as if you're running this task you're unlikely to care about ended subscriptions anyway.

ChrisBAshton: Done 👍 Agreed.

PR #1239, opened 2020-05-20
Only include lint task in a development environment

thomasleese:
This is similar to #1045 as without it deploys aren't currently possible.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,7 +5,7 @@ require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
 
-# RSpec shoves itself into the default task without asking, which confuses the ordering.
-# https://github.com/rspec/rspec-rails/blob/eb3377bca425f0d74b9f510dbb53b2a161080016/lib/rspec/rails/tasks/rspec.rake#L6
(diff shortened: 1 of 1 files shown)


lib/tasks/lint.rake
```diff
@@ -1,4 +1,8 @@
-desc "Run RuboCop"
-task lint: :environment do
-  sh "bundle exec rubocop --format clang"
+unless Rails.env.production?
+  desc "Run RuboCop"
+  task lint: :environment do
+    sh "bundle exec rubocop --format clang"
+  end
+
+  task(:default).prerequisites << task(:lint)
```
benthorner, line 7: We're trying to avoid this behaviour, as it surprising when adding something else into the `default` task; the order will depend on the load order of these files.

We could just add `unless Rails.env.production?` [here](https://github.com/alphagov/email-alert-api/pull/1239/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffL10).

thomasleese: Yeah, I realised that's a less nice solution. I've amended the commit which basically wraps that in `unless Rails.env.production?`.

PR #1241, opened 2020-05-20
Update email copy for migrating subscriber lists

chao-xian:
Copy changes from our Content Designer to match messaging in other parts of journey.

Trello card: https://trello.com/c/QpH7iXqW

### DIFF
--- a/lib/subscriber_list_mover.rb
+++ b/lib/subscriber_list_mover.rb
@@ -70,11 +70,13 @@ class SubscriberListMover
     bulk_move_template = <<~BODY.freeze
       Hello,
 
-      You've subscribed to get email alerts about #{list_title} information.
+      You've subscribed to get emails about #{list_title}.
+      
+      GOV.UK is changing the way we send emails, so you may notice a difference in the number and type of updates you get.
+      
+      ## If you're getting too many emails
 
-      GOV.UK is changing the type of email alerts you get to make sure you find out when any #{list_title} guidance is added or updated.
-      You can [change how often you get updates or unsubscribe](#{email_redirect}) from GOV.UK email alerts.
-
+      You can [manage your subscription](#{email_redirect}) to choose how often you want to receive emails.
       Thanks,
       GOV.UK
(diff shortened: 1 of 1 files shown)


lib/subscriber_list_mover.rb
```diff
@@ -69,12 +69,12 @@ def email_change_to_subscribers(source_subscriber_list)
 
     bulk_move_template = <<~BODY.freeze
       Hello,
-
-      You've subscribed to get email alerts about #{list_title} information.
-
-      GOV.UK is changing the type of email alerts you get to make sure you find out when any #{list_title} guidance is added or updated.
-      You can [change how often you get updates or unsubscribe](#{email_redirect}) from GOV.UK email alerts.
-
+      You've subscribed to get emails about #{list_title}.
```
kevindew, line 72: As this is parsed as markdown (well a rather restricted one) you may well find that, due to the lack of empty line, that this renders as `<p>Hello, You've subscribed to get emails about #{list_title}.</p>`

lib/subscriber_list_mover.rb
```diff
@@ -69,12 +69,12 @@ def email_change_to_subscribers(source_subscriber_list)
...
+      
+      GOV.UK is changing the way we send emails, so you may notice a difference in the number and type of updates you get.
+      
+      If you're getting too many emails
+      You can [manage your subscription](#{email_redirect}) to choose how often you want to receive emails.
```
kevindew, line 77: Looks like we've got an Grammar issue here. I'm not sure if `If you're getting too many emails` is missing the end of a sentence or if the "You" is meant to be lowercase?

lib/subscriber_list_mover.rb
chao-xian: Good spot thanks!

lib/subscriber_list_mover.rb
chao-xian: Ah it was supposed to be a heading!

PR #1247, opened 2020-05-27
Add comment on migration rake task

chao-xian:


lib/tasks/manage.rake
```diff
@@ -91,6 +91,8 @@ namespace :manage do
     end
   end
 
+  # TODO: this currently has an issue where emails are sent before
+  #       the task completes and causes confusion.
```
vanitabarrett, line 95: Can we make it clear that the problem is **emails are sent before migrating users to the new subscriber list** rather than just being sent "before the task completes"?

chao-xian: @vanitabarrett [Done](https://github.com/alphagov/email-alert-api/pull/1247/commits/f74fd5ed08f0cab28f4abf4d494b3c644a169b4e)

PR #1250, opened 2020-05-29
test workaround to speed up deletion of emails during anonymization process

fredericfran-gds:


fredericfran-gds:
no longer needed

PR #1261, opened 2020-06-04
Fix sending emails to migrated subscribers

chao-xian:
The rake task for migrating subscribers to a different subscriber list
sent emails before the migration was completed. Especially for large
lists of users, this caused a confusing experience as emails were
received while the migration was still happening.

This change allows to to generate the emails to send, make the
migration, and then send the emails after the migration has
completed.

https://trello.com/c/HYqY4ZHw

### DIFF
--- a/app/services/bulk_email_sender_service.rb
+++ b/app/services/bulk_email_sender_service.rb
@@ -1,6 +1,6 @@
 class BulkEmailSenderService
-  def self.call(subject:, body:, subscriber_lists:, queue: :delivery_immediate)
-    email_ids = BulkEmailBuilder.call(subject: subject, body: body, subscriber_lists: subscriber_lists).ids
+  def self.call(bulk_email_builder:, queue: :delivery_immediate)
+    email_ids = bulk_email_builder.ids
 
     email_ids.each do |email_id|
       DeliveryRequestWorker.perform_async(email_id, queue)
--- a/spec/lib/tasks/bulk_spec.rb
+++ b/spec/lib/tasks/bulk_spec.rb
@@ -1,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe "bulk" do
+  describe "email" do
+    before do
+      Rake::Task["bulk:email"].reenable
+    end
+
+    around(:each) do |example|
+      ClimateControl.modify(SUBJECT: "subject", BODY: "body") do
+        example.run
+      end
+    end
+
+    it "sends emails to a subscription list" do
+      subscriber_list = create(:subscriber_list)
+
+      expect(BulkEmailSenderService).to receive(:call).with(
+        bulk_email_builder: BulkEmailBuilder.call(
+          subject: "subject",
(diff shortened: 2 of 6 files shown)


app/services/bulk_email_sender_service.rb
```diff
@@ -1,6 +1,6 @@
 class BulkEmailSenderService
-  def self.call(subject:, body:, subscriber_lists:, queue: :delivery_immediate)
-    email_ids = BulkEmailBuilder.call(subject: subject, body: body, subscriber_lists: subscriber_lists).ids
+  def self.call(bulk_email_builder:, queue: :delivery_immediate)
```
benthorner, line 2: We also need to change [this task](https://github.com/alphagov/email-alert-api/blob/master/lib/tasks/bulk.rake). I was tempted to remove it, but [it's an official thing](https://docs.publishing.service.gov.uk/manual/bulk-email.html). Maybe we should add a test for it, so it won't break in the future.

chao-xian: Ah good spot - thanks! Yes will add a test.

chao-xian: @benthorner [added a test for this rake task](https://github.com/alphagov/email-alert-api/pull/1261/commits/4c9951f5050a65629cb53b5138733389fe7fb19f)

spec/lib/tasks/bulk_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe "bulk" do
+  describe "email" do
+    before do
+      Rake::Task["bulk:email"].reenable
+      allow(ENV).to receive(:fetch).with("SUBJECT").and_return("subject")
```
benthorner, line 7: This repo [has ClimateControl](https://github.com/alphagov/email-alert-api/blob/master/Gemfile#L29), so we should use that here.

chao-xian: [Switched to ClimateControl](https://github.com/alphagov/email-alert-api/commit/2ee6d188f09d57f52f547629c8f5f9864a518752)

spec/lib/tasks/bulk_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+RSpec.describe "bulk" do
+  describe "email" do
+    before do
+      Rake::Task["bulk:email"].reenable
+    end
+
+    it "sends emails to a subscription list" do
+      subscriber_list = create(:subscriber_list)
+
+      ClimateControl.modify(SUBJECT: "subject", BODY: "body") do
```
benthorner, line 12: (You can also do this in an `around` block.)

chao-xian: [👍 ](https://github.com/alphagov/email-alert-api/commit/500fd67f01c9d5005dd49870459ba8c195ef0b05)

PR #1271, opened 2020-06-09
Optimise finding a content change for an email

richardTowers:
The thinking here is that `first` will attempt to do some ordering based on the ID (which is a UUID, so the order has no meaning). `take` will simply go off and get the first one it finds.

I don't think we need to sort the `content_changes`, as we're excluding digests so they should all have approximately the same `created_at` (and we were sorting them by uuid before, so they've always been random).

A bit of explain analyze on staging suggests this might be a significant performance win (but I could be wrong - I'm not good with databases):

```sql
EXPLAIN ANALYZE
SELECT "content_changes".*
FROM "content_changes"
WHERE "content_changes"."id" IN
(
    SELECT "subscription_contents"."content_change_id"
    FROM "subscription_contents"
    WHERE "subscription_contents"."email_id" = 'REDACTED'
)
ORDER BY "content_changes"."id" ASC
LIMIT 1;
```

```
                                                                              QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=22.16..3393.12 rows=1 width=1314) (actual time=82.668..82.670 rows=1 loops=1)
   ->  Nested Loop Semi Join  (cost=22.16..2251820.27 rows=668 width=1314) (actual time=82.666..82.666 rows=1 loops=1)
         Join Filter: (content_changes.id = subscription_contents.content_change_id)
         Rows Removed by Join Filter: 661752
         ->  Index Scan using content_changes_pkey on content_changes  (cost=0.42..154618.22 rows=209036 width=1314) (actual time=0.005..4.430 rows=5211 loops=1)
         ->  Materialize  (cost=21.74..2663.01 rows=668 width=16) (actual time=0.000..0.006 rows=127 loops=5211)
               ->  Bitmap Heap Scan on subscription_contents  (cost=21.74..2659.67 rows=668 width=16) (actual time=0.022..0.043 rows=127 loops=1)
                     Recheck Cond: (email_id = '...'::uuid)
                     Heap Blocks: exact=4
                     ->  Bitmap Index Scan on index_subscription_contents_on_email_id  (cost=0.00..21.58 rows=668 width=0) (actual time=0.019..0.020 rows=127 loops=1)
                           Index Cond: (email_id = '...'::uuid)
 Planning time: 0.292 ms
 Execution time: 82.705 ms
(13 rows)
```

```sql
EXPLAIN ANALYZE
SELECT "content_changes".*
FROM "content_changes"
WHERE "content_changes"."id" IN
(
    SELECT "subscription_contents"."content_change_id"
    FROM "subscription_contents"
    WHERE "subscription_contents"."email_id" = 'REDACTED'
)
-- NOTE: order by is removed
LIMIT 1;
```

```
                                                                              QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=2661.76..2667.95 rows=1 width=1314) (actual time=0.074..0.075 rows=1 loops=1)
   ->  Nested Loop  (cost=2661.76..6801.08 rows=668 width=1314) (actual time=0.073..0.073 rows=1 loops=1)
         ->  HashAggregate  (cost=2661.34..2666.93 rows=559 width=16) (actual time=0.066..0.066 rows=1 loops=1)
               Group Key: subscription_contents.content_change_id
               ->  Bitmap Heap Scan on subscription_contents  (cost=21.74..2659.67 rows=668 width=16) (actual time=0.024..0.039 rows=127 loops=1)
                     Recheck Cond: (email_id = '...'::uuid)
                     Heap Blocks: exact=4
                     ->  Bitmap Index Scan on index_subscription_contents_on_email_id  (cost=0.00..21.58 rows=668 width=0) (actual time=0.020..0.020 rows=127 loops=1)
                           Index Cond: (email_id = '...'::uuid)
         ->  Index Scan using content_changes_pkey on content_changes  (cost=0.42..7.39 rows=1 width=1314) (actual time=0.005..0.005 rows=1 loops=1)
               Index Cond: (id = subscription_contents.content_change_id)
 Planning time: 0.277 ms
 Execution time: 0.127 ms
(13 rows)
```

### DIFF
--- a/app/services/metrics_service.rb
+++ b/app/services/metrics_service.rb
@@ -64,7 +64,7 @@ class MetricsService
     def store_time_to_send_content_change(email, time)
       return if SubscriptionContent.where(email: email).digest.exists?
 
-      content_change = ContentChangesForEmailQuery.call(email).first
+      content_change = ContentChangesForEmailQuery.call(email).take
       return unless content_change
 
       difference = (time - content_change.created_at) * 1000


kevindew:
Nice work 🚀  that is an amazing reduction in time. TIL on the `take` method too. Looks super handy for this project.

It's bananas how much of a difference the `ORDER BY` seems to make given we never anticipate it to ever return more than one record so would never need any ordering. 

Since the query plan is still a bit grim in what it expects to do, do we think this performance might fluctuate somewhat?

The approaches Chris and I looked at involved leveraging the [`SubscriptionContent.where(email: email).digest.exists?`](https://github.com/alphagov/email-alert-api/pull/1271/files#diff-0f17b24554bd2aa9302470840a10845eR65) query to avoid the need for a join altogether and switch the created_at pull to be a primary key lookup.

Roughly it comes out something like:

```
      content_change_id = SubscriptionContent.where(email: email).immediate.pick(:content_change_id)
      return unless content_change_id

      created_at = ContentChange.where(id: content_change_id).pick(:created_at)
      return unless created_at

      difference = (time - created_at) * 1000
      namespace = "content_change_created_to_first_delivery_attempt"
      timing(namespace, difference)
```

Plan wise the two queries come out at:

```
EXPLAIN ANALYZE SELECT content_change_id FROM subscription_contents WHERE subscription_contents.email_id = 'cbd0acc0-fb98-424d-8415-bb50ba5e117f' AND "subscription_contents"."digest_run_subscriber_id" IS NULL LIMIT 1;

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.57..9.39 rows=1 width=16) (actual time=0.012..0.012 rows=1 loops=1)
   ->  Index Scan using index_subscription_contents_on_email_id on subscription_contents  (cost=0.57..2507.16 rows=284 width=16) (actual time=0.012..0.013 rows=1 loops=1)
         Index Cond: (email_id = 'cbd0acc0-fb98-424d-8415-bb50ba5e117f'::uuid)
         Filter: (digest_run_subscriber_id IS NULL)
 Planning time: 0.059 ms
 Execution time: 0.022 ms
(6 rows)
```

and

```
EXPLAIN ANALYZE SELECT "content_changes"."created_at" FROM "content_changes" WHERE "content_changes"."id" = '8521d24b-79ec-4d83-b2ad-62f3ee013a02' LIMIT 1;
--------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..8.44 rows=1 width=8) (actual time=0.015..0.015 rows=1 loops=1)
   ->  Index Scan using content_changes_pkey on content_changes  (cost=0.42..8.44 rows=1 width=8) (actual time=0.013..0.013 rows=1 loops=1)
         Index Cond: (id = '8521d24b-79ec-4d83-b2ad-62f3ee013a02'::uuid)
 Planning time: 0.055 ms
 Execution time: 0.026 ms
(5 rows)
```

---

Tangentially I feel like at some point in this week we should move this query lookup out of the metrics service (if we need it, I know Chris is investigating removing need for a query altogether). It seems rather surprising that a call to a `MetricsService` talks to the db.

richardTowers:
Yeah, that approach looks sensible to me too - probably a bit easier to understand than this monstrosity.

I did a bit of research around the `first` / `take` difference - the ordering on first is useful when you're using auto-increment ids, as the ordering will get you the "first" entry that was inserted.  It's pointless if you're using `uuids` though. I think that might be a good general lesson for us, as we use uuids a lot.

Strongly agree with moving the query outside of the metric service too - it's super surprising for a metric write to be messing with the DB, which I think is a big part of why we didn't spot this performance issue earlier.

PR #1277, opened 2020-06-11
Remove import_ignoring_duplicates

cbaines:
Switch to using import! from the activerecord-import gem. We're now
using PostgreSQL 9.6, so we should be able to use the gem directly, as
described in the code comment.

I've updated the tests for this area, just to help demonstrate that
the behaviour is consistent. Maybe the tests are irrelevant now.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 
 library("govuk")
 
-node("postgresql-9.3") {
+node("postgresql-9.6") {
   govuk.setEnvar("TEST_DATABASE_URL", "postgresql://email-alert-api:email-alert-api@localhost/email-alert-api_test")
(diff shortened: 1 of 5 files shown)


benthorner:
I'm wondering if this is related to some test failures I'm seeing:

```
    ​# ./spec/workers/process_message_and_generate_emails_worker_spec.rb:47:in `block (4 levels) in <main>'
    ​# ------------------
    ​# --- Caused by: ---
    ​# PG::UniqueViolation:
    ​#   ERROR:  duplicate key value violates unique constraint "index_subscription_contents_on_subscription_id_and_message_id"
```

benthorner:
And now it's fine after rebasing. Ignore me.

PR #1286, opened 2020-06-15
Match content changes in a worker

kevindew:
This change is to resolve a long running inconsistency between content
changes and messages. Both of these need to match subscription lists to
their content model. Message has performed this action asynchronously in
the worker whereas ContentChange has performed this during the HTTP
request.

This is an action that is unnecessary for the HTTP request and thus
risks these being slower than necessary (From looking over last 24 hours
I can see the request maxed at 5s and averages 1.1s).

It also adds a check in both of the MatchContentChangeGenerationService
and MatchedMessageGenerationService to handle cases where the service
is called after already matching. This serves to allow us to transition from
web request to worker and also serves as a means to not re-run this when
resuming any retried workers.

### DIFF
--- a/app/services/matched_content_change_generation_service.rb
+++ b/app/services/matched_content_change_generation_service.rb
@@ -8,7 +8,7 @@ class MatchedContentChangeGenerationService
   end
 
   def call
-    # if we already have any records already we expect the process completed
+    # if we have records already, then we expect the process completed
     # successfully previously since the insert is an atomic operation
     return if MatchedContentChange.exists?(content_change: content_change)
 
--- a/app/services/matched_content_change_generation_service.rb
+++ b/app/services/matched_content_change_generation_service.rb
@@ -1,5 +1,5 @@
 class MatchedContentChangeGenerationService
-  def initialize(content_change:)
+  def initialize(content_change)
     @content_change = content_change
   end
 
--- a/app/services/matched_content_change_generation_service.rb
+++ b/app/services/matched_content_change_generation_service.rb
@@ -8,6 +8,10 @@ class MatchedContentChangeGenerationService
   end
 
   def call
+    # if we already have any records already we expect the process completed
+    # successfully previously since the insert is an atomic operation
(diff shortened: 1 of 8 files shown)


app/services/matched_content_change_generation_service.rb
```diff
@@ -8,6 +8,10 @@ def self.call(*args)
   end
 
   def call
+    # if we already have any records we assume this process was completed
+    # previously and doesn't need to run again.
+    return if MatchedContentChange.exists?(content_change: content_change)
```
benthorner, line 13: It looks like the `import!` operation isn't atomic - [not least because it maps over SubscriberLists](https://github.com/alphagov/email-alert-api/blob/master/app/services/matched_content_change_generation_service.rb#L26). Is there a risk here that we could abort `import!` halfway through, and miss out a load of subscribers on a re-run?

kevindew: I believe it is atomic. It does the map before trying to insert and then inserts all of those records in a single query so it either all succeeds or doesn't.

benthorner: Ah, OK. I see now from the docs that it should be a single `insert` statement, since we haven't set a `batch_size`. Could we tweak the comment slightly to explain? (The "assume" part was making me a little nervous.)

kevindew: Yup sure, I've got another small tweak to make too. Will ping you when it's up.

kevindew: I've changed the copy. I decided against my other change as being a bit too edge case.

app/services/matched_content_change_generation_service.rb
```diff
@@ -8,6 +8,10 @@ def self.call(*args)
   end
 
   def call
+    # if we already have any records already we expect the process completed
```
benthorner, line 11:
```suggestion
    # if we have records already, then we expect the process completed
```

Entirely optional 😄

PR #1287, opened 2020-06-15
Expand Living In Europe report

vanitabarrett:
## What
Expands the existing report for Living In guides to:

- Provide more detailed stats (new subscribers, unsubscribers, and a breakdown of immediate/daily/weekly)
- Ability to provide a date so the data is gathered from that point
- Adds Romania to the list of European countries

## Why
The FCO have requested this data on a weekly basis. As we were previously using this task to provide them with data, it makes sense to expand the existing task rather than create a new one.

### DIFF
--- a/lib/tasks/report.rake
+++ b/lib/tasks/report.rake
@@ -61,9 +61,9 @@ namespace :report do
     Reports::DataExporter.new.export_csv_from_slugs_at(args.date, args.extras)
   end
 
-  desc "Export the number of subscriptions for the 'Living in' taxons for European countries"
-  task csv_from_living_in_europe: :environment do
-    Reports::DataExporter.new.export_csv_from_living_in_europe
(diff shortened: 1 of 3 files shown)


lib/tasks/report.rake
```diff
@@ -61,9 +61,9 @@ namespace :report do
     Reports::DataExporter.new.export_csv_from_slugs_at(args.date, args.extras)
   end
 
-  desc "Export the number of subscriptions for the 'Living in' taxons for European countries"
-  task csv_from_living_in_europe: :environment do
-    Reports::DataExporter.new.export_csv_from_living_in_europe
+  desc "Export the number of subscriptions for the 'Living in' taxons for European countries as of a given date (format: 'yyyy-mm-dd')"
+  task :csv_from_living_in_europe, [:date] => :environment do |_, args|
```
benthorner, line 65: Can we update [the test](https://github.com/alphagov/email-alert-api/blob/master/spec/lib/tasks/report_spec.rb#L84) as well? Kind of surprised it's passing 🤔

PR #1290, opened 2020-06-15
Consolidate and reprioritise Sidekiq queues

benthorner:
Related: https://github.com/mperham/sidekiq/issues/3709

This fixes a bug I noticed while investigating something.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -3,18 +3,18 @@
 :concurrency: 32
 :logfile: ./log/sidekiq.json.log
 :timeout: 4
-
 # Any changes to the queues should be reflected in `docs/queues.md`.
+# See https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
 :queues:
-  - delivery_transactional
   - [delivery_immediate_high, 10]
+  - [delivery_transactional, 7]
+  - [delivery_immediate_high, 6]
   - [delivery_immediate, 5]
   - [process_and_generate_emails, 4]
   - [default, 3]
   - [delivery_digest, 2]
   - [email_generation_digest, 1]
-  - cleanup
-
+  - [cleanup, 1]
 :schedule:
   daily_digest_initiator:
     cron: '30 8 * * * Europe/London' # every day at 8:30am
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -5,15 +5,15 @@
 :timeout: 4
 # Any changes to the queues should be reflected in `docs/queues.md`.
 # See https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
+# Use powers of 2: higher priority groups are checked twice as often.
 :queues:
-  - [delivery_immediate_high, 10]
-  - [delivery_transactional, 7]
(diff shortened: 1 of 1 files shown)


config/sidekiq.yml
```diff
@@ -3,15 +3,15 @@
...
-  - [delivery_immediate_high, 10]
-  - [delivery_immediate, 5]
-  - [process_and_generate_emails, 4]
-  - [default, 3]
-  - [delivery_digest, 2]
-  - [email_generation_digest, 1]
-  - cleanup
+:queues: # See https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
+  - [delivery_transactional, 4]
+  - [delivery_immediate_high, 4]
```
kevindew, line 8: Do we want these two to be 10 still so that, if there are delivery_immediate_high, it does them at a more rapid rate?

As I understand here we've switched from checking delivery_immediate_high from 10 times as frequent as cleanup to only 4 times so would mean they end up less prioritised.

benthorner: Yeah, that's true. I was thinking the "10 times" setting was somewhat arbitrary, and it wasn't clear that "delivery_transactional" needed it as well. I think this indicates a general problem with using these weights, and I doubt we actually need them.

benthorner:
Closing in favour of https://github.com/alphagov/email-alert-api/pull/1291.

config/sidekiq.yml
benthorner: So here we are again! I think we should keep the numbers simple, unless there's a empirical reason why a specific number is needed. Perhaps a simpler formula would be to use powers of two, so it's clearer expressed as "twice as fast". How about that?

kevindew: Sure, that sounds good.

@thomasleese do you know the logic for the use of 10? was there anything particular?

thomasleese: I think my reasoning was that 10 is double the normal priority queue weight (5), so would be picked up at twice the rate of that queue. Other than that I don't think there was any particular reason why 10 specifically was chosen.

We were seeing normal priority and high priority emails being picked at a similar rate so we wanted a number much higher than the normal priority weighting.

benthorner: Thanks, that's good to know. I've updated the last commit to use powers of 2 and explain the reasoning. Could you take another look and let me know what you think?

config/sidekiq.yml
```diff
@@ -3,18 +3,17 @@
 :concurrency: 32
 :logfile: ./log/sidekiq.json.log
 :timeout: 4
-
 # Any changes to the queues should be reflected in `docs/queues.md`.
-:queues:
-  - delivery_transactional
-  - [delivery_immediate_high, 10]
-  - [delivery_immediate, 5]
+# See https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
```
thomasleese, line 7: I think we've lost the `queues:` key somewhere in this diff...

benthorner: Oops. It's back. This is "one of those" PRs, isn't it.

PR #1291, opened 2020-06-16
Replace Sidekiq queue weights with priorities

benthorner:
Previously we had an incompatible mixture of ordering [1] and weights,
which were often set arbitrarily and misread as priorities [2]. This
removes all the weights, so that the configuration simply prioritises
the queues based on the order they are listed.

This also groups similar queues together: for delivery and generation;
we should prioritise digest email above cleanup and 'default' jobs.

[1]: 9473c38a4e7eb989cc39c84fd15fe9e0ffc4c136
[2]: 21dcb06850bb6ccbe857bc04d05f5b476bba96ea

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -3,18 +3,17 @@
 :concurrency: 32
 :logfile: ./log/sidekiq.json.log
 :timeout: 4
-
 # Any changes to the queues should be reflected in `docs/queues.md`.
+# https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
 :queues:
   - delivery_transactional
-  - [delivery_immediate_high, 10]
-  - [delivery_immediate, 5]
-  - [process_and_generate_emails, 4]
-  - [default, 3]
-  - [delivery_digest, 2]
-  - [email_generation_digest, 1]
+  - delivery_immediate_high
+  - delivery_digest
+  - delivery_immediate
+  - process_and_generate_emails
+  - email_generation_digest
+  - default
   - cleanup
-
 :schedule:
   daily_digest_initiator:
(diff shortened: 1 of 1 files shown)


config/sidekiq.yml
```diff
@@ -3,18 +3,17 @@
...
-  - [default, 3]
-  - [delivery_digest, 2]
-  - [email_generation_digest, 1]
+  - delivery_immediate_high
+  - delivery_digest
+  - delivery_immediate
+  - process_and_generate_emails
+  - email_generation_digest
+  - default
   - cleanup
```
kevindew, line 16: Removing the queue weights would have the side effect that any backlog of emails would block any new emails being created. So we'd lose the current behaviour where we can still process a content change despite a backlog of emails to send.

config/sidekiq.yml
```diff
@@ -3,18 +3,17 @@
...
-  - [delivery_immediate_high, 10]
-  - [delivery_immediate, 5]
-  - [process_and_generate_emails, 4]
-  - [default, 3]
-  - [delivery_digest, 2]
-  - [email_generation_digest, 1]
+  - delivery_immediate_high
+  - delivery_digest
+  - delivery_immediate
+  - process_and_generate_emails
```
thomasleese, line 13: I think this means that if there are any non high priority emails in the queue to be sent, no content changes will be processed even if that content change is a high priority one (for travel advice for example). The weighting means we can still generate emails for high priority content as it comes in, and they can sort of skip the queue and go straight to the high priority queue (which delays non high priority content, but that's okay).

benthorner:
Oh dear, you're right @kevindew, @thomasleese. This comes from not reading all the documentation. Back to the original, then...

https://github.com/alphagov/email-alert-api/pull/1290

PR #1293, opened 2020-06-17
Combine metric collection workers into one

benthorner:
Previously we had a 'X' Worker for each model we wanted to collect periodic
metrics about. These names were confusing because they implied some actual
processing associated with 'X', and introduced noise into the list of
workers. This moves the code to a new 'Metrics' namespace in 'lib/', and
invokes each 'Exporter' from a single worker.


,
```diff
@@ -1,4 +1,4 @@
-RSpec.describe StatusUpdateWorker do
```
thomasleese, line 1: This file seems to have been renamed to `,` 😬

benthorner: Oh yeah, I remember forgetting to delete the comma file. (Like a dotfile, but more suspensful.) Updated.

config/sidekiq.yml
```diff
@@ -34,18 +34,6 @@
   digest_run_completion_marker:
     every: '1m'
```
thomasleese, line 35: While we're hear, I do wonder if we can reduce the frequency of some of these jobs. Perhaps metrics collection can stay at `1m`, but I can't imagine we'd lose much if digests run every 5 minutes instead.

benthorner: That's true. I noticed that previously and thought it a bit weird. That said, I'm going to leave it as-is for now. We've got a stream of work to stabilise the system and we'll be looking at what we can do to make it less surprising as part of that.

PR #1295, opened 2020-06-18
Increase rate limit to 21,000

vanitabarrett:
Trello: https://trello.com/c/OfF8KVKq/338-increase-the-notify-rate-limit-back-to-25-26k-min

We previously temporarily reduced the rate limit to give Notify more breathing space to an investigate an issue they were seeing.
They've given us the go-ahead to increase again.

Our original limit was 360/s (21,600). This was weirdly just above Notify's limit, when really we want it to be just under or the same. Setting to 350/s for now so it's the same as Notify's. We can chat about getting Notify to increase things on their side in the future once they've confirmed the issue they were seeing has been resolved.

### DIFF
--- a/app/workers/delivery_request_worker.rb
+++ b/app/workers/delivery_request_worker.rb
@@ -50,8 +50,8 @@ class DeliveryRequestWorker
   end
 
   def rate_limit_threshold
-    per_minute_to_allow_250_per_second = "15000"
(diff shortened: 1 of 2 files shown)


yahoopete:
Will confirm it's good to go shortly.... please don't deploy until then 🙏

yahoopete:
Release the hounds.

PR #1296, opened 2020-06-18
Improve resend troubleshooting tasks

thomasleese:
This expands on the existing tasks for resending failed emails to support providing either an email ID or a date range. The change came from a recent incident where a task accepting a date range would have been useful.

[Trello Card](https://trello.com/c/hBOhkenb/2026-3-investigate-improving-task-to-resend-email-when-theres-been-a-failure)

### DIFF
--- a/docs/support-tasks.md
+++ b/docs/support-tasks.md
@@ -117,12 +117,25 @@ To send a test email to an email address (doesn't have to be subscribed to anyth
 $ bundle exec rake deliver:deliver_to_test_email[<email_address>]
 ```
 
-## Resend emails
+## Resend failed emails
 
-This task takes an array of email ids and re-sends them.
+There are two Rake tasks available to resend emails which didn't send for
+whatever reason and ended up in the failed state. This is most useful after
+an incident to resend all emails that failed with a technical failure.
+
+### Using a date range
 
 ```bash
-bundle exec rake troubleshoot:resend_failed_emails:by_id[<email_one_id>, <email_two_id>]
+bundle exec rake 'troubleshoot:resend_failed_emails:by_date[<from_date>,<to_date>]'
+```
+
+The date format should be in ISO8601 format, for example `2020-01-01T10:00:00Z`.
+Depending on the number of emails to send, the Rake task can take a few minutes to run.
+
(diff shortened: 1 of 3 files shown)


lib/tasks/troubleshoot.rake
```diff
@@ -39,13 +39,26 @@ namespace :troubleshoot do
...
     end
   end
 end
+
+def resend_failed_emails(scope)
+  ids = scope.where(status: :failed).pluck(:id)
+  puts "Resending #{ids.length} emails"
+
+  ids.each do |id|
+    DeliveryRequestWorker.perform_async_in_queue(id, queue: :delivery_immediate)
```
1pretz1, line 62: Do we also potentially want to add an option to specify the `delivery_immediate_high` queue too?

docs/support-tasks.md
```diff
@@ -117,12 +117,25 @@ To send a test email to an email address (doesn't have to be subscribed to anyth
...
+```
+
+The date format should be in ISO8601 format, for example `2020-01-01T10:00:00Z`.
+Depending on the number of emails to send, the Rake task can take a few minutes to run.
+
+### Using email IDs
 
 ```bash
-bundle exec rake deliver:resend_failed_emails[<email_one_id>, <email_two_id>]
+bundle exec rake troubleshoot:resend_failed_emails:by_id[<email_one_id>, <email_two_id>]
```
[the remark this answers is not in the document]
thomasleese: Yes, good point! I've amended it and fixed it... by getting rid of the space and adding quotes.

lib/tasks/troubleshoot.rake
thomasleese: Thinking about it some more, I think it actually makes sense for these tasks to default to the high queue, as they're designed to be used to resolve an incident. I've made that change in 52e9d1b. I haven't added Rake tasks to add them to the normal priority queue as we can always add them later if we discover that we need them, but also it helps to avoid people having to decide which one to use in the context of an incident.

1pretz1: Sounds reasonable to me!

PR #1299, opened 2020-06-22
Gems cleanup

kevindew:
This removes a few gems that aren't needed and applies a consistent approach to the Gemfile.

Since all the gems for this repo are at the latest version I thought we could apply an approach of specifying no gem versions which makes this file a lot cleaner. I'm pondering this as a convention that could be applied across GOV.UK where we only specify a pinned version where we've got a reason to do so. It's early days on that idea though.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,27 +1,24 @@
 source "https://rubygems.org"
 
-gem "pg", "~> 1.2"
-gem "rails", "~> 6.0"
-
+gem "aws-sdk-s3"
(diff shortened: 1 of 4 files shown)


kevindew:
> Nice one - do we even need `startup.sh` anymore?

I still use it/them, I could live without but they are convenient. Could be a question for the dev tools survey?

benthorner:
@kevindew ah, fair enough. I think it's fine to leave them in - as long as they're adding some kind of value.

PR #1300, opened 2020-06-22
Replace Timecop gem

kevindew:
This extends the work in https://github.com/alphagov/email-alert-api/pull/1299 by doing a more involved replacement of the timecop gem. This gem is no longer necessary for a Rails application as ActiveSupport provides the same functionality in ActiveSupport::Testing::TimeHelpers. The Rails functionality is also preferable as it freezes time to the second, this resolves a common problem where time comparisons involving the database fail due to differences between Rubys and PostgreSQL time precision.

In doing this work I switched from a rails_helper.rb to a spec_helper.rb (as per [our Rails conventions](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#testing-utilities)).

I also had to apply this to some pretty grizzly feature tests. I did this in separate commits so they'd be easier to understand.


spec/features/unsubscribing_spec.rb
```diff
@@ -4,37 +4,31 @@
...
-      clear_any_requests_that_have_been_recorded!
+    email_data = expect_an_email_was_sent
+    id = extract_unsubscribe_id(email_data)
+    unsubscribe_from_subscriber_list(id, expected_status: 204)
+    clear_any_requests_that_have_been_recorded!
 
-      Timecop.freeze "2019-06-25 09:30:01" do
-        create_content_change(public_updated_at: "2019-06-25 10:00:00")
-      end
+    travel_to(Time.zone.today.midday) { create_content_change }
```
benthorner, line 23:
```suggestion
    travel_to(Time.zone.today.midnight) { create_content_change }
```

I think the purpose of the test is to check more recent content changes are excluded; this would be a little clearer if we could change the data, but keep the frozen runtime the same.

spec/features/create_and_deliver_a_weekly_digest_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+
+    stub_notify
+  end
+
+  scenario do
+    login_with_internal_app
+
+    # create two subscriber lists with different links
+    list_one_topic_id = "0eb5d0f0-d384-4f27-9da8-3f9e9b22a820"
+    list_one_taxon_id = "86db0cbd-a1f9-4218-b571-ca0550265e33"
```
benthorner, line 16:
```suggestion
    list_two_taxon_id = "86db0cbd-a1f9-4218-b571-ca0550265e33"
```

This appears to only be used for the second list.

kevindew: Oh good spot.

PR #1304, opened 2020-06-23
Refactor DeliveryRequestWorker and DeliveryRequestService

kevindew:
I wanted to continue the work of https://github.com/alphagov/email-alert-api/pull/1272 and found myself fighting against a number of inconsistencies and unconventional approaches. I decided for ease of review it'd be best to separate the refactor into a separate PR and open one that builds on this for the metrics change.

A brief summary of changes are:

- fixed a global exception being defined in the DeliveryRequestWorker file
- added a test for Sidekiq worker exhausted retries since that was untested
- upped the 30 second rate limit retry to 5 minutes as 30 seconds seemed suspiciously short for a job that has already been retrying for 2 hours 30 mins
- moved public methods in DeliveryRequestWorker to be private, removed unused instance variables/methods
- moved the DeliveryRequestService::EmailAddressOverrider to a separate file since it's unit tested
- gave DeliveryRequestService a more convention `.call` interface where all arguments are passed to the initialize method, moved various tested public methods that aren't used publicly to private

The easiest way to remove this is to do this commit by commit - not that much has actually changed despite what the overall diff looks like.

This work does leave two things dangling for follow up work. We're left with an unusued _queue argument to DeliveryRequestWorker, I resolve that in the next PR. The EmailAddressOverrider stuff all looks a bit of a mess with confusing environment variables and whitelist/blacklist terminology.


spec/services/delivery_request_service_spec.rb
```diff
@@ -1,212 +1,147 @@
...
-    it "records a #{status} statistic" do
-      expect(GovukStatsd).to receive(:increment).with("delivery_attempt.status.#{status.underscore}")
-      subject.call(email: email)
+      expect(described_class.call(email: email)).to be(true)
     end
-  end
 
-  describe "#call" do
-    let!(:email) { create(:email) }
+    it "returns false when the overrider can't provide an email address" do
```
benthorner, line 58:
```suggestion
    it "returns false when no email was sent due to filtering" do
```

It took a little effort to track down what the boolean is used for, so helpful to indicate this in the tests.

app/services/delivery_request_service.rb
```diff
@@ -5,31 +5,34 @@ class DeliveryRequestService
...
+    @attempt_id = SecureRandom.uuid
     @provider_name = config.fetch(:provider).downcase
-    @provider = PROVIDERS.fetch(provider_name)
     @subject_prefix = config.fetch(:email_subject_prefix)
     @overrider = EmailAddressOverrider.new(config)
   end
 
   def self.call(*args)
-    new.call(*args)
+    new(*args).call
```
benthorner, line 17: Out-of-scope: looking at the other services, we're quite close to being able to give them all a common superclass. Do you think an `ApplicationService` makes sense for this repo?

app/workers/delivery_request_worker.rb
```diff
@@ -10,7 +10,7 @@ class RateLimitExceededError < StandardError; end
       email_id = msg["args"].first
       queue = msg["args"].second
       GovukStatsd.increment("delivery_request_worker.rescheduled")
-      DeliveryRequestWorker.set(queue: queue).perform_in(30.seconds, email_id, queue)
+      DeliveryRequestWorker.set(queue: queue).perform_in(5.minutes, email_id, queue)
```
benthorner, line 13: I'm confused: why do we do this at all if Sidekiq is doing it for us?

app/services/delivery_request_service.rb
```diff
@@ -43,45 +46,36 @@ def call(email:)
 
 private
 
-  def call_provider(address, reference, email)
+  attr_reader :attempt_id, :email, :provider_name, :subject_prefix, :overrider
+
+  def address
+    @address ||= begin
+      overrider.destination_address(email.address).tap do |address|
+        next if address == email.address
```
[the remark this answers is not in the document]
kevindew: Fair, I'm quite used to them myself. After some juggling I decided to drop the logging behaviour as I'm not sure when any value is gained from it (or if it even works given it's a Rails logger running in a Sidekiq context).

spec/services/delivery_request_service_spec.rb
```diff
@@ -1,212 +1,147 @@
...
-
-        expect(MetricsService).to receive(:content_change_created_to_first_delivery_attempt)
-          .with(content_change.created_at, Time.now.utc)
+    context "when sending the email raises an error" do
+      before do
+        allow(default_provider).to receive(:call).and_raise("Ut oh")
+      end
 
-        subject.call(email: email)
+      it "sets the delivery attempt status to internal_failure" do
```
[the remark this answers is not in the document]
kevindew: Thanks, I've added in.

spec/services/delivery_request_service_spec.rb
```diff
@@ -1,212 +1,147 @@
...
+                              .and_return(:sending)
+      described_class.call(email: email)
     end
-  end
 
-  describe "#provider" do
-    it "defaults to the PseudoProvider" do
-      expect(subject.provider).to eq(PseudoProvider)
+    it "can send an email to a configured provider" do
+      config = EmailAlertAPI.config.email_service.merge(provider: "notify")
```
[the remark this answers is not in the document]
kevindew: Sure done, I was quite pleased to lose the shoutiness but alas makes sense.

spec/services/delivery_request_service_spec.rb
kevindew: Yeah, the boolean is a bit nasty. I've updated as per your suggestion.

app/services/delivery_request_service.rb
kevindew: Yeah, if it's consistent enough to do it that'd be a good idea to at least try prevent them going experimental again later.

app/workers/delivery_request_worker.rb
kevindew: This code only runs after sidekiq is given up. I'm assuming the original intent was that if Sidekiq is going to give up and it's given up because of our RateLimiting then we should keep on trying.

kevindew:
Thanks @benthorner, I've amended.

app/services/delivery_request_service.rb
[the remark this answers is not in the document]
benthorner: Ah, I did actually benefit from this logging recently, because it explained why no email was being sent: "Overriding email address foo@bar.com to <blank>". This is kind of covered by [the docs](https://docs.publishing.service.gov.uk/manual/load-test-email-alert-api.html#testing-without-notify) now, so I guess we're unlikely to get more benefit from it.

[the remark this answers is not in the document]
kevindew: Oh sure, that's interesting. One of my observations was that the logging doesn't capture the nil case which probably ends up looking confusing.

PR #1305, opened 2020-06-23
Record content change created at metrics without DB call

kevindew:
This continues the work of https://github.com/alphagov/email-alert-api/pull/1272 and is dependent on https://github.com/alphagov/email-alert-api/pull/1304

The recording of when a content change was created is still a rather spiky part in our metrics and in busy times can hit averages of 500ms with the upper 90s between 1 and 2 seconds. Applying this should reduce these figures so that this metric is much closer to being an insignificant part of the process.

Screenshot 2020-06-23 at 12 24 43

Doing this should allow us to reduce further a bottleneck in our email sending process allowing us to have better control over how many emails we're sending and maximising our resource usage.

### DIFF
--- a/app/workers/delivery_request_worker.rb
+++ b/app/workers/delivery_request_worker.rb
@@ -6,27 +6,35 @@ class DeliveryRequestWorker
   sidekiq_options retry: 9
 
   sidekiq_retries_exhausted do |msg, error|
-    if error.is_a?(RateLimitExceededError)
-      email_id = msg["args"].first
-      queue = msg["args"].second
-      GovukStatsd.increment("delivery_request_worker.rescheduled")
-      DeliveryRequestWorker.set(queue: queue).perform_in(5.minutes, email_id, queue)
-    end
+    next unless error.is_a?(RateLimitExceededError)
+
+    email_id, metrics = msg["args"]
+    GovukStatsd.increment("delivery_request_worker.rescheduled")
+    DeliveryRequestWorker.set(queue: msg["queue"])
+                         .perform_in(5.minutes, email_id, metrics)
   end
 
-  def perform(email_id, _queue)
+  def perform(email_id, metrics = {})
+    # existing jobs may have the second parameter set as a string, representing
+    # a queue and need their type changing. This can be removed once deployed
+    # and the queue is cleared
+    metrics = {} unless metrics.is_a?(Hash)
+
     check_rate_limit_exceeded
 
     email = MetricsService.delivery_request_worker_find_email do
       Email.find(email_id)
     end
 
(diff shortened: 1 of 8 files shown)


benthorner:
I'm struggling a bit with the motivation for this change:

- Our current performance is stable enough that it's not a team priority
- Our current work on SLOs may render this specific metric redundant
- Passing data through the pipeline seems to add more complexity
- I'm not convinced we get much diagnostic value from this metric

Sorry, I know it's frustrating to hear this. I'm tempted to completely overhaul a lot of this codebase, but I think that's lower value than redefining how we monitor it first. Some possible ways forward:

- We agree this is a significant improvement we need to get in now
- We pick it up again as part of our upcoming epics to reduce incidents

I'm open to other persuasive suggestions 😉.

kevindew:
Thanks for looking @benthorner.

So sounds like there are two areas: whether we want to continue recording this metric and whether we want to change the implementation of recording it.

On the former, it seems out of scope here to discuss whether to discuss recording this or not since we don't have a proposal on removing it and it seems to be wrapped up in personal views on its value. Myself, I've always found this to be one of our more useful metrics and would argue for keeping it while we still send a significant proportion of immediate emails. It's the only thing we have that shows the impacts of sustained high volume email sending. We can see queues go up, but what does that mean for users? This metrics tells us they may be waiting 4 hours for us to send the email (as per Friday).

On the latter, this is an operation we do millions of times a day and we can see it's a spiky part of our performance. Looking at Fridays data at the upper 90s of times we can frequently spend a whole extra seconds work on sending an email greater than the time requesting notify, and the metric recording can be a quarter of that. Looking again at Fridays peak data, we're not hitting our rate limits and are often a 100 emails a second below, so we're still not sending as fast as we can (of course this won't fully change that).

So I think my argument on these fronts for merging this would be: If we decide to remove the metric we can also remove this, it's not much code so hardly a bigger challenge. Then on the recording of metrics front, I'd suggest this arguably sets a better precedent that we'd want to inject metrics into this worker rather than add more responsibilities to it since it is a key area for bottlenecks.

app/workers/delivery_request_worker.rb
```diff
@@ -40,6 +48,15 @@ def check_rate_limit_exceeded
     raise RateLimitExceededError
   end
 
+  # Sidekiq uses JSON for a workers arguments, so richer objects are not
+  # available. This converts the scalar values to objects.
+  def parsed_metrics(metrics)
+    content_change_created_at = metrics["content_change_created_at"]
+      &.then { |t| Time.zone.iso8601(t) }
```
laurentqro, line 55: interesting idiom this `&.then` :)

PR #1311, opened 2020-06-29
Namespace rate limiter to support multiple tenants

benthorner:
https://trello.com/c/fCrWTqFj/372-incident-action-rate-limit-email-authentication-emails-by-address

Previously we assumed the rate limiter was unique to this app, with
no risk of name clashes. This prefixes the original key with the name
of this app, to avoid future, accidental name clashes [1].

[1]: https://github.com/alphagov/email-alert-frontend/pull/832

### DIFF
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -4,6 +4,6 @@ module Services
   end
 
   def self.rate_limiter
-    @rate_limiter ||= Ratelimit.new("deliveries")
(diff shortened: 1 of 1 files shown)


benthorner:
Note: this will cause us to temporarily lose track of our rate limit, so we should deploy this at a "low" period, when the limit is irrelevant.

PR #1312, opened 2020-06-30
Add comment around Notify's rate limit

1pretz1:
Trello:
https://trello.com/c/Hb7kL8pX/354-document-govuk-rate-limiting

### DIFF
--- a/app/workers/delivery_request_worker.rb
+++ b/app/workers/delivery_request_worker.rb
@@ -49,6 +49,8 @@ class DeliveryRequestWorker
     rate_limiter.add("delivery_request")
   end
 
+  # More information around the rate limit can be found here ->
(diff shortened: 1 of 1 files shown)


app/workers/delivery_request_worker.rb
```diff
@@ -49,6 +49,7 @@ def increment_rate_limiter
     rate_limiter.add("delivery_request")
   end
 
+  # Notify shares the same rate limit of 350r/s
```
benthorner, line 52: Instead of repeating the number, can we point to [the dev doc](https://github.com/alphagov/govuk-developer-docs/pull/2639) for more info? We should do that so any changes here are kept in sync.

1pretz1: Good point!

PR #1314, opened 2020-07-01
Fix typo in "non-existent"

brucebolt:
We were spelling "non-existent" as "non-existant", which is incorrect.


brucebolt:
Thanks @timblair.  I've also updated the ADR in case anyone is grepping specifically for this word.

PR #1315, opened 2020-07-02
Create worker to recover lost jobs

laurentqro:
Automatically re-enqueue content changes and messages that have been lost.

There is [a general problem with the queuing system](https://github.com/mperham/sidekiq/wiki/Error-Handling#process-crashes) that means in-progress jobs can be permanently lost, which leads to alerts and call-outs. We experience the most pain from the loss of [jobs to process content changes](https://github.com/alphagov/email-alert-api/blob/fdec17c576963f0f67e82bd02cd829ca4debe122/app/workers/process_content_change_worker.rb). We want to recover these lost jobs automatically.

https://trello.com/c/JD9CohHz/356-incident-action-automatically-requeue-content-changes-and-messages-that-have-been-lost

### DIFF
--- a/app/workers/recover_lost_jobs_worker.rb
+++ b/app/workers/recover_lost_jobs_worker.rb
@@ -1,0 +1,27 @@
+class RecoverLostJobsWorker
+  include Sidekiq::Worker
+
+  def perform(model, worker)
+    orphan_records(model).each do |record|
+      worker.perform_async(record.id)
+    end
+  end
+
+private
+
+  def orphan_records(model)
+    model.unprocessed.where("created_at <= ?", 1.hour.ago).select { |record| orphan?(record) }
+  end
+
+  def orphan?(record)
+    !queued?(record) || !retrying?(record)
+  end
+
+  def queued?(record)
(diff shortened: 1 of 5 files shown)


lib/recovery/recover_lost_jobs.rb
```diff
@@ -0,0 +1,8 @@
+module Recovery
```
benthorner, line 1: I think we should switch this around a bit: a single worker that calls a bunch of classes. We'll want to extend this to support other kinds of recovery.

app/workers/recover_lost_jobs_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+private
+
+  def orphan_records(model)
+    model.unprocessed.where("created_at <= ?", 1.hour.ago).select { |record| orphan?(record) }
+  end
+
+  def orphan?(record)
+    !queued?(record) || !retrying?(record)
```
benthorner, line 17: We should also check for processing jobs as well, especially for these ones.

laurentqro:
closed and replaced by https://github.com/alphagov/email-alert-api/pull/1318.

PR #1318, opened 2020-07-07
Requeue lost content change and email work

laurentqro:
## What and why

We had [an incident](https://docs.google.com/document/d/1GYiQr_06HsXO5Nm_zsoXCBRlhDJvTEzLsJX5rIeXqAc/edit) where travel advice emails were not sent due to the original Content Change jobs being lost from our queueing system. 

This is [a general problem with the queuing system](https://github.com/mperham/sidekiq/wiki/Error-Handling#process-crashes) across GOV.UK. Since we're not going to change the system, we need to make our software resilient to this flaw.

## Solution

This PR creates a sweeper job that requeues incomplete work that is over 1 hour old, for the following models:

* ContentChange (where `processed_at: nil`)
* Message (where `processed_at: nil`)

We chose 1 hour somewhat arbitrarily. That roughly matches the worst case time for some content change processing.

This PR makes use of the [sidekiq-unique-jobs](https://github.com/mhenrixon/sidekiq-unique-jobs#until-executed) gem to enforce job uniqueness, with `sidekiq_options lock: :until_executed`:

> Locks from when the client pushes the job to the queue. Will be unlocked when the server has successfully processed the job.

## Trello cards

https://trello.com/c/WYQDff8u/223-automatically-requeue-email-work-that-has-been-lost
https://trello.com/c/JD9CohHz/356-incident-action-automatically-requeue-content-changes-and-messages-that-have-been-lost

### DIFF
--- a/spec/workers/recover_lost_jobs_worker_spec.rb
+++ b/spec/workers/recover_lost_jobs_worker_spec.rb
@@ -1,0 +1,31 @@
+RSpec.describe RecoverLostJobsWorker do
+  describe ".perform" do
+    describe "ContentChange recovery" do
+      it "does not requeue incomplete content changes that are under 1-hour old" do
+        create(:content_change, created_at: 59.minutes.ago)
+        expect(ProcessContentChangeWorker).not_to receive(:perform_async)
+        subject.perform
+      end
+
+      it "requeues incomplete content changes that are over 1-hour old" do
+        create(:content_change, created_at: 1.hour.ago)
+        expect(ProcessContentChangeWorker).to receive(:perform_async)
+        subject.perform
+      end
+    end
+
+    describe "Message recovery" do
+      it "does not requeue incomplete messages that are under 1-hour old" do
+        create(:message, created_at: 59.minutes.ago)
+        expect(ProcessMessageWorker).not_to receive(:perform_async)
+        subject.perform
+      end
+
+      it "requeues incomplete messages that are over 1-hour old" do
+        create(:message, created_at: 1.hour.ago)
+        expect(ProcessMessageWorker).to receive(:perform_async)
+        subject.perform
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem "rails"
 gem "ratelimit"
 gem "redcarpet"
 gem "sidekiq-scheduler"
+gem "sidekiq-unique-jobs"
 gem "with_advisory_lock"
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -330,6 +330,10 @@ GEM
     sidekiq-statsd (2.1.0)
       activesupport
       sidekiq (>= 3.3.1)
+    sidekiq-unique-jobs (6.0.22)
+      concurrent-ruby (~> 1.0, >= 1.0.5)
+      sidekiq (>= 4.0, < 7.0)
+      thor (~> 0)
     sprockets (4.0.2)
       concurrent-ruby (~> 1.0)
       rack (> 1, < 3)
@@ -338,7 +342,7 @@ GEM
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
     statsd-ruby (1.4.0)
-    thor (1.0.1)
+    thor (0.20.3)
     thread_safe (0.3.6)
     thwait (0.1.0)
     tilt (2.0.10)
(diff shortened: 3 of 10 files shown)


lib/recovery/recover_lost_jobs.rb
```diff
@@ -0,0 +1,25 @@
+module Recovery
+  class RecoverLostJobs
+    STRATEGIES = [
```
benthorner, line 3: Minor: since we only have a pattern of one, could we just repeat the code for the time being? I'm not sure whether future recovery will fit this pattern.

lib/recovery/recover_lost_jobs.rb
```diff
@@ -12,14 +12,31 @@ class RecoverLostJobs
...
+      requeued_count = 0
+      skipped_count = 0
+
       STRATEGIES.each do |strategy|
         model = strategy[:model]
         worker = strategy[:worker]
 
         model.unprocessed.where("created_at <= ?", 1.hour.ago).map do |record|
-          worker.perform_async(record.id)
+          formatted_record = "{record.class}: #{record.slice(:id, :created_at)}"
```
benthorner, line 23: We shouldn't need this logging, except perhaps for the "success" one: the config for the unique jobs gem means we'll log if the job is a no-op. (And I don't think we have a need for knowing the number of recovered jobs.)

lib/tasks/recovery.rake
```diff
@@ -0,0 +1,6 @@
+namespace :recovery do
+  desc "Requeue content changes and messages over one-hour old for which jobs have been lost"
+  task recover_lost_jobs: :environment do
+    Recovery::RecoverLostJobs.run
```
benthorner, line 4: This should be a worker, that runs periodically: we're making this change so that no manual intervention is required when something is lost.

lib/tasks/recovery.rake
```diff
@@ -0,0 +1,6 @@
+namespace :recovery do
+  desc "Requeue content changes and messages over one-hour old for which jobs have been lost"
+  task recover_lost_jobs: :environment do
+    RecoverLostJobsWorker.perform_async
+  end
+end
```
kevindew, line 6: Do we need this task? I'm not sure under what scenarios we'd run it. It seems to be we'd need it if the sidekiq scheduling failed but if that happened we'd probably not trust putting another thing on sidekiq.

Anyway we should probably have a basic test for this too if this is something that adds value.

laurentqro: @benthorner the card suggest a rake task, but this is a fair point indeed. What is your opinion on this?

benthorner: It's referenced in the docs, and I think the reasoning still stands. However, the card could have been more specific: if someone on 2ndline is running this, it's probably because the work needs requeueing immediately (not after ~1.25 hours). Really, it depends on what the new documentation is going to look like: we auto-fix this alert for you, but if that doesn't work .... (we need some kind of debugging steps, although maybe a rake task isn't one of them).

kevindew: Makes sense. I've tried putting myself in the perspective of what I'd do if I saw the alert and tried to jot it down, maybe this will help the documentation.

1. Check Sentry and Kibana for any indication either this job is raising errors and failing or if the ProcessContentChangeWorker is raising errors (this seems most likely scenario)
2. Identify which changes are not processed: `ContentChange.unprocessed.where("created_at < ?", 2.hours.ago)`
3. Try run the process manually from a console and see if anything breaks. For example: `ContentChange.unprocessed.where("created_at < ?", 2.hours.ago).pluck(:id).map { |id| ProcessContentChangeWorker.new.perform(id) }`

Since the [documentation](https://docs.publishing.service.gov.uk/manual/alerts/email-alert-api-unprocessed-content-changes.html) is already rather focussed on manual console tasks it's probably worth just amending things to this list rather than the rake task.

I'd suggest adding some info on how to check if a job raised an error as the main thing then having bits on how to find the content changes and how to process them (which seems to already be there but could use some rephrasing.

spec/workers/recover_lost_jobs_worker_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.describe RecoverLostJobsWorker do
+  before do
+    Sidekiq::Worker.clear_all
+  end
```
kevindew, line 4: Minor, do we know if this block is needed? I don't think it actually does anything for the tests.

laurentqro: It's only necessary when making assertions on things like queue size, etc. It can go. 👌

PR #1323, opened 2020-07-14
Extend stats for immediate content changes

benthorner:
https://trello.com/c/hTAPWQfE/360-chart-email-backlog-by-publishing-app-and-document-type-in-grafana

Previously we could only see the realtime number of emails being
processed by the system, noting that almost all email we send is
due to a content change. This changes the processing of immediate
emails for a content change to send a breakdown of the generated
email by document type and publishing app, to improve visibility
of the effect of publishing on the volume of email we send, and to
help reassure ourselves about its source.

Note that this metric is not applicable for digest emails, because
of the many-to-one link between content changes and the email that
is sent. For digest emails, the number of subscriptions is a better
(maximum) indicator of the volume of email we expect to send.

## Example of what this could look like in a dashboard

Screenshot 2020-07-15 at 12 08 17

### DIFF
--- a/app/services/immediate_email_generation_service/batch.rb
+++ b/app/services/immediate_email_generation_service/batch.rb
@@ -17,6 +17,7 @@ class ImmediateEmailGenerationService
         SubscriptionContent.populate_for_content(content, records)
       end
 
+      MetricsService.content_change_emails(content_change, email_parameters.count) if content_change
       email_ids
     end
 
--- a/app/services/metrics_service.rb
+++ b/app/services/metrics_service.rb
@@ -1,5 +1,10 @@
 class MetricsService
   class << self
+    def content_change_emails(content_change, count)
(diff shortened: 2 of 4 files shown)


kevindew:
This seems like interesting data to collect. 

What types of visualisations are you hoping to make with it? It struck me as a somewhat unusual usage of gauge as it's not a relatively constant system number (such as memory usage) and thus would produce some odd results if two events of the same name happened twice within the sample (where an increment would be more representative).

If it's underlying data wanted rather than visualisations it might make sense to store a count of immediate emails created on a content change db record and then that can be grouped and queried in various ways.

benthorner:
@kevindew good point about the gauge - I was struggling with the StatsD docs and have since updated this to use a "count" metric type. Since the aim is to expose realtime stats, I think the metric approach is best.

I've added an example screenshot of how this could be shown on a dashboard. Could you take another look?

PR #1324, opened 2020-07-15
Ensure the correct sidekiq-unique-jobs option is used

laurentqro:
https://github.com/alphagov/email-alert-api/pull/1318 mistakenly shipped specifying the `lock_args` option from the upcoming version 7 of sidekiq-unique-jobs, rather than the currently used version's (6.0.22) `unique_args`.

Compare:

* https://github.com/mhenrixon/sidekiq-unique-jobs/tree/v6.0.22#finer-control-over-uniqueness
* https://github.com/mhenrixon/sidekiq-unique-jobs/tree/v7.0.0.beta22#finer-control-over-uniqueness

This ensures we use the version's correct sidekiq-unique-jobs option to enforce uniqueness.

Once the gem has been upgraded to version 7, it should be ok to remove the tests.

### DIFF
--- a/spec/workers/process_content_change_worker_spec.rb
+++ b/spec/workers/process_content_change_worker_spec.rb
@@ -56,4 +56,20 @@ RSpec.describe ProcessContentChangeWorker do
       described_class.new.perform(processed_content.id)
     end
   end
+
+  describe ".perform_async" do
+    # We're expecting redis-namespace to raise a warning unfortunately
+    before { allow_any_instance_of(Redis::Namespace).to receive(:warn) }
+
+    around do |example|
+      SidekiqUniqueJobs.use_config(enabled: true, logger: Logger.new("/dev/null")) do
+        example.run
+      end
+    end
+
+    it "enforces job uniqueness with the correct sidekiq-unique-jobs option" do
+      expect(described_class).to receive(:uniqueness_with).with([content_change.id])
+      described_class.perform_async(content_change.id)
+    end
+  end
 end
--- a/app/workers/process_content_change_worker.rb
+++ b/app/workers/process_content_change_worker.rb
@@ -3,8 +3,12 @@ class ProcessContentChangeWorker
 
   sidekiq_options queue: :process_and_generate_emails,
                   lock: :until_executed,
-                  lock_args: ->(args) { [args.first] },
+                  unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
                   on_conflict: :log
+
+  def self.uniqueness_with(args)
+    [args.first]
+  end
 
   def perform(content_change_id)
     content_change = ContentChange.find(content_change_id)
--- a/app/workers/process_message_worker.rb
+++ b/app/workers/process_message_worker.rb
@@ -3,8 +3,12 @@ class ProcessMessageWorker
 
   sidekiq_options queue: :process_and_generate_emails,
                   lock: :until_executed,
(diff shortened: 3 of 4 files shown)


laurentqro:
Ok @kevindew  I managed to silence all the logger info and the redis-namespace warning to stderr 👌

spec/workers/process_content_change_worker_spec.rb
```diff
@@ -55,5 +55,15 @@
...
+
+    it "enforces job uniqueness with the correct sidekiq-unique-jobs option" do
+      expect do
+        SidekiqUniqueJobs.use_config(enabled: true) do
+          allow(SidekiqUniqueJobs.logger).to receive(:info)
+          expect(described_class).to receive(:uniqueness_with).with([content_change.id])
+          described_class.perform_async(content_change.id)
+        end
+      end.to output.to_stderr # silence redis-namespace warning output during the tests
+    end
```
kevindew, line 67: Nice one sussing this out. It looks rather thorny, I can see your code has picked up some wounds along the way.

I'd suggest doing a refactoring. Instead of defining this test in `#perform` it probably should be in `.perform_asnyc` with something like this:

```
  describe ".perform_async" do
    around do |example|
      SidekiqUniqueJobs.use_config(enabled: true, logger: Logger.new("/dev/null")) do
        # We're expecting redis-namespace to raise a warning unfortunately
        expect { example.run }.to output.to_stderr
      end
    end

    it "enforces job uniqueness with the correct sidekiq-unique-jobs option" do
      expect(described_class).to receive(:uniqueness_with).with([content_change.id])
      described_class.perform_async(content_change.id)
    end
  end
```

It does feel a bit wrong having an `expect {}.to output.to_stderr` but I don't know a better approach and at least that will actually tell us if we manage to resolve the warning by dropping redis namespace.

laurentqro: I like it, very elegant. Thanks @kevindew

laurentqro: @kevindew When running the entire test suite, I am finding one of the two tests fails, because the redis warning is output only once, the first time.

What do you think about using a non-namespaced redis db in tests? :

```
Sidekiq.configure_server do |config|
  config.redis = { db: Rails.env.test? ? 1 : 0 }
end

Sidekiq.configure_client do |config|
  config.redis = { db: Rails.env.test? ? 1 : 0 }
end
```

laurentqro: > It does feel a bit wrong having an expect {}.to output.to_stderr but I don't know a better approach and at least that will actually tell us if we manage to resolve the warning by dropping redis namespace.

I guess that's a statement in favour of not using a redis db just for tests 🤔

kevindew: Oh darn, that's a pain.

This should sort it:

```
  describe ".perform_async" do
    around do |example|
      SidekiqUniqueJobs.use_config(enabled: true, logger: Logger.new("/dev/null")) do
        example.run
      end
    end

    # We're expecting redis-namespace to raise a warning unfortunately
    before { allow_any_instance_of(Redis::Namespace).to receive(:warn) }

    it "enforces job uniqueness with the correct sidekiq-unique-jobs option" do
      expect(described_class).to receive(:uniqueness_with).with([content_change.id])
      described_class.perform_async(content_change.id)
    end
  end
```

laurentqro: Works a charms, thanks @kevindew

PR #1325, opened 2020-07-16
Introduce ApplicationService to DRY-up services

benthorner:
This moves and refactors a few services in order to adopt
a consistent interface for classes in app/services.

UnsubscribeService was the main blocker for doing this, and
had some major issues with it:

- The interface was really unclear (the method names did not
indicate the outcome).

- The class had multiple responsibilities: single, bulk and
spam-related unsubscribes.

### DIFF
--- a/lib/tasks/manage.rake
+++ b/lib/tasks/manage.rake
@@ -108,7 +108,7 @@ namespace :manage do
       if active_subscriptions.empty?
         puts "Subscriber #{email_address} already unsubscribed from #{subscriber_list_slug}"
       else
-        UnsubscribeService.subscription!(active_subscriptions.last, :unsubscribed)
+        UnsubscribeService.unsubscribe!(subscriber, [active_subscriptions.last], :unsubscribed)
         puts "Unsubscribing from #{email_address} from #{subscriber_list_slug}"
       end
     end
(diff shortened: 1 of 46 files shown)


lib/tasks/manage.rake
```diff
@@ -108,7 +108,7 @@ namespace :manage do
       if active_subscriptions.empty?
         puts "Subscriber #{email_address} already unsubscribed from #{subscriber_list_slug}"
       else
-        UnsubscribeService.subscription!(active_subscriptions.last, :unsubscribed)
+        UnsubscribeService.unsubscribe!(subscriber, active_subscriptions.last, :unsubscribed)
```
kevindew: Is this passing through just one subscription (via the `.last`)? I'd have thought this only accepted an array?

benthorner: Oh yeah. Turns out none of these tasks are tested 🎉. I've fixed this, but I think adding tests is out of scope for this PR.

benthorner:
@kevindew I agree. I've added another commit for that. This brings into question `UnsubscribeService` taking an array. I tried to simplify the interface, but it turns out the deactivation behaviour [is expected, irrespective of subscriptions](https://github.com/alphagov/email-alert-api/blob/master/spec/lib/clean/invalid_subscribers_spec.rb#L8). I suspect that's just that the tests aren't using realistic data, but pursuing this further would require more effort.

Could you take another look?

PR #1327, opened 2020-07-17
Remove dangling reports (not listed in docs)

benthorner:
https://trello.com/c/hTAPWQfE/360-chart-email-backlog-by-publishing-app-and-document-type-in-grafana

Previously we added a "matched_content_changes" report that's only
been run once and is undocumented. In general, we a reporting task
should only exist for a documented (i.e. regular use) reason, and
otherwise we can recover the code if a new need arises in future.
This removes all tasks that are not referenced in documentation.

Note that, while the "csv_from_living_in_europe" is not referenced,
we are using it until some undetermined point. So it doesn't seem
appropriate to document it, but also we can't delete it.


benthorner:
@kevindew nope, nothing in the dev docs 🚀

PR #1332, opened 2020-07-21
Add rake task to fix /world email subscriptions

benthorner:
https://trello.com/c/iwhQbAaE/383-incident-action-fix-the-world-of-email

In an attempt to migrate world location email signups out
of Whitehall, a number of subscriptions became corrupted [1].
While we've stopped the issue for new subscriptions [2], the
corrupted ones remain and we need to fix them.

Original pages:

1. /world/spain (world location taxon)
2. /world/living-in-spain (world location child taxons)
3. /world/spain/news (world location)

Corruption:

- 1 + 3 => "4" (impossible to untangle)
- 2 => 2a (possible to auto-fix)

In order to cease the user impact and officially close the
incident, we have agreed there are two things for us to do:

1. Report the number of subscriptions for each corrupted list.
This can just be the subscriber list title, with the active subscription count.

2. Write a rake task (to be safe) to delete "4"-type subscriptions,
and auto-fix "2a"-type subscriptions.

The ~1300 corrupted lists we are interested in have links like:

```
"world_locations"=>{"any"=>["<content_id>"]}
```

We need to disambiguate between "4"-type and "2a"-type lists,
based on the Content ID in their links:

- "2a"-type Content IDs will match a level one world location
taxon with the same slug as one of the world locations in Whitehall.*

- "4"-type Content IDs will match any other world "taxon" Content
ID. We can get the full list of world taxons from the Publishing API.

*the slugs are the same because world location taxons previously
took over from world locations in Whitehall, in a failed attempt to
replace them - they continue to coexist, with different purposes.

[1]: https://docs.google.com/document/d/1fL5mQjRX2ApoVd4rOIxrnwjhrC-w4pgNgNXaA93KeM8/edit
[2]: https://github.com/alphagov/email-alert-frontend/pull/714


lib/tasks/temp_fix_world_subs.rake
```diff
@@ -0,0 +1,79 @@
...
+      puts "#{dead_sl.title} (#{dead_sl.slug}, #{dead_sl.subscriptions.active.count} subscriptions) - fixable, merging into '#{alive_sl.slug}'"
+    end
+
+    # So we can do a dry run
+    next unless args[:for_reals]
+
+    # End all subscriptions to the dead lists we can't fix, so they don't appear to the user
+    dead_4_type_lists.each do |dead_list|
+      dead_list.subscriptions.active.each do |dead_subscription|
+        UnsubscribeService.call(dead_subscription.subscriber, [dead_subscription], :subscriber_list_changed)
```
kevindew, line 68: I wonder if unpublished is a more appropriate reason than subscriber_list_changed, admittedly it's rather ambiguous.

benthorner: Yeah, I decided that "subscriber list changed" (since it did when we fixed it back in March) was the most accurate description, but none of the reasons are very accurate.

benthorner:
Thanks @kevindew - migration of broken data is definitely on the trickier end of the spectrum of code reviews. I think you missed the rather lengthy commit/PR description, though, which does define all the terms (like "2a").

> ie if /world/bhutan content_id is 123 then you have subscriptions to "world_locations" => { "any" => ["123"] }. Is that correct?

Yes, that's right. And they're not migratable because the user's intent could have been to subscribe to a world location, or a world location taxon.

> I did wonder whether you wanted to edit the 2a lists in place rather than duplicate them

Could you explain what more here? Perhaps you mean alter the list "links" so they become valid, leaving all the subscriptions as-is. That would be easier, but a little confusing because we'd end up with two identical lists in most cases.

> Are we going to delete the subscriber lists at all or leave them dormant afterwards?

I thought a little about this, and decided not to: we don't normally delete subscriber lists (even when they become dead for genuine reasons), and our approach for subscriptions is they are immutable. Leaving them as-is means we have a similar audit trail to the rest of the system, and we can always clean them (along with many others) in future if we decide this is a desirable behaviour for the system - I think we should!

Does that answer all your questions?

kevindew:
> Thanks @kevindew - migration of broken data is definitely on the trickier end of the spectrum of code reviews. I think you missed the rather lengthy commit/PR description, though, which does define all the terms (like "2a").

Ah no, that's the first thing I read and I didn't get that they were definitions, I read them as consequences. So I was wondering what properties a 2a list has that means it matches a level one taxon. I expect I didn't interpret them as definitions since the terminology was used so freely before.

> > ie if /world/bhutan content_id is 123 then you have subscriptions to "world_locations" => { "any" => ["123"] }. Is that correct?
> 
> Yes, that's right. And they're not migratable because the user's intent could have been to subscribe to a world location, or a world location taxon.
> 
> > I did wonder whether you wanted to edit the 2a lists in place rather than duplicate them
> 
> Could you explain what more here? Perhaps you mean alter the list "links" so they become valid, leaving all the subscriptions as-is. That would be easier, but a little confusing because we'd end up with two identical lists in most cases.

Sure, in this code: https://github.com/alphagov/email-alert-api/pull/1332/files#diff-44b2410e03a5d5fe0ed18a42413c2966R43-R48

Where we create a new list and then leave the broken one behind. If for most cases they already exist don't worry about the suggestion.

> > Are we going to delete the subscriber lists at all or leave them dormant afterwards?
> 
> I thought a little about this, and decided not to: we don't normally delete subscriber lists (even when they become dead for genuine reasons), and our approach for subscriptions is they are immutable. Leaving them as-is means we have a similar audit trail to the rest of the system, and we can always clean them (along with many others) in future if we decide this is a desirable behaviour for the system - I think we should!

This seems to be historically unclear for the project. You're right that nothing deletes them under every day circumstances though there seems to have been various work put into removing invalid or empty ones in the past (https://github.com/alphagov/email-alert-api/pull/912, https://github.com/alphagov/email-alert-api/pull/1109), so precedence seems to be towards removing. Though I don't think there's been any strategy put into it so I don't mind if you don't do it.



> Does that answer all your questions?

Yes, thank you

benthorner:
@kevindew ah, sorry the description wasn't so clear. It sounds like this is ready to go in?

> Where we create a new list and then leave the broken one behind. If for most cases they already exist don't worry about the suggestion.

Yes, my memory is that most lists have a working counterpart. Since it's so hard to check anything to do with the corrupted data in advance, I thought I'd put the code in for it anyway.

> Though I don't think there's been any strategy put into it so I don't mind if you don't do it.

I'll stick with it as-is, which has the small advantage that we can still look at the corrupted data again if we really, really need to.

PR #1335, opened 2020-07-22
Refresh README and a few supporting docs

benthorner:


README.md
```diff
@@ -47,36 +44,27 @@ subscribers to those tags via the external services.
...
-```
-
-* email-alert-api runs on port 3088
-* sidekiq-monitoring for email-alert-api uses 3089
+Using any email address that ends with `@notifications.service.gov.uk`
+will not create a subscriber or a subscription, however will return a `201 Created` response.
 
 ### Running the test suite
 
 * Run `RAILS_ENV=test bundle exec rake db:setup` to load the database
```
kevindew, line 54: Might want to lose the bullet point here, now this is a lone bullet.

docs/matching-content-to-subscriber-lists.md
```diff
@@ -27,6 +24,11 @@ similar to the links hash on a content item. For example:
 }
 ```
 
+Only one of these should be used on a given subscriber list; `tags` is
+deprecated. They can also both be empty `{}` if they aren't needed.
```
kevindew, line 28: We should probably drop the terminology about tags being deprecated. They don't seem to be going anywhere and they're the only thing we've got to inform about emails that is not related to a content item (which seems to be an ongoing need).

If it's not opening a fresh can of worms would be able to quietly remove the deprecation reference?

README.md
benthorner: Oops - meant to delete that. Updated.

docs/matching-content-to-subscriber-lists.md
benthorner: Sure, that makes sense - it's used by the Brexit Checker. I've added another commit.

README.md
```diff
@@ -47,36 +44,26 @@ subscribers to those tags via the external services.
...
 ### Running the application
 
-```bash
-$ ./startup.sh
-```
-
-* email-alert-api runs on port 3088
-* sidekiq-monitoring for email-alert-api uses 3089
+Using any email address that ends with `@notifications.service.gov.uk`
+will not create a subscriber or a subscription, however will return a `201 Created` response.
```
kevindew, line 50: Sorry - I didn't notice this on my last review but this text seems to have appeared inadvertently in this section.

docs/matching-content-to-subscriber-lists.md
kevindew: Thanks 👍

README.md
benthorner: Oh that's deliberate: it's in the section about "running the app", and I think it would only be useful here - we should be stubbing stuff in the tests. Underlying aim: put some context around the statement, so its utility is clear. 

Does that make more sense?

kevindew: Right I see what you mean. I find the section reads confusingly because there isn't a command to tell you how to run the application. This contrasts with the "Running the test suite" section which tells you how to do it, whereas this just seems to be a gotcha that you might experience in a particular application scenario. 

It might be clearer if you had a quick instruction on how to run the app and then put this in as an aside. 

For what it's worth I think this information seems rather niche for the prominence it has in the readme, it might be better to add it to the API docs for the related endpoints.

benthorner: Sure, that makes sense. I've added another commit for that. How does it look now?

PR #1336, opened 2020-07-22
Report count of emails marked as spam

laurentqro:
* Report count of emails marked as spam
* Report unsubscribe reason (Grafana will chart the ones `:marked_as_spam` specifically)
* Stop sending spam stats to Athena

https://trello.com/c/alIJoEDk/408-chart-emails-marked-as-spam-and-subscriptions-ended-due-to-this


app/models/email.rb
```diff
@@ -23,4 +23,11 @@ def self.timed_bulk_insert(records, batch_size)
 
     Metrics.email_bulk_insert(batch_size) { insert_all!(records) }
   end
+
+  def mark_as_spam!
+    unless marked_as_spam?
```
benthorner, line 28: Minor: we could do a short-circuit return here.

app/models/email.rb
```diff
@@ -23,4 +23,11 @@ def self.timed_bulk_insert(records, batch_size)
 
     Metrics.email_bulk_insert(batch_size) { insert_all!(records) }
   end
+
+  def mark_as_spam!
```
benthorner, line 27: Although this is a relevant method for `Email`, it's only used once. The trouble with models is that they can contain most methods; in general, we try to locate code elsewhere if possible. Could we keep this logic in the service?

app/models/email.rb
```diff
@@ -23,4 +23,11 @@ def self.timed_bulk_insert(records, batch_size)
...
+    unless marked_as_spam?
+      update!(marked_as_spam: true)
```
kevindew, line 29: Do we gain value storing this on an email model? I guess we don't expect it to be used for anything and deleted within a few days. 

It bugs me a bit as, aside from this, the model shouldn't be amended once we've finished sending and could be treated as immutable.

app/controllers/spam_reports_controller.rb
```diff
@@ -2,8 +2,8 @@ class SpamReportsController < ApplicationController
   wrap_parameters false
 
   def create
-    delivery_attempt = DeliveryAttempt.find(params[:reference])
-    SpamReportService.call(delivery_attempt)
+    email = DeliveryAttempt.find(params[:reference]).email
```
kevindew, line 5: Do we get a parameter of the email address? since we only hold on to DeliveryAttempts for a week this might 404 a lot. However if we used the email address we could remove the subscriptions even once we've deleted all the delivery attempts.

app/models/email.rb
laurentqro: Yeah I was in two minds about this. Let's keep the model slim 👌

app/models/email.rb
laurentqro: Moving this to the service 👌

laurentqro: @kevindew Did you mean the `marked_as_spam` attribute itself?

kevindew: Yup, the attribute itself.

laurentqro: Indeed we don't do much with that attribute. We now use it only to check whether a given email has already been marked as spam(https://github.com/alphagov/email-alert-api/pull/1336/files#diff-7bc42ad20bdcd55a897f3905ef82e794R12). Do you think it's ok if we remove that check @kevindew ?

benthorner: Sorry for wading in: it does have a slight value, because it de-dupes a user marking the same email as spam multiple times. Otherwise, we could skip this metric and just use the request logs from Kibana.*

*when I looked at the request logs for the spam endpoint, the data seemed a bit off to me. Perhaps it's worth another look, as it's still kind of relevant if a user marks the same email as spam multiple times.

kevindew: Sure, that makes sense. Bummer for the immutable aspect of the model mind. 

Do we know how it works at AWS SES/Notify if the same email gets marked multiple times? Might be worth checking that Notify don't retry us each time we return a 404.

laurentqro: Regarding those 404s, it does indeed look like it's caused by users unsubscribing after the email being marked as spam has been archived. The way around this would be for `/spam-reports` endpoint to use a unique identifier for the subscriber (e.g. subscriber ID, or email address) and use that to fetch the subscriber record and unsubscribe it from everything.

kevindew: Are you confident this happens more than the subscriber_id being nil? 

The bit that intrigues me is whether Notify matches the complaint to their data or not. I think they also have a 7 day retention period of emails so if they're matching the email to their database in theory our database should be pretty much the same. Whereas they could be sending them through whether they have the email or not which could mean it comes from any time.

app/services/spam_report_service.rb
```diff
@@ -1,14 +1,29 @@
...
+    @delivery_attempt_id = delivery_attempt_id
+    @email_address = email_address
   end
 
   def call
-    subscriber_id = delivery_attempt.email.subscriber_id
-    subscriber = Subscriber.find(subscriber_id)
+    return if email.marked_as_spam?
+
+    UnsubscribeAllService.call(subscriber, :marked_as_spam)
```
kevindew, line 12: This call doesn't look safe as subscriber could be nil and I don't think this service anticipates it to be nil.

app/services/spam_report_service.rb
```diff
@@ -1,14 +1,29 @@
...
+    Metrics.marked_as_spam
+    email.update!(marked_as_spam: true)
+  rescue ActiveRecord::RecordNotFound
     UnsubscribeAllService.call(subscriber, :marked_as_spam)
-    delivery_attempt.email.update!(marked_as_spam: true)
+    Metrics.marked_as_spam
```
kevindew, line 17: It's great that we're handling the 404s but I think this metric becomes somewhat confused if we'll populate it for all delivery attempts that 404 but only for emails we've not already marked as spam.

app/services/spam_report_service.rb
```diff
@@ -1,14 +1,11 @@
...
-  def initialize(delivery_attempt)
-    @delivery_attempt = delivery_attempt
+  def initialize(subscriber)
+    @subscriber = subscriber
   end
 
   def call
-    subscriber_id = delivery_attempt.email.subscriber_id
-    subscriber = Subscriber.find(subscriber_id)
     UnsubscribeAllService.call(subscriber, :marked_as_spam)
```
kevindew, line 9: I don't think there is much value in keeping this as a class if it only has one method.

app/queries/email_archive_query.rb
```diff
@@ -19,7 +19,6 @@ def fields
       message_ids,
       :finished_sending_at,
       :id,
-      :marked_as_spam,
```
kevindew, line 22: We should also mark this column as ignored in the model so that it can be safely removed. For example: https://github.com/alphagov/email-alert-api/pull/1348

Otherwise when you run the migration in https://github.com/alphagov/email-alert-api/pull/1353 there's a good chance the application will start erroring until the processes restart.

laurentqro: Ah yes, good catch, thanks

app/controllers/spam_reports_controller.rb
```diff
@@ -2,8 +2,8 @@ class SpamReportsController < ApplicationController
   wrap_parameters false
 
   def create
-    delivery_attempt = DeliveryAttempt.find(params[:reference])
-    SpamReportService.call(delivery_attempt)
+    subscriber = Subscriber.find_by(address: params[:to])
```
kevindew, line 5: Sorry I missed this before, this should use [the find_by_address method](https://github.com/alphagov/email-alert-api/blob/57c1abf9d512331d674cba60055b8e23c4c80c3f/app/models/subscriber.rb#L25)

PR #1343, opened 2020-07-28
Resolve concurrent request issues that cause 500 responses

kevindew:
Trello: https://trello.com/c/koo6gJov/355-increase-visibility-of-sentry-alerts-for-email-apps-and-services

This card was inspired by a Sentry error that occurred: https://sentry.io/organizations/govuk/issues/356403617/?project=202221&query=&statsPeriod=14d which was due to Email Alert API accepting concurrent requests for the same new subscriber. This raised an error due to a find or create section of code that can fail on a unique key violation if executed concurrently.

It also looks at what other errors could occur for subscriptions creation when concurrent and prevents those by locking the request based on the subscriber.

Finally it edits the update subscription endpoint to check whether a subscription has changed before doing an update to save unnecessary work.

I found these endpoints rather baffling in their shared purpose (hopefully I didn't write them originally) so put in a [pain card](https://trello.com/c/Z8tTO9pr/368-the-subscription-create-endpoint-does-basically-everything-that-the-update-endpoint-does) for it.

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -2,20 +2,21 @@ class SubscriptionsController < ApplicationController
   def create
     return render json: { id: 0 }, status: :created if smoke_test_address?
 
-    subscription = send_email = status = nil
+    subscriber = Subscriber.resilient_find_or_create(
+      address,
+      signon_user_uid: current_user.uid,
+    )
 
-    Subscription.transaction do
-      if (subscriber_was_deactivated = subscriber.deactivated?)
-        subscriber.activate!
-      end
+    subscription, email, status = subscriber.with_lock do
+      deactivated_subscriber = subscriber.deactivated?
+      subscriber.activate! if deactivated_subscriber
 
-      existing_subscription = Subscription.active.lock.find_by(
+      existing_subscription = Subscription.active.find_by(
         subscriber: subscriber,
         subscriber_list: subscriber_list,
       )
 
-      create_new_subscription = existing_subscription.nil? ||
-        (frequency.to_s != existing_subscription.frequency.to_s)
+      create_new_subscription = frequency.to_s != existing_subscription&.frequency.to_s
 
       if create_new_subscription
         existing_subscription&.end(reason: :frequency_changed)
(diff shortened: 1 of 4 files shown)


app/models/subscriber.rb
```diff
@@ -26,6 +26,23 @@ def self.find_by_address!(address)
...
+      subscriber = find_by_address(address)
+      subscriber || create!({ address: address }.merge(create_params))
+    end
+  rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid
+    # if we have concurrent requests trying to find or create the same
+    # subscriber both can fail to find by address and then the both try
+    # create the record. This retries the first time an
+    # ActiveRecord::RecordNotUnique error is raised with the expectation
+    # that this occurring more than once is a bigger problem.
+    (retries += 1) == 1 ? retry : raise
```
1pretz1, line 43: Ohh nice, haven't seen `retry` used before

app/controllers/subscriptions_controller.rb
```diff
@@ -2,20 +2,21 @@ class SubscriptionsController < ApplicationController
...
 
-      existing_subscription = Subscription.active.lock.find_by(
+      existing_subscription = Subscription.active.find_by(
         subscriber: subscriber,
         subscriber_list: subscriber_list,
       )
 
-      create_new_subscription = existing_subscription.nil? ||
-        (frequency.to_s != existing_subscription.frequency.to_s)
+      create_new_subscription = frequency.to_s != existing_subscription&.frequency.to_s
```
1pretz1, line 19: Do we need to be calling `to_s` on these when they will already be strings (or nil)?

app/controllers/subscriptions_controller.rb
```diff
@@ -54,6 +54,11 @@ def update
       subscription_params.require(:id),
     )
 
+    if frequency.to_s == existing_subscription.frequency.to_s
```
1pretz1, line 57: Same point here regarding `to_s`

app/controllers/subscriptions_controller.rb
kevindew: I'll check if it's necessary or not.

kevindew: I managed to remove them 🎉

PR #1345, opened 2020-07-29
Brexit report task

hannako:
Create rake tasks to report on the number of SubscriberLists relating to Brexit content. 

https://trello.com/c/xIXJJM8p/355-produce-email-stats-for-paul-cronk-for-specific-date-range
https://trello.com/c/yHKowGks/346-produce-regular-email-stats-for-paul-cronk

### DIFF
--- a/lib/reports/brexit_subscribers_report.rb
+++ b/lib/reports/brexit_subscribers_report.rb
@@ -1,23 +1,51 @@
 require "csv"
 
 class Reports::BrexitSubscribersReport
-  CSV_HEADERS = %w[title slug tags subscribed unsubscribed immediately daily weekly]
+  CSV_HEADERS = %w[title slug tags subscribed unsubscribed immediately daily weekly].freeze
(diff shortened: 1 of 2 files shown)


kevindew:
@hannako  Sorry to comment post-merge, I didn't see this PR until the incident.

There's some things in here that'd be good to resolve where this is inconsistent with the existing report tasks.

- There don't seem to be any tests for this
- Looks like the previous CSV generation task outputted the CSV to stdout when the task was run whereas this one writes to a temp file, any reason this needs to be different? I imagine the temp file approach is quite a bit harder to work with

Some minor code aspects:

- it seems unconventional that everything is a private method and that there isn't a `#call` method to do the action to correspond with the `.call` method
- line 38 looks super long.

Would you be able to have a look at these?

hannako:
Hi @kevindew - sorry for delay, have been on AL since you left the comments. Will take a look, thanks 👍

PR #1346, opened 2020-07-30
Fix Zeitwerk loading EmailAlertAPI constant

kevindew:
This resolves a problem where the Email Alert API application was not
starting when using the Zeitwerk autoloader. The application wasn't
starting because it couldn't find the EmailAlertAPI constant and was
triggering this error:

```
/Users/kevindew/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/zeitwerk-2.4.0/lib/zeitwerk/loader.rb:392:in `const_get': uninitialized constant EmailAlertApi::Config (NameError)
```

The autoloader would expect to find a class named EmailAlertApi rather
than EmailAlertAPI at the location of email_alert_api.

### DIFF
--- a/config/initializers/zeitwerk.rb
+++ b/config/initializers/zeitwerk.rb
@@ -1,5 +1,6 @@
 Rails.autoloaders.each do |autoloader|
   autoloader.inflector.inflect(
     "symbolize_json" => "SymbolizeJSON",
+    "email_alert_api" => "EmailAlertAPI",
   )
 end


richardTowers:
Do we not have any tests that cover the autoloader?

kevindew:
> Do we not have any tests that cover the autoloader?

We can get caught out with the differences in loading behaviour between test environments and production. When production starts it eager loads all the classes where this error can occur. Whereas in test we avoid eager loading the whole application and as this constant gets loaded by something else: 

https://github.com/alphagov/email-alert-api/blob/5864d586c39428907448a2332e0197832dc915f5/config/application.rb#L37
(which I'm going to see if I can now remove with this change)

so the auto-loader isn't needed. 

This situation is actually one of the ones that the publishing-e2e-tests suite can catch as it runs the application in a production environment, which makes it good for catching the parts of code that the test suite doesn't exercise but are essential for basic functionality.

kevindew:
Thanks @richardTowers, I was busy editing my earlier reply as I learnt more from digging further. It's probably easier to just write a fresh reply (and a reminder that I should probably slow down 😅 )

So the issue wasn't actually that it fails to autoload the class we want, it's actually that when the new Zeitwerk autoloader runs it expects to find a constant with the name `EmailAlertApi::Config` at the path it loaded and errors when one isn't defined. So rather this actually being a problem with it not finding the class we want, it instead errors for this different reason.

There's a couple of smells about this to look into further. This `EmailAlertAPI::Config` is a somewhat odd pattern, which we should see about removing. It also might not be too big a deal to `eager_load = true` in tests, particularly with trying out [spring](https://github.com/alphagov/email-alert-api/pull/1334), since in my quick tests it doesn't seem to make much time difference - however this would be unconventional for a Rails app.

PR #1352, opened 2020-08-04
Add task to switch users to daily digest emails

benthorner:
https://trello.com/c/RPBoOye6/440-start-the-switch-immediate-to-digest-experiment

This adds a new task as part of an experiment to switch immediate
subscribers of specific lists to daily digest emails. For each
subscriber, we send a summary email of the changes made. In order to
track the changes, we use a new reason for start/ending a subscription.

We process each subscriber individually, using the same locking approach
as when a subscriber is manually updating their subscriptions. I decided
to avoid doing a single, bulk change, as it's not clear that a roll back
would be possible in the presence of one or more race conditions.

The task is best effort: if a race condition causes the change to fail
for a particular subscriber, we can always rerun the task.

## Things left to do

- [ ] Update the content (separate PR, if this is ready to go in)
- [x] Test this works on Integration
- [ ] Decouple *Request*Helpers from requests
- [x] Rename BulkEmailBuilder to be more specific

### DIFF
--- a/app/builders/switch_to_daily_digest_email_builder.rb
+++ b/app/builders/switch_to_daily_digest_email_builder.rb
@@ -1,0 +1,44 @@
+class SwitchToDailyDigestEmailBuilder
+  def initialize(subscriber, subscriptions)
+    @subscriber = subscriber
+    @subscriptions = subscriptions
+  end
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def call
+    Email.create!(
+      address: subscriber.address,
+      subject: "Your GOV.UK email subscriptions",
+      body: body,
+      subscriber_id: subscriber.id,
+    ).id
+  end
+
+  private_class_method :new
+
+private
+
+  attr_reader :subscriber, :subscriptions
+
+  def body
+    <<~BODY
+      From now on, we'll group all the emails you get together into one digest when there's a change to content covered by one of these subscriptions:
+
+      #{subscriptions.map { |subscription| "- #{subscription.subscriber_list.title}" }.join("\n")}
+
+      If you need to keep receiving immediate emails you can change this by [managing your email subscriptions](#{manage_subscriptions_link}).
+    BODY
+  end
+
+  def manage_subscriptions_link
+    utm_source = "switch-to-daily-digest"
+    utm_medium = "email"
+    utm_campaign = "govuk-notifications"
+    utm_content = "initial-experiment"
+    base_url = PublicUrls.authenticate_url(address: subscriber.address)
+    "#{base_url}&utm_source=#{utm_source}&utm_medium=#{utm_medium}&utm_campaign=#{utm_campaign}&utm_content=#{utm_content}"
+  end
+end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -5,8 +5,8 @@ class Subscription < ApplicationRecord
   has_many :subscription_contents, dependent: :destroy
 
   enum frequency: { immediately: 0, daily: 1, weekly: 2 }
-  enum source: { user_signed_up: 0, frequency_changed: 1, imported: 2, subscriber_list_changed: 3 }, _prefix: true
-  enum ended_reason: { unsubscribed: 0, non_existent_email: 1, frequency_changed: 2, subscriber_list_changed: 3, marked_as_spam: 4, unpublished: 5 }, _prefix: :ended
+  enum source: { user_signed_up: 0, frequency_changed: 1, imported: 2, subscriber_list_changed: 3, bulk_immediate_to_digest: 4 }, _prefix: true
+  enum ended_reason: { unsubscribed: 0, non_existent_email: 1, frequency_changed: 2, subscriber_list_changed: 3, marked_as_spam: 4, unpublished: 5, bulk_immediate_to_digest: 6 }, _prefix: :ended
 
   validates :subscriber, uniqueness: { scope: :subscriber_list, conditions: -> { active } }
 
--- a/lib/tasks/bulk.rake
+++ b/lib/tasks/bulk.rake
@@ -11,4 +11,81 @@ namespace :bulk do
       DeliveryRequestWorker.perform_async_in_queue(id, queue: :delivery_immediate)
     end
   end
+
+  desc "Switch immediate subscribers of the specified list slugs to daily digest"
+  task switch_to_daily_digest: :environment do |_t, args|
+    lists = SubscriberList.where(slug: args.extras)
+    raise "One or more lists were not found" if lists.count != args.extras.count
+
+    subscriptions = Subscription.active.immediately.where(subscriber_list: lists)
+    raise "No subscriptions to change" if subscriptions.none?
+
+    subscribers = Subscriber.where(id: subscriptions.pluck(:subscriber_id)).index_by(&:id)
+    subscriptions_by_subscriber = subscriptions.group_by(&:subscriber_id).transform_keys { |k| subscribers[k] }
+
+    subscriptions_by_subscriber.each do |subscriber, immediate_subscriptions|
+      email_id = nil
+
+      subscriber.with_lock do
+        immediate_subscriptions.each do |subscription|
+          subscription.end(reason: :bulk_immediate_to_digest)
+
+          Subscription.create!(
+            subscriber_id: subscription.subscriber_id,
+            subscriber_list_id: subscription.subscriber_list_id,
+            frequency: :daily,
+            source: :bulk_immediate_to_digest,
+          )
+        end
+
+        email_id = SwitchToDailyDigestEmailBuilder.call(subscriber, immediate_subscriptions)
+      end
+
+      DeliveryRequestWorker.perform_async_in_queue(email_id, queue: :default)
+    rescue StandardError => e
+      puts "Skipping subscriber: #{e}"
+    end
+  end
+
+  desc "Switch immediate subscribers of the following lists to daily digest (experiment)"
+  task switch_to_daily_digest_experiment: :environment do
+    Rake::Task["bulk:switch_to_daily_digest"].invoke(
+      "news-and-communications-2",
+      "guidance-and-regulation-2",
+      "guidance-and-regulation",
+      "guidance-about-all-topics-by-all-organisations",
+      "news-and-communications-3",
+      "news-and-communications",
+      "all-announcements-about-all-topics-by-all-organisations",
+      "business-and-industry",
+      "crime-justice-and-law",
+      "statistics",
+      "government-efficiency-transparency-and-accountability-2",
+      "press-releases-about-all-topics-by-all-organisations",
+      "environment-agency",
+      "corporate-information",
+      "all-types-of-document-about-all-topics-by-department-for-environment-food-rural-affairs",
+      "department-for-environment-food-rural-affairs",
+      "hm-revenue-customs",
+      "animal-and-plant-health-agency",
+      "ministry-of-defence",
+      "car-driving-tests",
+      "car-motorcycle-and-van-mot-tests",
+      "business-tax-self-employment",
+      "driving-and-motorcycle-tests",
+      "ofsted",
+      "rail-accident-investigation-branch",
+      "correspondence-related-to-education-and-education-and-skills-funding-agency-2",
+      "news-stories-about-all-topics-by-department-of-health-and-social-care",
+      "department-for-education",
+      "international-development-funding",
+      "the-charity-commission",
+      "theory-tests",
+      "standards-and-testing-agency",
+      "personal-tax-self-assessment",
+      "planning-and-development-planning-officer-guidance",
+      "civil-service-fast-track-apprenticeship",
+      "further-education-and-skills-apprenticeships",
+    )
+  end
 end
--- a/spec/lib/tasks/bulk_spec.rb
+++ b/spec/lib/tasks/bulk_spec.rb
@@ -1,4 +1,6 @@
 RSpec.describe "bulk" do
+  include RequestHelpers
+
   describe "email" do
     before do
       Rake::Task["bulk:email"].reenable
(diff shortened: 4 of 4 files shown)


lib/tasks/bulk.rake
```diff
@@ -11,4 +11,80 @@ namespace :bulk do
...
+
+  desc "Switch immediate subscribers of the specified list slugs to daily digest"
+  task switch_to_daily_digest: :environment do |_t, args|
+    lists = SubscriberList.where(slug: args.extras)
+    raise "One or more lists were not found" if lists.count != args.extras.count
+
+    subscriptions = Subscription.active.immediately.where(subscriber_list: lists)
+    raise "No subscriptions to change" if subscriptions.none?
+
+    subscribers = Subscriber.where(id: subscriptions.pluck(:subscriber_id)).index_by(&:id)
```
kevindew, line 23: Do you know what size the `subscriptions.pluck(:subscriber_id)` array might reach? I have a recollection of us hitting an issue on this project when we exceeded 2^16 (~65,000) items in an `IN` clause, but can't find the project. Googling for it seems to conflict between things [saying no limit](https://stackoverflow.com/questions/1009706/postgresql-max-number-of-parameters-in-in-clause#:~:text=There%20is%20no%20limit%20to,in%20the%20array%20or%20not.) and [others saying ~32,000](https://github.com/jOOQ/jOOQ/issues/5701).

lib/tasks/bulk.rake
```diff
@@ -11,4 +11,80 @@ namespace :bulk do
       DeliveryRequestWorker.perform_async_in_queue(id, queue: :delivery_immediate)
     end
   end
+
+  desc "Switch immediate subscribers of the specified list slugs to daily digest"
+  task switch_to_daily_digest: :environment do |_t, args|
```
thomasleese, line 16: I'm just commenting here based on https://github.com/alphagov/email-alert-api/pull/1354.

I feel like this doesn't belong in the `bulk` namespace because in my mind, the "bulk email" is a very specific thing which sends a single email out to multiple subscriber list. The task being added here is a different kind of email which also sends to multiple subscribe lists, but isn't a "bulk email".

I actually think having a `bulk` namespace is a bit confusing and I wonder if restructuring the tasks to be like this would make the distinction clearer (and perhaps avoid needing to do https://github.com/alphagov/email-alert-api/pull/1354):

- `emails:send_bulk`
- `emails:send_switch_to_daily_digest`
- ... any other new emails builders go here

spec/lib/tasks/bulk_spec.rb
```diff
@@ -40,4 +42,84 @@
...
+        allow(Subscription)
+          .to receive(:create!)
+          .with(hash_including(subscriber_id: subscription2.subscriber_id))
+          .and_raise("An error")
+      end
+
+      it "does not send an email to the subscriber" do
+        expect { Rake::Task["bulk:switch_to_daily_digest"].invoke(list1.slug) }
+          .to output.to_stdout
+          .and change { Email.count }.by(1)
```
kevindew, line 113: This seems to conflict with the description which caught me out. Perhaps you want it to say `it "only sends an email for the subscription that changed"`?

app/builders/switch_to_daily_digest_email_builder.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      If you need to keep receiving immediate emails you can change this by [managing your email subscriptions](#{manage_subscriptions_link}).
+    BODY
+  end
+
+  def manage_subscriptions_link
+    utm_source = "switch-to-daily-digest"
+    utm_medium = "email"
+    utm_campaign = "govuk-notifications"
+    utm_content = "experiment-001"
```
kevindew, line 42: Can we name the experiment at all? this feels like a rather leading number that implies a future whereas this even might be rather isolated.

app/builders/switch_to_daily_digest_email_builder.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def call
+    Email.create!(
+      address: subscriber.address,
+      subject: "Your GOV.UK email subscriptions",
+      body: body,
+      subscriber_id: subscriber.id,
+      created_at: Time.zone.now,
+      updated_at: Time.zone.now,
```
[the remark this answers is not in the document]
benthorner: Oops - relic of a previous approach. Removed.

app/builders/switch_to_daily_digest_email_builder.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  private_class_method :new
+
+private
+
+  attr_reader :subscriber, :subscriptions
+
+  def body
+    <<~BODY
+      From now on, we'll group all the emails you get together into one digest when there's a change to content covered by one of these subscriptions:
```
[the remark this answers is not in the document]
benthorner: Content is in progress. I've got a checklist item to cover that ☝️

app/builders/switch_to_daily_digest_email_builder.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      #{subscriptions.map { |subscription| "- #{subscription.subscriber_list.title}\n" }}
+
+      If you need to keep receiving immediate emails you can change this by [managing your email subscriptions](#{manage_subscriptions_link}).
+    BODY
```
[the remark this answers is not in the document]
benthorner: I'm not sure. Again, this is dependent on ongoing content design, which we could incorporate later.

app/models/subscription.rb
```diff
@@ -5,8 +5,8 @@ class Subscription < ApplicationRecord
   has_many :subscription_contents, dependent: :destroy
 
   enum frequency: { immediately: 0, daily: 1, weekly: 2 }
-  enum source: { user_signed_up: 0, frequency_changed: 1, imported: 2, subscriber_list_changed: 3 }, _prefix: true
-  enum ended_reason: { unsubscribed: 0, non_existent_email: 1, frequency_changed: 2, subscriber_list_changed: 3, marked_as_spam: 4, unpublished: 5 }, _prefix: :ended
+  enum source: { user_signed_up: 0, frequency_changed: 1, imported: 2, subscriber_list_changed: 3, bulk_change: 4 }, _prefix: true
+  enum ended_reason: { unsubscribed: 0, non_existent_email: 1, frequency_changed: 2, subscriber_list_changed: 3, marked_as_spam: 4, unpublished: 5, bulk_change: 6 }, _prefix: :ended
```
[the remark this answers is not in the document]
benthorner: Sure, I've come around to your suggestion of "bulk_immediate_to_digest".

lib/tasks/bulk.rake
```diff
@@ -11,4 +11,80 @@ namespace :bulk do
...
+          Subscription.create!(
+            subscriber_id: subscription.subscriber_id,
+            subscriber_list_id: subscription.subscriber_list_id,
+            frequency: :daily,
+            source: :bulk_change,
+          )
+        end
+      end
+
+      email_id = SwitchToDailyDigestEmailBuilder.call(subscriber, immediate_subscriptions)
```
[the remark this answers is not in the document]
benthorner: Sure, I've done that. I think originally I was worrying about an actually-non-existent scenario where the email did send successfully, but some error occurred after (the point being the email isn't critical).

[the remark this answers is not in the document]
benthorner: Putting it inside the lock has the advantage that we're even less likely to send a confusing email in the presence of a race condition.

spec/lib/tasks/bulk_spec.rb
benthorner: Yeah, you have to read it together with the context. I've changed it to "only sends an email to switched subscribers".

app/builders/switch_to_daily_digest_email_builder.rb
benthorner: Sure, it was arbitrary: I've gone with "initial-experiment".

lib/tasks/bulk.rake
benthorner: Hmm...tricky, I see what you mean. If this is an issue, it should come up when testing on Integration (TODO).

lib/tasks/bulk.rake
benthorner: @thomasleese it sounds like you're seeing the "bulk" namespace and the "email" task as a single entity. Whereas I'm seeing them as separate things: a namespace to put bulk tasks (we have to put them somewhere, after all), and a task that describes what kind of bulk operation it does.

> https://github.com/alphagov/email-alert-api/pull/1354 is making a similar distinction: a generic name for specific code makes it harder to understand its purpose.

The "email" task could do with a slightly longer name to describe what it does, but for now it's not in competition with any other bulk email task. I don't think "emails" would be appropriate namespace for this new task, because the email sending bit is only part of its behaviour.

Does that help?

thomasleese: Yeah, that makes sense. I think this might just be a personal preference thing.

To me, we have plenty of tasks in Email Alert API which perform bulk operations, but their defining characteristic is not the fact that they are bulk, it's _what_ they actually do: whether that be sending emails or managing subscriptions. For example, [we've got three tasks in the `manage` namespace](https://github.com/alphagov/email-alert-api/blob/master/lib/tasks/manage.rake#L117-L144) which are definitely bulk operations and therefore if we were going to stick with this naming scheme, we should probably move them into the `bulk` namespace.

Perhaps my complaint is actually with the naming of the `manage` namespace (and by extension the `troubleshoot` namespace) in that it's very generic, it doesn't really explain what the tasks do and makes it difficult to place new tasks. Thinking about it further, I agree with you that the task in this PR doesn't just send emails and therefore doesn't belong in an `emails` namespace. I'm actually thinking that my ideal structure would be something like this:

- `emails:send_bulk`
- [`emails:test:send_to_subscriber`](https://github.com/alphagov/email-alert-api/blob/4ba89df82a105ce2d88c93058fa9b6f84cf40ce2/lib/tasks/troubleshoot.rake#L20-L30)
- [`subscribers:change_email_address`](https://github.com/alphagov/email-alert-api/blob/4ba89df82a105ce2d88c93058fa9b6f84cf40ce2/lib/tasks/manage.rake#L79)
- `subscribers:switch_to_digest`

I realise though that this is kind of derailing your PR and beyond the scope. If you're interested I'm happy to talk about it further, but I won't block your PR going in as it is now!

lib/tasks/bulk.rake
benthorner: Looks like it works 🎉

lib/tasks/bulk.rake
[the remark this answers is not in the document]
benthorner: Update: turns out the async sending of the email can be in a race with the creation of it inside the transaction. I've moved the last line out of the transaction/lock, to avoid this.

benthorner:
@kevindew this should be ready for another look.

lib/tasks/bulk.rake
```diff
@@ -11,4 +11,81 @@ namespace :bulk do
...
+    subscriptions = Subscription.active.immediately.where(subscriber_list: lists)
+    raise "No subscriptions to change" if subscriptions.none?
+
+    subscribers = Subscriber.where(id: subscriptions.pluck(:subscriber_id)).index_by(&:id)
+    subscriptions_by_subscriber = subscriptions.group_by(&:subscriber_id).transform_keys { |k| subscribers[k] }
+
+    subscriptions_by_subscriber.each do |subscriber, immediate_subscriptions|
+      email_id = nil
+
+      subscriber.with_lock do
```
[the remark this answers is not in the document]
benthorner: Oh yeah. Juuust in case that doesn't work for some reason, I'm going to leave as-is.

benthorner:
@kevindew sorry, I didn't spot your PR-level comment before merging. Happy to raise another if we agree to change it.

I agree the parameters are a bit wonky - for some reason I assumed we used the same ones for different emails, but [we don't](https://github.com/alphagov/email-alert-api/blob/16505e2d3bac948f1ba0394be6005aee22b8e501/app/builders/subscription_confirmation_email_builder.rb#L59). I'll raise another PR with something like your proposal.

kevindew:
@benthorner It's cool, it wasn't a blocker.

lib/tasks/bulk.rake
benthorner: Thanks @thomasleese. Good to know this isn't a blocker. I agree that two of those tasks you mentioned are indeed bulk operations. However, "unsubscribe all" reflects an action a user can do themselves - why do we even need that?

Anyway, I think there's still some ambiguity with the namespaces, even with your suggestion. When we're writing and finding rake tasks, I think we're more concerned with the context (e.g. "2ndline" or "data_migration"), rather than the scope ("subscribers" or "bulk"). How about this?

## 2ndline support tasks

- `2ndline:view_subscriptions`
- `2ndline:view_emails`
- ~`manage:change_email_address`~ (users can do this themselves?)
- ~`manage:unsubscribe_single_subscription`~ ([the definite need for this](https://github.com/alphagov/email-alert-api/pull/1226) is unclear, user can do this themselves?)
- ~`manage:unsubscribe_all_subscriptions`~ (users can do this themselves?)
- `2ndline:get_notifications_from_notify_by_email_id`
- ~`troubleshoot:get_notifications_from_notify`~ (not sure why we need this if we have the above)
- `2ndline:delivery_to_test_email`
- ~`troubleshoot:delivery_to_subscriber`~ (not sure why we need this if we have the above)
- `2ndline:resend_failed_emails`
- `2ndline:send_bulk_to_subscriber_list`

## Data migration tasks

- `data_migration:move_all_subscribers`
- `data_migration:find_subscriber_list_by_title` ([precursor to the next one](https://github.com/alphagov/email-alert-api/pull/799))
- `data_migration:update_subscriber_list`
- `data_migration:switch_to_daily_digest[_experiment]`
- ~`manage:unsubscribe_bulk_from_csv`~ (seems redundant, based on the [original PR](https://github.com/alphagov/email-alert-api/pull/511))

thomasleese: Sorry for the delay, finally got some free time to reply to this!

I think your proposal for classifying the Rake tasks is interesting, although I'm not 100% convinced. For example when I ran the job to send the bulk travel advice email, I wasn't doing it while on 2nd line. I think we might be trying to second guess the context that people are running these tasks in whereas grouping the tasks by _what_ they do feels more natural to me. For example, I struggle to see a big difference between what we have now with `manage:` and your proposal of `2ndline:` in terms of clarity of where to put new tasks and where to find existing ones.

In terms of the tasks you've identified as ones we can remove, there have certainly been times when I've used the tasks to manually unsubscribe people who are having problems (and just solving the problem for them seemed easier than helping them solve it for themselves), but I agree that ultimately it's something users can do so I'm also unsure about their need.

benthorner: Ah, the merged PR comment saga continues. I kind of like it.

I think we've both got some good points, and are slowly converging on something agreeable.

## Pointless tasks

It would help to cut out the cruft. Could you tick/comment on the ones you agree we can get rid of?

- [ ] manage:change_email_address
- [ ] manage:unsubscribe_single_subscription
- [ ] manage:unsubscribe_all_subscriptions
- [ ] troubleshoot:get_notifications_from_notify
- [ ] troubleshoot:delivery_to_subscriber
- [ ] manage:unsubscribe_bulk_from_csv

## Namespaces

- You didn't say anything about `data_migration`, so I'm going to assume that's OK. It incorporates both the context and the behaviour of the tasks within it, so seems to meet both our expectations.

- `2ndline` does seem slightly better for some of the tasks. For example, `view_emails` isn't really managing anything. Some of these tasks appear to be defined specifically for 2ndline, so their purpose is "support" or "troubleshoot". I don't really mind about the word here, as long as it's more specific than "manage". What do you prefer?

- `2ndline:send_bulk_to_subscriber_list` seems to be an outlier. Perhaps it could be in its own namespace, which is specific to `custom_email` or `send_email`, or something?!

What do you think? Hopefully that gets us a bit closer to a conclusion.

thomasleese: It's been a few days so I've found time to reply to this 😆 

I've ticked the tasks I don't think we need. In terms of the naming, I think `support` is a better name than `manage` and I prefer that it's not tied specifically to `2ndline`.

I agree that perhaps a good solution here to to put `send_bulk_to_subscriber_list` in its own namespace, I think either `emails` or `send_email` is good.

benthorner: Cool, thanks for the review.

First part of the change here: https://github.com/alphagov/email-alert-api/pull/1367

PR #1353, opened 2020-08-04
Remove marked as spam field on email

laurentqro:
The limited retention period for emails make recording this not
very useful, especially that we now chart spam reports to a
Grafana dashboard.

Dependent on https://github.com/alphagov/email-alert-api/pull/1336.

Related:

* https://github.com/alphagov/email-alert-api/pull/1336
* https://github.com/alphagov/govuk-puppet/pull/10529

Trello:

https://trello.com/c/alIJoEDk/408-chart-emails-marked-as-spam-and-subscriptions-ended-due-to-this

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -17,9 +17,6 @@ class Email < ApplicationRecord
 
   enum status: { pending: 0, sent: 1, failed: 2 }
 
-  # This can be removed once the column is deleted.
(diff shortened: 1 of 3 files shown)


kevindew:
We can merge https://github.com/alphagov/email-alert-api/pull/1361 into this since the migration will run first before the code is deployed.

PR #1355, opened 2020-08-05
Avoid overlapping experiment UTM with other emails

benthorner:
https://trello.com/c/RPBoOye6/440-start-the-switch-immediate-to-digest-experiment

This is more consistent with the way these parameters should be used
[1]. Note that 'utm_content' is an optional parameter, but may be
useful for distinguishing other links we add in future.

- [ ] Send test email and check tracking works on Integration

[1]: https://en.wikipedia.org/wiki/UTM_parameters#UTM_parameters

### DIFF
--- a/app/builders/switch_to_daily_digest_email_builder.rb
+++ b/app/builders/switch_to_daily_digest_email_builder.rb
@@ -34,10 +34,10 @@ private
   end
 
   def manage_subscriptions_link
-    utm_source = "switch-to-daily-digest"
+    utm_campaign = "govuk-notifications-switch-to-daily-experiment"
     utm_medium = "email"
-    utm_campaign = "govuk-notifications"
-    utm_content = "initial-experiment"
+    utm_source = "gov.uk"
+    utm_content = "manage-subscriptions"
     base_url = PublicUrls.authenticate_url(address: subscriber.address)
     "#{base_url}&utm_source=#{utm_source}&utm_medium=#{utm_medium}&utm_campaign=#{utm_campaign}&utm_content=#{utm_content}"
   end


app/builders/switch_to_daily_digest_email_builder.rb
```diff
@@ -34,10 +34,10 @@ def body
   end
 
   def manage_subscriptions_link
-    utm_source = "switch-to-daily-digest"
+    utm_campaign = "govuk-notifications-switch-to-daily-digest"
     utm_medium = "email"
-    utm_campaign = "govuk-notifications"
-    utm_content = "initial-experiment"
+    utm_source = "initial-experiment"
```
kevindew, line 39: Sorry nitpick, this should be the site name so should be "gov.uk"

benthorner: Hmm, while I agree that's what Wikipedia says, it's inconsistent with how we use it in the repo:

- https://github.com/alphagov/email-alert-api/blob/16505e2d3bac948f1ba0394be6005aee22b8e501/app/presenters/content_change_presenter.rb
- https://github.com/alphagov/email-alert-api/blob/16505e2d3bac948f1ba0394be6005aee22b8e501/app/builders/subscription_confirmation_email_builder.rb
- https://github.com/alphagov/email-alert-api/blob/16505e2d3bac948f1ba0394be6005aee22b8e501/app/services/unpublish_handler_service.rb

I see that as more useful, as we retain the advantage of being able to distinguish between experiments / subsequent uses of this task, so we can distinguish between different groups of lists.

Would it be OK if we continue to deviate?

kevindew: Yup, our usage has been very inconsistent.

I don't mind you leaving it as is, though would suggest that we try focus on analysing this by using utm_content and utm_campaign to at least try not to increase reliance on the field.

benthorner: Oh, that's true. So how about a longer campaign of "govuk-notifications-switch-to-daily-experiment".

How about that?

kevindew: Sounds good 👍

benthorner: Updated.

PR #1359, opened 2020-08-06
Add temporary rake task digest switch experiment

benthorner:
https://trello.com/c/NvjAZOC8/454-get-the-data-needed-to-support-the-experiment

This is primarily so it can be retrieved by our PA on Integration.

### DIFF
--- a/lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
+++ b/lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
@@ -1,0 +1,20 @@
+class Reports::SubscriptionChangesAfterSwitchToDailyDigestReport
+  def self.call
+    updated_subscriptions = Subscription.where(source: 4).where.not(ended_at: nil)
+
+    newer_subscriptions = updated_subscriptions.all.flat_map do |s|
+      Subscription
+        .includes(:subscriber_list)
+        .where(subscriber_id: s.subscriber_id, subscriber_list_id: s.subscriber_list_id)
+        .where("updated_at > ?", s.created_at).order(updated_at: :desc).first
+    end
+
+    puts(CSV.generate do |csv|
+      csv << %w[created_at ended_at frequency subscriber_id subscriber_list_slug]
+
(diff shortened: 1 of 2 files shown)


lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
```diff
@@ -0,0 +1,21 @@
+class Reports::SubscriptionChangesAfterSwitchToDailyDigest
+  def self.call
+    updated_subscriptions = Subscription.where(source: 4).where.not(ended_at: nil)
+
+    newer_subscriptions = updated_subscriptions.all.to_a.flat_map do |s|
+      Subscription
```
1pretz1, line 6: Could we do `Subscription.includes(:subscriber_list)` here and use the loaded relation to populate the subscriber list title on line 17? Slightly simplifies the code and means we can remove line 11.

1pretz1: Thinking about it don't know how well the above suggestion would scale

lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
```diff
@@ -0,0 +1,21 @@
+class Reports::SubscriptionChangesAfterSwitchToDailyDigest
+  def self.call
+    updated_subscriptions = Subscription.where(source: 4).where.not(ended_at: nil)
+
+    newer_subscriptions = updated_subscriptions.all.to_a.flat_map do |s|
```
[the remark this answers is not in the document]
benthorner: Sure, I've removed that.

lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
benthorner: Yes we can/should 👍.

(I learnt about `index_by` recently, and it seems to have overridden my memory of `includes`!)

benthorner:
Adding a "DO NOT MERGE" caveat until we're sure we can get access.

PR #1361, opened 2020-08-10
Remove ignore marked_as_spam column

laurentqro:
Now that the marked_as_spam column has been removed in
https://github.com/alphagov/email-alert-api/pull/1353, we can safely delete this code.


laurentqro:
merged into https://github.com/alphagov/email-alert-api/pull/1353.

PR #1362, opened 2020-08-10
Stop populating subscriber count for digest runs

benthorner:
This data isn't used anywhere, and we have no current need for it.
Once we have stopped populating the field, we can remove the column.

### DIFF
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -10,6 +10,9 @@ class DigestRun < ApplicationRecord
 
   enum range: { daily: 0, weekly: 1 }
 
+  # TEMPORARY (to remove once this column is dropped)
+  self.ignored_columns = %w[subscriber_count]
+
   def mark_complete!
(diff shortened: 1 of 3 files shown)


benthorner:
> It'd be good to also ignore the column in this PR so we don't hit errors when the column is removed.

Could you explain this more? I'm assuming it pertains to archiving, but the column isn't part of the archive.

kevindew:
> > It'd be good to also ignore the column in this PR so we don't hit errors when the column is removed.
> 
> Could you explain this more? I'm assuming it pertains to archiving, but the column isn't part of the archive.

No not related to archiving. ActiveRecord keeps a cache of which columns it expects a database table to have. When a column is removed any queries, for example `DigestRun.find(id)` will query `SELECT<ActiveRecords cached columns> FROM digest_run`. Thus when you don't use `ignored_columns` you have a window between when the column is removed until the application restarts that any queries will fail.

benthorner:
Ahh, that's new to me - nobody ever suspects the cache!

I've added a commit to ignore it. Thanks for pointing it out.

PR #1363, opened 2020-08-10
Remove redundant "subscriber_count" for digest run

benthorner:
We no longer populate this data [1].

[1]: https://github.com/alphagov/email-alert-api/pull/1362

### DIFF
--- a/db/migrate/20200810121128_remove_subscriber_count_from_digest_run.rb
+++ b/db/migrate/20200810121128_remove_subscriber_count_from_digest_run.rb
@@ -1,0 +1,5 @@
+class RemoveSubscriberCountFromDigestRun < ActiveRecord::Migration[6.0]
+  def up
+    remove_column :digest_runs, :subscriber_count
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20200810121128_remove_subscriber_count_from_digest_run.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveSubscriberCountFromDigestRun < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :digest_runs, :subscriber_count # rubocop:disable Rails/ReversibleMigration
```
kevindew, line 3: no biggie, but it'd have been less characters to add `, :integer` here than to ignore the rule.

benthorner: I see what you're saying. While that would make the column replacement, that's not true for the data it holds i.e. this is also a data migration to delete everything. It's in that sense that I'm saying it's irreversible (there's not much point in reversing it).

thomasleese: Would the linting error go away if this was in a `def up` block rather than a `def change` block (and then there wasn't a `def down` block).

benthorner: So it does - updated, thanks for the suggestion.

PR #1364, opened 2020-08-11
Refactor of brexit reporting task

hannako:
Relates to https://github.com/alphagov/email-alert-api/pull/1345

Following some post merge review comments from @kevindew, made some improvements:

- write to stdout instead of to a tmp file
- improve test coverage

### DIFF
--- a/lib/reports/brexit_subscribers_report.rb
+++ b/lib/reports/brexit_subscribers_report.rb
@@ -4,16 +4,18 @@ class Reports::BrexitSubscribersReport
   CSV_HEADERS = %w[title slug tags subscribed unsubscribed immediately daily weekly].freeze
   attr_reader :date
 
-  def self.call(*args)
-    new(args).export_csv
+  def call
+    export_csv
   end
+
+private
 
   def initialize(date = nil)
     @date = date
   end
 
   def parsed_date
-    unless date.empty?
+    unless date.nil?
       @parsed_date ||= Date.parse(date.to_s)
     end
   end
@@ -34,31 +36,19 @@ class Reports::BrexitSubscribersReport
     subscriptions.any?
   end
 
-  def report_for_list(list, csv)
-    csv << [list.title, list.slug, list.tags, list.subscriptions.active.count, list.subscriptions.ended.count, list.subscriptions.active.immediately.count, list.subscriptions.active.daily.count, list.subscriptions.active.weekly.count]
-  end
-
-  def file_name
-    date_identifier = `date +%d%m%y_%H%M%S`.chomp
-    prefix =
-      date.empty? ? "brexit_report" : "brexit_report_subs_before_#{parsed_date}"
-    "/tmp/#{prefix}_#{date_identifier}.csv"
+  def row_data(list)
+    active = list.subscriptions.active
+    inactive = list.subscriptions.ended
+    [list.title, list.slug, list.tags, active.count, inactive.count, active.immediately.count, active.daily.count, active.weekly.count]
   end
 
   def export_csv
-    puts "building csv..."
-    csv = build_csv
-    puts "finished building"
-    File.write(file_name, csv)
-    puts "Done! It's in #{file_name}"
-  end
-
-  def build_csv
     subscriber_lists =
-      date.empty? ? brexit_lists : brexit_lists_before_date
-    CSV.generate do |csv|
-      csv << CSV_HEADERS
-      subscriber_lists.each { |list| report_for_list(list, csv) }
+      date.nil? ? brexit_lists : brexit_lists_before_date
+    CSV($stdout, headers: CSV_HEADERS, write_headers: true) do |csv|
+      subscriber_lists.each do |list|
+        csv << row_data(list)
+      end
     end
   end
 end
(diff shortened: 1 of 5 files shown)


spec/lib/reports/brexit_subscribers_report_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        STR
+        expect { described_class.new(date).call }.to output(empty).to_stdout
+      end
+    end
+
+    context "no date is given" do
+      it "returns all brexit lists" do
+        csv = <<~CSV
+          title,slug,tags,subscribed,unsubscribed,immediately,daily,weekly
+          brexit,/brexit,"{:brexit_checklist_criteria=>{:any=>[""visiting-eu""]}}",1,0,0,1,0
```
hannako, line 26: @sihugh not the double quotes on visiting-eu!

lib/reports/brexit_subscribers_report.rb
```diff
@@ -4,16 +4,18 @@ class Reports::BrexitSubscribersReport
   CSV_HEADERS = %w[title slug tags subscribed unsubscribed immediately daily weekly].freeze
   attr_reader :date
 
-  def self.call(*args)
-    new(args).export_csv
+  def call
+    export_csv
   end
```
kevindew, line 9:
```suggestion
  def self.call(*args)
    new(*args).call
  end
  
  def call
    subscriber_lists = date.nil? ? brexit_lists : brexit_lists_before_date

    CSV($stdout, headers: CSV_HEADERS, write_headers: true) do |csv|
      subscriber_lists.each do |list|
        csv << row_data(list)
      end
  end
```

This would make this a consistent interface with the common use of `.call` in this app. And would mean you could skip the new(date) when calling. An example this is: https://github.com/alphagov/email-alert-api/blob/cd0330133f6ead2f9d1db02e069eff3c6d9ccaaf/app/builders/content_change_email_builder.rb#L6-L13

lib/reports/brexit_subscribers_report.rb
```diff
@@ -4,16 +4,18 @@ class Reports::BrexitSubscribersReport
...
 
+private
+
   def initialize(date = nil)
     @date = date
   end
```
kevindew, line 15: This seems a little surprising and unconventional to be private - I'd suggest moving to be public. 

Also a today-I-learnt moment that this works as private 😮

spec/lib/reports/brexit_subscribers_report_spec.rb
```diff
@@ -0,0 +1,32 @@
+RSpec.describe Reports::BrexitSubscribersReport do
+  describe "#call" do
+    let!(:date) { Time.zone.local(2020, 7, 12) }
+    let(:list) { create(:subscriber_list, :brexit, title: "brexit", slug: "/brexit") }
```
kevindew, line 4: Since we only use the :brexit trait once I think it'd be preferable to set the tags here rather than creating a new trait. It'd save having to look up what the trait means when trying to understand this code.

spec/lib/reports/brexit_subscribers_report_spec.rb
```diff
@@ -0,0 +1,32 @@
+RSpec.describe Reports::BrexitSubscribersReport do
+  describe "#call" do
+    let!(:date) { Time.zone.local(2020, 7, 12) }
```
kevindew, line 3: The bang appears redundant here.

spec/lib/reports/brexit_subscribers_report_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    let(:subscriber) { create(:subscriber) }
+    let!(:subscription) do
+      create(:subscription, :daily,
+             subscriber: subscriber,
+             subscriber_list: list,
+             created_at: (date + 1.day))
+    end
+
+    context "date is given" do
+      it "returns all brexit lists subscribed to before a date" do
```
kevindew, line 14: This it condition doesn't seem to quite match with what is being tested - since the test is checking that an empty output.

I think you might want to have 3 scenarios tested with 2 subscriptions existing in the system: it returns all subscriptions when not given a date (returns the list with 2 results), it returns all subscriptions created before a date (returns the list with 1 result), and it returns an empty CSV when no subscriptions were created before the date. 

What do you think?

spec/lib/tasks/report_spec.rb
```diff
@@ -23,4 +23,18 @@
...
+        .to output.to_stdout
+    end
+  end
+
+  describe "csv_brexit_subscribers_on_or_before" do
+    it "outputs a report of subscriptions to brexit lists" do
+      expect { Rake::Task["report:csv_brexit_subscribers_on_or_before"].invoke("2020-12-07") }
+        .to output.to_stdout
+    end
+  end
```
kevindew, line 39: Super 👍

lib/reports/brexit_subscribers_report.rb
hannako: Ah ok, I misunderstood your comments on the previous pr! I thought your comment there was to remove the `self.call` method altogether and replace it with an instance method, but I see what you mean was to rename the export_csv task ==> call. Thanks for explaining.

spec/lib/reports/brexit_subscribers_report_spec.rb
hannako: Sounds good - thanks!

PR #1365, opened 2020-08-11
Index digest_run_subscribers by subscriber

benthorner:
https://trello.com/c/NvjAZOC8/454-get-the-data-needed-to-support-the-experiment

Currently it's impossible to query this model for a large number of
subscribers, since these records are not naturally ordered by this
column. This adds an index to facilitate a query we want to run, to
determine the number of digest emails sent to a set of subscribers.

Note that, although we wish to run this query of a time period, an
index on "created_at" has little added benefit, since the records
are naturally ordered by this column as-is.


benthorner:
Closing, as @kevindew found a more efficient way to get the data we need (see comment on card).

PR #1372, opened 2020-08-18
Add processed_at field to DigestRun

kevindew:

This is one of the early steps to adapt DigestRun models to support retrying a partially successful one. This introduces a processed_at value which will be used to record whether a DigestRun had completed processing or not (in a similar manner to ContentChange and Message). This initial PR introduces the column and starts populating it for new records. Once we've done a migration to populate this on old records we can then write the code to make use of it.

### DIFF
--- a/db/migrate/20200817151547_add_processed_at_to_digest_runs.rb
+++ b/db/migrate/20200817151547_add_processed_at_to_digest_runs.rb
@@ -1,0 +1,5 @@
+class AddProcessedAtToDigestRuns < ActiveRecord::Migration[6.0]
+  def change
+    add_column :digest_runs, :processed_at, :datetime
+  end
+end
--- a/spec/services/digest_initiator_service_spec.rb
+++ b/spec/services/digest_initiator_service_spec.rb
@@ -10,6 +10,12 @@ RSpec.describe DigestInitiatorService do
           expect { described_class.call(range: range) }
             .to change { DigestRun.daily.count }.from(0).to(1)
         end
+
+        it "marks the digest run as processed" do
+          described_class.call(range: range)
+          digest_run = DigestRun.last
+          expect(digest_run.processed_at).to eq(Time.zone.now)
+        end
       end
 
       context "when a DigestRun already exists" do
--- a/spec/services/digest_initiator_service_spec.rb
+++ b/spec/services/digest_initiator_service_spec.rb
@@ -1,103 +1,50 @@
 RSpec.describe DigestInitiatorService do
   describe ".call" do
-    around { |example| travel_to(Time.zone.parse("08:30")) { example.run } }
+    let(:subscribers) { create_list(:subscriber, 2) }
 
-    context "daily" do
-      let(:range) { Frequency::DAILY }
+    before do
+      allow(DigestRunSubscriberQuery).to receive(:call).and_return(subscribers)
+      allow(DigestEmailGenerationWorker).to receive(:perform_async)
+    end
 
-      context "when there is no daily DigestRun for the date" do
-        it "creates one" do
-          expect { described_class.call(range: range) }
-            .to change { DigestRun.daily.count }.from(0).to(1)
-        end
+    context "when a digest run doesn't exist" do
+      it "creates a digest run for the current data with the specified range" do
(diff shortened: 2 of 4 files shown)


db/migrate/20200817151547_add_processed_at_to_digest_runs.rb
```diff
@@ -0,0 +1,5 @@
+class AddProcessedAtToDigestRuns < ActiveRecord::Migration[6.0]
```
benthorner, line 1: Minor: the commit message could be more precise:

> we have no evidence that the run was incomplete

Suggest: "not processed", since we do know it's "incomplete" (assuming it was created).

spec/services/digest_initiator_service_spec.rb
```diff
@@ -1,103 +1,48 @@
...
-        it "enqueues a DigestEmailGenerationWorker job" do
-          expect(DigestEmailGenerationWorker)
-            .to receive(:perform_async)
-            .exactly(2).times
-
-          described_class.call(range: range)
-        end
+      it "creates a DigestRunSubscriber for each subscription" do
+        expect { described_class.call(range: Frequency::DAILY) }
+          .to change { DigestRunSubscriber.exists?(subscriber_id: subscribers.map(&:id)) }
```
benthorner, line 31: Minor: you could do `subscriber: subscribers`.

spec/services/digest_initiator_service_spec.rb
```diff
@@ -1,103 +1,48 @@
...
       end
 
-      it "records a metric for the delivery attempt" do
-        expect(Metrics).to receive(:digest_initiator_service)
-          .with("daily")
-
-        described_class.call(range: range)
+      it "enqueus DigestEmailGenerationWorker for each DigestRunSubscriber" do
+        described_class.call(range: Frequency::DAILY)
+        expect(DigestEmailGenerationWorker).to have_received(:perform_async).twice
```
benthorner, line 37: Minor: we could/should check the arguments, like we do [here](https://github.com/alphagov/email-alert-api/blob/master/spec/workers/process_content_change_worker_spec.rb#L39).

spec/services/digest_initiator_service_spec.rb
```diff
@@ -1,103 +1,48 @@
...
+          .to change { DigestRunSubscriber.exists?(subscriber_id: subscribers.map(&:id)) }
+          .to(true)
       end
 
-      it "records a metric for the delivery attempt" do
-        expect(Metrics).to receive(:digest_initiator_service)
-          .with("daily")
-
-        described_class.call(range: range)
+      it "enqueus DigestEmailGenerationWorker for each DigestRunSubscriber" do
```
benthorner, line 35:
```suggestion
      it "enqueues DigestEmailGenerationWorker for each DigestRunSubscriber" do
```

benthorner (approved):
🇹🇳

db/migrate/20200817151547_add_processed_at_to_digest_runs.rb
kevindew: Sure, I've updated the message

spec/services/digest_initiator_service_spec.rb
kevindew: d'oh! hopefully that's one I copied and pasted from the existing tests.

spec/services/digest_initiator_service_spec.rb
kevindew: Sure, I've added something in for this. I didn't think it was too necessary really as the ids we test are created as part of the test but not too concerned about it being in there either.

spec/services/digest_initiator_service_spec.rb
kevindew: Whoops thanks.

kevindew:
Should be good for another look @benthorner

spec/services/digest_initiator_service_spec.rb
benthorner: Ah, I was thinking more about the interface boundary: checking we're passing the arguments as expected (vs. the specific IDs).

benthorner (approved):

PR #1373, opened 2020-08-18
Remove superfluous model methods and make remaining more conventional

kevindew:

This is a bit of a tangential exercise as part of the Trello card, when doing https://github.com/alphagov/email-alert-api/pull/1372 I noticed it was a unusual how we had `mark_processed!` bang methods [since we didn't have safe equivalents](https://rubystyle.guide/#dangerous-method-bang). I then realised most of these were rather superfluous so I did some trimming.


app/models/digest_run.rb
```diff
@@ -10,17 +10,11 @@ class DigestRun < ApplicationRecord
...
 
-  def mark_complete!
-    completed_at = digest_run_subscribers.maximum(:completed_at) || Time.zone.now
-    update!(completed_at: completed_at)
-  end
-
-  def check_and_mark_complete!
-    mark_complete! unless has_incomplete_subscribers?
-  end
+  def check_if_completed
```
1pretz1, line 13: Could this be renamed to `check_and_mark_complete`? I wouldn't expect a method called `check_if_completed` to actually update anything

1pretz1 (commented):
Nice bit of refactoring Kev, only one minor comment.

app/models/digest_run.rb
kevindew: Sure, I didn't think `check_and_mark_complete?` seemed right since this didn't return a boolean. I also found `check_and_mark_complete` didn't sound great as it didn't capture the conditional nature if the mark complete. I understand the concern with `check_if_completed` it does sound more like a query than a state change.

I've addressed this by moving the conditional to the worker for this.

benthorner (approved):
Much better to be testing the actual values set, than a `?` method 👍

PR #1374, opened 2020-08-18
Unique index on digest_run_subscribers

kevindew:

This applies a unique index to digest_run_subscribers columns (at the time of writing there are no duplicates so it can be safely applied).

This is done so that we can resume the processing of a digest run were one to fail part way through.

I was worried this might slow down inserts for digest_run_subscribers significantly but much to my surprise it showed a performance improvement. Before todays DigestRun took 1291 seconds on integration whereas after it took 918 seconds.


1pretz1 (approved):
Looks good to me! 🎲

PR #1376, opened 2020-08-19
Tweak migration and reporting for Experiment 2.0

benthorner:
https://trello.com/c/QF69iYHE/489-start-experiment-20

    This updates the migration and reporting tasks so they're scoped to
    the lists we're migrating in the second experiment to switch from
    individual emails to daily digests. Note that the criteria to select
    the lists for the experiment are complex, involving knowledge of
    previously migrated lists (which is about to change) and external
    data from the Publishing API.


lib/experiment_2_slugs
```diff
@@ -0,0 +1,4880 @@
+guidance-and-regulation-with-1-topic-ce3f6f1b86
```
kevindew, line 1: I'm a bit surprised about this file being in lib, it feels more like config? perhaps useful to give it a .txt extension too.

lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
```diff
@@ -1,6 +1,13 @@
 class Reports::SubscriptionChangesAfterSwitchToDailyDigestReport
   def self.call
-    updated_subscriptions = Subscription.where(source: 4).where.not(ended_at: nil)
+    experiment_2_list_ids = SubscriberList.where(
+      slug: File.readlines(Rails.root.join("lib/experiment_2_slugs")),
+    )
+
+    updated_subscriptions = Subscription
+      .where(source: 4)
```
kevindew, line 8: minor, you can make this a bit more readable with either Subscription.where(source: :bulk_immediate_to_digest)` or even `Subscription.source_bulk_immediate_to_digest`

lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
```diff
@@ -1,6 +1,13 @@
...
+      .where(subscriber_list_id: experiment_2_list_ids)
```
kevindew, line 9: Maybe I've missed something but it looks like this a collection of models and not the ids? Is there a test that executes this code?

kevindew (commented):
Wow beasty list. Looks like it'll be impactful. What proportion of subscriptions will be digest after this?

lib/experiment_2_slugs
benthorner: Sure, I've done that. I've got a few checklist items to remove it soonish anyway.

lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
benthorner: Oh yeah - for some reason originally this didn't work. Updated.

lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb
benthorner: Oh yes, I forgot to `pluck`. Updated.

I've only tested this manually, since it's only a short-lived report we're running most days.

lib/tasks/data_migration.rake
```diff
@@ -36,43 +38,8 @@ namespace :data_migration do
...
-      "international-development-funding",
-      "the-charity-commission",
-      "theory-tests",
-      "standards-and-testing-agency",
-      "personal-tax-self-assessment",
-      "planning-and-development-planning-officer-guidance",
-      "civil-service-fast-track-apprenticeship",
-      "further-education-and-skills-apprenticeships",
+    Rake::Task["data_migration:switch_to_daily_digest"].invoke(
+      *CSV.read(Rails.root.join("config/experiment_2_slugs.txt")).flatten,
```
kevindew, line 42: The CSV confused me a little, I think you can achieve the same with `File.read(path).split("\n")`

lib/tasks/data_migration.rake
```diff
@@ -10,6 +10,8 @@ namespace :data_migration do
     subscribers = Subscriber.where(id: subscriptions.pluck(:subscriber_id)).index_by(&:id)
     subscriptions_by_subscriber = subscriptions.group_by(&:subscriber_id).transform_keys { |k| subscribers[k] }
 
+    puts 'hi'
```
kevindew, line 13: Some debugging caught behind - or a cryptic message to someone 🙂

lib/tasks/data_migration.rake
benthorner: Sure, thanks - I was rushing this a bit.

lib/tasks/data_migration.rake
benthorner: Yeah, I was investigating a performance problem. It's gone now.

benthorner:
@kevindew thanks for your comments so far. Although we have tests for the migration task, running it on Integration showed there was a significant performance problem. I've added another commit to address that. 

Could you take another look? All your original comments should also be addressed.

lib/tasks/data_migration.rake
```diff
@@ -1,78 +1,61 @@
...
+              subscriber_id: subscription.subscriber_id,
+              subscriber_list_id: subscription.subscriber_list_id,
+              frequency: :daily,
+              source: :bulk_immediate_to_digest,
+              created_at: now,
+              updated_at: now,
+            }
+          end
+
+          Subscription.where(id: immediate_subscriptions).update_all(
```
kevindew, line 36: This looks like there's a type mismatch where you're missing a pluck?

lib/tasks/data_migration.rake
```diff
@@ -1,78 +1,61 @@
...
-            source: :bulk_immediate_to_digest,
+          email_id = SwitchToDailyDigestEmailBuilder.call(
+            subscriber, immediate_subscriptions
           )
         end
 
-        email_id = SwitchToDailyDigestEmailBuilder.call(subscriber, immediate_subscriptions)
+        DeliveryRequestWorker.perform_async_in_queue(email_id, queue: :default)
+      rescue StandardError => e
+        puts "Skipping subscriber: #{e}"
```
kevindew, line 50: Minor, this is unlikely to be a single subscriber anymore.

kevindew (commented):
Looks good, just a small issue I think I've caught.

lib/tasks/data_migration.rake
benthorner: ActiveRecord was converting them automatically. It's not a query anymore, so no `pluck`. Still good to be explicit, though 👍.

lib/tasks/data_migration.rake
benthorner: I think it's right: the `rescue` is around the `subscriptions_with_subscriber.each` block.

kevindew: Oh sorry! misread where it was rescuing

kevindew (approved):

lib/tasks/data_migration.rake
kevindew: Oh cool TIL that you can do that.

PR #1377, opened 2020-08-19
Stop suggesting individual emails are "immediate"

benthorner:
https://trello.com/c/QF69iYHE/489-start-experiment-20


kevindew (approved):
pre-emptive tick assuming no-one has concerns about the content.

PR #1381, opened 2020-08-20
Adapt DigestEmailGenerationWorker for allowing retries

kevindew:

This adapts this worker so that it can abort elegantly should it be called in a situation it be re-queued after completing by checking whether the item is processed or not.

It applies a FOR UPDATE database lock on the DigestRunSubscriber object when this runs, this is to create isolation that will prevent any concurrent attempts to run this worker - without this if we ended up with this worker running twice, creating two emails and leaving the app in a somewhat inconsistent state.

I also refined the tests a little by adding a more specific clause to creating an email and removing a superfluous one that tests we don't change digest run completed_at.

Finally, I did some renaming after being confused by some references to `subscription_content`.


benthorner (approved):
I like the renaming - the previous one was definitely confusing! Have you considered something that implies the value is a list? I was wondering about `digest_items` - only a minor suggestion, though.

kevindew:
> I like the renaming - the previous one was definitely confusing! Have you considered something that implies the value is a list? I was wondering about `digest_items` - only a minor suggestion, though.

Sure, that sounds a good suggestion it does portray a list more. I did wonder if "item" is perhaps too generic a term, but we're essentially using content in the same way. I've updated the PR.

spec/workers/digest_email_generation_worker_spec.rb
```diff
@@ -104,7 +102,7 @@
     end
 
     context "when there are no content changes to send" do
```
benthorner, line 104:
```suggestion
    context "when there are no digest items to send" do
```

This refers to messages now, so slightly misleading.

PR #1382, opened 2020-08-20
Make the DigestInitiatorService retriable

kevindew:

This updates this service so that it can support retrying a DigestRun
that didn't complete. This changes the approach for deciding whether to
abort a digest initiation from checking whether the digest run model
exists to instead checking whether that digest run model has a
populated processed_at field.

To prevent risks that could occur from this code being run concurrently
I've moved all of the service code inside the database lock.


1pretz1 (approved):
Looks good! 🦖

PR #1384, opened 2020-08-24
Rake task to bulk email all Brexit subscribers

hannako:
## What

A rake task that will email all subscribers of the Brexit checker.

trello https://trello.com/c/o8NIyi3H/322-check-if-we-can-email-existing-subscribers-about-new-campaign-actions-to-encourage-users-to-revisit-the-checker

### DIFF
--- a/lib/tasks/bulk_email.rake
+++ b/lib/tasks/bulk_email.rake
@@ -6,9 +6,14 @@ namespace :bulk_email do
       body: ENV.fetch("BODY"),
       subscriber_lists: SubscriberList.where(id: args.extras),
     )
-
     email_ids.each do |id|
       DeliveryRequestWorker.perform_async_in_queue(id, queue: :delivery_immediate)
     end
+  end
+
+  desc "Email all subscribers of the Brexit checker"
+  task brexit_subscribers: :environment do
+    brexit_lists = SubscriberList.where("subscriber_lists.tags->>'brexit_checklist_criteria' IS NOT NULL")
+    Rake::Task["bulk_email:for_lists"].invoke(*brexit_lists)
   end
 end
--- a/spec/lib/tasks/bulk_email_spec.rb
+++ b/spec/lib/tasks/bulk_email_spec.rb
@@ -1,4 +1,45 @@
 RSpec.describe "bulk_email" do
+  describe "brexit_subscribers" do
+    let(:tags) { { brexit_checklist_criteria: { any: %w[visiting-eu] } } }
+    let!(:list) { create(:subscriber_list, tags: tags) }
+    before do
+      Rake::Task["bulk_email:brexit_subscribers"].reenable
+      Rake::Task["bulk_email:for_lists"].reenable
+    end
+
+    around(:each) do |example|
+      ClimateControl.modify(SUBJECT: "subject", BODY: "body") do
+        example.run
+      end
+    end
+
+    it "builds emails for a subscriber list" do
+      expect(BulkSubscriberListEmailBuilder)
+        .to receive(:call)
+        .with(subject: "subject",
+              body: "body",
+              subscriber_lists: [list])
+        .and_call_original
+
+      Rake::Task["bulk_email:brexit_subscribers"].invoke
+    end
+
+    it "enqueues the emails for delivery" do
+      allow(BulkSubscriberListEmailBuilder).to receive(:call)
+        .and_return([1, 2])
+
+      expect(DeliveryRequestWorker)
+        .to receive(:perform_async_in_queue)
+        .with(1, queue: :delivery_immediate)
+
+      expect(DeliveryRequestWorker)
+        .to receive(:perform_async_in_queue)
+        .with(2, queue: :delivery_immediate)
+
+      Rake::Task["bulk_email:brexit_subscribers"].invoke
+    end
+  end
+
   describe "for_lists" do
     before do
       Rake::Task["bulk_email:for_lists"].reenable


lib/tasks/bulk_email.rake
```diff
@@ -11,4 +11,25 @@ namespace :bulk_email do
...
+
+  desc "Email all subscribers of the Brexit checker"
+  task :brexit_subscribers, [:dry_run] => :environment do |_t, args|
+    brexit_list_ids = SubscriberList.where("subscriber_lists.tags->>'brexit_checklist_criteria' IS NOT NULL").pluck(:id)
+    email_ids =
+      brexit_list_ids.map do |id|
+        BulkSubscriberListEmailBuilder.call(
+          subject: ENV.fetch("SUBJECT"),
+          body: ENV.fetch("BODY"),
+          subscriber_lists: SubscriberList.where(id: id),
```
benthorner, line 23: You could just pass the lists directly here, instead of plucking.

lib/tasks/bulk_email.rake
```diff
@@ -11,4 +11,25 @@ namespace :bulk_email do
...
+        )
+      end
+    if args.dry_run == "not_dry_run"
```
benthorner, line 26: I would avoid this, as it will generate alerts about unsent emails, and they might accidentally get sent (since they exist in the system). How about just testing it on Integration or Staging, instead of a dry run feature?

lib/tasks/bulk_email.rake
```diff
@@ -11,4 +11,25 @@ namespace :bulk_email do
...
+      email_ids.flatten.each do |id|
```
benthorner, line 27: Minor: using `flat_map` above would DRY up `flatten` here.

benthorner (commented):
Sorry for the delay in reviewing this (I've been on leave). The actual functionality looks good to me. I've got one more major comment about the "dry run" functionality, though.

lib/tasks/bulk_email.rake
hannako: Thanks 👍

lib/tasks/bulk_email.rake
hannako: ok will do, thanks

lib/tasks/bulk_email.rake
hannako: lovely thank you

spec/lib/tasks/bulk_email_spec.rb
```diff
@@ -1,4 +1,46 @@
 RSpec.describe "bulk_email" do
+  describe "brexit_subscribers" do
+    let(:tags) { { brexit_checklist_criteria: { any: %w[visiting-eu] } } }
+    let!(:list) { create(:subscriber_list, tags: tags) }
+    let(:task) { Rake::Task["bulk_email:brexit_subscribers"].invoke }
```
benthorner, line 5: Do we need this? We seem to be [invoking it manually later anyway](https://github.com/alphagov/email-alert-api/pull/1384/files#diff-a7e3b6edd154b4653f447b46ea4c18feR25).

lib/tasks/bulk_email.rake
```diff
@@ -11,4 +11,22 @@ namespace :bulk_email do
...
+
+  desc "Email all subscribers of the Brexit checker"
+  task brexit_subscribers: :environment do
+    brexit_lists = SubscriberList.where("subscriber_lists.tags->>'brexit_checklist_criteria' IS NOT NULL")
+    email_ids =
+      brexit_lists.flat_map do |list|
+        BulkSubscriberListEmailBuilder.call(
+          subject: ENV.fetch("SUBJECT"),
+          body: ENV.fetch("BODY"),
+          subscriber_lists: list,
```
benthorner, line 23: We should [pass all of the lists here](https://github.com/alphagov/email-alert-api/blob/master/app/builders/bulk_subscriber_list_email_builder.rb#L2), instead of iterating over them: if we don't, then e could end up sending multiple emails if someone has multiple checker subs. You could actually simplify the whole task by doing:

```
Rake::Task["bulk_email:for_lists"].invoke(*brexit_lists)
```

lib/tasks/bulk_email.rake
```diff
@@ -11,4 +11,22 @@ namespace :bulk_email do
       DeliveryRequestWorker.perform_async_in_queue(id, queue: :delivery_immediate)
     end
   end
+
+  desc "Email all subscribers of the Brexit checker"
+  task brexit_subscribers: :environment do
+    brexit_lists = SubscriberList.where("subscriber_lists.tags->>'brexit_checklist_criteria' IS NOT NULL")
+    email_ids =
+      brexit_lists.flat_map do |list|
```
benthorner, line 19: Minor: the collection of lists should already be flat at this point.

benthorner (commented):
Thanks for taking a look @hannako. In reviewing again, I found another more major issue that I didn't spot last time - sorry about that!

lib/tasks/bulk_email.rake
hannako: Thanks Ben, that is way nicer.

spec/lib/tasks/bulk_email_spec.rb
hannako: I've removed it thanks

hannako:
@benthorner Thank you Ben, have made the changes and squashed the commits.

benthorner (approved):
Looks good - thanks for following-up on the suggestions. I reckon there should be a way to DRY-up the test too e.g. by stubbing the call to the `for_lists` task. But it's nothing major.

PR #1386, opened 2020-09-01
[DNM] Update dfid_research_output subscriber lists

1pretz1:
This temp rake tasks is to migrate dfid_research_output subscriber lists
over to research_for_development_output.

In email-alert-api there are a few different types of
dfid_research_output lists. We're migrating the dfid_research_output
finder and all lists associated with the dfid_research_output document
type.

I've tested this in integration and all the lists have been updated
successfully with no mention of dfid_research_output in any other
subscriber list. The rake task also only takes a few seconds to complete.

Trello:
https://trello.com/c/OxoDZj59/523-migrate-subscribers-for-dfid-finder

### DIFF
--- a/lib/tasks/update_dfid_subscriber_lists.rake
+++ b/lib/tasks/update_dfid_subscriber_lists.rake
@@ -1,0 +1,34 @@
+desc "Update DFID subscriber lists"
+task update_dfid_subscriber_lists: :environment do
+  # Update all subscriber lists subscribed to dfid_research_output document types
+  SubscriberList.where("links->'content_store_document_type'->>'any' LIKE '%dfid_research_output%'").each do |sl|
+    new_links = sl.links.merge(
+      content_store_document_type: {
+        any: (sl.links[:content_store_document_type][:any] - %w[dfid_research_output] + %w[research_for_development_output]),
+      },
+    )
+
+    sl.update!(links: new_links)
+  end
+
+  # Update all subscriber lists subscribed to dfid_research_output document types (no idea what makes these different to the links equivalent)
+  SubscriberList.where("tags->'content_store_document_type'->>'any' LIKE '%dfid_research_output%'").each do |sl|
+    new_tags = sl.tags.merge(
+      content_store_document_type: {
+        any: (sl.tags[:content_store_document_type][:any] - %w[dfid_research_output] + %w[research_for_development_output]),
+      },
+    )
+
+    sl.update!(tags: new_tags)
+  end
+
+  # Update subscriber list associated with the dfid_research_output finder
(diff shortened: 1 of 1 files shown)


edwardkerry (commented):
I'm not familiar enough with email-alert-api to approve but this looks to swap the correct document type 👍

lib/tasks/update_dfid_subscriber_lists.rake
```diff
@@ -0,0 +1,34 @@
...
+      content_store_document_type: {
+        any: (sl.tags[:content_store_document_type][:any] - %w[dfid_research_output] + %w[research_for_development_output]),
+      },
+    )
+
+    sl.update!(tags: new_tags)
+  end
+
+  # Update subscriber list associated with the dfid_research_output finder
+  sl = SubscriberList.find_by("tags->'format'->>'any' LIKE '%dfid_research_output%'")
```
kevindew, line 26: Any reason why this approach is done differently to the others?

lib/tasks/update_dfid_subscriber_lists.rake
```diff
@@ -0,0 +1,34 @@
...
+      content_store_document_type: {
+        any: (sl.links[:content_store_document_type][:any] - %w[dfid_research_output] + %w[research_for_development_output]),
+      },
+    )
+
+    sl.update!(links: new_links)
+  end
+
+  # Update all subscriber lists subscribed to dfid_research_output document types (no idea what makes these different to the links equivalent)
+  SubscriberList.where("tags->'content_store_document_type'->>'any' LIKE '%dfid_research_output%'").each do |sl|
```
kevindew, line 15: It is a bit baffling about this, there seems to be only 2 of these whereas 131 of the link variant. The link one is perplexing too given that they are not links (i.e. not a content_id) - there's probably a grim cleanup job waiting in the future.

lib/tasks/update_dfid_subscriber_lists.rake
kevindew: Ah I see it's that there is just 1 entry.

kevindew (approved):
Looks good to me, have you tried it out on integration?

I did wonder if you might accidentally squash anything with a `sl.tags[:content_store_document_type][:all]` but it doesn't look like there is any uses of that in this dataset.

lib/tasks/update_dfid_subscriber_lists.rake
1pretz1: It's very baffling 🤷‍♂️ I think as we've discovered in previous work that our subscriber lists are well due a cleanup!

1pretz1:
Thanks for the reviews both @kevindew & @edwardkerry!

I've tested in integration and all works as expected so I'm going to merge away.

PR #1388, opened 2020-09-02
Close subscriptions for FCO/DFID lists

kevindew:
Trello: https://trello.com/c/b5TF5mWv/499-migrate-fco-and-dfid-subscriptions-to-fcdo

This should not be merged until 8th September as per the schedule in the Trello card.

This adds a rake task which identifies lists that are associated with
FCO and DFID organisations and closes any subscriptions to them. It
filters out any lists that are associated with additional organisations.

This task has been put together to reflect that these lists will no
longer receive any emails now that the FCO and DFID organisations are
closed and that these subscriptions have become superfluous.

Once this has task has been run on September 8th (as per the current
schedule) it can be deleted from the code.

### DIFF
--- a/lib/tasks/clean.rake
+++ b/lib/tasks/clean.rake
@@ -5,4 +5,35 @@ namespace :clean do
     cleaner = Clean::EmptySubscriberLists.new
     cleaner.remove_empty_subscriberlists(dry_run: dry_run)
   end
+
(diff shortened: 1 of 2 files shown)


benthorner (approved):
Looks good to me. I'm assuming the list query is the same as what we've used previously to determine the lists associated with these organisations.

PR #1265, opened 2020-06-08
Add report for matched content changes

benthorner:
https://trello.com/c/IAj6XWFo/309-create-a-report-to-show-the-content-changes-that-have-triggered-notifications

    This adds a report to output content changes by matching subscriber
    list over a given period, calculating the number of emails each match
    would have generated using a similar technique to the "active_on" scope
    on the Subscription model. We want to use this report to see clusters of
    changes that contribute to high email volume.

PR #1392, opened 2020-09-07
Reinstate simplified report for content changes

benthorner:
https://trello.com/c/bKAemKqV/517-review-the-emails-that-users-have-received-for-driving-subscription-lists

Previously we had a report for content changes matched against each
subscriber list with its immediate subscribers, as a way to discover
the origin of email volume [1]. Since then, we have begun reporting
email volume by publishing app and document type [2], and have done
migrations to increase the use of daily digest emails, to the extent
that immediate subscriber numbers are no longer meaningful as a way
to estimate volume, and are time consuming to include. This reinstates
a report of just the matched content changes, in the spirit of making
this data more easily available [3], and to support our research into
the behaviour of specific lists.

[1]: https://github.com/alphagov/email-alert-api/pull/1265
[2]: https://github.com/alphagov/email-alert-api/pull/1323
[3]: https://trello.com/c/Izm4d6NM/454-epic-make-data-about-email-notifications-easily-available

### DIFF
--- a/lib/reports/matched_content_changes_report.rb
+++ b/lib/reports/matched_content_changes_report.rb
@@ -1,0 +1,33 @@
+class Reports::MatchedContentChangesReport
+  OUTPUT_ATTRIBUTES = {
+    created_at: :content_change,
+    base_path: :content_change,
+    change_note: :content_change,
+    document_type: :content_change,
+    publishing_app: :content_change,
+    priority: :content_change,
+    title: :subscriber_list,
+    slug: :subscriber_list,
+  }.freeze
+
+  def call(start_time: nil, end_time: nil)
+    start_time = start_time ? Time.zone.parse(start_time) : 1.week.ago
+    end_time = end_time ? Time.zone.parse(end_time) : Time.zone.now
+
+    query = MatchedContentChange
+      .includes(:subscriber_list, :content_change) # for efficient "row.content_change"
+      .joins(:content_change) # for the next query
+      .where("content_changes.created_at": start_time..end_time)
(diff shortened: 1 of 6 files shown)


lib/reports/matched_content_changes_report.rb
```diff
@@ -0,0 +1,32 @@
...
+  }.freeze
+
+  def call(start_time: nil, end_time: nil)
+    start_time = start_time ? Time.zone.parse(start_time) : 1.week.ago
+    end_time = end_time ? Time.zone.parse(end_time) : Time.zone.now
+
+    query = MatchedContentChange
+      .includes(:subscriber_list, :content_change) # for efficient "row.content_change"
+      .joins(:content_change) # for the next query
+      .where("content_changes.created_at": start_time..end_time)
```
kevindew, line 20: It'd be good to have an ordering on this, otherwise the report might be rather surprising with same content change spread out across the results.

kevindew (commented):
Looks good, just a single suggestion.

Also are we documenting these report tasks anywhere?

lib/reports/matched_content_changes_report.rb
benthorner: Sure, I've added one.

benthorner:
> Also are we documenting these report tasks anywhere?

I was wondering about that. My original thought was that we're (now) intending to automate away all these ad-hoc tasks, so that developer-level access isn't necessary to get basic data out of the system. However, until that glorious moment, I agree it's worth writing it down. Perhaps the change will be larger than you were expecting, though ;-).

kevindew (approved):

PR #1393, opened 2020-09-07
Revert "Try speeding things up with the spring gem"

benthorner:
This reverts commit 35e69853ecead13d0d9c080af65b84cdfc36dc6a.

Related to: https://github.com/alphagov/govuk-docker/pull/395

In practice I didn't find myself using this much. On the occasions
I did, one problem I encountered was that the code in the spring
server took some time to reload, such that the tests would report
some strange system-level error if I ran them too quickly after.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,6 +34,4 @@ group :development, :test do
   gem "pry-byebug"
   gem "rspec-rails"
   gem "rubocop-govuk"
-  gem "spring"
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Shame it didn't work out so good, I never found myself getting into the swing of using it. Nice one for cleaning it up though.

PR #1395, opened 2020-09-08
Upgrade to ruby 2.7.1

cdccollins:
These are the changes necessary for the app to run on Ruby 2.7.1.

I have left the version of Ruby the same, since there is a script that can be run after all the apps have been updated to update the version all at once. 

Still some warnings coming from Rspec-mocks, the fix has not been merged yet

### DIFF
--- a/app/builders/bulk_subscriber_list_email_builder.rb
+++ b/app/builders/bulk_subscriber_list_email_builder.rb
@@ -5,8 +5,8 @@ class BulkSubscriberListEmailBuilder
     @subscriber_lists = subscriber_lists
   end
 
-  def self.call(*args)
-    new(*args).call
+  def self.call(**args)
+    new(**args).call
   end
 
   def call
--- a/app/builders/digest_email_builder.rb
+++ b/app/builders/digest_email_builder.rb
@@ -6,8 +6,8 @@ class DigestEmailBuilder
     @subscriber_id = subscriber_id
   end
 
-  def self.call(*args)
-    new(*args).call
+  def self.call(**args)
+    new(**args).call
   end
 
   def call
--- a/app/builders/message_email_builder.rb
+++ b/app/builders/message_email_builder.rb
@@ -22,12 +22,12 @@ private
     @records ||= begin
       now = Time.zone.now
 
-      recipients_and_messages.map do |address:, message:, subscriptions:, subscriber_id:|
+      recipients_and_messages.map do |x|
         {
-          address: address,
-          subject: subject(message),
-          body: body(message, subscriptions, address),
-          subscriber_id: subscriber_id,
+          address: x[:address],
+          subject: subject(x[:message]),
+          body: body(x[:message], x[:subscriptions], x[:address]),
+          subscriber_id: x[:subscriber_id],
           created_at: now,
           updated_at: now,
         }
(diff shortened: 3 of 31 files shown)


.ruby-version
```diff
@@ -1 +1 @@
-2.6.6
+2.7.0
```
thomasleese, line 1: Did you mean for this change to be in the PR?

cdccollins: Oops, no, changing it now :)

kevindew:
I think it'd be beneficial to include the .ruby-version change as part of this for a few reasons: allows using syntax for 2.7, makes it easier to double check in review the quantity of warnings we'll have to deal with, and one less thing to review/deploy when ready.

For the places where we're doing argument forwarding it'd be good if we could use the `...` operator in 2.7 (sidenote, I'm not sure that operator has a name? I call it the ellipsis one). For example we could change:

https://github.com/alphagov/email-alert-api/blob/8c948193b1c2b07f699664c188b8b723cb7595c5/app/services/application_service.rb#L2-L4

to

```
  def self.call(...)
    new(...).call
  end
```

which would mean we could avoid adding the double splats in: app/services/immediate_email_generation_service.rb, app/services/matched_content_change_generation_service.rb and a few more.

I notice we still get about 5 warnings still about kwargs in 2.7, is that something we're experiencing in most repos? I see CSV and RSpec are the culprits here, do we know if they have upgrade plans?

pixeltrix:
> For the places where we're doing argument forwarding it'd be good if we could use the `...` operator in 2.7 (sidenote, I'm not sure that operator has a name? I call it the ellipsis one). For example we could change:

Operator forwarding was only added in Ruby 2.7 so we'd need to deploy either both the Ruby upgrade and the app at the same time or deploy the Ruby upgrade first and then the app upgrade to fix it with Ruby 2.7. I'd be nervous about using a version specific feature during the deployment process and maybe that's something that should be addressed via a [Rubocop rule][1] in a later deploy?

> I notice we still get about 5 warnings still about kwargs in 2.7, is that something we're experiencing in most repos? I see CSV and RSpec are the culprits here, do we know if they have upgrade plans?

Looks like the CSV error is because the app is using the `CSV()` singleton method which is a pretty dated pattern - we can probably re-write the code to avoid the method. The [RSpec fix] is yet to be merged.

[1]: https://github.com/rubocop-hq/rubocop/pull/7646
[2]: https://github.com/rspec/rspec-mocks/pull/1324

pixeltrix:
@kevindew seems like you suggested to use `CSV()` in this [comment][1] - any preference whether we should replace it with `CSV.instance` or `CSV.new` ?

kevindew:
> @kevindew seems like you suggested to use `CSV()` in this [comment](https://github.com/alphagov/email-alert-api/pull/1364#discussion_r469179526) - any preference whether we should replace it with `CSV.instance` or `CSV.new` ?

Very sorry for the long delay, I started writing a reply but must have never finished.

I just suggested using CSV directly for consistency with [our existing approach](https://github.com/alphagov/email-alert-api/blob/0e04a7537f3302ab010dd13153c19d1529a56068/lib/reports/living_in_europe_report.rb#L70) - I don't know enough about the choice to have a preference, though `.new` sounds more idiomatic. Happy with whatever feels most appropriate.

benbaumann95 (approved):

PR #1396, opened 2020-09-08
Stop sending finished_sending_at & sent to athena

1pretz1:
The reason we are stopping sending these is they may change after an
email has been archived, which makes them misleading. For archiving
purposes, it should be enough to know when the email was created, since
this is usually similar to when it was sent.

Also removes an unused let in EmailArchiveQuery.

Related PR:
https://github.com/alphagov/govuk-aws/pull/1337

Trello:


benthorner (approved):
Nice 🇹🇭. I was interested to see the grouping functionality is unchanged in the tests, since we were using the same timestamp for everything.

From my own experience, it's safe to change the data being sent, since the schema / Athena is more like an "overlay" on the files in S3.

PR #1399, opened 2020-09-10
Make it possible to retry digests on different days

benthorner:
https://trello.com/c/6NVuDOF5/525-make-it-possible-to-fix-digest-emails-if-they-break-over-the-weekend

### DIFF
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -24,13 +24,13 @@ private
   end
 
   def ends_at_is_in_the_past
-    errors.add(:ends_at, "must be in the past") if ends_at >= Time.zone.now
+    errors.add(:date, "must be after #{DIGEST_RANGE_HOUR}:00") if ends_at >= Time.zone.now
   end
 
   def weekly_digest_is_on_a_saturday
-    return if daily? || ends_at.saturday?
+    return if daily? || date.saturday?
 
-    errors.add(:ends_at, "must be a Saturday for weekly digests")
+    errors.add(:date, "must be a Saturday for weekly digests")
   end
 
   def starts_at_time
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -18,17 +18,9 @@ class DigestRun < ApplicationRecord
 
 private
 
-  def starts_at=(value)
-    super
-  end
-
-  def ends_at=(value)
-    super
-  end
-
   def set_range_dates
-    self.starts_at = configured_starts_at
-    self.ends_at = configured_ends_at
+    self.starts_at = Time.zone.parse("#{DIGEST_RANGE_HOUR}:00", starts_at_time)
+    self.ends_at = Time.zone.parse("#{DIGEST_RANGE_HOUR}:00", date)
   end
 
   def ends_at_is_in_the_past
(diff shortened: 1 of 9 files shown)


app/models/digest_run.rb
```diff
@@ -40,18 +42,14 @@ def weekly_digest_is_on_a_saturday
...
+    Time.zone.parse("#{DIGEST_RANGE_HOUR}:00", date)
   end
 
   def starts_at_time
     (daily? ? date - 1.day : date - 1.week)
   end
-
-  def digest_range_hour
-    ENV.fetch("DIGEST_RANGE_HOUR", 8).to_i
-  end
```
kevindew, line 49: Nice clean-up

spec/models/digest_run_spec.rb
```diff
@@ -31,20 +31,20 @@
...
           expect(instance.starts_at).to eq(
-            Time.zone.parse("08:00", (date - 1.week)),
+            Time.zone.parse("08:00", (saturday - 1.week)),
           )
         end
 
         it "sets ends_at to 8am on date" do
-          date = Date.current
-          instance = described_class.create!(date: date, range: "weekly")
+          saturday = Date.new(2020, 9, 5)
```
kevindew: It looks like this fixes a flaky test I saw pop up this morning where it seems our test suite fails if it [runs before 8am](https://ci.integration.publishing.service.gov.uk/job/email-alert-api/job/dependabot%252Fbundler%252Frails-6.0.3.3/1/console) 🤦  So that's a nice bonus

spec/services/digest_initiator_service_spec.rb
```diff
@@ -9,31 +9,33 @@
 
     context "when a digest run isn't processed" do
```
kevindew, line 10: If you don't mind going a little bit further this file could benefit from a frozen time as I believe most of the tests in this fail before 8am https://ci.integration.publishing.service.gov.uk/job/email-alert-api/job/dependabot%252Fbundler%252Frails-6.0.3.3/1/console

app/models/digest_run.rb
```diff
@@ -18,17 +18,9 @@ def mark_as_completed
 
 private
 
-  def starts_at=(value)
-    super
-  end
-
-  def ends_at=(value)
-    super
-  end
```
kevindew, line 24: I wonder why these were ever a thing 🤔

kevindew (commented):
Looks great, just a suggestion and a favour.

app/models/digest_run.rb
```diff
@@ -32,6 +33,12 @@ def ends_at_is_in_the_past
     errors.add(:ends_at, "must be in the past") if ends_at >= Time.zone.now
   end
 
+  def weekly_digest_is_on_a_saturday
+    return if daily? || ends_at.saturday?
```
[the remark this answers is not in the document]
benthorner: I thought about this, and while this is consistent with the existing validation, the error for that one is unhelpful. I've added another commit to rewrite them both.

spec/services/digest_initiator_service_spec.rb
benthorner: Sure, I've added another commit to do that for this class and the DigestRun model specs.

benthorner:
@kevindew thanks for the comments - I'm pleased we've spotted the time issues, which would have been painful to address as part of other work. This should be ready for another look now.

kevindew (approved):
Nice one

PR #1402, opened 2020-09-11
Populate when we send an email request to Notify

1pretz1:
This PR is the first stage of work to do with reducing our reliance on status updates from notify to archive emails. 

Included in this PR:

* Created a new `sent_at` column on Email model
* Updated the `sending` enum to `sent` for the DeliveryAttempt model
* We now populate the new column (sent_at) and mark the email as sent when we successfully make an email request to Notify. 
* Stop updating the email's status from the status update

Future work to do:

* Remove sent_at and completed_at from the DeliveryAttempt model
* Remove finished_sending_at from the Email model
* Delete all emails older than 14 days old

Trello:

### DIFF
--- a/app/services/delivery_request_service.rb
+++ b/app/services/delivery_request_service.rb
@@ -37,9 +37,11 @@ class DeliveryRequestService < ApplicationService
         email.update!(status: :sent, sent_at: Time.zone.now)
       when :delivered
         record_metric_and_update_attempt(attempt, status)
+        # Deprecated: finished_sending_at is deprecated and soon to be removed due to the introduction of sent_at
         email.update!(status: :sent, finished_sending_at: attempt.finished_sending_at, sent_at: Time.zone.now)
       when :undeliverable_failure
         record_metric_and_update_attempt(attempt, status)
+        # Deprecated: finished_sending_at is deprecated and soon to be removed due to the introduction of sent_at
         email.update!(status: :failed, finished_sending_at: attempt.finished_sending_at)
       when :provider_communication_failure
         record_metric_and_update_attempt(attempt, status)
--- a/app/services/delivery_request_service.rb
+++ b/app/services/delivery_request_service.rb
@@ -31,13 +31,19 @@ class DeliveryRequestService < ApplicationService
 
     status = Metrics.email_send_request(provider_name) { send_email }
 
-    return attempt if status == :sent
-
     ActiveRecord::Base.transaction do
-      attempt.update!(status: status, completed_at: Time.zone.now)
-      Metrics.delivery_attempt_status_changed(status)
-      email.mark_as_sent(attempt.finished_sending_at) if attempt.delivered?
-      email.mark_as_failed(attempt.finished_sending_at) if attempt.undeliverable_failure?
+      case status
+      when :sent
+        email.update!(status: :sent, sent_at: Time.zone.now)
+      when :delivered
+        record_metric_and_update_attempt(attempt, status)
+        email.update!(status: :sent, finished_sending_at: attempt.finished_sending_at, sent_at: Time.zone.now)
+      when :undeliverable_failure
+        record_metric_and_update_attempt(attempt, status)
+        email.update!(status: :failed, finished_sending_at: attempt.finished_sending_at)
+      when :provider_communication_failure
+        record_metric_and_update_attempt(attempt, status)
+      end
     end
 
     attempt
@@ -75,5 +81,10 @@ private
       metrics[:content_change_created_at],
       now,
     )
+  end
+
+  def record_metric_and_update_attempt(attempt, status)
+    Metrics.delivery_attempt_status_changed(status)
+    attempt.update!(status: status, completed_at: Time.zone.now)
   end
 end
--- a/app/services/delivery_request_service.rb
+++ b/app/services/delivery_request_service.rb
@@ -25,13 +25,13 @@ class DeliveryRequestService < ApplicationService
     attempt = Metrics.delivery_request_service_create_delivery_attempt do
       DeliveryAttempt.create!(id: attempt_id,
                               email: email,
-                              status: :sending,
+                              status: :sent,
                               provider: provider_name)
     end
 
     status = Metrics.email_send_request(provider_name) { send_email }
 
-    return attempt if status == :sending
+    return attempt if status == :sent
 
     ActiveRecord::Base.transaction do
       attempt.update!(status: status, completed_at: Time.zone.now)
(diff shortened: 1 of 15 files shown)


app/services/delivery_request_service.rb
```diff
@@ -31,13 +31,21 @@ def call
...
 
     ActiveRecord::Base.transaction do
       attempt.update!(status: status, completed_at: Time.zone.now)
       Metrics.delivery_attempt_status_changed(status)
-      email.mark_as_sent(attempt.finished_sending_at) if attempt.delivered?
-      email.mark_as_failed(attempt.finished_sending_at) if attempt.undeliverable_failure?
+      sent_at = attempt.finished_sending_at
+
+      if attempt.delivered?
+        email.update!(status: :sent, finished_sending_at: sent_at, sent_at: sent_at)
```
benthorner, line 45:
```suggestion
        email.update!(status: :sent, finished_sending_at: sent_at, sent_at: Time.zone.now)
```

We can be consistent with the line above here

app/services/delivery_request_service.rb
```diff
@@ -31,13 +31,21 @@ def call
 
     status = Metrics.email_send_request(provider_name) { send_email }
 
-    return attempt if status == :sent
+    if status == :sent
```
benthorner, line 34: Could we simplify this code a bit? It's confusing that we're:

- Doing some of the updates in a transaction.
- Sometimes checking `attempt` vs. `status`.
- Implicitly handling the other (comms) status.

We could have something like:

```
  case status
  when :sent ...
  when :delivered ...
  when :undeliverable_failure ...
  when :provider_communication_failure ...
end
```

app/services/status_update_service.rb
```diff
@@ -70,9 +70,9 @@ def delivery_attempt_status
   end
 
   def update_email_status(delivery_attempt)
-    finished_sending_at = delivery_attempt.finished_sending_at
-    email.mark_as_sent(finished_sending_at) if delivery_attempt.delivered?
-    email.mark_as_failed(finished_sending_at) if delivery_attempt.undeliverable_failure?
+    if delivery_attempt.status.in? %w[delivered undeliverable_failure]
```
benthorner, line 73: Do we need this `if` anymore?

app/services/delivery_request_service.rb
```diff
@@ -31,13 +31,21 @@ def call
...
       attempt.update!(status: status, completed_at: Time.zone.now)
       Metrics.delivery_attempt_status_changed(status)
-      email.mark_as_sent(attempt.finished_sending_at) if attempt.delivered?
-      email.mark_as_failed(attempt.finished_sending_at) if attempt.undeliverable_failure?
+      sent_at = attempt.finished_sending_at
+
+      if attempt.delivered?
+        email.update!(status: :sent, finished_sending_at: sent_at, sent_at: sent_at)
+      elsif attempt.undeliverable_failure?
+        email.update!(status: :failed, finished_sending_at: sent_at, sent_at: sent_at)
```
benthorner, line 47: We shouldn't populate `sent_at` if we haven't sent the email successfully.

app/services/delivery_request_service.rb
benthorner: Although this may introduce a little duplication, I think it's worth it for the clarity.

spec/models/delivery_attempt_spec.rb
```diff
@@ -2,7 +2,7 @@
   describe ".finished_sending_at" do
     subject { delivery_attempt.finished_sending_at }
 
-    context "when email is sent" do
+    context "when email is delivered" do
```
benthorner, line 5: Nice catch!

benthorner (commented):
Thanks for breaking this into separate commits 👍. I found an issue with populating `sent_at` in the last commit, and also have a suggestion to make the code a bit easier to understand, as I struggled a bit.

Since this PR is something of a halfway point for the next two weeks, I also think it's worth adding a "DEPRECATED" comment in places where we populate `finished_sending_at`. The reason is that this field no longer makes much sense: we can be "finished sending" but have `sent_at: nil`!

app/services/status_update_service.rb
1pretz1: Nope, good spot!

app/services/delivery_request_service.rb
1pretz1: Sure thing, I've rejigged this bit around and added some missing tests around recording metrics. I did actually start writing a case statement but then changed my mind, have added said case.

app/services/delivery_request_service.rb
1pretz1: Ah for some reason I thought only for a communication failure we shouldn't update sent_at, got my wires crossed!

1pretz1:
Thanks for the review @benthorner, I've addressed your comments and added come deprecation comments.

It also looks like the CI hooks are pointing to Carrenza again so we have passing tests! 🎉

spec/services/delivery_request_service_spec.rb
```diff
@@ -155,10 +168,18 @@
...
       end
 
-      it "marks the email as failed" do
-        expect { described_class.call(email: email) }
-          .to change { email.reload.status }
-          .to("failed")
+      it "marks the email as failed and sets sent_at time" do
+        freeze_time do
+          expect { described_class.call(email: email) }
+            .to change { email.status }.to("failed")
```
benthorner, line 174: We're missing the actual check for `sent_at` here.

spec/services/status_update_service_spec.rb
```diff
@@ -26,18 +26,13 @@
...
-
     context "when provided a 'delivered' status" do
       let(:status) { "delivered" }
 
       it "sets the delivery attempt status to delivered" do
         expect { described_class.call(args) }
-          .to change { delivery_attempt.reload.status }
-          .to("delivered")
+          .to change { delivery_attempt.reload.status }.to("delivered")
+          .and change { email.reload.finished_sending_at }.to(sent_at)
```
benthorner, line 35: Minor: the use of `sent_at` is (now) confusing here, and seems redundant since we also have this `completed_at` variable, which is set to the same value. Can we just say `timestamp` or `status_update_time` (or something)?

benthorner (approved):
Thanks for adding the deprecation warnings @1pretz1 - looks good! I found a couple of minor issues that would be good to fix before this goes in.

spec/services/delivery_request_service_spec.rb
1pretz1: That was old text describing the test - have updated it accordingly!

PR #1405, opened 2020-09-16
Automatically retry incomplete digest runs

benthorner:

This extends the existing RecoverLostJobsWorker to also
cope with missing or unprocessed work for digest runs.

### DIFF
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -11,6 +11,8 @@ class DigestRun < ApplicationRecord
 
   enum range: { daily: 0, weekly: 1 }
 
+  scope :unprocessed, -> { where(processed_at: nil) }
+
   def mark_as_completed
     completed_time = digest_run_subscribers.maximum(:processed_at) || Time.zone.now
     update!(completed_at: completed_time)
--- a/app/models/digest_run.rb
+++ b/app/models/digest_run.rb
@@ -11,8 +11,6 @@ class DigestRun < ApplicationRecord
 
   enum range: { daily: 0, weekly: 1 }
 
-  scope :unprocessed, -> { where(processed_at: nil) }
-
   def mark_as_completed
     completed_time = digest_run_subscribers.maximum(:processed_at) || Time.zone.now
     update!(completed_at: completed_time)
--- a/app/services/digest_initiator_service.rb
+++ b/app/services/digest_initiator_service.rb
@@ -37,7 +37,7 @@ private
   end
 
   def run_with_advisory_lock
-    DigestRun.with_advisory_lock("#{range}_digest_initiator", timeout_seconds: 0) do
+    DigestRun.with_advisory_lock("#{range}_digest_initiator-#{date}", timeout_seconds: 0) do
       yield
     end
   end
--- a/app/workers/daily_digest_initiator_worker.rb
+++ b/app/workers/daily_digest_initiator_worker.rb
@@ -1,5 +1,12 @@
-class DailyDigestInitiatorWorker < DigestInitiatorWorker
-  sidekiq_options lock: :until_executed,
+class DailyDigestInitiatorWorker
+  include Sidekiq::Worker
+
+  sidekiq_retry_in do |count|
+    60 * (count + 1)
+  end
+
+  sidekiq_options retry: 3,
+                  lock: :until_executed,
                   unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
                   on_conflict: :log
 
--- a/app/workers/daily_digest_initiator_worker.rb
+++ b/app/workers/daily_digest_initiator_worker.rb
@@ -1,5 +1,5 @@
 class DailyDigestInitiatorWorker < DigestInitiatorWorker
-  def perform(date = Date.current)
-    DigestInitiatorService.call(date: date, range: Frequency::DAILY)
+  def perform(date = Date.current.to_s)
+    DigestInitiatorService.call(date: Date.parse(date), range: Frequency::DAILY)
   end
 end
--- a/app/workers/daily_digest_initiator_worker.rb
+++ b/app/workers/daily_digest_initiator_worker.rb
@@ -1,4 +1,12 @@
 class DailyDigestInitiatorWorker < DigestInitiatorWorker
+  sidekiq_options lock: :until_executed,
+                  unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
+                  on_conflict: :log
+
+  def self.uniqueness_with(args)
+    [args.first || Date.current.to_s]
+  end
+
   def perform(date = Date.current.to_s)
     DigestInitiatorService.call(date: Date.parse(date), range: Frequency::DAILY)
   end
--- a/app/workers/digest_email_generation_worker.rb
+++ b/app/workers/digest_email_generation_worker.rb
@@ -1,7 +1,14 @@
 class DigestEmailGenerationWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: :email_generation_digest
+  sidekiq_options queue: :email_generation_digest,
+                  lock: :until_executed,
+                  unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
+                  on_conflict: :log
+
+  def self.uniqueness_with(args)
+    [args.first]
+  end
 
   def perform(digest_run_subscriber_id)
     email = nil
--- a/app/workers/recover_lost_jobs_worker.rb
+++ b/app/workers/recover_lost_jobs_worker.rb
@@ -3,6 +3,6 @@ class RecoverLostJobsWorker
 
   def perform
     RecoverLostJobsWorker::UnprocessedCheck.new.call
-    RecoverLostJobsWorker::DigestRunsCheck.new.call
+    RecoverLostJobsWorker::MissingDigestRunsCheck.new.call
   end
 end
--- a/app/workers/recover_lost_jobs_worker.rb
+++ b/app/workers/recover_lost_jobs_worker.rb
@@ -2,30 +2,20 @@ class RecoverLostJobsWorker
   include Sidekiq::Worker
 
   def perform
-    old_unprocessed(ContentChange).pluck(:id).each do |content_change_id|
-      ProcessContentChangeWorker.perform_async(content_change_id)
-    end
-
-    old_unprocessed(Message).pluck(:id).each do |message_id|
-      ProcessMessageWorker.perform_async(message_id)
-    end
-
-    old_unprocessed(DigestRunSubscriber).pluck(:id).each do |digest_run_subscriber_id|
-      DigestEmailGenerationWorker.perform_async(digest_run_subscriber_id)
-    end
-
-    old_unprocessed(DigestRun.daily).pluck(:date).each do |date|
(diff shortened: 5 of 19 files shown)


benthorner:
Adding "DO NOT MERGE", as this depends on removing ancient, unprocessed digest run (/subscriber) records.

app/workers/recover_lost_jobs_worker.rb
```diff
@@ -9,5 +9,17 @@ def perform
     Message.unprocessed.where("created_at <= ?", 1.hour.ago).pluck(:id).each do |message_id|
       ProcessMessageWorker.perform_async(message_id)
     end
+
+    DigestRunSubscriber.unprocessed.where("created_at <= ?", 1.hour.ago).pluck(:id).each do |digest_run_subscriber_id|
+      DigestEmailGenerationWorker.perform_async(digest_run_subscriber_id)
+    end
+
+    DigestRun.daily.unprocessed.where("created_at <= ?", 1.hour.ago).pluck(:date).each do |date|
```
1pretz1: I'm wondering whether an hour is enough time to allow the digest runs to process now we're moving towards a default model of digest emails as the burden on these workers will only increase.

I ran a query over the last 20 digest runs and the average processing time was over 31 minutes with a maximum of 52 minutes and minimum of 1 minute.

Maybe increase the threshold of the daily and weekly processing time to 2 hours?

app/workers/recover_lost_jobs_worker.rb
```diff
@@ -15,6 +17,23 @@ def old_unprocessed(scope)
...
+
+  def non_existent_weekly_digests
+    [DigestRun.new(date: expected_digest_week.find(&:saturday?), range: :weekly)]
+      .reject { |digest_run| DigestRun.exists?(date: digest_run.date, range: :weekly) }
+  end
+
+  def expected_digest_week
+    digestion_time = Time.zone.parse("#{DigestRun::DIGEST_RANGE_HOUR}:00")
+    cutoff_with_delay = digestion_time + 1.hour
+    end_date = Time.zone.now > cutoff_with_delay ? Time.zone.today : Time.zone.yesterday
```
1pretz1: Hmm are we intending to wait an hour or 30 minutes before trying to create a missing digest run?

I notice that `DigestRun::DIGEST_RANGE_HOUR` is set for 8am but we don't actually start any digests until [8:30am]. This would mean the the `cutoff_with_delay` would be 9am leaving a 30 minute window for the digest run to be created which should be fine but just checking that was intended.

[8:30am]: https://github.com/alphagov/email-alert-api/blob/master/config/sidekiq.yml#L18-L23

1pretz1 (commented):
Looks great @benthorner, I like the implementation of the missing digest runs, it's clever but simple!

Just left a few comments.

kevindew (commented):
Looking good to me, Pete seems to have caught all the interesting bits already.

app/workers/recover_lost_jobs_worker.rb
```diff
@@ -15,6 +17,23 @@ def old_unprocessed(scope)
     scope.where(processed_at: nil).where("created_at <= ?", 1.hour.ago)
   end
 
+  def non_existent_daily_digests
+    expected_digest_week.map { |date| DigestRun.new(date: date, range: :daily) }
+      .reject { |digest_run| DigestRun.exists?(date: digest_run.date, range: :daily) }
```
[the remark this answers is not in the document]
benthorner: I wondered the same. I don't think we can use `persisted?`, because it [doesn't actually query the DB](https://apidock.com/rails/ActiveRecord/Persistence/persisted%3F).

app/workers/digest_email_generation_worker.rb
```diff
@@ -1,7 +1,10 @@
 class DigestEmailGenerationWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: :email_generation_digest
+  sidekiq_options queue: :email_generation_digest,
+                  lock: :until_executed,
+                  unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
```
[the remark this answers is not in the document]
benthorner: Oh yes, thanks for spotting 😬. I was too keen to get to the complicated bit later on. Updated to add the missing method.

app/workers/recover_lost_jobs_worker.rb
benthorner: Yeah, I did wonder about that. I checked in production and the timings are similar to what you're suggesting here. My conclusion is that they're similar enough to those for processing content changes that we can retry them similarly after 1 hour. The main reason we would want to increase the grace period would be if it would otherwise cause a large / alarming build up on the queue, which won't happen for digest runs.

app/workers/recover_lost_jobs_worker.rb
benthorner: Haha, no it wasn't! I completely missed the half hour discrepancy. However, I do think it's fine: half an hour is still plenty of time, and I think it's actually a good thing that the recovery is ignorant of the original scheduling: we really _should_ have a new digest run if we're one hour in to a new digest day.

> Hmm are we intending to wait an hour or 30 minutes before trying to create a missing digest run?

In conclusion: kind of both 😄.

app/workers/daily_digest_initiator_worker.rb
```diff
@@ -1,4 +1,7 @@
 class DailyDigestInitiatorWorker < DigestInitiatorWorker
+  sidekiq_options lock: :until_executed,
+                  on_conflict: :log
+
   def perform(date = Date.current)
```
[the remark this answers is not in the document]
benthorner: Oh yeah, thanks for spotting. I remember thinking about this potential issue when I changed the interface, but neglected to account for the future way we'd be executing the job.

I've added a commit (near the beginning) the arg to a string.

app/workers/daily_digest_initiator_worker.rb
```diff
@@ -1,4 +1,7 @@
 class DailyDigestInitiatorWorker < DigestInitiatorWorker
+  sidekiq_options lock: :until_executed,
+                  on_conflict: :log
```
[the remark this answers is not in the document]
benthorner: Oops, that was a bit of an oversight. Both of these flaws are unlikely to cause a problem for digest runs, as we'd have to have one failing for over a day. However, it just makes more sense for the locking to be in-sync with the arguments, so I've modified the initial commit to fix it for Sidekiq, and appended one to fix it for Postgres.

app/models/digest_run.rb
```diff
@@ -11,6 +11,8 @@ class DigestRun < ApplicationRecord
 
   enum range: { daily: 0, weekly: 1 }
 
+  scope :unprocessed, -> { where(processed_at: nil) }
```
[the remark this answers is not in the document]
benthorner: Sure, in hindsight I would have gone the other way. I thought it easier to leave as-is, since I only later concluded the scopes were't very helpful.

benthorner:
@kevindew @1pretz1 thanks very much for the reviews, spotting some key missing pieces ⭐. I've responded to all your comments, so could you take another look?

app/workers/recover_lost_jobs_worker.rb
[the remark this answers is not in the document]
kevindew: It doesn't need to query the db as that is already as part of `find_or_initialize_by` the presence of an id column is sufficient to determine the record is new.

app/workers/daily_digest_initiator_worker.rb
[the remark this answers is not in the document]
kevindew: Great, thanks

app/workers/recover_lost_jobs_worker.rb
1pretz1: > The main reason we would want to increase the grace period would be if it would otherwise cause a large / alarming build up on the queue, which won't happen for digest runs.

True! It just seems this code may be run unnecessarily in the not so distant future as processing times increase, not that it particularly matters as there will be no adverse side effects, as you pointed out.

app/workers/recover_lost_jobs_worker.rb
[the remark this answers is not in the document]
benthorner: Aha, I missed `find_or_initialize_by`. That reads a bit better - updated 👍.

app/services/digest_initiator_service.rb
```diff
@@ -37,7 +37,7 @@ def enqueue_jobs(digest_run_subscriber_ids)
   end
 
   def run_with_advisory_lock
-    DigestRun.with_advisory_lock("#{range}_digest_initiator", timeout_seconds: 0) do
+    DigestRun.with_advisory_lock("#{range}_digest_initiator-#{date}", timeout_seconds: 0) do
```
kevindew, line 40: ⭐

app/workers/recover_lost_jobs_worker.rb
1pretz1: Yeah `DigestRun::DIGEST_RANGE_HOUR` threw me a bit that it wasn't the same as our actual scheduled digest run time but hey, it does say just 'hour' in the name. 

Half an hour seems fine! 👍

app/workers/recover_lost_jobs_worker.rb
kevindew: The somewhat benign adverse effect is that our stats could show multiple successful digest runs for a day if this kick off happens before the digest finishes when we hit our first 1 hour one (which is surely looming with the next experiment?)

A way we could resolve that while keeping this at 1 hour is change the uniqueness args in the worker to also use current date e.g: `[args.first || Date.current.to_s]` - bit hacky mind.

app/workers/daily_digest_initiator_worker.rb
```diff
@@ -1,5 +1,13 @@
 class DailyDigestInitiatorWorker < DigestInitiatorWorker
-  def perform(date = Date.current)
-    DigestInitiatorService.call(date: date, range: Frequency::DAILY)
+  sidekiq_options lock: :until_executed,
```
kevindew, line 2: Working on something different I noticed sidekiq_options are already set in DigestInitiatorWorker and so figured this might override the other options (not sure if they merge or not) might be good to consolidate either repeating them all setting them all in the parent class either way.

app/workers/recover_lost_jobs_worker.rb
benthorner: That's a good thought. I don't think it'll be a problem, though: our conflict strategy is to `log`, so the duplicate job would never actually make it on to the queue i.e. would never be reported as a metric.

kevindew: It wouldn't recognise it as a conflict as I understand as the one that starts automatically would have arguments of [nil] whereas the one that retried would have arguments of `["2020-09-18"]`

app/workers/daily_digest_initiator_worker.rb
benthorner: Another good find - you're both on a roll today. I think it's clearer to have everything in one place, so I've added another commit to make this so.

app/workers/recover_lost_jobs_worker.rb
benthorner: Oh, I see what you mean now. Yes, that's another surprise we may as well avoid. It makes sense to me, seeing the same defaults next to each other in the worker code. I've updated the second commit to do that - thanks!

benthorner:
> The name of RecoverLostJobsWorker::DigestRunsCheck implied to me that all digest run checks would be there

Fair enough. I've prepended "Missing", as in "MissingDigestRunsCheck".

PR #1406, opened 2020-09-16
Co-locate metrics worker code with the worker

benthorner:

This repeats a similar refactoring of the RecoverLostJobsWorker [1],
which has the benefit of keeping the 'lib' directory lean.

[1]: https://github.com/alphagov/email-alert-api/pull/1405/commits/482153f3b2d4d5c2a6f5700046bde96b1c26cf04

### DIFF
--- a/app/workers/metrics_collection_worker.rb
+++ b/app/workers/metrics_collection_worker.rb
@@ -2,8 +2,8 @@ class MetricsCollectionWorker
   include Sidekiq::Worker
 
   def perform
-    Metrics::ContentChangeExporter.call
-    Metrics::DigestRunExporter.call
-    Metrics::MessageExporter.call
+    ContentChangeExporter.call
+    DigestRunExporter.call
+    MessageExporter.call
   end
 end
--- a/app/services/delivery_request_service/delay_provider.rb
+++ b/app/services/delivery_request_service/delay_provider.rb
@@ -1,4 +1,4 @@
-class DelayProvider
+class DeliveryRequestService::DelayProvider
   def self.call(*args)
     new.call(*args)
   end
--- a/app/services/delivery_request_service/email_address_overrider.rb
+++ b/app/services/delivery_request_service/email_address_overrider.rb
@@ -1,21 +1,19 @@
-class DeliveryRequestService
-  class EmailAddressOverrider
-    attr_reader :override_address, :whitelist_addresses, :whitelist_only
+class DeliveryRequestService::EmailAddressOverrider
+  attr_reader :override_address, :whitelist_addresses, :whitelist_only
 
-    def initialize(config)
-      @override_address = config[:email_address_override]
-      @whitelist_addresses = Array(config[:email_address_override_whitelist])
-      @whitelist_only = config[:email_address_override_whitelist_only]
-    end
+  def initialize(config)
+    @override_address = config[:email_address_override]
+    @whitelist_addresses = Array(config[:email_address_override_whitelist])
+    @whitelist_only = config[:email_address_override_whitelist_only]
+  end
 
-    def destination_address(address)
-      return address unless override_address
+  def destination_address(address)
+    return address unless override_address
 
-      if whitelist_addresses.include?(address)
-        address
-      else
-        whitelist_only ? nil : override_address
-      end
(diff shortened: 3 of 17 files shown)


1pretz1:
Love a bit of refactoring @benthorner!

One comment - if we're moving these classes under app shouldn't we be refactoring these `*Exporter` classes into their own `exporters` directory and dropping the namespace? From having a nosey through the code this seems to be the common practice for the other workers and their called upon classes. Some examples are the [Archive worker] and it's [archive query class] or the [process messages worker] and it's [matched message generation service].

What do you think?

[Archive worker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/email_archive_worker.rb
[archive query class]: https://github.com/alphagov/email-alert-api/blob/bc0e6fdd586cdb59c92edb2704cb96ce83d7663e/app/queries/email_archive_query.rb
[process messages worker]: https://github.com/alphagov/email-alert-api/blob/bc0e6fdd586cdb59c92edb2704cb96ce83d7663e/app/workers/process_message_worker.rb
[matched message generation service]: https://github.com/alphagov/email-alert-api/blob/16505e2d3bac948f1ba0394be6005aee22b8e501/app/services/matched_message_generation_service.rb

benthorner:
Thanks @1pretz1 - we can but try!

I think I see where you're coming from: there's a precedent for grouping unconventional code like "queries" by type. That makes sense when it's an application-wide concern: many parts of the app need to do complex queries. And "services" is a well-established concept in many of our apps.

Thinking about the exporters: they're only used in one place, for one purpose. I argue they shouldn't have their own top-level namespace. While we can always shove "misc" code in `lib/`, it's helpful if we can bring related code closer together, without making a mess. We take a similar approach for some of the code [in Content Publisher](https://github.com/alphagov/content-publisher/tree/master/app/services/create_image_blob_service).

How are you seeing it now?

1pretz1:
Okay I see where you're coming from @benthorner, thanks for explaining. 

By that thinking should we also move the classes in the `providers` [directory] to the [delivery_request_service dir] as they're only used by DeliveryRequestService and similarly only serve one purpose (giving us the provider when sending emails)?

[directory]: https://github.com/alphagov/email-alert-api/tree/master/app/providers
[delivery_request_service dir]: https://github.com/alphagov/email-alert-api/tree/master/app/services/delivery_request_service

benthorner:
@1pretz1 sure, good spot (/ what have I started 😱). I've added another commit to do that - slowly this app gets less special.

app/workers/metrics_collection_worker.rb
```diff
@@ -2,8 +2,8 @@ class MetricsCollectionWorker
   include Sidekiq::Worker
 
   def perform
-    Metrics::ContentChangeExporter.call
-    Metrics::DigestRunExporter.call
-    Metrics::MessageExporter.call
+    MetricsCollectionWorker::ContentChangeExporter.call
+    MetricsCollectionWorker::DigestRunExporter.call
+    MetricsCollectionWorker::MessageExporter.call
```
1pretz1, line 7: Oh and I just realised that you don't actually need to explicitly navigate to MetricsCollectionWorker, just calling the classes straight up should be enough, e.g.

`MetricsCollectionWorker::ContentChangeExporter.call` -> `ContentChangeExporter.call`

1pretz1 (approved):
Looks good, just left one more comment! Approving as it's basically complete! 💯 

Oh and while you're in this neck of the woods, do you mind making the namespacing syntax consistent between the new nested classes and the [email address overrider].

[email address overrider]: https://github.com/alphagov/email-alert-api/blob/master/app/services/delivery_request_service/email_address_overrider.rb#L1-L2

app/workers/metrics_collection_worker.rb
benthorner: Indeed. At the time I thought it might be helpful to be explicit, but it's also nice to be concise, and [somewhat consistent](https://github.com/alphagov/email-alert-api/blob/master/app/services/delivery_request_service.rb#L15). Updated.

benthorner:
Sure, I've done that @1pretz1 - unfortunately this isn't something we can lint for, as it depends on the point when the parent class gets required.

PR #1408, opened 2020-09-16
Archive emails pt2

1pretz1:
This is part 2 of the long awaited archiving emails story.

In this PR we remove all references to columns we plan to remove. The columns are `finished_sending_at` for emails and `sent_at` & `completed_at` for delivery_attempts.

We can't merge and deploy this PR until at least 2 weeks have past from the deployment of [this PR] which occurred on the 15th September at 3:52pm. This is so all non deleted emails will have a populated `sent_at` value on the Email model (emails get deleted after 2 weeks).

Trello:

[this PR]: https://github.com/alphagov/email-alert-api/pull/1402


spec/factories.rb
```diff
@@ -231,7 +228,9 @@
     subscriber_list
   end
 
-  factory :user
+  factory :user do
+    uid { SecureRandom.uuid }
```
benthorner, line 220: How come this change is here?

benthorner (approved):
Nice one - especially remembering to ignore the columns 💯. And some nice cleanup in the factory. I have just on comment, but I don't think it's important, so approving now.

spec/factories.rb
1pretz1: Due to the updated test [here](https://github.com/alphagov/email-alert-api/pull/1408/files#diff-9066f9057bbc4594aadaf7c321dd4506R19-R20).

After updating the test to a `change` block it started to fail as it was comparing nil (default signon_user_id value for delivery attempts) against nil (user's uid passed in). This seemed like a bad test so I created a uid value in the factory as I assumed all user's should have one.

benthorner: Oh, I see. Thanks for explaining!

PR #1412, opened 2020-09-18
Delete unique job digest when a job is marked as dead

kevindew:
Without this any unique jobs cannot be re-enqueued to retry them as they
will be marked as already existing.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,4 +1,9 @@
-Sidekiq.configure_server do
+Sidekiq.configure_server do |config|
   # Calls to Rails.logger in a sidekiq process will use Sidekiq's logger
   Rails.logger = Sidekiq::Logging.logger
+
+  config.death_handlers << lambda { |job, _error|
+    digest = job["unique_digest"]
+    SidekiqUniqueJobs::Digests.delete_by_digest(digest) if digest
+  }
 end


benthorner (commented):
I think it's worth noting this is because [we're using Sidekiq v5](https://github.com/alphagov/govuk_sidekiq/blob/master/govuk_sidekiq.gemspec#L24), for which [a manual approach is necessary](https://github.com/alphagov/govuk_sidekiq/blob/master/govuk_sidekiq.gemspec#L24). Could we add a comment about that, suggesting we remove if we've upgraded?

kevindew:
> I think it's worth noting this is because [we're using Sidekiq v5](https://github.com/alphagov/govuk_sidekiq/blob/master/govuk_sidekiq.gemspec#L24), for which [a manual approach is necessary](https://github.com/alphagov/govuk_sidekiq/blob/master/govuk_sidekiq.gemspec#L24). Could we add a comment about that, suggesting we remove if we've upgraded?

My understanding of https://github.com/mhenrixon/sidekiq-unique-jobs/tree/v6.0.22#cleanup-dead-locks is that we need a manual approach whether we're using version 5 or 6. The difference being that if you use sidekiq > 5.1 you can do a global configuration through a death_handler, whereas <= 5.1 you have to use the `sidekiq_retries_exhausted ` hook in your individual worker classes.

benthorner (approved):
Ah yes, sorry. I misread that part. Still worth a comment I think, as it's not obvious to me what the code is for, since we don't run it explicitly anywhere - not a major issue, though.

kevindew:
> Ah yes, sorry. I misread that part. Still worth a comment I think, as it's not obvious to me what the code is for, since we don't run it explicitly anywhere - not a major issue, though.

Cool thanks.

One of the things that hit me about this is that we're still vulnerable to problems for any jobs that have a lock "until executed" if a job is killed (e.g. out of memory) where this code will never run and the lock will be stuck eternally. We might want to resolve this with switching the lock type to "until_executing", experimenting with the expiry further or trying out the new reaper stuff in unique jobs 7 which is in beta. The reaper seems to be a feature developed exactly around this use case (though lacks documentation) as it is a common complaint: https://github.com/mhenrixon/sidekiq-unique-jobs/issues/362

kevindew:
Superseded by https://github.com/alphagov/email-alert-api/pull/1419

PR #1414, opened 2020-09-21
Revert "Add temporary task to remove dead records"

benthorner:

Reverts alphagov/email-alert-api#1413

Now that the task has been run.


kevindew (approved):
That was quick ;)

PR #1415, opened 2020-09-22
Replace sidekiq-unique-jobs with non-blocking lock

benthorner:

    Previously we added sidekiq-unique-jobs as a way to prevent duplicate
    jobs executing concurrently. However, we've since learned the gem is
    flawed for the main scenario we're trying to prevent: a worker being
    killed due to memory issues [1]. An non-blocking lock [2] can be used
    to get the same no-wait-noop effect, similarly to an advisory lock, but
    without the need to define a unique namespace [3]. This replaces the
    use of the gem with non-blocking locks, which are more robust.
    
    Note this replaces the use of specific models with ApplicationRecord
    where the specific model class is not relevant.
    
    Note that, while we could use "SKIP LOCKED" instead of "NOWAIT", this
    would still result in a "not found" error later on, due to the record
    being invisible to all but the process holding the lock [2]. I think
    it's better to use "NOWAIT", so the error we catch (to avoid a failure
    / retry) is explicitly related to locking.
    
    [1]: https://github.com/alphagov/email-alert-api/pull/1412#issuecomment-695999823
    [2]: https://www.postgresql.org/docs/9.5/sql-select.html
    [3]: https://github.com/alphagov/email-alert-api/pull/1405/commits/e2b85f9e1223d3af386a6c20029962445e6ac0c3

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,6 @@ gem "plek"
 gem "ratelimit"
 gem "redcarpet"
 gem "sidekiq-scheduler"
-gem "sidekiq-unique-jobs"
 gem "with_advisory_lock"
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -328,10 +328,6 @@ GEM
     sidekiq-statsd (2.1.0)
       activesupport
       sidekiq (>= 3.3.1)
-    sidekiq-unique-jobs (6.0.22)
-      concurrent-ruby (~> 1.0, >= 1.0.5)
-      sidekiq (>= 4.0, < 7.0)
-      thor (~> 0)
     simplecov (0.19.0)
       docile (~> 1.1)
       simplecov-html (~> 0.11)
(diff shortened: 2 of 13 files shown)


kevindew (commented):
Thanks for raising Ben, I think there's some fundamental problems here that need addressing based around the understanding of "FOR UPDATE" and where we can apply transactions.

I wonder if we can use `while_advisory_lock` for places where there isn't a transaction and just include an id as part of the lock. Whereas for places where we can do a transaction I think we can probably get away with rails explicit locking without the `NOWAIT` clause. I'm not sure any of them are things we'd expect them to take a long time.

I think it'd be useful for the commit to also explain why we're not concerned anymore about having the same item multiple times in the queues.

benthorner:
@kevindew thanks for taking a look. I see now this has suffered from one refactor too many. My starting point was to use advisory locks as planned, but this lead to a fair amount of duplication, and the arbitrary creation of ID strings for each lock. While it may be possible to get the NOWAIT locks working properly, I agree we shouldn't use them/transactions if the code is doing more than simple operations, which it is. This needs a rewrite.

I think I'll leave it until I'm back from 2ndline. To address my duplication concern, we can introduce an ApplicationWorker class, which seems appropriate for this common use case.

> I think it'd be useful for the commit to also explain why we're not concerned anymore about having the same item multiple times in the queues.

Sure, that can go into the next PR.

PR #1419, opened 2020-10-01
Swap sidekiq-unique-jobs for PG advisory locks

benthorner:

This PR makes a couple of changes to the way we avoid repeating work:

- Replaces the use of sidekiq-unique-jobs with advisory locks as the
primary mechanism to prevent duplicate work.

- Standardises the location and ID structure of our locks. This means
we need to deploy the change when affected worker code is not in use.

- DRYs-up the worker classes by introducing a new base class.

More detail is in the commits. We could consider making the changes to
the advisory lock IDs in a separate, smaller deployment. I think it's
more coherent if we keep all the changes together in one PR, unless this
proves hard to review.

### DIFF
--- a/app/workers/application_worker.rb
+++ b/app/workers/application_worker.rb
@@ -1,0 +1,9 @@
+class ApplicationWorker
+  include Sidekiq::Worker
+
+private
+
+  def run_with_advisory_lock(model, unique_ref)
+    ApplicationRecord.with_advisory_lock("#{model}-#{unique_ref}", timeout_seconds: 0) { yield }
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,6 @@ gem "plek"
 gem "ratelimit"
 gem "redcarpet"
 gem "sidekiq-scheduler"
-gem "sidekiq-unique-jobs"
 gem "with_advisory_lock"
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -328,10 +328,6 @@ GEM
     sidekiq-statsd (2.1.0)
       activesupport
       sidekiq (>= 3.3.1)
-    sidekiq-unique-jobs (6.0.22)
-      concurrent-ruby (~> 1.0, >= 1.0.5)
-      sidekiq (>= 4.0, < 7.0)
-      thor (~> 0)
     simplecov (0.19.0)
       docile (~> 1.1)
       simplecov-html (~> 0.11)
@@ -402,7 +398,6 @@ DEPENDENCIES
   rspec-rails
   rubocop-govuk
   sidekiq-scheduler
-  sidekiq-unique-jobs
   simplecov
   webmock
   with_advisory_lock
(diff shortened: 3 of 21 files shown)


app/workers/application_worker.rb
```diff
@@ -0,0 +1,7 @@
+class ApplicationWorker
+  include Sidekiq::Worker
+
+  def run_with_advisory_lock(model, id, &block)
+    ApplicationRecord.with_advisory_lock("#{model}-#{id}", timeout_seconds: 0) { block.call }
+  end
```
kevindew, line 6: Any reason you've gone with using block.call rather than yield? It feels less idiomatic and provides a more cryptic error if a block is missing, though I was wondering if I missed a nuace?

By the way I think this should be a private method

app/workers/application_worker.rb
```diff
@@ -0,0 +1,7 @@
+class ApplicationWorker
+  include Sidekiq::Worker
+
+  def run_with_advisory_lock(model, id, &block)
```
kevindew, line 4: I wonder if you need two arguments of this, it feels that it works as a good abstraction where we're running a particular id of an item but feels a bit cumbersome for things like `DigestRun, "#{date}-#{Frequency::DAILY}"` where the id argument doesn't refer to the common understanding of an id of that model.

I wonder if you could get away with just having an optional id field and have the lock automatically use the class name.

for example:

```
def run_with_advisory_lock(id = nil, &block)
ApplicationRecord.with_advisory_lock("#{self.class}-#{id}", timeout_seconds: 0) { block.call }
end
```

then you'd not need to provide any arguments for most cases and be able to just pass in a date for daily/weekly digest initiators.

kevindew (commented):
Looks good and I agree that it seems reasonable to disable those workers while this goes out. All I've done is bikeshedded away at the lock name process.

benthorner (commented):
Thanks for looking - just need to remember to re-enable them after (if only we had an alert for that...).

I agree that `id` should be reserved for actual IDs, although I'm not so keen on making it optional: as an abstraction, this could lead to multiple workers accidentally using the same key (e.g. the email archive/delete ones). I've gone with a compromise of `unique_ref` - does that read more accurately to you?

kevindew:
> Thanks for looking - just need to remember to re-enable them after (if only we had an alert for that...).
> 
> I agree that `id` should be reserved for actual IDs, although I'm not so keen on making it optional: as an abstraction, this could lead to multiple workers accidentally using the same key (e.g. the email archive/delete ones). I've gone with a compromise of `unique_ref` - does that read more accurately to you?

Sure that looks better. You wouldn't hit the workers using the same key if you used the worker class name, but no biggie.

kevindew (approved):

benthorner:
> You wouldn't hit the workers using the same key

🤭. On the whole I prefer the explicit consistency, for greater visibility of the lock. There's are already [a weirdness](https://github.com/alphagov/email-alert-api/pull/932) around renaming Sidekiq worker classes, and coupling that to the lock would compound that.

PR #1425, opened 2020-10-06
Add an endpoint to get a subscriber's details from an account token

barrucadu:
For at least our initial experiment, we want to be able to manage
email subscriptions from the account manager.  All email-alert-api
interaction is done in terms of subscriber IDs, rather than addresses,
and the only way to turn an address into a subscriber ID is by sending
a verification email with a magic link in it.

We don't want to have a "double verification email" scenario, where we
ask a user to verify their account and then also to verify we can
manage their subscriptions.  So add this endpoint which:

1. Accepts a JWT over POST containing the fields `email` and `email_verified`
2. Checks that the JWT has a valid signature from the account system
3. Checks that the `email_verified` is true
4. Returns either an error or the subscriber details

With the ID and the address all the other endpoints work.  We can:

- List a user's subscriptions
- Change a user's address
- Unsubscribe a user from one or all of their subscriptions
- Subscribe a user to a list (which sends them a confirmation, but not
  verification, email)
- Update the frequency of a subscription

---

[Trello card](https://trello.com/c/VOHkLg9s/319-figure-out-what-changes-we-need-in-email-alert-api)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -26,6 +26,8 @@ Rails.application.routes.draw do
     post "/subscribers/auth-token", to: "subscribers_auth_token#auth_token"
     post "/subscriptions/auth-token", to: "subscriptions_auth_token#auth_token"
 
+    post "/subscribers/account-jwt", to: "subscribers_account_jwt#account_jwt"
+
     post "/unsubscribe/:id", to: "unsubscribe#unsubscribe"
 
     get "/healthcheck", to: "healthcheck#index"
--- a/app/controllers/subscribers_account_jwt_controller.rb
+++ b/app/controllers/subscribers_account_jwt_controller.rb
@@ -1,0 +1,57 @@
+class SubscribersAccountJwtController < ApplicationController
+  before_action :validate_params
+
+  class TokenInvalid < RuntimeError; end
+
+  rescue_from TokenInvalid do |exception|
+    render json: { error: "Could not fetch subscriber",
+                   details: exception.message },
+           status: :forbidden
+  end
+
+  def account_jwt
+    jwt = expected_params.require(:jwt)
+    address = validate_token(jwt)
+    subscriber = find_subscriber(address)
+
+    render json: { subscriber: subscriber }
+  end
+
+private
+
+  def validate_token(jwt)
+    payload, = JWT.decode jwt, public_key, true, { algorithm: "ES256" }
+    raise TokenInvalid, "email address not verified" unless payload["email_verified"]
+
+    payload["email"]
+  rescue JWT::DecodeError
+    raise TokenInvalid, "could not decode jwt"
+  end
+
+  def public_key
+    OpenSSL::PKey::EC.new(Rails.application.secrets.accounts_jwt_public_key)
+  end
+
+  def find_subscriber(address)
+    Subscriber.resilient_find_or_create(
+      address,
+      signon_user_uid: current_user.uid,
+    ).tap do |subscriber|
+      subscriber.activate if subscriber.deactivated?
+    end
+  end
+
+  def expected_params
+    params.permit(:jwt)
+  end
+
+  def validate_params
+    ParamsValidator.new(expected_params).validate!
+  end
+
+  class ParamsValidator < OpenStruct
+    include ActiveModel::Validations
+
+    validates :jwt, presence: true
+  end
+end
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -22,6 +22,7 @@
 development:
   secret_key_base: 68c093e39d8ac8ecbaca8cc09dc573c3fd7d6942a9cd1a88474c50f028cb85f653a2d2d7c0579da6e172519518e77fc557d5dca7920bae925a6e2e8f3ad7c514
   email_alert_auth_token: b132e4dde18854890516866be3ef3a2885addd6a47fc1d6173f3069290b6753ec28ccbe9dee580837f9027be18b36f8983dc03d1230e4d23818049b1777c8a2b
+  accounts_jwt_public_key: <%= ENV["ACCOUNTS_JWT_PUBLIC_KEY"] %>
 
 test:
   secret_key_base: 8bcfe54ee0999f12fc3a57eb369fc010788e3e4335ea8a7d85dea34859265260b9bfdce6bc92296c2e8d4af1ca534276759b14de510de9347b67b0b1c8532bee
@@ -35,3 +36,4 @@ test:
 production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
   email_alert_auth_token: <%= ENV["EMAIL_ALERT_AUTH_TOKEN"] %>
+  accounts_jwt_public_key: <%= ENV["ACCOUNTS_JWT_PUBLIC_KEY"] %>
--- a/spec/integration/subscribers_account_jwt_spec.rb
+++ b/spec/integration/subscribers_account_jwt_spec.rb
@@ -1,0 +1,72 @@
+RSpec.describe "Subscribers from account JWT", type: :request do
+  let(:private_key) { OpenSSL::PKey::EC.new("prime256v1").tap(&:generate_key) }
+  let(:public_key) { OpenSSL::PKey::EC.new(private_key).tap { |key| key.private_key = nil } }
+
+  let(:email) { "text@example.com" }
+  let(:email_verified) { true }
+
+  let(:jwt) do
+    payload = { email: email, email_verified: email_verified }
+    JWT.encode payload.compact, private_key, "ES256"
+  end
+
+  let(:params) { { jwt: jwt }.compact }
+  let(:path) { "/subscribers/account-jwt" }
+
+  before do
+    allow(Rails.application.secrets).to receive(:accounts_jwt_public_key).and_return(public_key)
+  end
+
+  context "with authentication and authorisation" do
+    before do
+      login_with_internal_app
+    end
+
+    it "returns the subscriber details" do
+      post path, params: params
+      expect(response.status).to eq(200)
+      expect(JSON.parse(response.body)["subscriber"]).to_not be_nil
+    end
+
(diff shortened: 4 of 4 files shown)


app/controllers/subscribers_from_account_token_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+private
+
+  def authorise
+    authorise_user!("manage_subscriptions_from_account")
+  end
+
+  def validate_token!(address, token)
+    account_manager = Plek.find("account-manager")
```
brucebolt, line 37: You'll need to set the `PLEK_SERVICE_ACCOUNT_MANAGER` environment variable for this app in each environment, since it's not using the typical `app-name.publishing.service.gov.uk` structure.

brucebolt (approved):
👍

app/controllers/subscribers_from_account_token_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    render json: { subscriber: subscriber }
+  end
+
+private
+
+  def authorise
+    authorise_user!("manage_subscriptions_from_account")
+  end
+
+  def validate_token!(address, token)
```
kevindew, line 36: Minor, this isn't a conventional use of a bang method: https://rubystyle.guide/#dangerous-method-bang

app/controllers/subscribers_from_account_token_controller.rb
```diff
@@ -0,0 +1,79 @@
+class SubscribersFromAccountTokenController < ApplicationController
```
kevindew, line 1: There seems to be a heck of a lot going on in a controller here which looks like a bit of a smell. Would we be able to abstract the account API client to a separate class? To do so you can probably trim a bunch of boilerplate by having it inherit from GdsApi::Base.

app/controllers/subscribers_from_account_token_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    subscriber = find_subscriber(address)
+
+    render json: { subscriber: subscriber }
+  end
+
+private
+
+  def authorise
+    authorise_user!("manage_subscriptions_from_account")
```
kevindew, line 33: Do we need a special permission for this? I found it a bit confusing because this code is both authorising the requesting app and the user. Since the call to accounts is to validate a token I don't think we'd care too much if this end point was called by a different authenticated user would we?

app/controllers/subscribers_from_account_token_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def validate_params
+    ParamsValidator.new(expected_params).validate!
+  end
+
+  class ParamsValidator < OpenStruct
+    include ActiveModel::Validations
+
+    validates :address, presence: true
+    validates :address, email_address: true
```
kevindew, line 75: Minor, this can just be `validates :address, presence: true, email_address: true`

config/routes.rb
```diff
@@ -26,6 +26,8 @@
     post "/subscribers/auth-token", to: "subscribers_auth_token#auth_token"
     post "/subscriptions/auth-token", to: "subscriptions_auth_token#auth_token"
 
+    post "/subscribers/from-account-token", to: "subscribers_from_account_token#from_token"
```
kevindew, line 29: Minor but I wonder if the from is superfluous here and this can be named more consistently as per auth_token?

kevindew (commented):
Thanks for raising, I've dropped a few comments in (I assumed it was ready for review 🤞 )

I understand this is all for a prototype so you're probably factoring speed here but there are a couple of things about this approach that feels rather peculiar that you might want to consider when iterating.

The first of these is that it feels a bit strange that we provide the payload needed (address) and then use the token to verify that payload with an API call. This seems an approach more commonly handled by a JWT token or similar where the integrity of the request can be determined by a shared secret rather than needing to make an API call.

The other thing that seemed a bit strange was that an external parameter pushed into this app is then used as the means to authenticate with the other app. This makes Email Alert API something of a silent partner where there isn't any record that it was actually Email Alert API that communicated with the accounts system and instead looks like the user made that request manually. It'd feel more conventional if Email Alert API was a client of accounts that authenticated with the app and asked accounts "is this token accurate?"

Hope these suggestions help.

barrucadu:
Thanks for reviewing Kevin,

> The first of these is that it feels a bit strange that we provide the payload needed (address) and then use the token to verify that payload with an API call. This seems an approach more commonly handled by a JWT token or similar where the integrity of the request can be determined by a shared secret rather than needing to make an API call.

My thinking with this approach was that currently email-alert-api enforces that you can't turn an address into a subscriber ID without sending a verification email; but this endpoint does let you do that.  We don't want to lose that security guarantee, so rather than just giving the accounts system a bearer token which lets it manage subscriptions however it wants, it should instead pass a token to email-alert-api which lets email-alert-api confirm that the user has validated their address.

However, I can see that this is a bit weird, I'll change this endpoint to accept the userinfo JSON as a JWT instead (and remove the extra `address` parameter).

> The other thing that seemed a bit strange was that an external parameter pushed into this app is then used as the means to authenticate with the other app. This makes Email Alert API something of a silent partner where there isn't any record that it was actually Email Alert API that communicated with the accounts system and instead looks like the user made that request manually. It'd feel more conventional if Email Alert API was a client of accounts that authenticated with the app and asked accounts "is this token accurate?"

:+1: removing the token and passing the userinfo directly will resolve this.

app/controllers/subscribers_from_account_token_controller.rb
barrucadu: With the switch to checking a JWT's signature, the controller is much smaller now.

barrucadu:
I think I've addressed all the comments now

kevindew:
Thank you Michael - fast work and looks good 👍 

Going back a step on this I'm wondering if we're actually creating more of a coupling to GOV.UK accounts than is needed (sorry I very much should have caught this first time around). As I understand what accounts need is the mechanism to create an account based on an email address and to look one up based on an email address - these feel like functionality Email Alert API can achieve without verifying that accounts made such a request? I wonder if we're creating a rather strange RPC call here for something that can be achieved by more generic API requests - since other API clients don't need this level of verification.

What has made me ponder this is mostly how we have the logic of email_verified in Email Alert API whereas I ponder why would the account manager make the request if the email wasn't verified? so I wonder if we're slipping into a distributed logic situation inadvertently here. 

I wonder if, to meet your needs, you need either a pair of endpoints a GET that can 404 and a POST that can create a subscriber, or a single POST that handles both a la:

```
  def create
    subscriber = Subscriber.resilient_find_or_create(
      params.require(:address),
    )
    created = subscriber.saved_change_to_attribute?(:id)
    subscriber.activate if subscriber.deactivated?
    render json: { subscriber: subscriber }, status: created ? :created : :ok
  end
```

One thing about the approach here that goes against the grain of the Email Alert API system is the concept of creating a subscriber without them having any subscriptions. The intention (that may have got a bit lost in some of the double opt-in) was that users without subscriptions have their [personal details removed after 30 days automatically](https://github.com/alphagov/email-alert-api/blob/a656389b1abdd46226ca37c1682c318f1c2eafee/app/workers/nullify_deactivated_subscribers_worker.rb#L11-L14) and with planned work will be fully deleted after a year. I imagine this would cause problems for you if you held onto an id for any longer than a short session - and given that email addresses can be changed, you really shouldn't hold onto that id for anything longer than a short session.

Sorry to have this discussion after your second implementation. I feel like I reviewed the implementation last time and now I'm reviewing more the concept.

barrucadu:
I started typing out a response, checked on some things, and realised I might have just not taken in some of the email-alert-api functionality.  So, just to confirm,

We can sign someone up for email notifications by email address (not by subscriber ID) with `SubscriptionsController#create`.  This sends an email saying that the user will receive updates, but it *doesn't* ask them to click a confirmation link (I think this is the bit I missed).  It returns the subscription ID, which we can store and use to update or delete.  All good.  We can't list a user's subscriptions without knowing their ID, but it looks like we can get that by immediately calling `SubscriptionsController#show` with the subscription ID we just got?

I'd somehow got it into my head that the only way to sign someone up for something through the API is to send them an email with a confirmation link, but that doesn't look to be the case after all.

kevindew:
> I started typing out a response, checked on some things, and realised I might have just not taken in some of the email-alert-api functionality. So, just to confirm,
> 
> We can sign someone up for email notifications by email address (not by subscriber ID) with `SubscriptionsController#create`. This sends an email saying that the user will receive updates, but it _doesn't_ ask them to click a confirmation link (I think this is the bit I missed). It returns the subscription ID, which we can store and use to update or delete. All good. We can't list a user's subscriptions without knowing their ID, but it looks like we can get that by immediately calling `SubscriptionsController#show` with the subscription ID we just got?
> 
> I'd somehow got it into my head that the only way to sign someone up for something through the API is to send them an email with a confirmation link, but that doesn't look to be the case after all.

Sure, the `SubscriptionsController#create` is a rather complex endpoint - we've got it flagged as [a source of pain](https://trello.com/c/Z8tTO9pr/368-dev-pain-the-subscription-create-endpoint-does-basically-everything-that-the-update-endpoint-does) feel free to comment on the card if you have anything to add.

You're right that people won't get a confirmation link they have to click. The email will be like this one:

Screenshot 2020-10-07 at 14 56 42

> We can't list a user's subscriptions without knowing their ID, but it looks like we can get that by immediately calling `SubscriptionsController#show` with the subscription ID we just got?

Yes, that'll work in a convoluted way. `SubscriptionsController#create` provides a subscription id, `SubscriptionsController#show` will return the details of that subscription, including the subscriber with their id:

```
irb(main):005:0> subscriber_id = GdsApi.email_alert_api.get_subscription("fffffff9-26e5-49a1-b863-40432a2ab136").dig("subscription", "subscriber", "id")
=> 609492
```

barrucadu:
Great, thanks Kevin.  We don't need to change email-alert-api after all :+1:

PR #1429, opened 2020-10-08
Fix test that fails before 8am

kevindew:
As DigestRun models only validate when they are created at 8am this test
will fail on any PR's raised before then (normally Dependabot, thankfully
none of us work through the night).

This also applies some minor refactoring where the class and test were
more fragmented than needed.


1pretz1 (approved):
Good catch!

PR #1437, opened 2020-10-15
Restrict deleting subscribers if emails exist

1pretz1:
We want to change the behaviour of the db to restrict the deletion of
subscribers if they have associated emails, the previous behaviour was
to cascade the emails.

We don't want to do that anymore as we introduce a job to cleanup old
[subscribers] as emails shouldn't exist at the point of subscriber
deletion and if they do then we should raise an error.

I'm not sure why this foreign key was specifically named as I find no mention of it 
in the code so have removed.

Trello:

[subscribers]: https://github.com/alphagov/email-alert-api/pull/1432

### DIFF
--- a/db/migrate/20201020153450_change_foreign_key_constraint_for_emails.rb
+++ b/db/migrate/20201020153450_change_foreign_key_constraint_for_emails.rb
@@ -1,0 +1,6 @@
+class ChangeForeignKeyConstraintForEmails < ActiveRecord::Migration[6.0]
+  def change
+    remove_foreign_key :emails, :subscribers
+    add_foreign_key :emails, :subscribers, on_delete: :restrict
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_10_19_154701) do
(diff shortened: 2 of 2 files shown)


kevindew (approved):
Looks good - do you know what the impacts of running this against a full database? I imagine it's a locking query so it'd be good to know how long it takes on staging to know if it's something that risks downtime.

1pretz1:
I was assuming it would create downtime and have to be run out of hours but I was going to verify this.

I'll test tomoz!

1pretz1:
So you're right in that it's a locking query @kevindew.

Tested in integration and removing the foreign key was super quick (1 second), adding it back not so much (81 seconds) for around 4.5 million records which made me slightly wary concerning larger sets.

However, running in staging with approx 20 million records resulted in a run time of around 23 seconds. Extra db juice making the difference:

```
11:25:07 -- remove_foreign_key(:emails, :subscribers)
11:25:07    -> 0.0200s
11:25:07 -- add_foreign_key(:emails, :subscribers, {:on_delete=>:restrict})
11:25:30    -> 22.9871s
```

kevindew:
@1pretz1 Cool thanks for trying it out. I guess that means we have to be rather delicate in when we deploy this for production then and should probably anticipate it taking around a minute to be on the safe side.

PR #1438, opened 2020-10-19
Remove usage of DeliveryAttempt model

kevindew:

This PR removes our usage of the DeliveryAttempt model. Once this is merged in we will no longer create DeliveryAttempt models when we send email and DeliveryAttempt models will no longer be updated when Notify sends a status update. This change is not backwards compatible, should we need to rollback this change aspects of data consistency will be lost - this seemed acceptable to me as we have to make this leap at some point to apply this change.

This PR contains the following things:

- Removal of a number of metrics stored as part of a DeliveryRequest worker that aren't used on any Grafana dashboards. Alternatives haven't been added but can be in the future should we want to get a closer insight into performance, the ones we had felt like temporary debugging data
- No longer creating DeliveryAttempt models as part of sending an email
- The status-update endpoint no longer interacts with DeliveryAttempt models, it also does not interact with Email anymore and uses the to field that Notify send as a means to determine subscriber. We no longer store whether an email was delivered or not in the DB, instead logging it and reflecting this data is available in Notify (see: https://trello.com/c/Xd47oJKZ/535-remove-the-deliveryattempt-model#comment-5f89a47789889d7095e93a17) 
- I renamed DeliveryRequestWorker and DeliveryRequestService to be SendEmailWorker and SendEmailService to reflect that DeliveryRequest isn't reflective of our current terminology and isn't particularly clear.
- It adapts the Rake task to look up emails from Notify to use the new id, noting that it won't work for items during this transition

It'd be easiest to review this commit by commit as the renaming ones swamp things a bit.

Related PRs:

- https://github.com/alphagov/govuk-puppet/pull/10787
- https://github.com/alphagov/govuk-puppet/pull/10789

Following this there are follow-up tasks:

1. Update govuk-developer-docs: https://github.com/alphagov/govuk-developer-docs/blame/afb4febac0c7cfe4dabf8e8a5728ade73642e636/source/manual/govuk-notify.html.md#L31 and https://github.com/alphagov/govuk-developer-docs/blame/003d358904638daa928f4fb67fa3200a705e10fa/source/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html.md#L32
2. Update puppet to adopt new metrics for the dashboard: https://github.com/alphagov/govuk-puppet/blob/edfaee70f90eda8957a9acfd3c523ea0a6d794b2/modules/grafana/files/dashboards/email_alert_api_technical.json#L540-L555
3. Remove transitory steps (duplicate metrics, aliasing, and old queue names)
4. Once we're happy, drop the DeliveryAttempts table


spec/integration/status_updates_spec.rb
```diff
@@ -1,69 +1,54 @@
 RSpec.describe "Receiving a status update", type: :request do
-  let(:delivery_attempt) { create(:delivery_attempt, status: "sent") }
-  let(:reference) { delivery_attempt.id }
-
-  let(:permissions) { %w[signin status_updates] }
-  let(:user) { create(:user, permissions: permissions) }
-  before { login_as(user) }
-
   describe "#create" do
```
benthorner, line 2: Out-of-scope: is there a reason why this directory is called "integration", instead of "requests"? The latter is more specific than the former, and seems to capture the current contents.

spec/integration/status_updates_spec.rb
```diff
@@ -1,69 +1,54 @@
...
     end
 
-    context "when a status attempt arrives for an already handled delivery attempt" do
-      before { delivery_attempt.update!(status: "delivered") }
+    context "when a status attempt arrives with an permanent-failure status" do
+      let(:status) { "permanent-failure" }
 
-      it "renders 409" do
-        post "/status-updates", params: params
+      it "can unsubscribe all of a subscribers subscriptions by email id" do
```
benthorner, line 36: We're missing a test for when the subscriber doesn't exist. (Looks like the previous class was missing it also, so I can understand why.)

benthorner (commented):
Nice cleanup! `+264 -550` is a good indicator this is simplification 💪. I also found the links to follow-up work helpful 💯. Just a few inline comments, although some of them are more major.

Some other related, but non-blocking things:

- I was wondering what happens if Notify gets a repeat request for the same "unique" reference e.g. if 2ndline are trying to debug someone not receiving emails. I tried this myself and it just sends the email again i.e. no error.

- Could you explain the reasoning behind waiting a week to drop the table?

- This seems to have gotten lost on the card, but previously we agreed an ADR makes sense after the culmination of this work, not least because it supersedes two previous ADRs [[1](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-003-data-retention.md)] [[2](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-001-notify-integration.md)]. Shall I add that as a separate card, so we can agree the gist of it before you go on leave?

kevindew:
Thanks for the review @benthorner - I'll go through it properly tomorrow but just thought I'd quickly reply to one comment.

> * This seems to have gotten lost on the card, but previously we agreed an ADR makes sense after the culmination of this work, not least because it supersedes two previous ADRs [[1](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-003-data-retention.md)] [[2](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-001-notify-integration.md)]. Shall I add that as a separate card, so we can agree the gist of it before you go on leave?

I had considered that part of the card so I started work on that while waiting for a review. So feel free to just add something to the card for it, should hopefully have a draft up tomorrow morning.

app/workers/delivery_request_worker.rb
```diff
@@ -16,12 +14,12 @@ def perform(email_id, metrics, queue)
       return
     end
 
-    attempt = DeliveryRequestService.call(
+    increment_rate_limiter
```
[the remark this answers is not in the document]
kevindew: Yup was intentional, not mentioning it was not - I'll amend the commit.

lib/metrics.rb
```diff
@@ -45,13 +45,15 @@ def email_bulk_insert(size, &block)
       time("email_bulk_insert.#{size}.timing", &block)
     end
 
-    def email_created_to_first_delivery_attempt(created_time, first_attempt_time)
-      difference = (first_attempt_time - created_time) * 1000
-      timing("email_created_to_first_delivery_attempt", difference)
+    def email_created_until_email_sent(created_time, sent_time)
```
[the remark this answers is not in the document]
kevindew: I don't think it'll be very useful on the product dashboard given the 7 day timeframe it has. I'll remove it.

app/workers/delivery_request_worker.rb
```diff
@@ -47,18 +52,6 @@ def increment_rate_limiter
     rate_limiter.add("delivery_request")
   end
 
-  # More information around the rate limit can be found here ->
-  # https://docs.publishing.service.gov.uk/manual/govuk-notify.html under "GOV.UK Emails".
-  def rate_limit_threshold
-    per_minute_to_allow_350_per_second = "21000"
```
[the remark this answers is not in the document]
kevindew: Oh good catch - I must have totally messed up my puppet search 🤦 . Yeah seems worth removing. I'll open a PR for puppet.

lib/services.rb
```diff
@@ -2,8 +2,4 @@ module Services
   def self.content_store
     @content_store ||= GdsApi::ContentStore.new(Plek.new.find("content-store"))
   end
-
-  def self.rate_limiter
-    @rate_limiter ||= Ratelimit.new("email-alert-api:deliveries")
```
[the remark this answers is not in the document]
kevindew: Another good catch, I had expected it uses Redis.connection like most of these utilities. I can swap it over to use the same redis sidekiq uses (which will mean it benefits from the dubious redis-namespace utility)

spec/integration/status_updates_spec.rb
kevindew: It seems to be part of a broader testing strategy that didn't really take off: https://github.com/alphagov/email-alert-api/pull/300. I can see that spec/requests was renamed to spec/integration. I'd be happy for us to change it back.

kevindew:
> * I was wondering what happens if Notify gets a repeat request for the same "unique" reference e.g. if 2ndline are trying to debug someone not receiving emails. I tried this myself and it just sends the email again i.e. no error.

Yeah, as far as I could tell they accept it. It's actually a shame that it does that as it'd be nice if they prevented us sending same email twice (as if we have a situation like a 502 or 504 response from them we don't know if it was successful or not)

> * Could you explain the reasoning behind waiting a week to drop the table?

That was based on the suggestion in the card which suggested waiting until the data was deleted automatically, although that incorrectly referencing 14 days as the period until data is automatically deleted. There isn't anything to really stop us deleting it sooner though just will be a relatively slow migration I believe.

app/workers/delivery_request_worker.rb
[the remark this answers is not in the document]
benthorner: I had a quick look, but can't see anything about it. Perhaps it got lost in another amend?

spec/integration/status_updates_spec.rb
kevindew: Thanks - it was one of those rather annoying negative tests, I'm always in two minds about adding tests that test code doesn't do something.

app/workers/delivery_request_worker.rb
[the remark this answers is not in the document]
kevindew: I've added details about this in the commit message. It was actually a bigger can of worms than I originally realised because of me missing that darn puppet env var. As we didn't increment the rate limit unless we made an attempt staging/integration would have very low numbers here (as must of their attempts are aborted due to https://github.com/alphagov/email-alert-api/blob/166834e3a1b78a3f427f7684ce98121f89a80bf6/app/services/delivery_request_service.rb#L19. Therefore with this being pre-emptive rather than retrospective we need staging/integration to have similar rate limit otherwise they'll be rate limiting on each publishing. 

My takeaway from this was that the rate limit was rather oddly coupled, as if it applies to notify only it should be part of the NotifyProvider whereas if it is at worker level it should be for all emails regardless of whether they're sent or provider.

lib/metrics.rb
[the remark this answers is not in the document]
kevindew: I've removed it now.

app/workers/delivery_request_worker.rb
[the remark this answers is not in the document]
kevindew: Opened a PR here: https://github.com/alphagov/govuk-puppet/pull/10787

lib/services.rb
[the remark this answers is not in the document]
kevindew: I've switched to using Sidekiq's redis.

kevindew:
Thanks @benthorner I've made amends for all your comments. I've then gone and done something annoying and expanded the scope a bit after feeling that the queue names were now a bit strange "delivery_immediate" and have included them too. Do shout if that makes reviewing this a pain and I'll cherry pick them out for a different PR.

kevindew:
> Thanks @benthorner I've made amends for all your comments. I've then gone and done something annoying and expanded the scope a bit after feeling that the queue names were now a bit strange "delivery_immediate" and have included them too. Do shout if that makes reviewing this a pain and I'll cherry pick them out for a different PR.

I'm now just going to look at the impact of queue name changes in puppet and dev docs.

spec/integration/status_updates_spec.rb
benthorner: Hmm...interesting. Looking again I forgot this repo also has `spec/features`, which are also `type: :request`. It also turns out there are [non-request specs](https://github.com/alphagov/email-alert-api/blob/master/spec/integration/anonymise_email_addresses_spec.rb) in this directory, and [a README](https://github.com/alphagov/email-alert-api/blob/master/spec/integration/README.md) that conflicts with the tighter scope I was suggesting. While it's a shame, I think it would introduce more confusion to rename it; if it starts causing pain, then we should consider revising the test strategy. Thanks for the chat about it!

lib/services.rb
benthorner, line 7: Note to self: double-checked this prefix matches [the automatic namespace from govuk_sidekiq](https://github.com/alphagov/govuk_sidekiq/blob/master/lib/govuk_sidekiq/railtie.rb#L16).

spec/integration/status_updates_spec.rb
benthorner: Well, at least it's in good spirit: we could have code that would error otherwise.

app/workers/send_email_worker.rb
```diff
@@ -13,15 +13,15 @@ class DeliveryRequestWorker < ApplicationWorker
   def perform(email_id, metrics, queue)
     if rate_limit_exceeded?
       logger.warn("Rescheduling email #{email_id} due to exceeding rate limit")
-      GovukStatsd.increment("delivery_request_worker.rescheduled")
-      DeliveryRequestWorker.set(queue: queue || "delivery_immediate")
+      GovukStatsd.increment("send_email_worker.rescheduled")
```
benthorner, line 16: Minor: we don't use this, and it's inconsistent with the (apparent) pattern of proxying metrics through [the monolith](https://github.com/alphagov/email-alert-api/blob/master/lib/metrics.rb). Since we're renaming it, we could just remove it.

app/workers/send_email_worker.rb
```diff
@@ -34,7 +34,7 @@ def self.perform_async_in_queue(email_id, metrics = {}, queue:)
 private
 
   def rate_limit_exceeded?
-    rate_limiter.exceeded?("delivery_request",
+    rate_limiter.exceeded?("requests",
```
benthorner, line 37: This change will need a bit of careful handling: we'll temporarily reset our counter and stop rate limiting.

lib/tasks/support.rake
```diff
@@ -131,9 +131,9 @@ namespace :support do
   end
 
   desc "Send a test email to an email address"
-  task :deliver_to_test_email, [:test_email_address] => :environment do |_t, args|
+  task :send_test_email, [:email_address] => :environment do |_t, args|
```
benthorner, line 134: This will need an update [here](https://github.com/alphagov/email-alert-api/blob/master/docs/support-tasks.mdl).

benthorner (approved):
Thanks, all looks good to me 👍. I do agree this has turned into rather a large PR, with several potential areas for breaking: rate limiting, queues, class names. Reflecting, I would have found it helpful if it were split up: it was easy enough to follow the commits, but it did feel like a long review, and it would be a shame if the whole PR got held up by one commit.

The only major issue I see is the change to the rate limiting namespace. It makes sense, but we'll want to be careful when we deploy this change.

> There isn't anything to really stop us deleting it sooner though just will be a relatively slow migration I believe.

Wonders why he wrote that and can't remember any good reason - let's drop it ASAP!

app/workers/send_email_worker.rb
kevindew: Yup, we already have to be careful because of the change of queue names too

lib/tasks/support.rake
kevindew: That is already part of https://github.com/alphagov/email-alert-api/pull/1438/commits/3075e03d8ae4f3f612112ec687ca36d03acc9477

app/workers/send_email_worker.rb
kevindew: Sure, removed.

lib/tasks/support.rake
benthorner: Oh sorry - I confused myself going in a loop through the dev docs!

lib/services.rb
kevindew: This is the namespace it comes out with:

```
irb(main):069:0> Redis.new.exists("email-alert-api:ratelimit:send_email:requests")
=> true
```

Note the key only lasts as long as the rate limit is running so you'll find in quiet times that this returns false.

benthorner: Thanks for clarifying 👍

1pretz1:
Only just got a chance to look through this, looks great! 💯

PR #1439, opened 2020-10-19
Make emails deleteable after a week

1pretz1:
This is so we don't end up with lingering emails and subscription
contents in the system which will never be deleted or recovered.

Trello:

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -5,16 +5,6 @@ class Email < ApplicationRecord
   COURTESY_EMAIL = "govuk-email-courtesy-copies@digital.cabinet-office.gov.uk".freeze
   has_many :delivery_attempts
 
-  scope :archivable,
-        lambda {
-          where(archived_at: nil).where.not(status: :pending)
-        }
-
-  scope :deleteable,
-        lambda {
-          where.not(status: :pending).where("archived_at < ?", 7.days.ago)
-        }
-
   enum status: { pending: 0, sent: 1, failed: 2 }
 
   def self.timed_bulk_insert(records, batch_size)
--- a/app/queries/email_archive_query.rb
+++ b/app/queries/email_archive_query.rb
@@ -4,7 +4,7 @@ class EmailArchiveQuery
   end
 
   def call
-    Email.archivable.select(fields)
+    Email.where(archived_at: nil).select(fields)
   end
 
   private_class_method :new
--- a/app/workers/email_deletion_worker.rb
+++ b/app/workers/email_deletion_worker.rb
@@ -2,7 +2,7 @@ class EmailDeletionWorker < ApplicationWorker
   def perform
     run_with_advisory_lock(Email, "delete") do
       start_time = Time.zone.now
-      deleted_count = Email.deleteable.delete_all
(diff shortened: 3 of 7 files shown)


benthorner (commented):
Much simpler 👍. As much as archiving is a bit of a side-show, we ought to try and preserve it's place in the timeline of an email, though. How about this?

- Emails can be archived immediately. This makes more sense now, because [we no longer send any sending-specific stats along with the email](https://github.com/alphagov/email-alert-api/commit/090df0e2221a3597f2d1d6f36a0b8d2dac7d27d7#diff-46c313c179f423a1fe30f6a9cd62600e7b1d5babec7c4c4894b0d45296e75620).

- Emails can be deleted once they're archived (after a week).

1pretz1:

> Emails can be archived immediately. This makes more sense now, because we no longer send any sending-specific stats along with the email.

Do you mean archiving the emails as soon as they've been sent as part of the [DeliveryRequestWorker]? Currently the archiving worker runs every 30 mins which is pretty soon after an email has been sent.

> Emails can be deleted once they're archived (after a week).

So this sounds like there needs to be some sort of logic where we delete pending emails regardless after a week but don't for emails that could be archived (not pending) but haven't been... yet. Does that capture what you mean?

[DeliveryRequestWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/delivery_request_worker.rb

benthorner:
Sorry @1pretz1, I fear I've made this more confusing than it is. Does this make more sense?

- We don't need a scope for archiving emails; they can be archived at any time.
- Emails can be deleted when `archived_at < 1.week.ago`.

So we always archive, but we can only delete once we've archived (which should always be true).

1pretz1:
Sorry if I'm being dense @benthorner (entirely possible).

> they can be archived at any time 

By 'any time' do you mean 'any status' @benthorner ?

If so, my original thinking was that we don't want to archive pending emails as it may be confusing that we have emails in our archives that we've never attempted to send and our only there due to a system error but now as I'm writing this I'm thinking that this may actually be good as it gives us some visibility of an issue that may vanish if these emails just disappear after a week without being pushed to s3.

> Emails can be deleted when archived_at < 1.week.ago.

This makes sense in the context of pending emails being archived otherwise `archived_at` would never be populated for these and so the emails would never be deleted.

benthorner:
@1pretz1 that's OK, it's good to clarify.

> By 'any time' do you mean 'with any status' @benthorner ?

Yes, I do. The archiving has never made much sense to me, and we plan to get rid of it. Until then, though, we should try to keep it working as best we can. Archive doesn't have to mean "old" or "finished" - it's just a historical record - so it's consistent to put any and all emails in it, irrespective of what happens to them (noting that we don't archive any status information anymore).

1pretz1:
Sure, thanks for explaining @benthorner !

Have updated accordingly.

benthorner (approved):
Nice works cleaning up the factories 👍.

One minor thing: the first commit could explain that we weren't previously able to change the archive strategy because of the timing fields that we sent to Athena.

1pretz1:
Sure, I've added that to the commit message!

PR #1441, opened 2020-10-20
ADR on email sending and delivery responsibilities

kevindew:

This reflects a number of recent changes made to the system with how it
communicates with Notify.

👉 [Rendered](https://github.com/alphagov/email-alert-api/blob/email-sent-adr/docs/adr/adr-005-email-delivery-responsibilities.md)

### DIFF
--- a/docs/adr/adr-006-email-delivery-responsibilities.md
+++ b/docs/adr/adr-006-email-delivery-responsibilities.md
@@ -1,0 +1,104 @@
+# 6. Email Delivery Responsibilities
+
+Date: 2020-10-23
+
+## Context
+
+Email, as an asynchronous communication medium, inherently has two boolean
+scenarios that need to be true for an email to be received. An email has to be
+_sent_ to a mail server and the mail server then needs to _deliver_ the email to the
+recipient's mail server. Resolving whether an email is sent is a rather simple
+question that can be resolved synchronously - "did the mail server accept the
+email?" - however, determining whether an email is received is more complex
+as delivery is not synchronous and may be subject to automated retrying should
+any problems occur. Typically, in the medium of email, clients only consider
+the first scenario in reporting an email's status, this is the point an email
+becomes sent. The latter scenario, delivery, isn't typically reflected in
+email clients, it is normal to assume the email was received
+successfully unless later you receive an email indicating a bounce has
+occurred.
+
+The [initial Email Alert API Notify integration][adr-1] was designed to
+consider both the sending and delivery of email. It had systems to monitor
+whether Notify had managed to deliver an email to a recipient and an automatic
+retry mechanism for when Notify had problems delivering an email.
+
+In September 2020 we decided to re-evaluate this concept conflation as we were
+concerned that Notify functionality was duplicated in Email Alert API and
+that this had an adverse effect on the complexity of the system.
+
+[adr-1]: adr-001-notify-integration.md
+
+## Decision
+
+We decided that Email Alert API has a responsibility to send email, but does
+not have a responsibility to deliver it. That is the responsibility of Notify.
+
+We defined that, in the context of Email Alert API, the process of sending
+email was the ability to successfully send a request to Notify to perform
+this action - we consider this the equivalent of a mail client successfully
+sending an email to an [SMTP][] server. We then consider the
+[Notify callback][notify-callback] as the mechanism to learn if an email was
+delivered or not.
+
+To reflect this we have [switched][switched-to-sent-success] the
+meaning of an [email's status][email-status]. A status of "sent" now means that
+the email was accepted by Notify and not that the email was delivered. A
+status of "failed" now means we weren't able to send the email to Notify,
+instead of the previous meaning where it meant that it may not have been sent
+to Notify or Notify failed to deliver the email.
+
+[SMTP]: https://en.wikipedia.org/wiki/Simple_Mail_Transfer_Protocol
+[notify-callback]: https://docs.notifications.service.gov.uk/ruby.html#delivery-receipts
+[switched-to-sent-success]: https://github.com/alphagov/email-alert-api/commit/c457f62c3b6f1eaadf47e6596223cc0fdcffa853
+[email-status]: https://github.com/alphagov/email-alert-api/blob/0b87e62288ddb1653451f84e1f36e17ce4e8e9dc/app/models/email.rb#L7
+
+## Status
+
+Accepted
+
+## Consequences
+
+### We no longer retry delivering emails ourselves
+
+Notify, via Amazon SES, will automatically [retry][ses-bounce] the delivery of
+emails that fail for transitory reasons (such as the recipient's mail server
+being full or offline). If it doesn't succeed in a reasonable period
+of time Notify will inform us by telling us an email has experienced a
+["temporary-failure"][temporary-failure-status].
+
+Email Alert API [no longer][no-retries-commit] has its own retry
+mechanism to re-attempt emails that have failed to be delivered. This
+resolves an aspect of duplicated functionality between the two systems and
+an area of ambiguity as to how long Email Alert API should retry for. With
+this removed there are no longer attempts to resend an email should the
+first delivery fail, this allowed us to [delete the
+`DeliveryAttempt` model][delete-delivery-attempt] as the purpose of this was
+to disambiguate between different Notify requests.
+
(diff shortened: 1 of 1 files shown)


docs/adr/adr-005-email-sending-and-delivery-responsibilities.md
```diff
@@ -0,0 +1,120 @@
...
+We reflected that Email Alert API had diverged from the common pattern of email
+clients by conflating the sending and delivery of email. This resulted
+in the sending of email to be somewhat pessimistic - we considered an email to
+only be sent when it was confirmed to be delivered - and to be
+somewhat confusing, with terminology about sending emails mixed with
+terminology about delivering email.
+
+To resolve these concerns we have switched to considering an email to
+be _sent_ when the request to Notify is successful, which will end the period
+where an email is in a transitory `pending` status. We will consider it the
```
benthorner, line 48:
```suggestion
where an email is in a transitory `pending` status that could mean anything from not sent, to delivered but the callback got lost. We will consider it the
```

docs/adr/adr-005-email-sending-and-delivery-responsibilities.md
```diff
@@ -0,0 +1,120 @@
...
+responsibility of Notify to manage and monitor the delivery of an email,
+which is already part of the Notify service.
+
+[notify-callback]: https://docs.notifications.service.gov.uk/ruby.html#delivery-receipts
+
+### Consequences
+
+#### Emails will leave their `pending` state when sent to Notify
+
+Email Alert API would consider an email to be in a `pending` status until
```
benthorner, line 58:
```suggestion
Email Alert API would consider a sent email to be in a `pending` status until
```

This isn't quite true, if we couldn't send it to Notify, or the delivery failed (depending on what period in time we're thinking about).

benthorner (commented):
Not a grammatical error in sight 💯. It's also a helpful RFC, and thanks for putting effort into the extensive discussion about _sending_ vs. _delivery_, as that's quite a subtle distinction that is likely the origin of the mess we were in!

I've got a few comments, but they're mostly just suggestions on two themes: 

- There are two tenses in use, which was confusing when I read this from top-to-bottom. I think it's most useful to write in the past tense, as then it makes sense to see some links to PRs and commits, which will help future readers follow along.

- The word "send" is used in a few contexts, and sometimes I was confused about the scope of what it's referring to. We could try and define the scope for "send", or maybe be consistently specific e.g. "send to Notify" vs. "send all the way to end user".

docs/adr/adr-005-email-sending-and-delivery-responsibilities.md
```diff
@@ -0,0 +1,120 @@
...
+concluded that this was unnecessary as it duplicated functionality that
+Notify provided. We chose to simplify the system and re-define the
+responsibilities of email sending and delivery. We decided
+that Email Alert API's responsibility in sending email was ensuring a request
+was sent to Notify and that the onward responsibility of email delivery
+was a concern that belonged to Notify.
+
+[adr-1]: adr-001-notify-integration.md
+
+## Specifics
```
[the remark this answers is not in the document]
kevindew: Thanks - I've flipped that a bit. That's me struggling with the rather unusual ADR format we have here (that I'm pretty sure I'm to blame for 🤦 )

docs/adr/adr-005-email-sending-and-delivery-responsibilities.md
```diff
@@ -0,0 +1,120 @@
...
+email to be no longer pending and be in a final `sent` status.
+
+This reflects a more optimistic outlook towards email sending. Data suggests
+that it is very rare that an email fails to be delivered (On Friday 16th
+October 5,338,956 were delivered, against 21,966 failures, providing a 0.4%
+failure rate). By making the state of an email dependent on a callback we were
+leaving a problem that any lost or missed callbacks left the system in an
+[inconsistent state][archive-pr].
+
+This will mean that emails are eligible for archiving and deletion sooner,
```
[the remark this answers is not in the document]
kevindew: This seemed an unnecessary detail - particularly as we think archiving isn't used often.

kevindew:
Thanks @benthorner - I've addressed your comments. I put the changes in a separate commit for easier reviewing but will squash before merge.

benthorner (commented):
Thanks @kevindew! I've had another read through, and it was easier this time so thanks for making those changes. There's still one more major comments about the archiving, which I didn't connect with this work initially. I also struggled a bit with the structure still, and think there are few simple changes that would make it clearer and reduce some repetition:

- With only a few movements, we could re-partition the paragraphs into "Context", "Decision" and "Consequences" sections, which I've found more helpful in [the ADRs for Content Publisher](https://github.com/alphagov/content-publisher/blob/master/docs/adr/0001-record-architecture-decisions.md). Let's depart from the old layout if it leads to confusing results! We could make ADR 5 change the layout, and this one could become ADR 6.

- We mention how we're redefining the send/delivery distinction three times (in each section). The one I find confusing is in "Specifics", because it's really a consequence. Much of what is said in that section could be folded into other sections, also without too much effort (I hope).

I'm hoping those suggestions resonate with you, as they're not really about the content itself. Let's have a chat if the "movements" I'm suggesting aren't clear.

kevindew:
@benthorner it's a bit tedious having you ask for a restructure after it's already been through one round of amends. Do you want to open a PR to change the ADR structure for 5 and I'll amend this to be ADR 6?

benthorner:
Sure, happy to help: https://github.com/alphagov/email-alert-api/pull/1445.

I agree it's not fun to get a second review with major changes. I only suggested these changes because: I think they're quite easy, yet they'll have a big impact on readability. In my original review, I was struggling to picture the outcome of my suggestions and wrap my head around all the content, so I was reading each paragraph in isolation from the rest. The changes you've made have made it much easier for me to read them all together, hence the new comments.

I agree it's not worth spending much more time on this, beyond restructuring 👍.

docs/adr/adr-006-email-delivery-responsibilities.md
```diff
@@ -0,0 +1,110 @@
...
+The [initial Email Alert API Notify integration][adr-1] was designed to
+consider both the sending and delivery of email. It had systems to monitor
+whether Notify had managed to deliver an email to a recipient and an automatic
+retry mechanism for when Notify had problems delivering an email.
+
+In September 2020 we decided to re-evaluate this concept conflation as we were
+concerned that Notify functionality was duplicated in Email Alert API and
+that this had an adverse effect on the complexity of the system.
+
+In September 2020, we re-evaluated the need for this level of functionality and
```
benthorner, line 29:
```suggestion

```

Minor: we've said this first sentence already in the previous paragraph.

docs/adr/adr-006-email-delivery-responsibilities.md
```diff
@@ -0,0 +1,110 @@
...
+concluded that this was unnecessary as it duplicated features that
+Notify provided. We chose to simplify the system and re-define the
+responsibilities of email delivery. We decided that Email Alert API
```
benthorner, line 32:
```suggestion
responsibilities of email delivery.
```

Minor: we repeat this decision in the next paragraph (in the decision section), so we don't need to say it here.

docs/adr/adr-006-email-delivery-responsibilities.md
```diff
@@ -0,0 +1,110 @@
...
+A question we pondered with these changes was "what should do with the data
+we receive from Notify about the delivery of emails?". Now that this no longer
+plays a role in the state modelling of an email it risked confusion
+to include it in an email's status (it conflated what a failure to send
+an email meant - was this failure to _send_ or failure to _deliver_?).
+
+We decided that we did not need to store delivery success or failure in the
+database with the email. This is because there is not currently
+a known use of this data and, since the email table only has a 7-day retention
+period, can be added later to meet any later needs that are identified. The
```
benthorner, line 101:
```suggestion
period, it can be added later to meet any later needs that are identified. The
```

benthorner (approved):
Thanks, I find it much easier to read with the new structure 🏅. I think there are a couple of sentences we could loose, as we seem to be saying the same thing twice, but that's a minor issue.

docs/adr/adr-006-email-delivery-responsibilities.md
kevindew: Oh darn - that's a replacement paragraph that got left behind. I recall this is the second time I've done this recently 🤦

PR #1442, opened 2020-10-20
Add indexes for historical data deletion

1pretz1:
These foreign key indexes are needed so we can efficiently delete [historic] data from the database. Previously we tried running the new worker without these indexes and certain queries took an obscene amount of time due to the constraints we have setup on certain tables.

One example of the improved run time was the total time of deleting 15 messages including its constraints (matched messages and subscription contents). Without an index for the message_id on the subscription contents table the query took 77 seconds vs 0.22 seconds which is a sizeable speed boost.

More info (before and afters) are found within the commits ->

Trello:

[historic]: https://github.com/alphagov/email-alert-api/pull/1432


kevindew (approved):
Awesome benchmarks Pete - thanks for working all this out. 💯

benthorner (approved):
Same here 🥇. Always helpful to be reminded of the importance of good index management.

PR #1447, opened 2020-10-21
Remove transitionary code for #1438

kevindew:

This removes code that was left for backwards compatibility during the
migration of the queue names, worker names and metrics in https://github.com/alphagov/email-alert-api/pull/1438


benthorner (approved):
🇹🇿

PR #1449, opened 2020-10-21
Use Ruby 2.7 argument forwarding operator

kevindew:
This applies the `...` operator in this app. This feature was introduced in Ruby 2.7 and is the equivalent to `def method(*args, **kwargs, &block)`. This has been done to reduce the clutter we've acquired in places where we do in argument forwarding with various usages of empty double splats to satisfy method definitions.

More info in the commits.


kevindew:
Ugh - this looks like a big can of worms. Brakeman doesn't understand the [argument forward operator](https://github.com/presidentbeef/brakeman/issues/1449). I'll tentatively close this until I think of a good path forward.

PR #1452, opened 2020-10-23
Catch Net::ReadTimeout errors from Notify

kevindew:
We saw one of these come through to our error handling today. I had
thought we had covered timeouts previously with checking for
Net::OpenTimeout and didn't realise it had a sibling.

I did consider whether to catch the superclass of these, Timeout::Error,
but decided that this wasn't the best approach as Timeout::Error seems
unrelated to Net communications and I felt that it would be strange to
either a) not test the exceptions we expect or b) test against different
exceptions than the code.

### DIFF
--- a/app/services/send_email_service/notify_provider.rb
+++ b/app/services/send_email_service/notify_provider.rb
@@ -21,7 +21,7 @@ class SendEmailService::NotifyProvider
 
     Metrics.sent_to_notify_successfully
     :sent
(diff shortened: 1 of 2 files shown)


benthorner (approved):
Thanks for spotting and fixing - it's great to see the longterm commitment to keeping our error at bay ⭐. I agree the more specific error makes sense, since it's only a total of 3.

PR #1456, opened 2020-10-28
Add reporting list task

1pretz1:
Add task to report on subscriber lists. This task is needed so non devs (and devs tbf) can freely access metrics around subscriber lists without having to concoct queries every-time. This will also be a feed of information our new product dashboard.

Took ~5 mins to run in integration:
https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/190098/console

The task outputs as a CSV to stdout.

More info in the commits ->

Trello:
https://trello.com/c/PRXBoZBu/576-create-a-one-stop-data-shop-report-task-for-all-subscriber-lists

### DIFF
--- a/lib/reports/subscriber_lists_report.rb
+++ b/lib/reports/subscriber_lists_report.rb
@@ -1,0 +1,69 @@
+class Reports::SubscriberListsReport
+  attr_reader :date
+
+  CSV_HEADERS = %i[title
+                   slug
+                   matching_criteria
+                   created_at
+                   individual_subscribers
+                   daily_subscribers
+                   weekly_subscribers
+                   unsubscriptions
+                   matched_content_changes_for_date
+                   matched_messages_for_date].freeze
+
+  def initialize(date)
+    @date = Time.zone.parse(date)
+  end
+
+  def call
+    CSV do |csv|
+      csv << CSV_HEADERS
+
+      SubscriberList.where("created_at < ?", date.end_of_day).find_each do |list|
+        csv << export_list_row(list)
+      end
+    end
+  end
+
+private
+
+  def export_list_row(list)
+    unsubscriptions_count = list.subscriptions.where("ended_at <= ?", date.end_of_day).count
+    scope = list.subscriptions.active_on(date.end_of_day)
+
+    [list.title,
+     list.slug,
+     criteria(list),
+     list.created_at,
+     scope.immediately.count,
+     scope.daily.count,
+     scope.weekly.count,
+     unsubscriptions_count,
+     matched_content_changes_count(list),
+     matched_messages_count(list)]
+  end
+
+  def criteria(list)
+    list.as_json(root: false,
+                 only: %i[links
+                          tags
+                          email_document_supertype
+                          government_document_supertype
+                          document_type]).to_json
+  end
+
+  def matched_content_changes_count(list)
+    MatchedContentChange
+      .where(subscriber_list: list)
+      .where("created_at > ? AND created_at <= ?", date.beginning_of_day, date.end_of_day)
+      .count
+  end
+
+  def matched_messages_count(list)
+    MatchedMessage
+      .where(subscriber_list: list)
+      .where("created_at > ? AND created_at <= ?", date.beginning_of_day, date.end_of_day)
+      .count
+  end
+end
--- a/lib/reports/subscriber_lists_report.rb
+++ b/lib/reports/subscriber_lists_report.rb
@@ -17,6 +17,8 @@ class Reports::SubscriberListsReport
   end
 
   def call
+    validate_date
+
     CSV do |csv|
       csv << CSV_HEADERS
 
@@ -27,6 +29,12 @@ class Reports::SubscriberListsReport
   end
 
 private
+
+  def validate_date
+    raise "Invalid date" if date.blank?
+    raise "Date must be in the past" if date >= Time.zone.today
+    raise "Date must be within a year old" if date <= 1.year.ago
+  end
 
   def export_list_row(list)
     unsubscriptions_count = list.subscriptions.where("ended_at <= ?", date.end_of_day).count
--- a/lib/tasks/report.rake
+++ b/lib/tasks/report.rake
@@ -40,4 +40,9 @@ namespace :report do
     end
     Reports::BrexitSubscribersReport.call(args[:date])
   end
+
+  desc "Outputs a CSV of information for each subscriber list within a year for a past date, format: 'yyyy-mm-dd'"
+  task :csv_subscriber_lists, [:date] => :environment do |_t, args|
+    Reports::SubscriberListsReport.new(args[:date]).call
+  end
 end
--- a/spec/lib/reports/subscriber_lists_report_spec.rb
+++ b/spec/lib/reports/subscriber_lists_report_spec.rb
@@ -1,0 +1,31 @@
+RSpec.describe Reports::SubscriberListsReport do
+  before do
+    created_at = Time.zone.parse("2020-06-15").midday
+    list = create(:subscriber_list, created_at: created_at, title: "list 1", slug: "list-1")
+
+    create(:subscription, :immediately, subscriber_list: list, created_at: created_at)
+    create(:subscription, :daily, subscriber_list: list, created_at: created_at)
+    create(:subscription, :weekly, subscriber_list: list, created_at: created_at)
+    create(:subscription, :ended, ended_at: created_at, subscriber_list: list, created_at: created_at)
+
+    create(:matched_content_change, subscriber_list: list, created_at: created_at)
+    create(:matched_message, subscriber_list: list, created_at: created_at)
+  end
+
+  it "returns data around active lists for the given date" do
+    csv = <<~CSV
+      #{Reports::SubscriberListsReport::CSV_HEADERS.join(',')}
+      list 1,list-1,"{""document_type"":"""",""tags"":{""topics"":{""any"":[""motoring/road_rage""]}},""links"":{},""email_document_supertype"":"""",""government_document_supertype"":""""}",2020-06-15 12:00:00 +0100,1,1,1,1,1,1
+    CSV
+
+    expect { described_class.new("2020-06-15").call }.to output(csv).to_stdout
+  end
+
+  it "returns empty csv if there are no active subscriber lists for the given date" do
+    empty_csv = <<~CSV
+      #{Reports::SubscriberListsReport::CSV_HEADERS.join(',')}
+    CSV
+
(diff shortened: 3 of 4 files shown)


lib/reports/subscriber_lists_report.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  CSV_HEADERS = %i[title
+                   slug
+                   matching_criteria
+                   created_at
+                   individual_subscribers
+                   daily_subscribers
+                   weekly_subscribers
+                   individual_updates_for_date
+                   unsubscribed].freeze
```
benthorner, line 12: This isn't necessary: it can be calculated from running the report over multiple days.

lib/reports/subscriber_lists_report.rb
```diff
@@ -0,0 +1,70 @@
...
+  attr_reader :start_of_day, :end_of_day
+
+  CSV_HEADERS = %i[title
+                   slug
+                   matching_criteria
+                   created_at
+                   individual_subscribers
+                   daily_subscribers
+                   weekly_subscribers
+                   individual_updates_for_date
```
benthorner, line 11: This could be confusing: it could refer to the number of emails sent (depends on subscription changes through the day), or the number of content changes and messages. Can we split it into (matched) content changes and messages?

lib/reports/subscriber_lists_report.rb
```diff
@@ -0,0 +1,70 @@
...
+                   unsubscribed].freeze
+
+  def initialize(date)
+    @start_of_day = Time.zone.parse(date)
+    @end_of_day = Time.zone.parse(date).end_of_day
```
benthorner, line 16: We should validate this is in the past.

lib/reports/subscriber_lists_report.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def call
+    non_empty_lists = SubscriberList.where(
```
benthorner, line 20: I don't think we should filter out empty lists (see card comment on 15th October). There's value in knowing a list exists, with no subscribers, at a point in time.

lib/reports/subscriber_lists_report.rb
```diff
@@ -0,0 +1,70 @@
...
+      Subscription.active_on(end_of_day)
+                  .where("subscriber_lists.id = subscriptions.subscriber_list_id")
+                  .arel.exists,
+    )
+
+    CSV.instance($stdout, headers: CSV_HEADERS, write_headers: true) do |csv|
```
benthorner, line 26: Minor: I think it's more conventional to use `.generate`.

lib/reports/subscriber_lists_report.rb
```diff
@@ -0,0 +1,70 @@
...
+     list.created_at,
+     scope.immediately.count,
+     scope.daily.count,
+     scope.weekly.count,
+     individual_updates(list),
+     unsubscribed_count]
+  end
+
+  def criteria(list)
+    list.as_json(root: false,
```
benthorner, line 51: Nice!

lib/reports/subscriber_lists_report.rb
benthorner: Also, given our new retention policy, we should error if it's over a year ago.

spec/lib/reports/subscriber_lists_report_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe Reports::SubscriberListsReport do
+  let(:created_at) { Time.zone.parse("2020-06-15").midday }
+  let(:list) { create(:subscriber_list, created_at: created_at, title: "list 1", slug: "list-1") }
```
benthorner, line 3: Minor: looks like these aren't needed outside of the `before` block.

lib/tasks/report.rake
```diff
@@ -40,4 +40,17 @@ namespace :report do
     end
     Reports::BrexitSubscribersReport.call(args[:date])
   end
+
+  desc "Outputs a CSV of information for each subscriber list for a past date, format: 'yyyy-mm-dd'"
+  task :csv_subscriber_lists, [:date] => :environment do |_t, args|
+    date = Time.zone.parse(args[:date])
+
+    if date.blank?
+      puts 'Invalid date. Please use "yyyy-mm-dd"'
```
benthorner, line 49: While I agree with the principle of `puts`ing stuff at the task level, I think it would be simpler for the report to do the validation - it parses the time anyway - and throw an error instead of `puts`ing - the task description says it all!

benthorner (commented):
Nice one getting this ready so quickly 🏃. We should share a sample of the output with Andrew to make sure it can meet his needs, while it's easy to make adjustments.

I like the solution you've got for the "criteria bits". There are some more major comments about the data, and others about the validation. While _where_ we do validation is kind of minor, I think it's important to try and keep our rake tasks on the slim side.

lib/reports/subscriber_lists_report.rb
1pretz1: True, you could but that would only show you the difference between the two dates after you totalled up the other subscriber types (bit of a chore).

However, mainly I'm seeing the value in a scenario where a list had lots of subscribing and unsubscribing then the amount of subscriptions could stay roughly the same even though theres lots of events. You would see this if you had an unsubscribe count but wouldn't without it.

lib/reports/subscriber_lists_report.rb
1pretz1: Sure thing 👍

lib/reports/subscriber_lists_report.rb
1pretz1: Good point!

lib/reports/subscriber_lists_report.rb
1pretz1: Okay dokay, I've removed this filter so now we'll return all lists (dead or alive) that were created on/before the date given.

lib/reports/subscriber_lists_report.rb
1pretz1: `.generate` doesn't actually output to stdout (it would using puts) but it turns out using `CSV do ...` is suffice.

spec/lib/reports/subscriber_lists_report_spec.rb
1pretz1: Good spot, this was left over from some refactoring.

lib/tasks/report.rake
1pretz1: Sure! Have converted these to raises and shifted them inside the report class.

lib/reports/subscriber_lists_report.rb
benthorner: That's a good point. I was going to suggest this is less preferable, as you can't see what _kind_ of subscriptions were ended, but that applies equally based on what you said. 

Let's keep it then: hopefully we won't need to add more granular fields!

lib/reports/subscriber_lists_report.rb
```diff
@@ -0,0 +1,69 @@
...
+      SubscriberList.where("created_at < ?", date.end_of_day).find_each do |list|
+        csv << export_list_row(list)
+      end
+    end
+  end
+
+private
+
+  def export_list_row(list)
+    unsubscribed_count = list.subscriptions.where("ended_at <= ?", date.end_of_day).count
```
benthorner, line 40: If we're keeping this, we should clarify that it refers to "unsubscriptions" (over all time). Based on [our discussion](https://github.com/alphagov/email-alert-api/pull/1456#discussion_r513576736), I actually think it would be more useful to say "unsubscribes_for_date" or similar.

benthorner (approved):
Thanks for responding to all my comments. I've left one more minor comments about the behaviour of the unsubscribes column, which I think we could be clearer about.

I realise I could have explained my `.generate` suggestion better: it actually simplifies testing if the class return a string, which the rake task then `puts` ([example](https://github.com/alphagov/email-alert-api/blob/cbf50c6cf93959897e3df1db20962d995d9bceb5/lib/tasks/report.rake#L15)).

lib/reports/subscriber_lists_report.rb
1pretz1: I think it would be worse to add 'unsubscribes_for_date' as the "for_date" suffix is used for matched content changes / messages to mean **only** on a specific date.  

You could do something like "for_all_time" but this isn't strictly true either due to our data retention policy.

I think just 'unsubscribed' is okay as it follows the same pattern as individual/daily/weekly subscribers by not having any time related suffix. I could update them all so we are consistent if you can think of a viable option?

benthorner: > is used for matched content changes / messages to mean only on a specific date.

Indeed. I was being a bit implicit there: my suggestion was to change the behaviour to be like that, which I imagine is more useful in practice: all time unsubscribes is a pretty useless metric by itself.

Alternatively, the wording could change to make it clearer this is a counter: using the noun ("unsubscriptions") instead of an unqualified adjective ("unsubscribed" ... when?).

Up to you - I think the first option is more useful.

1pretz1: > Indeed. I was being a bit implicit there: my suggestion was to change the behaviour to be like that, which I imagine is more useful in practice: all time unsubscribes is a pretty useless metric by itself.

I think it's more useful as 'all time' as you could extract more information from it. Example being if you wanted the unsubscribers for a day then you could just run the task twice over two days and you'd see the difference. The benefit of counting 'all' unsubscriptions is you could see the amount between a larger range whereas you can't if you only have a "for_date" method unless you wanted to run it for every single day. I suppose it depends how you use the report to which is more useful.

> Alternatively, the wording could change to make it clearer this is a counter: using the noun ("unsubscriptions") instead of an unqualified adjective ("unsubscribed" ... when?).

Good point, I've updated the name to your suggestion.

> Up to you - I think the first option is more useful.

Appears we agree to disagree 😄 Doesn't happen often! If daily unsubscribes turns out to be wanted then I have no qualms updating the code.

1pretz1:
Ah interesting (regarding the ease of testing). I'll merge this for now as I'll be coming back to the tests anyways with the reduction in subscriber lists AC.

lib/reports/subscriber_lists_report.rb
benthorner: > Example being if you wanted the unsubscribers for a day then you could just run the task twice over two days and you'd see the difference

This is amusing: option 1 does exactly that, but with one step 😆.

> I suppose it depends how you use the report to which is more useful.

Yes, that's true. I'm struggling to come to terms with "total number of unsubscriptions over all time", just because it's not a quantity we've ever referred to before. However, we can also easily change it at the moment.

PR #1458, opened 2020-10-30
Update historic worker to delete empty subscriber lists sooner

1pretz1:
Previously empty subscriber lists were deleted after a year of creation.

This alters the behaviour so they are deleted after 7 days since this is
the [limit] for a user confirming their subscription.

Trello:

[limit]: https://github.com/alphagov/email-alert-api/blob/f0c61539f7244bdb52db54d277a46412b728d642/app/services/auth_token_generator_service.rb#L7


app/workers/historical_data_deletion_worker.rb
```diff
@@ -35,7 +35,7 @@ def delete_and_log(model)
 
   def empty_subscriber_lists
     SubscriberList.left_outer_joins(:subscriptions)
-                  .where("subscriber_lists.created_at < ?", retention_period)
+                  .where("subscriber_lists.created_at < ?", 7.days.ago)
```
benthorner, line 38: Could we clarify what the "retention_period" now refers to? I'd suggest moving it and this time period to a couple of constants:

- `MAX_RETENTION_PERIOD`
- `LIST_RETENTION_PERIOD`

benthorner (commented):
Thanks @1pretz1 🚀. Would be good to record to the number of lists this purges, as it should be quite substantial! I've left one comment where I think we should clarify something.

app/workers/historical_data_deletion_worker.rb
1pretz1: Good point on the naming clarification!

I've used methods instead of constants due to the fact constants don't change once declared. This is an issue if using dynamic logic like whats being used.

Also changed your suggestion for the list retention to 'empty_list_retention_period` as that is more reflective of what it is.

1pretz1:
Ta for the review @benthorner.

I've added the amount of lists to be deleted to the relevant commit which is 8433 empty lists to be deleted!

app/workers/historical_data_deletion_worker.rb
benthorner: Oops, good point 👍. And thanks for fixing up.

benthorner (approved):
🇫🇷

PR #1459, opened 2020-10-30
Create data retention adr

1pretz1:
ADR documenting our current approach to data retention.

Trello:


1pretz1:
ADR should be ready for review @benthorner when you have a mo.

benthorner (commented):
Hurray, your first ADR is on its way 👍. They can take some getting used to, and the main thing is that I think you've captured almost everything we need to say in here. Good start!

My comments range for very minor suggestions to moving or rewording a paragraph or two. I've avoided commenting on subscribers, as the behaviour for this is currently [in flux](https://github.com/alphagov/email-alert-api/pull/1462). One of the themes in the comments is that we can be more concise in a few places. The other is where some of the paragraphs go in the document.

One of the tricky things with an ADR is how to organise the content. If possible, we should try to push content into the "context" section, if it's not a necessary part of the decision or the consequence sections. The "decision" section is quite long, but the actual decisions we make are few and concise to describe:

- We decided on a max retention of 1 year for everything that's unused by the system.
- We decided on a shorter retention of 7 days for virgin subscriber lists (essentially unused).
- We decided to modify the deactivate/nullify process, but the retention is unchanged.
- We decided not to modify the retention for emails (and subscription contents)

So much of the "decision" section could move to "consequences". Thinking about "consequences" as "implementation" (and other, non-implementation, consequences!) has helped me to distinguish the sections in the past.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+A general maximum retention period was suggested of 1 year which seems a
+reasonable compromise between collecting analytics / maintaing a healthy
+database. This retention period was implemented differently for each model
+depending on the model. E.g for content changes we compare against `created_at`
+but for subscriptions we compare against `ended_at`.
+
+The benefit of removing unused data which is deemed as historic is that we can
+keep the db tidy and relevant. Perpetually keeping this data results in a
+detrimental effect on the db performance and complicates our auditing and
+analysis.
```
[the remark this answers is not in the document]
1pretz1: Sure, have expanded on the _why_ .

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+On the first run of deleting historical data we removed over 30 million
```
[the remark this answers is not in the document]
1pretz1: Moved to the consequences section 👍

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+To deem data as historic we had to be sure that firstly, the system no longer
+uses the data and secondly, we don't need the data for future analytics.
+
+A general maximum retention period was suggested of 1 year which seems a
+reasonable compromise between collecting analytics / maintaing a healthy
+database. This retention period was implemented differently for each model
+depending on the model. E.g for content changes we compare against `created_at`
+but for subscriptions we compare against `ended_at`.
+
+The benefit of removing unused data which is deemed as historic is that we can
```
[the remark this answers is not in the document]
1pretz1: Moved to the context section and explained the primary motivation.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+[midday]: https://github.com/alphagov/email-alert-api/blob/master/config/sidekiq.yml#L23-L25
+[worker]: https://github.com/alphagov/email-alert-api/blob/a62abc85453b723d683c2dc13f3bf0065fb86d5f/app/workers/historical_data_deletion_worker.rb
+
+### ContentChange & Message
+
+We decided to remove content changes / messages which have a `created_at` value
+of over a year old.
+
+This was decided as we have no use for them existing past the point of their
```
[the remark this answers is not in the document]
1pretz1: Yep, correct, have updated to use 'system'.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+### SubscriberList
+
+Subscriber lists are deleted when they're over 7 days old and have no
+subscriptions at all (e.g. when a user quits signup journey midway through), or when
+they have no active subscriptions with their most recent subscription being
+over a year old.
+
+Subscriber lists which have no subscriptions at all have very little value for
+our analytics and actually can have a determental impact when we analyze the
```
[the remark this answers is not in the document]
1pretz1: I've made reference to this in the 'context' section of the ADR so have removed this paragraph.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+After the digest run subscriber email has been sent we no longer need these
+records.
+
+### Subscription
+
+Subscriptions are deleted when they've been ended for over a year, based on
+their `ended_at` value. Once a subscription is ended it can't currently be made
+active again and so can be deleted.
+
+### SubscriberList
```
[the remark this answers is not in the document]
1pretz1: Sure thing. I've split them into two distinct policies within the `SubscriberList` section which should make it easier to read.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
+# Decision Record: Data Retention
+
+---
+
+## Context
+
+In November 2020 we wanted to document our changes to data retention.  The
```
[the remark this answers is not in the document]
1pretz1: I've removed the reference to the current date but left the date for the original ADR as I think it adds some nice context on when the last iteration of data retention was.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+[deleted]: ...
+[nullified]: ...
+[initial policy]: https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-003-initial-data-retention.md
+
+---
+
+## Decision
+
+The decision we came to was to implement a [worker] which ran periodically
+every day at [midday] to remove data we classify as historic.
```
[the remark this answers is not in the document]
1pretz1: Sure, added an explanation.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,161 @@
...
+
+Accepted
+
+---
+
+## Consequences
+
+### Only retaining historic data for a year or under
+
+We don't believe we need the data we're cleaning up but we may have to tweak
```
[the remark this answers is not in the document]
1pretz1: Okay dokay, I've removed that paragraph.

1pretz1:
Thanks for the review and for giving a thorough explanation of how to move this ADR forward @benthorner.

I think I've addressed most of your comments and tried to actively thin down the decision section. In doing this I've moved the implementation details of each model to the consequences section (as per your comment of `Thinking about "consequences" as "implementation"`) as that section is purely around implementation.

The Subscriber section has also been updated to the current approach in [this PR](https://github.com/alphagov/email-alert-api/pull/1462/commits/6bb41fa37bd96fb177a0ae42c27c74090b0b4a10) as I think we've consensus about this but can easily be updated if needed.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,186 @@
...
+suffice.
+
+To deem data as historic we had to be ensure that the system no longer
+uses the data, and we don't need the data for future analytics.
+
+We decided:
+
+* On a max retention of 1 year for everything that's unused by the system.
+
+* On a shorter retention of 7 days for subscriber lists with no subscriptions,
```
benthorner, line 57: We should explain why here. (We want to get rid of unused lists ASAP to avoid them polluting our analytics.)

benthorner (commented):
No worries @1pretz1. 

We're definitely making progress: the structure is now much clearer to me, overall and in each section 👍 . Although I've left a few comments, they're mostly about the same thing: explanation. 

Whenever we make a claim, like "performance would get worse", or a decision, like "data is useless after 1 year", we should try to explain or justify it; if we can't, then we need to revise our claim.

Most of my comments are just pointers towards being tighter with the argument we're trying to make, in the "decision" section the "context". Once we have a strong argument, I think this will be good to go 🚀.

docs/adr/adr-005-data-retention.md
1pretz1: Have added an explanation, pointing back to the context where I've given an example.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,186 @@
...
+  active or ended.
+
+* To [modify the nullify process] and retire the [deactivated] concept while
```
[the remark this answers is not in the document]
1pretz1: Certainly.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,186 @@
...
+Accepted
+
+---
+
+## Consequences
+
+On the first run of deleting historical data we removed over 30 million
+records. The biggest bloat of records belongs to the digest run subscribers
+which produces ~400,000 records per day, followed by matched content changes
+~10,000 and ended subscriptions ~4000, as of 05/11/2020.
```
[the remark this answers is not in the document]
1pretz1: I've adjusted the structure / these numbers to make a bit more sense which should answer your question.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,186 @@
...
+subscriptions we compare against `ended_at`. Below lists each model in
+EmailAlertApi and explains how we determined whether the associated records
+should be classified as historic.
+
+### ContentChange & Message
+
+We decided to remove content changes / messages which have a `created_at` value
+of over a year old.
+
+This was decided as the system no longer has use for them existing past the
```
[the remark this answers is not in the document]
1pretz1: Yep, agreed.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,186 @@
...
+
+* On a shorter retention of 7 days for subscriber lists with no subscriptions,
+  active or ended.
+
+* To [modify the nullify process] and retire the [deactivated] concept while
+  keeping the retention unchanged.
+
+* Not to modify the retention for emails (and subscription contents).
+
+A general maximum retention period was suggested of 1 year which seems a
```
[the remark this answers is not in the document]
1pretz1: Sure, I've removed that sentence and been more specific around what we need for analytics!

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,186 @@
...
+we should keep and what we should remove from the system completely.
+
+Our primary motivation is to increase the amount of digest subscriptions we
+have versus individual subscriptions. This prompted a general review of data
+retention as we knew that the system already produced a sizeable amount of
+digest run subscribers (400k a day) which don't get cleaned up and just reside
+in the db.
+
+The benefit of removing unused data which is deemed as historic is that we can
+keep the db tidy and relevant. Perpetually keeping this data results in a
```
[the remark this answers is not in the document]
1pretz1: I've gone with your suggestion although, personally, I don't think you need evidence to state that an ever growing unmanaged database will suffer from reduced performance, eventually.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,186 @@
...
+We originally retained all data produced by EmailAlertApi (apart from emails
+which are [archived] to S3 & [deleted] after a week). This included subscribers
+which have been [nullified], ended subsciptions and subscriber lists with no
+subscriptions to name a few.
+
+These records were originally kept forever as we didn't know what data was
+useful. Now, nearly 3 years on, we have a pretty good understanding of what data
+we should keep and what we should remove from the system completely.
+
+Our primary motivation is to increase the amount of digest subscriptions we
```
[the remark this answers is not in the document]
1pretz1: Haha oh no, we deffo don't want to do that 😄 Corrected!

1pretz1:
Sorry if this PR has been frustrating to review @benthorner due to its continually changing structure. 

I've taken on board your latest comments and updated accordingly. As well as the english comments which have been fixed. I've focused on:

> A lot of the consequences still seem to contain explanation for this decision, but that should reside here. As a way of DRYing it up, we could mention there are essentially two categories of data:
> Event data, which are only used for a short period after they're created, and subsequently for analytics.
> Subscriber data, which is used indefinitely until it's marked as "ended".

This really helped structure the ADR. I've moved parts around and I think I'm happy with the current state of it but feel free to comment if there are still holes (probs are)! 😆

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+
+In doing this we've also [modified the nullify process] which now only
+nullifies subscriber's addresses if they have no active subscriptions and their
+most recent is over 28 days old. We also nullify subscribers which have no
+subscriptions and are older than 28 days which prevents us from storing user's
+personal data longer than needed, if for whatever reason, a subscriber ends up
+with no subscriptions. See examples of where this happened [here][sub bug] and
+[here][sub bug two].
+
+### Not to modify the retention for emails (and subscription contents)
```
benthorner: Minor: we could move this section earlier, to help a future reader understand why we don't mention these models in previous sections.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+* SubscriberList - removed when they have no active subscriptions and their
+  latest, ended subscription is over a year old.
+
+* Subscriber - removed once their `created_at` is over a year old and they
+  don't have any subscriptions (active or ended).
+
+### Shorter retention of 7 days for subscriber lists with no subscriptions
+
+We also remove subscriber lists which have no subscriptions (active or ended)
+and are over 7 days old. As mentioned in the context section, dead subscriber
```
benthorner:
```suggestion
and are over 7 days old. As mentioned in the context section, unused subscriber
```

We use the term "dead" elsewhere to refer to lists that won't ever have matched changes against them.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+  their `ended_at` value. Once a subscription is ended it [can't currently be
+  made active again] and so can be deleted.
+
+* SubscriberList - removed when they have no active subscriptions and their
+  latest, ended subscription is over a year old.
+
+* Subscriber - removed once their `created_at` is over a year old and they
+  don't have any subscriptions (active or ended).
+
+### Shorter retention of 7 days for subscriber lists with no subscriptions
```
benthorner:
```suggestion
### Shorter retention of 7 days for subscriber lists that have never been used
```

Minor: we should avoid ambiguity between inactive subscriptions (we'd also say "no subscriptions" in this case) and having none at all.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+A scenario of when this can happen is when a user quits their signup journey
+midway through. 7 days was the chosen retention period to allow users to complete
+their signup journey as they have a [7 day limit] to confirm their
+subscription.
+
+### Modified the nullify process and retired the deactivated concept
+
+The [deactivated] concept was a point of confusion in the email system
+and wasn't actually needed so has been removed. It revolved around the
+`deactivated_at` field on the subscriber model. This field was mainly used as a
```
benthorner:
```suggestion
`deactivated_at` field on the subscriber model. This field was only used as a
```

Minor: to avoid sounding like we're not sure if we broke something 😄.

benthorner (commented):
That's OK @1pretz1. I'm not actually frustrated 😄. This is your first ADR and I hope you're learning something with each round of review - for a normal PR this would be quite excessive. I'm trying to keep my comments as objective as possible, so we're focussed on making this a solid argument about the overall change.

A general suggestion you may find helpful when writing these things is to consider, for each paragraph: what is the point I'm trying to make, and how does this compare to the previous paragraph?

Another theme is the structure of the "consequences" section. It's not clear what the individual (groups of) consequences are, or whether they are consequences at all. Using headings, like you've done in the "decision" section, would help a lot there, since the heading should say what the consequence is e.g. "Delete lots of data for various models". This may help clarify whether some things are really (part of) a decision, and not a consequence.

Here's another attempt to explain those sections:

- "decisions" are the main or "in-scope" things we decided to do, and why
- "consequences" are the effect it had (and why if it was unexpected), or could/will have

Finally, one potential consequence we're missing is that we restrict our data for analytics to a period of at most 1 year. We mentioned earlier we have no evidence of need for this, so this consequence can just acknowledge the effect, and reiterate that we don't expect it to be a problem. We should also update [this doc](https://github.com/alphagov/email-alert-api/blob/master/docs/analytics.md#how-data-is-organised), by the way.

As always, let me know if you'd like to chat about any of this 💬.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+useful. Now, nearly 3 years on, we have a pretty good understanding of what data
+we should keep and what we should remove from the system completely.
+
+Recent work to [migrate subscribers to daily digest emails] has lead to an
+increase in the amount of digest subscriptions we have versus individual
+subscriptions. This prompted a general review of data retention as we knew that
+the system already produced a sizeable amount of `DigestRunSubscriber` records
+(~430,000 a day) which don't get cleaned up and just reside in the db.
+
+The benefit of removing unused data which is deemed as historic is that we can
```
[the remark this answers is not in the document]
1pretz1: I've restructured this section which means it should flow better.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+lists pollute our analytics and should be removed regularly.
+
+A scenario of when this can happen is when a user quits their signup journey
+midway through. 7 days was the chosen retention period to allow users to complete
+their signup journey as they have a [7 day limit] to confirm their
+subscription.
+
+### Modified the nullify process and retired the deactivated concept
+
+The [deactivated] concept was a point of confusion in the email system
```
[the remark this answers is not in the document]
1pretz1: Yup, moved it to a more fitting spot.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+* Subscription - 1,046,576
+
+* SubscriberList - 11470
+
+* Subscriber - 196,376
+
+We have some model specific consequences concerning the removal of historic
+data which are captured below:
+
+* SubscriberList
```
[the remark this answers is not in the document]
1pretz1: Yep!

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+
+It is possible for subscriber lists to gain active subscriptions after being
```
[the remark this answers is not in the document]
1pretz1: Sure - I've acknowledged the decision in the.... decision section! 😄

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+deemed as historic and deleted, this seems unlikely if they haven't had any
+active subscriptions in over a year, or any subscriptions at all for more than
+7 days. EmailAlertApi would create a new subscriber list if this was to happen.
+
+* Subscriber
```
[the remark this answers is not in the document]
1pretz1: I was thinking one of the consequences is that we can't delete subscribers we should be able to delete due to the app potentially changing going forward.

I suppose this is part of the decision though and not a consequence, arghh! 😄

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+We also remove subscriber lists which have no subscriptions (active or ended)
+and are over 7 days old. As mentioned in the context section, dead subscriber
+lists pollute our analytics and should be removed regularly.
+
+A scenario of when this can happen is when a user quits their signup journey
+midway through. 7 days was the chosen retention period to allow users to complete
+their signup journey as they have a [7 day limit] to confirm their
+subscription.
+
+### Modified the nullify process and retired the deactivated concept
```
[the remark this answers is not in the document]
1pretz1: I've added some context on how this related to the data retention work and moved it the the consequence section as you're right this does feel more like a consequence than part of the active decision.

docs/adr/adr-005-data-retention.md
```diff
@@ -0,0 +1,201 @@
...
+We originally retained all data produced by EmailAlertApi (apart from emails
+which are [archived] to S3 & [deleted] after a week). This included subscribers
+which have been [nullified], ended subsciptions and subscriber lists with no
+subscriptions to name a few.
+
+These records were originally kept forever as we didn't know what data was
+useful. Now, nearly 3 years on, we have a pretty good understanding of what data
+we should keep and what we should remove from the system completely.
+
+Recent work to [migrate subscribers to daily digest emails] has lead to an
```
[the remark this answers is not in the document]
1pretz1: I've rejigged this section which should flow better.

docs/adr/adr-005-data-retention.md
1pretz1: Sure, this is now the first section.

docs/adr/adr-005-data-retention.md
1pretz1: 😆

1pretz1:
Cheers for the review @benthorner!

> A general suggestion you may find helpful when writing these things is to consider, for each paragraph: what is the point I'm trying to make, and how does this compare to the previous paragraph?

Thanks for pointing that out as I wasn't really paying attention to the general flow and was more concerned about getting all the info down in some form.

> Using headings, like you've done in the "decision" section, would help a lot there

I've added headings to the restructured consequence section which improves the structure! 👍 

> Finally, one potential consequence we're missing is that we restrict our data for analytics to a period of at most 1 year. 
   We mentioned earlier we have no evidence of need for this, so this consequence can just acknowledge the effect, and 
   reiterate that we don't expect it to be a problem.

Added a brief summary to the consequence section.

> We should also update this doc, by the way.

Yup it's on my radar, I was waiting until we had this doc sorted before linking it to the [analytics](https://github.com/alphagov/email-alert-api/blob/master/docs/analytics.md#how-data-is-organised) doc.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+We remove subscribers once their `created_at` is over a year old and they don't
+have any subscriptions (active or ended).
+
+Subscribers should only be without subscriptions (active or ended) when their
+historic subscriptions have been deleted via the historic deletion worker. This
+means we should be able to delete all subscribers (under a year old too) that
+don't have any subscriptions (active or ended).
+
+However, we should be cautious about doing this so we don't accidently remove
+new subscribers when the application changes. Some examples of where it was
```
benthorner, line 123:
```suggestion
new subscribers that lack subscriptions, due to flaws in our code. Some examples of where it was
```

We should be more specific about what we're worried about here.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Data which has never been used
+
+We remove subscriber lists which have never had any subscriptions and are over
+seven days old. A scenario of when this can happen is when a user quits their
+signup journey midway through. Seven days was the chosen retention period to
+allow users to complete their signup journey as they have a [seven day limit]
+to confirm their subscription.
+
+As mentioned in the context section, subscriber lists pollute our analytics and
```
benthorner, line 135:
```suggestion
As mentioned in the context section, unused subscriber lists pollute our analytics and
```

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+
+* DigestRunSubscriber - 30,858,770
+
+* Subscription - 1,046,576
+
+* SubscriberList - 11,470
+
+* Subscriber - 196,376
+
+### Presence of emails during historic deletion
```
benthorner, line 177: The presence of emails isn't something new. I think this is really an implementation detail, and a pretty minor one at that. We could combine it with the section about the worker, and maybe comment on how this may have once been a problem, but [ADR 006](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-006-email-delivery-responsibilities.md) means it's basically impossible now.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+
+If an old email exists and belongs to a historic subscriber then the historic
+worker will throw an error when the subscriber deletion is attempted. This is
+due to a db restriction as these emails should've been cleaned up already via
+the [email deletion mechanism]. We want to produce an error for this scenario
+rather than just delete them so bugs don't unintentionally creep in. This also
+applies to subscription contents as their deletion is cascaded when deleting
+emails.
+
+### Restricts analytics to 1 year of unused data
```
benthorner, line 187:
```suggestion
### Restrict analytics to 1 year of unused data
```

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+
+We restrict our unused data for analytics to a period of at most 1 year. This
```
benthorner, line 189:
```suggestion
We now restrict our unused data for analytics to a period of at most 1 year. This
```

Minor, but this should help the reader understand what's changing.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+means any queries looking for this data (e.g. ended subscriptions over a year
+old) will be met with no results. We don't expect this to be an issue but is
+something worth mentioning.
```
benthorner, line 192: Why is it worth mentioning? Perhaps we could emphasise why we don't expect it to be an issue.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Modify the nullify process and retire the deactivated concept
+
+When documenting our data retention policies we re-discovered a concept which
```
benthorner, line 196: We should clarify why this section is in the ADR. It's true that we re-discovered the concept, but that alone doesn't merit a section here. As I recall, the problem was that we found it hard to write documentation for the retention policy as described in this ADR, due to a tangential concept of nullifying / deactivating subscribers, which turned out to be hard to understand, and thus, to document.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+didn't make much sense regarding the activation / deactivation of subscribers.
+This concept was a point of confusion in the email system and wasn't actually
+needed so has been [removed].
+
+The concept revolved around the `deactivated_at` field on the subscriber model. This
+field was used as a means of determining when a subscriber could be nullified.
+We can however obtain this information via the most recent ended subscription
+of the subscriber (based on the subscriptions `ended_at`) which means
```
benthorner, line 204:
```suggestion
of the subscriber (based on the subscription's `ended_at`) which means
```

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+lists sooner than a year but that will be elaborated on in a later section.
+
+We also concluded that we don't need to change the retention for emails (and
+subscription contents) as we already clean them up via the [email deletion
+mechanism] which is sufficient.
+
+The following sections document our approach to the decision for different data
+in the system.
+
+### Data which has previously been used
```
benthorner, line 72: I'm finding the formatting in this section quite hard to follow, especially with the hrule, which cuts the whole document in half.

I appreciate there are quite a few levels at this point. Something which might work better would be to use a list, like this:

- **ContentChange, Message & DigestRun**. Remove once their created_at value is over a year old.

...

- **SubscriberList**. We remove lists when they have no active subscriptions and their latest, ended subscription is over a year old. It is possible for a list we deem as historic to be subscribed to again, however this seems unlikely due to the list not having any active subscriptions for over a year. If this was to happen after the list has been deleted then a new SubscriberList record will be generated.

benthorner (commented):
Nice work @1pretz1 - the decision, and the argument for it, is definitely clearer than it was. I've got basically no comments about the "context" section, but we need to tighten the argument in the "decision" section - I've left some comments about that.

The "consequences" section is getting better, but the headings are still quite inconsistent and it's unclear why a coupe of them are really consequences. I've left some comments on the sub-sections that I think we need to make clearer.

Thanks again for chatting about this last week - I realise it's a bit of a learning curve for you, and we could have done more upfront work together to make it easier. But we're nearly there now - remember, the aim is to persuade or "sell" our decision 💰.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Context
+
+The original ADR concerning data retention resides [here][initial policy] and
+outlines our approach from February 2018. At that point in time it was decided
+we should retain all data produced by EmailAlertApi with the exception of
+emails which are [archived] to S3 & [deleted] after a week.
+
+Since then we've come to realise that a revision of our data retention policy
+is needed as much of the data we retaine will never be used by us, or the
```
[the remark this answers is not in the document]
1pretz1: Sneaky 'e' suffix trying to make an appearance 😆 Thanks for spotting!

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+[archived]: ...
+[deleted]: ...
+[nullified]: ...
+[initial policy]: https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-003-data-retention.md
+[migrates subscribers to daily digest emails]: https://github.com/alphagov/email-alert-api/pull/1352
+
+## Decision
+
+We decided that historic (unused) data should be removed periodically from the
+system entirely, instead of retaining it through archiving. Archiving was
```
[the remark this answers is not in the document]
1pretz1: Hmm I didn't say it did mean useless. I also said that the data wasn't used after a point in the next paragraph.

However, after your comment I've reshuffled both paragraphs which should make it clearer that we don't need historic data after a certain point.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,221 @@
...
+deemed as unnecessary because if we've deemed data as being historic then we
+don't see a use for it (analytical or otherwise) going forward and so retaining
+it via an external data store seems a waste of resources and energy in
+maintaining the process.
+
+As part of the decision we concluded that the data we, and the system no longer
```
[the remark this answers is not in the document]
1pretz1: I mentioned that we don't have purpose for this data beyond a year and thought that implicitly implied that we do have a purpose for data under a year old. 

I've been more explicit anyways which should make it clearer. 👍

docs/adr/adr-007-data-retention.md
1pretz1: > I think this is really an implementation detail, and a pretty minor one at that

Hmm true, I've chose not to include it at all given it's deemed as 'minor'.

docs/adr/adr-007-data-retention.md
1pretz1: Yep true we did figure out it was hard to document. I've updated the intro to include this.

docs/adr/adr-007-data-retention.md
1pretz1: Arghglhjsncdp I did this originally. 😆 

I've used a list which does make the section easier to read.

Let me know if it's easier to follow! 👍

1pretz1:
Thanks for the review @benthorner. Your comments have been really helpful! It has been a learning curve but one I'm thankful of doing (and thankful to you for persevering 😆). 

Should be ready for a re-review.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,190 @@
...
+behaviour for subscriber lists where we delete certain lists sooner than a year
+but that will be elaborated on in a later section.
+
+We chose to remove the historic data entirely instead of retaining it through
+archiving as the latter option seems unnecessary. If we've deemed the data as
+historic and don't see a use for it (analytical or otherwise) going forward,
+then retaining it via an external data store seems a waste of resources and
+energy in maintaining the process.
+
+It was also concluded that we don't need to change the retention for emails
```
benthorner, line 64:
```suggestion
Emails (and their associated SubscriptionContents) are the one place where we already manage retention - these are held locally for 7 days, and also archived to Athena indefinitely. We still need this low retention for Emails in order to manage the size of the table. However, the archiving to Athena is inconsistent with the above decision to delete unused data entirely. While we originally thought Athena might serve as a data source for analytics, this has not happened in practice, so the archive data stored there is entirely unused. We therefore decided to remove the archiving mechanism for Emails, since it had no value.
```

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,190 @@
...
+  currently be made active again] and so can be deleted.
+
+* **Previously subscribed to SubscriberList**. Remove lists when they have no
+  active subscriptions and their latest, ended subscription is over a year old.
+  It is possible for a list we deem as historic to be subscribed to again,
+  however this seems unlikely due to the list not having any active
+  subscriptions for over a year.  If this was to happen, after the list has
+  been deleted then a new `SubscriberList` record will be generated.
+
+* **Unused SubscriberList**. Remove lists which have never had any
```
benthorner, line 88: Minor: we could make this a separate paragraph, since it's different to the other items in the list, which are all about retaining data for 1 year.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,190 @@
...
+
+We now restrict our unused data for analytics to a period of at most 1 year.
+This means any queries looking for this data (e.g. ended subscription's over a
+year old) will be met with no results. We don't expect this to be an issue due
+to our current expectation of the data we'll need for future analytics.
+
+### Modify the nullify process and retire the deactivated concept
+
+When documenting our data retention policy as described in this ADR, we found
+it difficult to understand and document the tangential concept of nullifying /
```
benthorner, line 156: We could also link to [the original explanation](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-003-data-retention.md#subscriber-and-subscription) here.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,190 @@
...
+* Nullifying subscribers means to update the `address` of `Subscriber` records
+  which have been unsubscribed for more than 28 days to `nil`.
+
+* Deactivating subscribers is a concept that revolves around the
+  `deactivated_at` field on the subscriber model. This field is populated when
+  a subscriber unsubscribes from their last active subscription. It is used as
+  a means to determine when a subscriber can be nullified.
+
+We can however obtain the information held within `deactivated_at` through the
+`ended_at` field of the most recent ended subscription of the subscriber,
```
benthorner, line 170:
```suggestion
`ended_at` field across all subscriptions of a subscriber,
```

We need to check that all subscriptions are ended, as well as determine the latest end time.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,190 @@
...
+meaning the deactivated concept is superfluous.
+
+Subsequently we didn't want to spend energy documenting a process that didn't
+make much sense to begin with so it was decided that it would be easier (and
+beneficial as a whole) to just [remove] the deactivated concept completely.
+
+In removing the deactivated concept we [modified the nullify process] to use
+the `ended_at` field of the last subscription, instead of `deactivated_at`. We
```
benthorner, line 178:
```suggestion
the `ended_at` field, as described above, instead of `deactivated_at`. We
```
 
As above, we need to be more accurate here.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,190 @@
...
+also updated the process to nullify subscribers which have no subscriptions
+(active or ended) and are older than 28 days. This prevents the system from
+storing user's personal data longer than needed, if for whatever reason, a
+subscriber ends up with no subscriptions. See examples of where this happened
```
benthorner, line 182:
```suggestion
subscriber is created without any subscriptions. See examples of where this happened
```

We should be more precise to distinguish between a subscriber with "no (active) subscriptions" and a subscriber that has never had any subscriptions, which is what we mean here.

docs/adr/adr-007-data-retention.md
benthorner: Saying this is sufficient contradicts the above decision and the context, where we've said that we won't hold unused data indefinitely. Although we could just be more specific and say "sufficient to keep the local database manageable", this is also the best time to clarify the future of archiving in Athena.

Sorry I didn't spot this earlier 😬. We'll need an extra consequence to reflect this. I'd suggest something like the following:

---

### Deprecate archiving Email to Athena

We planned work to remove it in the future. In the meantime, we consider archiving to Athena to be deprecated, and have updated [the documentation around it](https://github.com/alphagov/email-alert-api/blob/1a25d44126709bf8a98955d7609ced439095dd9c/docs/analytics.md) to reflect this status.

benthorner (commented):
No worries @1pretz1! I'm glad we're both getting something out of this process. It's definitely a learning curve for me, as well.

I was hoping to approve this time round, as everything is basically in place. However, on re-reading one last time, I found a couple of major errors that have eluded me before. I know it's frustrating when that happens, and I'm sorry I didn't find these in earlier reviews. I do think we need to address them, though, so we have correct documentation for the future.

Since this is coming late on, I've tried to be as helpful as I can and suggest exact changes, so we don't end up in further discussions on those. You're welcome to diverge from that, though, and I'm open to working on these fixes offline with you, so we can get to a ✅ on this faster.

Otherwise, just a few more minor comments on the changes you made last time.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,185 @@
...
+
+We decided that historic, unused data should be removed periodically from the
+system entirely. It was concluded that the data the system no longer uses
+should only exist for a maximum of one year before removal. We sometimes
+analyse historic data from up to a year ago but we have no evidence of needing
+this data for analytics beyond one year. There is however some bespoke
+behaviour for subscriber lists where we delete certain lists sooner than a year
+but that will be elaborated on in a later section.
+
+We chose to remove the historic data entirely instead of retaining it through
```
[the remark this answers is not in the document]
1pretz1: I've removed the paragraph in favour of the one you suggested in a later comment which sums up our approach to Athena.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,185 @@
...
+archiving as the latter option seems unnecessary. If we've deemed the data as
+historic and don't see a use for it (analytical or otherwise) going forward,
+then retaining it via an external data store seems a waste of resources and
+energy in maintaining the process.
+
+It was also concluded that we don't need to change the retention for emails
+(and subscription contents) as we already clean them up via the [email deletion
+mechanism] which is sufficient.
```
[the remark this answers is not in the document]
1pretz1: This paragraph has been replaced by your suggestion later on so this comment is void.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,185 @@
...
+  active again] and so can be deleted.
+
+* **Previously subscribed to SubscriberList**. Remove lists when they have no
+  active subscriptions and their latest, ended subscription is over a year old.
+  It is possible for a list we deem as historic to be subscribed to again,
+  however this seems unlikely due to the list not having any active
+  subscriptions for over a year.  If this was to happen, after the list has
+  been deleted then a new `SubscriberList` record will be generated.
+
+* **Unused SubscriberList**. Remove subscriber lists which have never had
```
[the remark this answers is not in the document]
1pretz1: Hm we could, however, I think it reads nicer in a list as the list isn't about data with strictly one year retention.

> The following list documents our approach to the decision for different data
    in the system.

docs/adr/adr-007-data-retention.md
1pretz1: > Saying this is sufficient contradicts the above decision and the context, where we've said that we won't hold unused data indefinitely.

This was referring to the email deletion mechanism, rather than the archiving aspect. Although they're interlinked so I can see this could be seen as contradictory.

I've updated to your recommendation and added the consequence.

docs/adr/adr-007-data-retention.md
1pretz1: Think this is a duplicate of a previous comment, I've responded to that one so resolving this one! 🤔

docs/adr/adr-007-data-retention.md
1pretz1: I thought this was implied due to the nullify explanation referencing 'unsubscribed' subscribers. I've updated that explanation and this sentence to make it clearer.

1pretz1:
Thanks for the review @benthorner and for catching some discrepancies.

I've updated the ADR based on your current round of comments. 👍

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Decision
+
+We decided that historic, unused data should be removed periodically from the
+system entirely. It was concluded that the data the system no longer uses
+should only exist for a maximum of 1 year before removal. We sometimes
+analyse historic data from up to a year ago but we have no evidence of needing
+this data for analytics beyond a year.
+
+Emails (and their associated SubscriptionContents) are the one place where we
```
benthorner, line 56: Minor: this paragraph could go after the following list, which relates to the paragraph above.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
+# Decision Record: Data Retention
+
+## Context
+
+The original ADR concerning data retention resides [here][initial policy] and
+outlines our approach from February 2018. At that point in time it was decided
+we should retain all data produced by EmailAlertApi with the exception of
```
benthorner, line 7:
```suggestion
we should retain all data in EmailAlertApi with the exception of
```

Minor: we still retained the data, but it was in a different place.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+
+* **MatchedContentChange, MatchedMessage & DigestRunSubscriber**. Cascaded when
+  their parent content change / message / digest run is deleted.
+
+* **Subscription**. Remove subscriptions when they've been ended for over a
+  year, based on their `ended_at` value. Once a subscription is ended it [can't
+  currently be made active again] and so can be deleted.
+
+* **Previously subscribed to SubscriberList**. Remove lists when they have no
+  active subscriptions and their latest, ended subscription is over a year old.
```
benthorner, line 80:
```suggestion
  subscriptions (after removing historic ones) and they are over a year old.
```

Minor: this is different to [the actual implementation](https://github.com/alphagov/email-alert-api/blob/master/app/workers/historical_data_deletion_worker.rb#L56).

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+`ended_at`, as described above, instead of `deactivated_at`. We also updated
+the process to nullify subscribers which have no subscriptions (active or
+ended) and are older than 28 days. This prevents the system from storing user's
+personal data longer than needed, if for whatever reason, a subscriber is
+created without any subscriptions. See examples of where this happened
+[here][sub bug] and [here][sub bug two].
+
+### Deprecate archiving Email to Athena
+
+We planned work to remove it in the future. In the meantime, we consider
```
benthorner, line 185:
```suggestion
We have planned work to remove it in the future. In the meantime, we consider
```

benthorner (approved):
And... ✅.

One of my comments is a little major, but I've missed it so many times previously it seems unfair to point it out now. Let's amend before this goes in though!

We should also update [the card](https://trello.com/c/tfu9zXuy/534-delete-unused-data-after-a-year) to make the final consequence true:

- Deprecate Athena in analytics.md
- Write a follow-up card to remove Athena / archiving

Otherwise, I think @kevindew should have a read of this before it goes in. Hopefully there isn't anything major left to change in here. Thanks again for sticking with it ⭐ 🏅.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
+# Decision Record: Data Retention
+
```
kevindew, line 2: Looks like this is missing Date from the template: https://github.com/alphagov/email-alert-api/blame/master/docs/adr/adr-005-record-architecture-decisions.md#L3

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
+# Decision Record: Data Retention
```
kevindew, line 1: Looks like this is missing a number at the start, eg `7. Data Retention` as per [the template](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-005-record-architecture-decisions.md) and it's not necessary to state "Decision Record:".

I also notice [my ADR](https://github.com/alphagov/email-alert-api/pull/1476) is also flagged as number 7 🤦  I'll update mine to 8. Gosh it's been a while since I've been a project with concurrent ADR's - it's like being back on Publishing Workflow 😁 .

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Context
+
+The original ADR concerning data retention resides [here][initial policy] and
```
kevindew, line 5: I imagine it's worth adding a note in that ADR to say something like `This decision was revisited in [ADR-7](link) which supersedes this decision.`. I don't think you need to rename that file.

docs/adr/adr-007-data-retention.md
kevindew: EmailAlertApi should be Email Alert API.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+
+Contrary to what was previously thought, storing lots of unused data
+indefinitely actually complicates and pollutes our auditing and analysis. An
+example of this can be seen if we look at subscriber lists. We have over
+~73,000 lists with active subscribers but in the last week alone, we've
+accumulated 1139 lists which don't have any subscriptions at all (ended or
+active). These unused lists prevent us from doing simple auditing tasks
+easily, like finding which lists to report on.
+
+Knowing the above, we thought about how we could cleanup this unused data to
```
kevindew, line 36:
```suggestion
Knowing the above, we thought about how we could clean-up this unused data to
```

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+We therefore decided to remove the archiving mechanism for Emails, since it has
+no value.
+
+The following list documents our approach to the decision for different data
+in the system.
+
+* **ContentChange, Message & DigestRun**. Remove once their `created_at` value is
+  over a year old.
+
+* **MatchedContentChange, MatchedMessage & DigestRunSubscriber**. Cascaded when
```
kevindew, line 72: I'm not sure about the use of Cascaded as a verb - perhaps: Removed automatically

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+  their parent content change / message / digest run is deleted.
```
kevindew, line 73: ContentChange/Message/DigestRun would be consistent with the terminology you've used above.

docs/adr/adr-007-data-retention.md
```diff
@@ -0,0 +1,196 @@
...
+  retention period to allow users to complete their signup journey as they have
+  a [7 day limit] to confirm their subscription. As mentioned in the context
+  section, unused subscriber lists pollute our analytics and so we should
+  remove them regularly.
+
+* **Subscriber**. Remove subscribers once their `created_at` is over a year old
+  and they don't have any subscriptions (active or ended). Subscribers should
+  only be without subscriptions when their historic subscriptions have been
+  deleted. This means we should be able to delete all subscribers (under a year
+  old too) that don't have any subscriptions. However, we should be cautious
```
kevindew, line 98: Minor - this area of caution seems somewhat superfluous to the decision. I'd suggest stopping the paragraph at "subscriptions."

kevindew (commented):
Looks great @1pretz1, nice job. I've just got a few comments.

docs/adr/adr-007-data-retention.md
1pretz1: Haha ADR's everywhere 👀 😆

docs/adr/adr-007-data-retention.md
1pretz1: The implementation you've linked to is concerning historic subscribers, rather then historic subscriber lists so I've the wording as is. 👍

benthorner: Oh sorry. I meant [here](https://github.com/alphagov/email-alert-api/blob/master/app/workers/historical_data_deletion_worker.rb#L46) - it is different.

docs/adr/adr-007-data-retention.md
1pretz1: Okay dokay, I've removed that commit and added a note in the previous ADR.

docs/adr/adr-007-data-retention.md
1pretz1: I've reworded the sentence so it relates more to the decision (as it was part of the decision). Let me know if thats any better?

docs/adr/adr-007-data-retention.md
1pretz1: The ordering of the deletions was factored into the implementation to simply the code.

We decided that we should only remove subscriber lists which have no active subscriptions and the newest subscription is over a year old. As we delete the historic ended subscriptions prior to subscriber lists we don't have to include this extra logic in the code.

I could explain the way it was implemented but that seems to detract from the decision that was made. Does that make sense?

benthorner: Yes, that's fair. It is just a minor point, and I wanted to make sure you were aware of the difference. I agree it's a valid perspective on the policy.

1pretz1:
Thanks for the reviews @kevindew & @benthorner.

Could you have another peek at the ADR @kevindew to check the changes I've made are inline with your comments.

kevindew (approved):
Nice one @1pretz1 - I approve of the renames, that helps these be distinguishable.

PR #1460, opened 2020-11-02
Explain data retention mechanisms

1pretz1:
We have a number of mechanisms regarding data management in
EmailAlertApi but no doc that explains the concepts. This adds one.

Trello:

### DIFF
--- a/docs/data-cleanup-mechanisms.md
+++ b/docs/data-cleanup-mechanisms.md
@@ -1,0 +1,83 @@
+# Data cleanup mechanisms
+
+There are several mechanisms regarding the cleanup of data in Email Alert Api.
+This doc gives an overview of each.
+
+## Nullifying email addresses
+
+Updates the `address` of old unsubscribed `Subscriber` records to `nil`. We do
+this to comply with GDPR, specifically so we're not retaining user's data for
+longer than needed.
+
+We nullify subscribers in two different scenarios.
+
+Firstly, once they've been unsubscribed from all subscriptions for over 28
+days. We could nullify user's email addresses immediately after unsubscribing
+but this would limit our ability to assist in user support queries. E.g.
+restoring a user's subscriptions after they've unsubscribed by accident which
+would be impossible to do if the system had nullified their address
+immediately after unsubscribing.
+
+Secondly, subscribers are nullified which are created more than 28 days ago and
+don't have any subscriptions at all (active or ended). It shouldn't be possible
+for subscribers to exist without subscriptions but these are queried and
+removed anyway to prevent storing user's data indefinitely if, for whatever
+reason, a subscriber ends up with no subscriptions. This has happened before,
+unintentionally, see [here][sub bug] and [here][sub bug two].
+
+The [NullifySubscribersWorker] runs [every hour].
+
+[every hour]: https://github.com/alphagov/email-alert-api/blob/cefdfa76b13915ea96b131490fd3186b6d52cf05/config/sidekiq.yml#L27
+[NullifySubscribersWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/nullify_subscribers_worker.rb
+[sub bug]: https://github.com/alphagov/email-alert-api/pull/1462/commits/053859c5962eef104256661f28727b08a43e3d31
+[sub bug two]: https://github.com/alphagov/email-alert-api/pull/1462/commits/a4b3e82801d79abd3989b1dd60ffd499e7ce82ba
+
+## Archiving and deleting emails
+
+**Note**
+> Archiving emails is deprecated and will be removed in the near future, you
+> can read more about this in [ADR-7]. In the process of removing archiving we
+> will update the email deletion policy so emails are deleted 7 days after
+> creation. We will also combine the [EmailDeletionWorker] into the
+> [HistoricalDataDeletionWorker] as both workers will soon have similar
+> responsibilities.
+
+We archive emails to S3 using the [EmailArchiveWorker]. This is done to support
+long term analytics as we can query the S3 bucket through [Athena]. Emails are
+eligible for archiving immediately after creation as long as they haven't
+already been archived (based on `archived_at` value).
+
+The [EmailDeletionWorker] deletes emails after they've been archived for over
+7 days. These recently archived emails get deleted regularly to keep the
+database performant as we produce a large quantity which the system only uses
+for a short period of time. In deleting these emails we also delete their
+associated subscription contents through a [db cascade]. As of 05/11/2020 we
+generate around ~3 million emails per day.
+
+The [archive and deletion workers] run every hour.
+
+[archive and deletion workers]: https://github.com/alphagov/email-alert-api/blob/ec075c7198b965d3771726946dcd37783b9de9c5/config/sidekiq.yml#L29-L34.
+[Athena]: https://aws.amazon.com/athena
+[EmailArchiveWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/email_archive_worker.rb
+[EmailDeletionWorker]: https://github.com/alphagov/email-alert-api/blob/a62abc85453b723d683c2dc13f3bf0065fb86d5f/app/workers/email_deletion_worker.rb
+[db cascade]: https://github.com/alphagov/email-alert-api/blob/11fb84542e6c7f3995f419e4affaf56aa759ec6c/db/schema.rb#L206
+
+## Deleting historic data
+
+The [HistoricalDataDeletionWorker] deletes data we deem as historic everday at
+[midday]. Midday was chosen as removing historic data is deemed non urgent work
+and so running once a day, in hours, seems sufficient.
+
+In general, we remove historic, unused data after a year. One exception to this
+is that we remove subscriber lists which have been created over seven days ago
+and never had any subscriptions.
+
+We remove this data so we don't pollute our analytics or run into future
+capacity issues.
+
+You can read in more detail about the decision to implement the historic
+deletion of data in [ADR-7].
+
+[midday]: https://github.com/alphagov/email-alert-api/blob/cefdfa76b13915ea96b131490fd3186b6d52cf05/config/sidekiq.yml#L24
+[HistoricalDataDeletionWorker]: https://github.com/alphagov/email-alert-api/blob/a62abc85453b723d683c2dc13f3bf0065fb86d5f/app/workers/historical_data_deletion_worker.rb
+[ADR-7]: https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-007-retain-data-for-up-to-one-year.md#decision


docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
+# Data retention mechanisms
+
+There are several mechanisms regarding the retention of data in EmailAlertApi.
+This doc gives a brief overview of each.
+
+## Deactivating subscribers
+
+We deactivate subscribers for a couple of different scenarios. One being where
```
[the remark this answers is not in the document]
benthorner: We could also describe what "deactivation" actually means, and why we do it: this appears to be so the nullifying worker can do its thing.

docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
...
+unpublished rabbitMQ message which is eventually handled by the
+[UnpublishHandleService]. This is only applicable for subscriptions to the
+topic taxonomy.
+
+[UnpublishHandleService]: https://github.com/alphagov/email-alert-api/blob/master/app/services/unpublish_handler_service.rb
+
+## Nullifying deactivated subscribers
+
+We nullify the email addresses of subscribers after they've been deactivated
+for over 28 days. This means that the address field on the subscriber record is
```
benthorner, line 23:
```suggestion
for over 28 days, so that we're not storing their email address for longer than necessary.
```

I think it's worth briefly explaining why we do this. The second sentence appears to be a repeat of the first, so I've suggested we remove that too.

docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
...
+archived (based on `archived_at` value).
+
+The [archive worker runs every hour].
+
+[EmailArchiveWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/email_archive_worker.rb
+[archive worker runs every hour]: https://github.com/alphagov/email-alert-api/blob/ec075c7198b965d3771726946dcd37783b9de9c5/config/sidekiq.yml#L29-L31
+
+## Deleting emails
+
+We delete emails after they've been archived for a minimum of 7 days. Emails get
```
benthorner, line 45:
```suggestion
We delete emails 7 days after they've been archived. Emails get
```

Minor: the "minimum" wording here sounded like there would be other days in the future where we might archive an email.

docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
...
+archived immediately (as soon as they're generated) and then deleted using the
```
benthorner, line 46: Minor: we could remove this sentence, as we've already talked about archiving in the previous section.

docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
...
+[EmailDeletionWorker].
+
+The [deletion worker runs every hour].
+
+[deletion worker runs every hour]: https://github.com/alphagov/email-alert-api/blob/ec075c7198b965d3771726946dcd37783b9de9c5/config/sidekiq.yml#L32-L34
+[EmailDeletionWorker]: https://github.com/alphagov/email-alert-api/blob/a62abc85453b723d683c2dc13f3bf0065fb86d5f/app/workers/email_deletion_worker.rb
+
+## Deleting historic data
+
+We now delete data we deem as historic everday at [midday] using the
```
benthorner, line 56: Minor: why midday?

docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
...
+[EmailArchiveWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/email_archive_worker.rb
+[archive worker runs every hour]: https://github.com/alphagov/email-alert-api/blob/ec075c7198b965d3771726946dcd37783b9de9c5/config/sidekiq.yml#L29-L31
+
+## Deleting emails
+
+We delete emails after they've been archived for a minimum of 7 days. Emails get
+archived immediately (as soon as they're generated) and then deleted using the
+[EmailDeletionWorker].
+
+The [deletion worker runs every hour].
```
benthorner, line 49: A future reader may be confused about why we have a specific worker for deleting emails, and why it runs hourly. Can we add more detail here?

docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
...
+
+[deletion worker runs every hour]: https://github.com/alphagov/email-alert-api/blob/ec075c7198b965d3771726946dcd37783b9de9c5/config/sidekiq.yml#L32-L34
+[EmailDeletionWorker]: https://github.com/alphagov/email-alert-api/blob/a62abc85453b723d683c2dc13f3bf0065fb86d5f/app/workers/email_deletion_worker.rb
+
+## Deleting historic data
+
+We now delete data we deem as historic everday at [midday] using the
+[HistoricalDataDeletionWorker].
+
+The data we delete is deemed historic if its now redundant and has no use
```
benthorner, line 59: We can be clearer here: we're _asserting_ that data over a year old has no use in the system, or analytical purposes. I think it's also important to mention the motivations: it makes analytics harder; and some of it quickly fills the DB.

docs/data-retention-mechanisms.md
```diff
@@ -0,0 +1,65 @@
...
+
+We use the [NullifyDeactivatedSubscribersWorker] to do this which runs [every
+hour].
+
+[every hour]: https://github.com/alphagov/email-alert-api/blob/master/config/sidekiq.yml#L26-L28
+[NullifyDeactivatedSubscribersWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/nullify_deactivated_subscribers_worker.rb
+
+## Archiving emails
+
+We archive emails to S3 using the [EmailArchiveWorker]. Emails are eligible for
```
benthorner, line 34: We can mention that we do this to support longterm analytics, and link to the [doc with more info about it](https://github.com/alphagov/email-alert-api/blob/master/docs/analytics.md#analytics-through-athena).

benthorner (commented):
Good start @1pretz1 - great idea to document all the weird cleanup jobs we have! It's definitely not obvious 🎖️. I've left a few comments, which are mostly on a single theme: explaining _why_ we do a particular sort of cleanup.


With deactivation and nullification, I'm wondering if it would actually be quicker to just rewrite the whole mess, so that we just have a single worker that nullifies emails 28 days after the subscriber was created, or the last subscription was ended. Summary:

- Rewrite the nullify worker as described.
- Delete the deactivation worker.
- Drop the "deactivated_at" for subscribers (and all associated code).

What do you think? It's possible we won't revisit this part of the system for a while. The whole "deactivation" thing is a low, but continual source of pain, so I think it would be valid for us to cure it, if that takes a similar time to documenting it fully!

1pretz1:

As for the deactivation and nullification worker mess. I agree it's quite confusing and could be improved and so I think it would be a good idea to sort it out as you describe. My only thought was that removing 'deactivated_at` may incur reduced performance (we'd have to do a join where we wanted to know this information) but thinking about it this wouldn't be an issue as we wouldn't need to know this info apart from when we're nullifying so that's fine! 👍 

I'll address your comments after I've resolved the worker mess as it makes sense to do that first.

docs/data-retention-mechanisms.md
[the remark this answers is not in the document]
1pretz1: We don't need this section anymore, yay! 🎉

docs/data-retention-mechanisms.md
1pretz1: Sure 👍

docs/data-retention-mechanisms.md
1pretz1: Why indeed, explanation added.

docs/data-retention-mechanisms.md
1pretz1: Sure, added more detail!

docs/data-retention-mechanisms.md
1pretz1: I've added a bit more detail and the motivations which improves the section. 

I think we need to be careful though using 'over a year' as that implies an out and out rule but thats not the case for some cases.

1pretz1:
I think I've addressed all your comments @benthorner and I've updated the doc to take into account that we're retiring the activate / deactivate concept.

Most of the 'why' comments were explained in the first attempt of the ADR, I just popped them in the wrong place! 😄

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,65 @@
+# Data cleanup mechanisms
+
+There are several mechanisms regarding the cleanup of data in EmailAlertApi.
+This doc gives a brief overview of each.
+
+## Nullifying subscribers
```
gavinwye, line 6: I wonder if it's worth changing the word subscribers to email addresses. The language of nullifying a subscriber feels a bit odd and it's the email address we're doing this to not the subscriber.

gavinwye:
There are some interesting terms in here that I think it would benefit the wider team to understand. At some point, it may be a good idea to build a concept model of how the service works.

docs/data-cleanup-mechanisms.md
1pretz1: Hmm, that's a good point!

I always thought of it through database goggles as a subscriber record has an email address and so we're effectively nullifying the subscriber. However I can see that doesn't make much sense to non devs so I've updated it to `Nullifying email addresses`.

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,106 @@
...
+
+The [archive worker runs every hour].
+
+[Athena]: https://github.com/alphagov/email-alert-api/blob/master/docs/analytics.md#analytics-through-athena
+[EmailArchiveWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/email_archive_worker.rb
+[archive worker runs every hour]: https://github.com/alphagov/email-alert-api/blob/ec075c7198b965d3771726946dcd37783b9de9c5/config/sidekiq.yml#L29-L31
+
+## Deleting emails
+
+The [EmailDeletionWorker] deletes emails after they've been archived for over
```
benthorner, line 51:
```suggestion
The [EmailDeletionWorker] deletes emails after 
```

Related to the above comment about archiving being deprecated: we'll just be deleting emails 7 days after they were created.

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,106 @@
...
+is that we remove subscriber lists which have been created over seven days ago
+and never had any subscriptions.
+
+We remove this data so we:
+
+* Don't pollute our analytics.
+
+* Don't run into future capacity issues.
+
+The following list explains how we deem each model's records as historic:
```
benthorner, line 81: We could just refer to the ADR for this, as it's the same content and someone can also just read the worker code, now they know it's there.

docs/data-retention-mechanisms.md
benthorner: Thanks. I actually wonder if we could combine the `EmailDeletionWorker` and the `HistoricalDataDeletionWorker` together, if the latter runs more frequently. Is there a reason we can't do that?

That's what I think we should explain here. It may be there is no reason and it's purely historical. If so, we could say we're planning to combine them as part of removing the archiving feature (as an idea).

benthorner (commented):
Thanks for looking at those comments @1pretz1. The ground has shifted again with the (proposed) deprecation of archiving, so I've left a more major comment this time round, since we should make sure this doc is in sync with that consequence.

docs/data-retention-mechanisms.md
1pretz1: Theres no reason why we can't combine the workers, technically speaking. However when the historic worker was created it was decided that the deletion of emails was solving a separate issue to what the historic worker was intended for, hence the different scheduled worker run times.

E.g: 

* Email deletion worker - to keep the db performant (as v large quantity of emails produced)

* Historical deletion worker - keeping the database tidy and manageable (not perpetually keeping unused data)

I think both workers responsibilities are now converging though so it may make sense to combine them. An example being is that we may have to start deleting digest_run_subscribers sooner and sooner as we produce more and more and at that point the historic worker will have v similar responsibilities to the email deletion worker.

I'll add a bit in to explain that's what we're planning to do when retiring archiving. I guess this means I'll have to update the ADR too.

benthorner: Thanks @1pretz1, a caveat in this section sounds good. Once archiving is gone I agree it makes sense to look at combining them, as long as no one part of the deletion takes a long time (on its own).

I don't think you need to change the ADR: the policy is unchanged; we're just talking about the location of the code, which is always up for change in the future. ADRs don't need to be permanently accurate.

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,106 @@
+# Data cleanup mechanisms
+
+There are several mechanisms regarding the cleanup of data in EmailAlertApi.
+This doc gives an overview of each.
+
+## Nullifying email addresses
+
+Updates the `address` of old unsubscribed `Subscriber` records to `nil`. We do
+this to comply with GDPR, specifically so we're not retaining users data for
```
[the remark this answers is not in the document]
1pretz1: I used GDPR as this is the regulation used in the initial data retention policy. Have updated to your more general suggestion.

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,106 @@
...
+but this would limit our ability to assist in user support queries. E.g.
+restoring a users subscriptions after they've unsubscribed by accident which
+would be impossible to do if the system had nullified their address
+immediately after unsubscribing.
+
+Secondly, subscribers are nullified which are created more than 28 days ago and
+don't have any subscriptions at all (active or ended). It shouldn't be possible
+for subscribers to exist without subscriptions but these are queried and
+removed anyway to prevent storing user's data indefinitely if, for whatever
+reason, a subscriber ends up with no subscriptions. This has happened before,
```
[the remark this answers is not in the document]
1pretz1: I've left the wording as is for the reason that we did have this happen before for subscribers that already existed ([when deactivating](https://github.com/alphagov/email-alert-api/pull/1462/commits/053859c5962eef104256661f28727b08a43e3d31)). I know that concept is gone now but it's still true that we could remove existing subscribers subscriptions somehow in the future.

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,106 @@
...
+unintentionally, see [here][sub bug] and [here][sub bug two].
+
+The [NullifySubscribersWorker] runs [every hour].
+
+[every hour]: https://github.com/alphagov/email-alert-api/blob/master/config/sidekiq.yml#L26-L28
+[NullifySubscribersWorker]: https://github.com/alphagov/email-alert-api/blob/master/app/workers/nullify_subscribers_worker.rb
+[sub bug]: https://github.com/alphagov/email-alert-api/pull/1462/commits/053859c5962eef104256661f28727b08a43e3d31
+[sub bug two]: https://github.com/alphagov/email-alert-api/pull/1462/commits/a4b3e82801d79abd3989b1dd60ffd499e7ce82ba
+
+## Archiving emails
```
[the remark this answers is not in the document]
1pretz1: I've combined archiving and deleting into one section and added a note on our plan for both of these mechanisms.

docs/data-cleanup-mechanisms.md
1pretz1: I've left how it works as is but have added the future suggestion to the note as the top of the section.

I'd find it strange if I looked at this documentation and it was telling me that the code did something different to what it actually does, currently.

1pretz1:
Thanks for the review @benthorner. I've updated the doc inline with the changes to the ADR and your comments. 

I've tried to keep the doc current (as in how the system works now) as I feel it should reflect the current state of the system rather than script what it will do in the future as this may be slightly confusing for readers in the window where the changes aren't live. We can always update it again when we change the various mechanisms.

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,106 @@
+# Data cleanup mechanisms
+
+There are several mechanisms regarding the cleanup of data in EmailAlertApi.
```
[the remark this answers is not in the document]
benthorner: This appears to be unchanged.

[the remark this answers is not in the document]
1pretz1: Sorry I did change it but removed my change when I pushed it up. Now checking for more mistakes I made with the old 'undo' key.

benthorner (approved):
Looks good - nice one keeping this going in tandem with the ADR ☯️.

I spotted a few minor things that would be good to fix before this goes in.

docs/data-cleanup-mechanisms.md
```diff
@@ -0,0 +1,83 @@
...
+
+We remove this data so we don't pollute our analytics or run into future
+capacity issues.
+
+You can read in more detail about the decision to implement the historic
+deletion of data in [ADR-7].
+
+[midday]: https://github.com/alphagov/email-alert-api/blob/master/config/sidekiq.yml#L23-L25
+[HistoricalDataDeletionWorker]: https://github.com/alphagov/email-alert-api/blob/a62abc85453b723d683c2dc13f3bf0065fb86d5f/app/workers/historical_data_deletion_worker.rb
+[ADR-7]: https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-007-data-retention.md#decision
```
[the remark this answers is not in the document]
1pretz1: Doh, good spot!

PR #1461, opened 2020-11-04
Prepare to migrate non-migrated subscribers

benthorner:
https://trello.com/c/mZjpPawg/605-switch-remaining-80-of-coronavirus-and-transition-subscribers

This resurrects and modifies the task to migrate subscribers
to daily digest emails for a specified bunch of lists.

### DIFF
--- a/app/builders/switch_to_daily_digest_email_builder.rb
+++ b/app/builders/switch_to_daily_digest_email_builder.rb
@@ -42,7 +42,7 @@ private
   end
 
   def manage_subscriptions_link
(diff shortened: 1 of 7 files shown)


benthorner:
Tested on Integration:

- Took around 15 mins to migrate 88344 subscribers ✅.
- Double checked it respects people who've switched back ✅ .

1pretz1 (approved):
Nice work @benthorner, all makes sense to me! 

My only comment was going to be it seems a bit superfluous including a CSV for two lists but you've already mentioned that in the commit.

PR #1462, opened 2020-11-04
Remove deactivated concept - part 1

1pretz1:
Removes the activated / deactivated concept from email-alert-api. This was quite confusing and isn't actually needed.

These concepts revolve around the `deactivated_at` field on the subscriber model. This field was mainly used as a means of determining when a subscriber could be `nullified`. We can however obtain this information via the most recent ended subscription of the subscriber (based on the `ended_at` field) so we can remove the application code referencing this field.

Part 2 will include the actual removal of the column.

Trello:


app/models/subscriber.rb
```diff
@@ -4,8 +4,6 @@ class Subscriber < ApplicationRecord
     validates :address, uniqueness: { case_sensitive: false }
   end
 
-  validate :not_nullified_and_activated
```
benthorner, line 7: Minor: it was puzzling me why this was the first commit, so you could explain it's required for the next one (where we nullify without deactivating).

benthorner (commented):
Fabulous cleanup @1pretz1 - this is so much simpler 🏅. The whole "active" vs. "activated" thing has always confused and irritated me. Apart from hard logic, we've got a couple of supporting statements to get rid of it:

- [ADR 3 on Data Retention](https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-003-data-retention.md#subscriber-and-subscription) has an aside saying this concept may not be necessary (too right!).
- A [discussion with @kevindew](https://github.com/alphagov/email-alert-api/pull/1284#issuecomment-646804004) also reaffirms how checking for active subscriptions is sufficient.

It could still surprise future devs to find records with `nil` email addresses. This doesn't change that, although it does make the reason for it slightly less visible, with the intent to remove `deactivated_at`. I think that's OK.

Thanks for finding all the places this concept has leaked over the years! I've got a couple of more major comments about the policy for nullifying and deleting subscribers. 

I was also surprised to see we've got 90K non-nullified subscribers. Apart from the whole thing being a bit of a mess (i.e. drift was inevitable), I think I found one of the culprits: the query [here](https://github.com/alphagov/email-alert-api/blob/856b76236dd58085f5b12f7d98f63aaedb9bf7be/app/services/unsubscribe_service.rb#L25) was not scoped to **active** subscriptions, so would return false for any subscriber who'd previously modified their subscriptions. The 90K are all from over a year ago because we only delete all the ended subscriptions after a year. So, to summarise, you've identified a major bug in the current deactivate/nullify process, which means we've been holding many people's personal data indefinitely. Nice catch!

(All the more reason to get this in!)

app/workers/nullify_subscribers_worker.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+private
+
+  def nullifyable_subscriber_ids
+    Subscriber.joins(:subscriptions)
+              .not_nullified
+              .where.not(active_subscriptions_exist)
+              .group("subscribers.id")
+              .having("MAX(ended_at) < ?", 28.days.ago)
```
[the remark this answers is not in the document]
1pretz1: I don't think this is necessary as the query won't return these subscribers as they wont have have any ended subscriptions at that point (this query wont return subscribers with no subscriptions).

spec/services/unsubscribe_service_spec.rb
```diff
@@ -7,32 +7,5 @@
       described_class.call(subscriber, [subscriptions.first], :unsubscribed)
       expect(subscriber.active_subscriptions.count).to eq 1
     end
-
-    context "when the email address is invalid" do
```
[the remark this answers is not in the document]
1pretz1: I've added a comment to the commit to clarify this.

app/services/unsubscribe_service.rb
```diff
@@ -12,16 +12,6 @@ def call
       subscriptions.each do |subscription|
         subscription.end(reason: reason)
       end
-
-      if !subscriber.deactivated? && no_other_subscriptions?(subscriber, subscriptions)
```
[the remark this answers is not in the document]
1pretz1: I did think about doing this originally although decided against it. I've added an alternative solution in [this commit](https://github.com/alphagov/email-alert-api/pull/1462/commits/8aae9cc5da790531ec7dd8f4fcc414848585db44) if you want to have a gander.

app/workers/historical_data_deletion_worker.rb
```diff
@@ -42,4 +42,10 @@ def empty_subscriber_lists
                   .where("subscriber_lists.created_at < ?", empty_list_retention_period)
                   .where("subscriptions.id": nil)
   end
+
+  def historic_nullified_subscribers
+    Subscriber.nullified
```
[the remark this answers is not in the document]
1pretz1: I've gone with your approach of creating both within a transaction.

app/models/subscriber.rb
1pretz1: I've combined this commit into the one after so there isn't the initial confusion. 😄

1pretz1:
Thanks for the review @benthorner! 😄 

I've made some changes off the back of your comments and it should be ready for a re-review.

app/workers/historical_data_deletion_worker.rb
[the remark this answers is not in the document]
benthorner: Ah, so by "the above" I meant "the previous/first" 😬. Sorry, I should have been more explicit. It still seems reasonable to create the subscriber in a transaction 👍. The reason the first approach is safer is because we can't predict what future code may come along, which might create subscribers without a subscription. In general, it's a bit odd to be deleting subscribers soon after they've been created, so waiting a year (in place of "nullified").

app/services/unsubscribe_service.rb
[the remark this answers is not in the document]
benthorner: Thanks - that's one less service, so I'm happy about that 🔢.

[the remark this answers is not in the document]
1pretz1: I've actually inlined this now as it made more sense than my first attempt!

app/workers/nullify_subscribers_worker.rb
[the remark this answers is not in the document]
benthorner: Oh yes, that's true. Thanks for explaining.

Looking at it another way, should we also nullify subscribers with no subscription after a certain point in time? Our [other discussion about creating subscriptions](https://github.com/alphagov/email-alert-api/pull/1462#discussion_r518017863) shows it's currently possible to have an orphaned subscriber, and while we might be fixing that, it's a reminder that the rest of the system may not always behave as we expect.

It's a relatively minor point, so I'm not fussed if we change this 👍.

app/workers/historical_data_deletion_worker.rb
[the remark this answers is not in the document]
1pretz1: Ah okay I get where you're coming from.

We can't trust the system... 😢

Well I've added the filter so the historic worker will only delete subscribers over a year old.

app/workers/nullify_subscribers_worker.rb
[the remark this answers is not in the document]
1pretz1: I did think about this too.

I think it's actually important to remove them as a system bug could mean we retain users data longer than needed. 

I've added an extra query to nullify all subscribers over 28 days old with no subscriptions. Let me know what you think!

app/workers/nullify_subscribers_worker.rb
```diff
@@ -0,0 +1,38 @@
+class NullifySubscribersWorker < ApplicationWorker
+  def perform
+    run_with_advisory_lock(Subscriber, "nullify") do
+      nullifyable_subscribers.update_all(address: nil)
```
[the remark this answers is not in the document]
benthorner: Alternatively, you could just have one query:

```
recently_active_subscriptions = Subscription
  .where("subscriptions.subscriber_id = subscribers.id")
  .where("ended_at IS NULL or ended_at > ?", NULLIFIABLE_PERIOD.ago)
  .arel.exists

Subscriber
  .where("created_at < ?", NULLIFIABLE_PERIOD.ago)
  .where.not(recently_active_subscriptions)
  .update_all(address: nil)
```

which also seems to be significantly faster than doing a `joins`. (It took me a while to come up with this - first time using `arel` 😮)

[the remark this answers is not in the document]
benthorner: Note that, with this query, I can a total of 287015 nullifiable subscribers, which hopefully matches roughly what you were expecting, given the bug we found.

spec/workers/nullify_subscribers_worker_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    it "doesn't nullify already nullified subscribers" do
+      create(:subscriber, :nullified)
+
+      expect { subject.perform }
+        .to_not(change { Subscriber.nullified.count })
+    end
+
+    it "doesn't nullify subscribers with recent active subscriptions" do
```
benthorner, line 26: Sorry I didn't notice this in my first review: this test doesn't actually check what it says it checks. It's essentially a duplicate of the next one. Can we just have one subscription that's only a few days old here instead?

app/controllers/subscriptions_controller.rb
```diff
@@ -23,14 +27,14 @@ def create
...
         )
       end
+    end
 
-      subscription = new_subscription || existing_subscription
-      email = if create_new_subscription
-                SubscriptionConfirmationEmailBuilder.call(subscription: subscription)
-              end
+    subscription = new_subscription || existing_subscription
+    email = if create_new_subscription
```
benthorner, line 33: I think we can do this in the transaction too: it's all inside the system. We take a similar approach [here](https://github.com/alphagov/email-alert-api/pull/1463/files#diff-547e678b8e52e4f0db33fa3d82cf46ed9ebbcca4aad50aefc889ce4ef2b168a8L33). If this becomes more involved, we could also separate this into a different PR, as it's tangential to the overall change.

app/workers/historical_data_deletion_worker.rb
```diff
@@ -15,7 +15,7 @@ def perform
     delete_and_log("subscriber lists") { empty_subscriber_lists }
 
     # restricts deletion if emails are present
-    delete_and_log("subscribers") { Subscriber.where("deactivated_at < ?", max_retention_period) }
+    delete_and_log("subscribers") { historic_nullified_subscribers }
```
benthorner, line 18:
```suggestion
    delete_and_log("subscribers") { historic_former_subscribers }
```

(This is no longer associated with nullification.)

spec/workers/historical_data_deletion_worker.rb
```diff
@@ -79,8 +79,10 @@ def perform
     end
 
     context "when deleting subscribers" do
-      it "should remove all old deactivated subscribers" do
-        create(:subscriber, deactivated_at: historic_date)
+      it "should remove all historic subscribers" do
```
benthorner, line 82: We're missing a test here for a historic subscriber with an active subscription.

app/controllers/unsubscribe_controller.rb
```diff
@@ -1,7 +1,7 @@
 class UnsubscribeController < ApplicationController
   def unsubscribe
     subscription = Subscription.active.find(id)
-    UnsubscribeService.call(subscription.subscriber, [subscription], :unsubscribed)
+    subscription.end(reason: :unsubscribed)
```
benthorner, line 4: Nice! The commit message could do with a "why" though.

benthorner (commented):
Thanks for looking through those @1pretz1! I'm conscious the scope of this is creeping a bit. Hopefully we'll reach a conclusion soon 🤞. Most of my comments are on changed stuff, but there are also a couple of more major issues I didn't spot the first time round (sorry about that!). I had an interesting time testing out the nullification queries 📲.

One general comment for future PRs: many of the removal commits don't have any associated description, and the messages themselves are quite vague. "Remove the concept of deactivated from subs worklow" also applies to other commits, for example. While small commits are often preferable, I would have found a single, larger one easier here.

Anyway, once we're confident about the query for the nullify worker, we should be basically good to go 🎉.

app/workers/historical_data_deletion_worker.rb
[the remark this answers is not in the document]
benthorner: Thanks! Looks good to me 👍.

app/workers/nullify_subscribers_worker.rb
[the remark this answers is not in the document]
benthorner: Oh yes, I see it. This had a few comments on it from my re-review, after testing out the queries. I can see you've added the condition, though, so I'll resolve this thread we can continue discussion on the other ones.

app/workers/nullify_subscribers_worker.rb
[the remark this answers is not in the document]
1pretz1: I've used your query cos it's speedy! Arel is cool, I stole the use from Kev. 😄  

I've also added a filter to make sure we aren't updating already nullified subscribers which should be a lil bit speedier. 

Side note: Every time I use a `joins` there always seems to be a more efficient approach (will remember this).

> Note that, with this query, I get a total of 287015 nullifiable subscribers, which hopefully matches roughly what you were expecting, given the bug we found.

Sounds about right.

app/controllers/subscriptions_controller.rb
1pretz1: I wasn't bothered about adding the email generation bits inside the transaction as we will be removing the use of this soon [here](https://trello.com/c/DkVHm6Gs/621-remove-subscription-confirmation-email). So I think is fine as is?

app/workers/historical_data_deletion_worker.rb
1pretz1: Good spot - I've just used `historic_subscribers`

spec/workers/historical_data_deletion_worker.rb
1pretz1: Good spot, have a test, or two.

1pretz1:
Thanks for the 2nd review @benthorner. I should've addressed your comments.

I've also switched to using sub queries instead of joins in the [historic deletion worker](https://github.com/alphagov/email-alert-api/pull/1462/commits/dd7ddedb20f1c948f32b792b0f64213a4b7efeb6#diff-fec774181493f87fd5050fc26b5f8b8760bd17e317b570337e6d9c445119402aR40-R57) as it's much quicker.

app/services/immediate_email_generation_service/batch.rb
```diff
@@ -61,9 +61,8 @@ def subscriptions_to_fulfill_by_subscriber
         end
     end
 
-    def activated_subscribers(subscriber_ids)
-      Subscriber.activated
-                .where(id: subscriber_ids)
+    def subscribers(subscriber_ids)
+      Subscriber.where(id: subscriber_ids)
                 .index_by(&:id)
```
kevindew, line 66: Minor, this doesn't look like it needs to be multiline anymore. Arguably this doesn't really need to be a method and could be just an extra line in `subscriptions_to_fulfill_by_subscriber`

app/services/unsubscribe_all_service.rb
```diff
@@ -7,6 +7,10 @@ def initialize(subscriber, reason, **)
   end
 
   def call
-    UnsubscribeService.call(subscriber, subscriber.active_subscriptions, reason)
+    ActiveRecord::Base.transaction do
+      subscriber.active_subscriptions.each do |subscription|
+        subscription.end(reason: reason)
```
kevindew, line 12: It feels a bit superfluous to be doing an n+1 to update each individual subscription here, given there isn't really any logic involved in the `end` method that is applicable.

You could do:

```
ended_count= subscriber.active_subscriptions.update_all(ended_at: Time.zone.now, ended_reason: reason)
```

To do it in one query (which would also save the need for a transaction as it would be atomic). You wouldn't need to worry about the check for ended or not since this would only be applying this to subscriptions that weren't ended.

Only thing that'd be missing would be the metrics. You could update `Metrics.unsubscribed` to accept a second parameter of the count.

app/workers/historical_data_deletion_worker.rb
```diff
@@ -37,9 +37,23 @@ def delete_and_log(model)
...
+
+    SubscriberList
+      .where("created_at < ?", empty_list_retention_period)
+      .where.not(subscriptions_exist)
+  end
+
+  def historic_subscribers
+    subscriptions_exist = Subscription.where(
+      "subscribers.id = subscriptions.subscriber_id",
+    ).arel.exists
```
kevindew, line 53: It'd be useful if the commit had metrics on the performance increase, saying "much quicker" leaves it rather ambiguous.

lib/tasks/support.rake
```diff
@@ -77,7 +77,7 @@ namespace :support do
       if active_subscriptions.empty?
         puts "Subscriber #{email_address} already unsubscribed from #{subscriber_list_slug}"
       else
-        UnsubscribeService.call(subscriber, [active_subscriptions.last], :unsubscribed)
+        active_subscriptions.last.end(:unsubscribed)
```
kevindew, line 80: Not your issue but it's a bit strange we do a `.last` here. It feels like https://github.com/alphagov/email-alert-api/blob/614fac637709cda9100fdca96e37e374d67081c9/lib/tasks/support.rake#L76

should be `subscription = Subscription.active.find_by(subscriber_list: subscriber_list, subscriber: subscriber)`

kevindew (commented):
Looks great - nice amount of code removal and really nice to no longer having to keep two things in sync.

app/workers/historical_data_deletion_worker.rb
benthorner: This may be in reference to [a comment I made](https://github.com/alphagov/email-alert-api/pull/1462#discussion_r518882513). I suspect it was actually the `group` that was causing the original query there to be so slow, but I only have the evidence of it not completing in a number of minutes.

I'm not sure it makes much difference for the queries here. That said, I do find the proposed change easier to read, as it's more consistent and matches the intuitive idea of checking each subscriber individually.

app/controllers/subscriptions_controller.rb
1pretz1: Okay sure, have added it back! 👍

app/services/immediate_email_generation_service/batch.rb
1pretz1: I've removed the method and updated `subscriptions_to_fulfill_by_subscriber` accordingly.

app/services/unsubscribe_all_service.rb
1pretz1: Sure, have done your recommendation and updated the `unsubscribed` method to take a count.

app/workers/historical_data_deletion_worker.rb
1pretz1: I've generally found joins to be quite slow in general versus multiple query approaches but I'll pop the exact benchmarks below.

With the `join` approach when finding historic subscriber lists the query ran in around 2700ms.

With the sub query approach it ran in around 167ms. 

Only 16 times quicker! 😆 

Have also added this info to the commit.

lib/tasks/support.rake
1pretz1: Good point, I've adjusted it to use `find_by`.

1pretz1:
Ta for the review @kevindew! Should've addressed your comments!

lib/metrics.rb
```diff
@@ -5,8 +5,8 @@ def content_change_emails(content_change, count)
       count("content_change_emails.document_type.#{content_change.document_type}.immediate", count)
     end
 
-    def unsubscribed(reason)
-      increment("unsubscribed.reason.#{reason}")
+    def unsubscribed(reason, count = 1)
+      count.times { increment("unsubscribed.reason.#{reason}") }
```
kevindew, line 9: It'd be more idiomatic to use `count("unsubscribed.reason.#{reason}", count)` here rather than a loop.

It'll do the same thing: https://github.com/reinh/statsd/blob/3863a1a341d21b3c81a0ab9aebc6bd901da45323/lib/statsd.rb#L351-L371

lib/tasks/support.rake
```diff
@@ -73,12 +73,12 @@ namespace :support do
     elsif !(subscriber.subscriptions.pluck(:subscriber_list_id).include? subscriber_list.id)
       puts "Subscriber #{email_address} does not appear to be signed up for #{subscriber_list_slug}"
     else
-      active_subscriptions = Subscription.active.where(subscriber_list: subscriber_list, subscriber: subscriber)
-      if active_subscriptions.empty?
-        puts "Subscriber #{email_address} already unsubscribed from #{subscriber_list_slug}"
-      else
-        UnsubscribeService.call(subscriber, [active_subscriptions.last], :unsubscribed)
+      active_subscription = Subscription.active.find_by(subscriber_list: subscriber_list, subscriber: subscriber)
```
kevindew, line 76: 🏅

lib/tasks/support.rake
```diff
@@ -73,12 +73,12 @@ namespace :support do
...
+      if active_subscription.present?
```
kevindew, line 77: Minor, `present?` is redundant here

spec/services/unsubscribe_all_service_spec.rb
```diff
@@ -0,0 +1,19 @@
+RSpec.describe UnsubscribeAllService do
+  describe ".call" do
+    let(:subscriber) { create(:subscriber, address: "foo@bar.com") }
```
kevindew, line 3: Minor, this email address seems redundant.

spec/workers/nullify_subscribers_worker_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    it "nullifies subscribers created after the nullifyable period that don't have any subscriptions" do
+      create(:subscriber, created_at: nullifyable_time)
+
+      expect { subject.perform }
+        .to change { Subscriber.nullified.count }.by(1)
+    end
+
+    it "doesn't nullify subscribers with recently ended subscriptions" do
+      subscriber = create(:subscriber, created_at: nullifyable_time)
```
kevindew, line 21: does setting nullifyable_time impact this test at all?

kevindew (commented):
Looks great @1pretz1, just a few minor comments left from me.

app/services/unsubscribe_all_service.rb
```diff
@@ -7,6 +7,10 @@ def initialize(subscriber, reason, **)
   end
 
   def call
-    UnsubscribeService.call(subscriber, subscriber.active_subscriptions, reason)
+    ended_count = subscriber.active_subscriptions
```
benthorner, line 10: Minor: this surprised me, as it doesn't seem to be documented behaviour.

benthorner (approved):
Hurray, we're at the end ✅.

Thanks for adding to the commit messages. Some of the commits have gained a bit of weight, but that's sort of expected for a PR that's gone on for a while, and for the number of comments. Nothing to change, but worth thinking about whether you would have found the whole thing easier if it were a sequence of smaller PRs.

lib/tasks/support.rake
1pretz1: Yep it is. 

I normally like being explicit in conditionals as I think it reads nicer / comparing against a boolean seems a more traditional approach. I'm not fussed though so have removed.

spec/services/unsubscribe_all_service_spec.rb
1pretz1: It is! Was a copy over from an existing test which has now been removed.

spec/workers/nullify_subscribers_worker_spec.rb
1pretz1: Yep it does otherwise the test would pass without actually testing the logic I wanted it to, see [here](https://github.com/alphagov/email-alert-api/pull/1462/files#diff-8ddc5cf4620c59f09c23573d63d8540c885cf3f6557988e8935dbd31ebaa0f01R18). I have rejigged the spec to more explicit which should clear things up.

app/services/unsubscribe_all_service.rb
1pretz1: I'm confused as to what you're referring to, could you explain?

benthorner: Oops, sorry - I can see it's a bit vague. The docs for `update_all` don't mention it returns the count of updated records (unless I missed that part). Since that's the actual behaviour, I think it's fine as-is (just a bit surprising).

1pretz1: Oh yes I see, not like rails to have incomplete docs! I assumed it was similar behaviour to `delete_all` which returns the amount of rows effected. 

There is a test at least which verifies the return of `update_all` so as you say, should be okay! 👍

kevindew: Looks like the documentation for it is relatively new: https://github.com/rails/rails/commit/b488d8e08950685ae978b118af813516ce71fecc

I wondered if it depended on the RDBMs but it seems to be universal.

PR #1463, opened 2020-11-05
Remove digest migration rake task

benthorner:
https://trello.com/c/mZjpPawg/605-switch-remaining-80-of-coronavirus-and-transition-subscribers

This is no longer needed since it was run. We can always reinstate
it when we come to do further migrations.


1pretz1 (approved):
💯

PR #1464, opened 2020-11-06
Handle SocketErrors for comms with Notify

benthorner:
https://sentry.io/organizations/govuk/issues/2010669159/?project=202220&referrer=slack

Previously we saw this raise a single error to Sentry, which is
likely due to a transient DNS failure. Although the SocketError
class may represent other kinds of error, this variant is most
likely [1], and we should avoid shouting about it, since it's not
actionable. This updates the NotifyProvider to handle these errors
graefully, noting we can always make the handling more specific if
in future we find we want to expose other variants.

SentryAPP Today at 12:46 AM
SocketError
Failed to open TCP connection to api.notifications.service.gov.uk:443 (getaddrinfo: Name or service not known)
environment
production
APP-EMAIL-ALERT-API-17Q via <https://sentry.io/organizations/govuk/alerts/rules/app-email-alert-api/283774/|Send a notification for new events>APP-EMAIL-ALERT-API-17Q via Send a notification for new events | Today at 12:46 AM

[1]: https://www.exceptionalcreatures.com/bestiary/SocketError.html

### DIFF
--- a/app/services/send_email_service/notify_provider.rb
+++ b/app/services/send_email_service/notify_provider.rb
@@ -21,7 +21,7 @@ class SendEmailService::NotifyProvider
 
     Metrics.sent_to_notify_successfully
     :sent
-  rescue Notifications::Client::RequestError, Net::OpenTimeout, Net::ReadTimeout => e
+  rescue Notifications::Client::RequestError, Net::OpenTimeout, Net::ReadTimeout, SocketError => e
     Metrics.failed_to_send_to_notify
 
     Rails.logger.warn(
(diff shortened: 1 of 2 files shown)


1pretz1 (approved):
Good spot! 👍

kevindew:
Argh good catch! - gosh so many sneaky errors. I wonder if we should see about raising an issue with Notify about them or opening a PR on the client to try have them handled there. I wonder if they should be using a higher level HTTP client that catches these things and makes them easier to wrap.

benthorner:
Well, it only happened once, so I'm not too worried about it. For all we know, they were doing some planned maintenance. I'm still a little uneasy about ignoring `SocketError`, because of its many meanings.

We now have [quite a large number of specific exception we have to handle at our level](https://github.com/alphagov/email-alert-api/blob/06b1ec355261dd6d4629f52d8a0ade070de113c4/app/services/send_email_service/notify_provider.rb#L24), and I agree that's much more in the domain of the library, given that it decides how to handle requests.

I've raised an issue about it: https://github.com/alphagov/notifications-ruby-client/issues/116

PR #1465, opened 2020-11-10
Stop sending two confirmation emails for new subs

benthorner:
https://trello.com/c/DkVHm6Gs/621-remove-subscription-confirmation-email

Previously these emails provided some reassurance that a new GOV.UK
subscription was working. However, the introduction of double opt-in makes
them largely redundant. This removes the second confirm email.


app/controllers/subscriptions_controller.rb
```diff
@@ -1,7 +1,5 @@
 class SubscriptionsController < ApplicationController
   def create
-    subscription, email, status = nil
```
1pretz1, line 3: Variables were short lived! 😄

1pretz1 (approved):
💯

PR #1468, opened 2020-11-11
Update `updated_at` when using `update_all`

1pretz1:
`update_all` doesn't update timestamps when used and so left the
'updated_at' column unchanged. This indirectly created a bug as we use
the `updated_at` field as filter for [one of our reports] which resulted
in nils being passed around.

[one of our reports]: https://github.com/alphagov/email-alert-api/blob/master/lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb#L17


benthorner (approved):
🔨

PR #1469, opened 2020-11-11
Backfill ended subscriptions updated_at

1pretz1:
In [this PR] a bug was introduced which stopped updating the
`updated_at` when all of a users subscriptions were ended. This had a
knock on effect of breaking a [report].

[this PR]: https://github.com/alphagov/email-alert-api/pull/1462#discussion_r519839598
[report]: https://github.com/alphagov/email-alert-api/blob/master/lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb#L17

### DIFF
--- a/db/migrate/20201111122823_backfill_updated_at_of_ended_subscriptions.rb
+++ b/db/migrate/20201111122823_backfill_updated_at_of_ended_subscriptions.rb
@@ -1,0 +1,13 @@
+class BackfillUpdatedAtOfEndedSubscriptions < ActiveRecord::Migration[6.0]
+  class Subscription < ApplicationRecord; end
+
+  disable_ddl_transaction!
+
+  def change
+    ended_subscriptions = Subscription.where("ended_at > updated_at")
+
+    ended_subscriptions.find_each do |sub|
+      sub.update!(updated_at: sub.ended_at)
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_10_163036) do
+ActiveRecord::Schema.define(version: 2020_11_11_122823) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"


db/migrate/20201111122823_backfill_updated_at_of_ended_subscriptions.rb
```diff
@@ -0,0 +1,12 @@
+class BackfillUpdatedAtOfEndedSubscriptions < ActiveRecord::Migration[6.0]
+  def change
+    ended_subscriptions = Subscription.where(
+      # The time the bug was deployed
+      "ended_at > ?", Time.zone.parse("2020-11-10 16:26:00 +0000")
```
benthorner, line 5: You could alternatively do:

```
Subscription.where("ended_at > updated_at")
```

benthorner: It looks like this gotcha has been a recurring them: https://github.com/alphagov/email-alert-api/search?q=update_all

We ought to fix it for the other occurrences: nullify and archive.

1pretz1: There seems to be quite a few subscriptions (before the bug was introduced) where that query returns true.

E.g

```
irb(main):002:0>     ended_subscriptions = Subscription.where("ended_at > updated_at").count
=> 629854
irb(main):003:2* (    ended_subscriptions = Subscription.where(
irb(main):004:2*     # The time the bug was deployed
irb(main):005:2*     "ended_at > ?", Time.zone.parse("2020-11-10 16:26:00 +0000")
irb(main):006:0>   ).count )
irb(main):007:0*
=> 1663
```

benthorner: Yup - whenever someone did something with `update_all` on it. Arguably we don't _need_ to fix them, but we may as well.

db/migrate/20201111122823_backfill_updated_at_of_ended_subscriptions.rb
```diff
@@ -0,0 +1,11 @@
+class BackfillUpdatedAtOfEndedSubscriptions < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
```
benthorner, line 2: Minor: you could mention why this is here in the commit - I'm assuming it's because the update is idempotent.

benthorner (approved):
Looks good - thanks for extending the scope a bit 👍.

I find the use of migrations to modify data a little strange, but I see we've [done it before](https://github.com/alphagov/email-alert-api/blob/master/db/migrate/20201019154701_delete_lingering_subscription_contents.rb), so it seems OK to continue the convention, which at least documents the change.

db/migrate/20201111122823_backfill_updated_at_of_ended_subscriptions.rb
1pretz1: I've updated the commit message. My thinking was as there is 629,854 records to update, if there is a transient failure along the way we won't have to start over and could just re-run the migration where it left off.

1pretz1:
Thanks for the review @benthorner. I've also redefined the `Subscription` class in the migration to make the migration resilient against future model changes.

benthorner:
> I've also redefined the Subscription class in the migration to make the migration resilient against future model changes.

This is an excellent argument for not using migrations for this sort of thing 😆. To that end, perhaps we should just cut the not-great practice now and just do it with a temporary rake task.

What do you think?

1pretz1:
I suppose the negative of doing it with a rake task is that it disappears when we remove it meaning we lose visibility of the change, although it would be present through Github with some searching.

However we've now done subscriber list migrations in rake tasks so it seems consistent with that approach so I'll close this PR and resurrect as a rake task @benthorner  😄

PR #1470, opened 2020-11-11
Update `updated_at` when we use `update_all`

1pretz1:
update_all doesn't update timestamps when used and so leaves the
'updated_at' timestamp unchanged. This indirectly created a bug as we
use the updated_at field as filter for one of [our reports] which resulted
in nils being passed around.

This updates the other uses of `update_all` in the email system.

[our reports]: https://github.com/alphagov/email-alert-api/blob/master/lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb#L17

### DIFF
--- a/app/workers/email_archive_worker.rb
+++ b/app/workers/email_archive_worker.rb
@@ -35,7 +35,7 @@ private
   end
 
   def mark_emails_as_archived(ids, archived_at)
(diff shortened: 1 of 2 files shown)


benthorner (approved):
Thanks for doing the extra fixes ⭐

PR #1471, opened 2020-11-11
Backfill ended subscriptions updated_at

1pretz1:
In [this PR] a bug was introduced which stopped updating the
`updated_at` when all of a users subscriptions were ended. This had a
knock on effect of breaking a [report].

Also updates all the other instances where the ended_at time does not
match updated_at.

[this PR]: https://github.com/alphagov/email-alert-api/pull/1462#discussion_r519839598
[report]: https://github.com/alphagov/email-alert-api/blob/master/lib/reports/subscription_changes_after_switch_to_daily_digest_report.rb#L17


benthorner (approved):
Thanks for your patience with discussion on this 🏅.

PR #1472, opened 2020-11-12
Fix backfill rake task

1pretz1:
We've had to use `update_attribute` to skip validations as the
[uniqueness validation] concerning subscribers on the subscriptions
model was triggerining causing the following error:

```
ActiveRecord::RecordInvalid (Validation failed: Subscriber has already been taken)
```

The task is _currently_ running in Staging -> https://deploy.blue.staging.govuk.digital/job/run-rake-task/104022/console

[uniqueness validation]: https://github.com/alphagov/email-alert-api/blob/master/app/models/subscription.rb#L11

### DIFF
--- a/lib/tasks/backfill_updated_at_for_ended_subscriptions.rake
+++ b/lib/tasks/backfill_updated_at_for_ended_subscriptions.rake
@@ -3,6 +3,6 @@ task backfill_updated_at_for_ended_subscriptions: :environment do
   ended_subscriptions = Subscription.where("ended_at > updated_at")
 
   ended_subscriptions.find_each do |sub|
-    sub.update!(updated_at: sub.ended_at)
+    sub.update_attribute(:updated_at, sub.ended_at) # rubocop:disable Rails/SkipsModelValidations
   end
 end


kevindew (approved):
Doh! that seems an annoying issue

benthorner:
Yeah. While I guess we can proceed, this suggests we've got some other invalid data: multiple active subscriptions associated with the same list, for the same subscriber.

1pretz1:
Ta for the reviews!

I thought that @benthorner. That doesn't appear to be the case though (for some instances anyway), see below:

```
irb(main):053:0> sub = Subscription.where("ended_at > updated_at").last
=> #<Subscription subscriber_id: 955851, subscriber_list_id: 10870, created_at: "2020-02-26 13:42:31", updated_at: "2020-02-26 13:42:31", frequency: "immediately", signon_user_uid: "ee9c3b40-b7df-013...

irb(main):054:0> sub.subscriber.subscriptions.pluck(:subscriber_list_id, :ended_at)
=> [[10870, nil], [18200, Wed, 26 Feb 2020 11:48:22 GMT +00:00], [10870, Thu, 05 Nov 2020 13:03:18 GMT +00:00]]

irb(main):055:0> sub.update!(updated_at: sub.ended_at)
Traceback (most recent call last):
        1: from (irb):55
ActiveRecord::RecordInvalid (Validation failed: Subscriber has already been taken)
```

If a subscriber has an active and ended subscription for the same list then the validation will fail when updating the ended subscription.

I'll merge the PR so Andrew can have his data but we can continue the chat here.

benthorner:
@1pretz1 I think I found the problem: https://github.com/alphagov/email-alert-api/pull/1474.

PR #1474, opened 2020-11-12
Remove broken / faulty validations

benthorner:
https://github.com/alphagov/email-alert-api/pull/1472#issuecomment-726011646


1pretz1 (approved):
Nice bit of cleanup and investigation!  

Seems strange that we ended up in a situation where we had validation in the db and with active record for the same thing.

PR #1476, opened 2020-11-13
Pending ADR for new approach to monitoring and alerting

kevindew:
Trello: https://trello.com/c/qHyEW6cE/585-record-the-architectural-decisions-for-email-alert-system-alerting

This is an ADR that tries to succinctly cover the major themes of the
23 page document on "How do we know the Email Alert System is broken"
[1]. It is a pending ADR as we have not yet started the development of
these new alerts which may lead to some of the consequences of the
decision being altered.

There's a few places where it could have appropriate to link to internal
Google Docs (such as incidents or the source document) however it felt
unconventional to link to internal documents in an ADR so I left this
out.

This was a tricky decision to portray as the work behind it was many
decisions. I tried to focus just on the major themes with the
expectation that future implementation documentation will convey more
granular information. In trying to make this succinct there were number
of avenues I chose not to go down, notably:

- Talking about why alerting is embedded in the system and the risks of that
  compared to entirely external alerting (it's hard to get sufficient
  information to call it).
- Aiming to alert about a problem with only a single alert even if it affects
  multiple machines. We're trying to avoid the situation where a problem occurs
  and an engineer gets ~10 alerts about the problem.
- Why it's not appropriate to monitor every content change to the granularity
  as is done in Email Alert monitoring.
- Talking about consistency across GOV.UK alerting approaches. They're mostly
  quite simplistic and there are pretty big gaps where things can break and
  no-one would get called automatically.
- Details as to why the stack boundaries are what they are, why Notify
  monitoring isn't our responsibility for instance.
- How we'd know if the monitoring machines stopped receiving data that powers
  alerts.
- How we may want to adjust retry schedules and attempts to reflect the
  new approaches.

As always in an ADR I ended up in a muddle over tense. I initially tried
to have this in past tense but that didn't flow right given the pending
nature of consequences, so I've instead opted for future tense.

[1]: https://docs.google.com/document/d/1HczXfKL6dJz1OqjJIqJSA8AohUiZJfZbVhsorCJdx0E/edit#

### DIFF
--- a/docs/adr/adr-007-monitoring-and-alerting.md
+++ b/docs/adr/adr-007-monitoring-and-alerting.md
@@ -7,11 +7,11 @@ Date: 2020-11-13
 Email Alert API has acquired the undesirable reputation of an application that
 produces a high volume of unactionable alerts and is a frequent source of
 GOV.UK incidents. The monitoring and alerting approach no longer appears to
-be sufficient for the application behaviour as the volume of work for the
-application has changed in the past 2 years. In previous years, the application
+be sufficient for the application behaviour as the application's volume of
+work has changed in the past 2 years. In previous years, the application
 would reach a peak of sending 1 million emails a day whereas now it can be
-sending over 10 million emails a day. This increased volume of work has
-increased system latency which has impacting alerts and contributed to
+sending over 10 million emails a day. This elevated volume of work has
+increased system latency which has impacted alerts and contributed to
 incidents.
 
 The approach used to monitoring the application and the wider email alert
@@ -19,43 +19,42 @@ system is two fold. One approach is to monitor Email Alert API's internal
 state to see if it is [completing tasks within a particular
 time frame][monitor-time-frame] or [whether there is a high quantity of work to
 do][monitor-workload]. The other approach is to use an [end-to-end
-test][email-alert-monitoring] which checks a Gmail account to identify that
-a particular address received an expected email within a time frame. Both of
-these approaches are flawed.
+check][email-alert-monitoring] which verifies that a Gmail account
+has received an email within a time frame. Both of these approaches are flawed.
 
-Monitoring for high quantities of work, and whether work was completed, in a
-time frame has left Email Alert API vulnerable to producing alerts when the
+Monitoring for high quantities of work, and whether work was completed in a
+time frame, has left Email Alert API vulnerable to producing alerts when the
 application is busy, yet fully operational. These alerts are unactionable for
 support engineers as the alerts resolve themselves once the work is completed
-and there is little to nothing an engineer can do to speed this process up.
-For example, if the application has accumulated 1,000,000 emails to send it
-raises a critical alert even if it is sending these at the fastest rate Notify
+and there is little an engineer can do to speed this process up. For example,
+if the application has accumulated 1,000,000 emails to send it raises a
+critical alert even if it is sending these at the fastest rate Notify
 would allow. These alerts, therefore, waste the time of support engineers and
 create a risk that a support engineer might miss an alert that requires their
 intervention.
 
-Since these alerts can't be relied upon to suggest the application is
-broken, they aren't used for any out-of-hours alerts. This means that if they
-did reflect a genuine problem no engineers would be contacted. An engineer
-instead learns of a problem out-of-hours due to the other type of monitoring
-approach: the end-to-end test. These alert when content published
-through [Travel Advice Publisher][] does not arrive in a Gmail inbox within a
-specified time period.
+Since these alerts can't be relied upon to reliably suggest the
+application is broken, they aren't used for any out-of-hours alerts. Therefore,
+even if they did reflect a genuine problem no engineers would be contacted.
+An engineer instead learns of a problem out-of-hours due to the other type
+of monitoring approach: the end-to-end checks. These alert when content
+published through [Travel Advice Publisher][] does not arrive in
+a particular Gmail inbox within a specified time period.
 
-Monitoring an inbox provided its own distinct flaws. If Email Alert API is
-broken out-of-hours engineers would not be notified unless travel advice
-content was published. This alert also monitor beyond the boundaries of
-the email alert system - considering whether GOV.UK publishing, Notify and
-Gmail are all working - leading to alerts representing problems outside the
-control of GOV.UK engineers. Finally, by nature of being time based, these
+Monitoring an inbox for a limited set of emails has its own distinct flaws.
+If Email Alert API breaks outside of office hours then engineers would not be
+notified unless travel advice content was published. This alert also monitors
+beyond the boundaries of the email alert system - considering whether
+GOV.UK publishing, Notify and Gmail are all working - leading to alerts
+representing problems outside the control of GOV.UK engineers (such as Gmail
+identifying an email as spam). Finally, by nature of being time based, these
 alerts are also vulnerable to the system being busy - where a sufficiently
-high volume of travel advice publishings guarantees that an alert will
-trigger.
+high volume of travel advice publishings guarantees that an alert will trigger.
 
-Reviewing these helped us understand the reputation for unactionable alerts and
-reviewing the GOV.UK incidents for Email Alert API helped us to identify
-patterns where incidents reflected ambiguity over whether the system
-was broken or not. We felt there was room for improvement.
+Reviewing this monitoring approach helped us understand the reputation for
+unactionable alerts and reviewing the GOV.UK incidents for Email Alert API
+helped us to identify patterns where incidents reflected ambiguity over whether
+the system was broken or not. We felt there was room for improvement.
 
 [monitor-time-frame]: https://github.com/alphagov/email-alert-api/blob/c70fcd7e5d64393074a3eebd83b3467fd8cc03b3/app/workers/digest_run_worker.rb#L6-L9
 [monitor-workload]: https://github.com/alphagov/govuk-puppet/blob/fac4b5163f10736bb19aa51a8eb7bc276dd2cb40/modules/govuk/manifests/apps/email_alert_api/sidekiq_queue_check.pp#L33-L44
@@ -64,7 +63,7 @@ was broken or not. We felt there was room for improvement.
 
 ## Decision
 
-Our decision is that the system should raise a critical alert
+We decided that the system should raise a critical alert
 [if and only if](https://en.wikipedia.org/wiki/If_and_only_if) the email alert
 system is broken. The majority of these alerts should contact engineers
 out-of-hours.
@@ -75,7 +74,7 @@ applications: Email Alert API,
 [Email Alert Service](https://github.com/alphagov/email-alert-service); this
 excludes the monitoring of whether GOV.UK publishing is working and whether
 Notify is working as they are independently monitored. The use of
-"if and only if" reflects a condition we're forming that the system will not
+"if and only if" reflects a condition we've formed that the system will not
 raise critical alerts unless we're highly confident it is broken, whereas
 warnings are appropriate for other situations. We decided that for the email
 alert system to be "broken" it represents a failing that will lead to a user
@@ -85,7 +84,8 @@ contact engineers out-of-hours for them as otherwise users will be
 experiencing issues.
 
 We decided that we will re-approach our alerting with this mindset and look to
-replace alerts that are vulnerable to reflecting a busy status.
+replace alerts that are vulnerable to reflecting a busy status, or which
+extend beyond the boundaries of the system.
 
 ## Status
 
@@ -96,9 +96,9 @@ Pending
 We will be changing the approaching to monitoring and alerting across the email
 alert system of applications in order to meet this new alerting criteria.
 
-## Monitoring Email Alert API
+### Monitoring Email Alert API
 
-### Sidekiq workers
+#### Sidekiq workers
 
 We will configure critical alerts, that contact engineers out-of-hours, if
 individual Sidekiq jobs fail multiple times for the following workers:
@@ -110,16 +110,25 @@ individual Sidekiq jobs fail multiple times for the following workers:
 - [SendEmailWorker][]
 - [WeeklyDigestInitiatorWorker][]
 
-These workers will use the database to store the amount of times we have
+These workers will use the database to store the number of times we have
 attempted to perform an action. An alert will be raised if the quantity of
-failed attempts reaches a threshold (the suggestion is 5).
+failed attempts reaches a threshold (the suggestion is 5). To ensure that these
+jobs actually run we will configure the [RecoverLostJobsWorker][] to identify
+any of these that need to run and attempt them, this protects from lost work
+problems such as [Sidekiq crashing][sidekiq-reliability].
 
-We will contact engineers out-of-hours if either [SendEmailWorker][] or
-[DigestEmailGenerationWorker][] have not completed at least one successful run
-within a time period (suggestion is 30 minutes) despite work existing. This
-reflects a concern that we would like support engineers to be notified quickly
-if all instances of these jobs are failing and the aforementioned retry method
-may take a long time to alert in a busy period.
+We expect [SendEmailWorker][] and [DigestEmailGenerationWorker][] jobs to run
+quickly (durations of up to a few seconds) and know that there can be high
+volumes of them to run (both can regularly reach volumes of > 100,000).
+Therefore, we are concerned about a scenario where these consistently fail
+because it could take a long time for sufficient retries to occur
+for an alert to be raised (as each retry is placed at the back of the queue).
+To reflect this concern we will contact engineers if either of these workers
+has not completed one successful run within a time period despite work
+existing (suggested time period is 30 minutes). The [Sidekiq retry
+cadence][sidekiq-retry] for these jobs will need to be updated to ensure they
+always retry within the time period, this is to ensure we don't alert while
+work isn't being attempted.
 
 Similarly, we will contact engineers out-of-hours if we determine that
 [DigestRuns][DigestRun] are not created within a reasonable time frame
@@ -132,7 +141,7 @@ We will also monitor that runs of [RecoverLostJobsWorker][] and
 (suggestion is 1 hour). These jobs should continue running even when the
 system is busy and the lack of them running indicates that we no longer have
 confidence the system is monitored or able to recover. Thus, an alert will be
-created that contacts engineers out-of-hours should these never succeed in a
+created that contacts engineers out-of-hours should these not succeed in a
 time frame.
 
 For Sidekiq workers that don't affect the user noticeable aspects of Email
@@ -140,14 +149,21 @@ Alert API, such as [EmailDeletionWorker][], we won't raise critical alerts if
 these are failing. Instead we will raise a warning if any Sidekiq workers
 exhaust all their retries and are placed in the [dead set][sidekiq-dead]. This
 approach was chosen as it was a pragmatic technique to monitor multiple workers
-without adding checks for individual workers.
+without adding alerts for individual workers. Workers will have their
+[retry cadence][sidekiq-retry] considered so that this will alert in a
+reasonable time frame of the retries being exhausted (the Sidekiq default of 21
+days is very long).
 
-We will modify the [latency checks for email queues][latency-checks] to be
-significantly less sensitive to latency for all email sending except
-[transactional emails][]. This is because transactional emails have an
-[expectation of a prompt delivery][transactional-delivery-expectation] and the
-lack of them blocks a user from progressing with their task. For the other
-queues we will change the time before a warning occurs to be in the measure of
+We will modify the [latency checks for email queues][latency-checks]. We
+will be treating [transactional emails][] different to other latency checks
+as these emails have have an [expectation of a prompt
+delivery][transactional-delivery-expectation] and the lack of them blocks a
+user from progressing with their task. These will raise a warning after 5 mins
+and go critical (with engineers contacted out-of-hours) should the latency
+reach 15 mins.
+
+The other latency checks will be made significantly less sensitive to latency
+and we will change the time before a warning occurs to be in the measure of
 multiple hours compared to minutes, with the intention that they serve to
 alert a support engineer only when the system is exceptionally busy and may be
 unhealthy. These alerts will only reach a critical status if they hit a 24 hour
@@ -155,7 +171,7 @@ point as this would indicate that we are consistently creating email faster
 than the application can send it, which may mean the system is perpetually
 backlogged.
 
-### Web application
+#### Web application
 
 We decided that there should be a mechanism that alerts people out-of-hours if
 the Email Alert API web application appears broken, noting that currently an
@@ -171,7 +187,7 @@ load balancer has no healthy instances of the application][aws-lb-check]. This
 represents that no instances of the web application are healthy and
 [no requests will succeed][aws-alb-health].
 
-### Alerts to be removed
+#### Alerts to be removed
 
 We will remove the following alerts that will become superseded:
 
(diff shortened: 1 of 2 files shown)


docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,234 @@
...
+(suggestion is 1 hour). These jobs should continue running even when the
+system is busy and the lack of them running indicates that we no longer have
+confidence the system is monitored or able to recover. Thus, an alert will be
+created that contacts engineers out-of-hours should these never succeed in a
+time frame.
+
+For Sidekiq workers that don't affect the user noticeable aspects of Email
+Alert API, such as [EmailDeletionWorker][], we won't raise critical alerts if
+these are failing. Instead we will raise a warning if any Sidekiq workers
+exhaust all their retries and are placed in the [dead set][sidekiq-dead]. This
```
benthorner, line 141: Minor: I think it's worth mentioning (at a high level) that the retries should be adjusted so they are exhausted in a short-ish period, as otherwise someone reading may wonder when exactly this alert would go off.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,234 @@
...
+- [Incomplete DigestRuns after 2 hours][incomplete-digest-runs]
+
+## Monitoring Email Alert Service
+
+Email Alert Service listens for [Publishing API][] events and communicates
+these to Email Alert API. We currently [only have alerts][rabbitmq-alert] that
+occur in-hours if this system fails, so there is no visibility for any
+out-of-hours problems.
+
+We intend to improve this by raising an alert that contacts engineers
```
benthorner, line 189:
```suggestion
We intend to improve and consolidate the existing alerts, and contact an engineer
```

Minor: thought it worth a mention that we're not intending to increase the number of RabbitMQ alerts.

benthorner (commented):
Great job @kevindew ⏰.

I don't have much to say on the approach, although that's not surprising given all the effort you've already put into documenting and discussing it. What comments I have are just around clarifying a few things, especially for someone who may not be so familiar with the system and some of the subtleties of alerting - we'll be sharing this more widely, after all.

I'd also like to clarify the proposed latency alert for transactional emails, since I think we were planning to call someone out if they are sufficiently delayed. However, the short timeframe will make such an alert vulnerable to false positives, even with job recovery and normal retries. Perhaps the simplest compromise here is to align the critical/call-out threshold with that for normal emails, at 30 minutes. We've still got a potential recovery issue, but at least both alerts have the same issue.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,234 @@
...
+
+[Email Alert Monitoring][email-alert-monitoring] is the aforementioned
+end-to-end test that monitors whether travel advice and drug advice content
+changes send email to a Gmail inbox. The changes to alerting described in this
+ADR will render Email Alert Monitoring redundant, as the new alerts should
+capture real problems faster and cover a broader range of scenarios - this
+would leave Email Alert Monitoring capturing false positives and out-of-scope
+problems.
+
+Once the new monitoring and alerting approaches have been implemented and
```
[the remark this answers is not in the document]
kevindew: Thanks I've added a section on this.

kevindew (commented):
Thanks Ben - I've gone through your comments and made some amends. I need to give this another proof read but I'm tired now so will do that tomorrow and let you know when it's ready for review again.

kevindew:
Ok @benthorner, this should be good for another look.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,234 @@
...
+- [SendEmailWorker][]
+- [WeeklyDigestInitiatorWorker][]
+
+These workers will use the database to store the amount of times we have
+attempted to perform an action. An alert will be raised if the quantity of
+failed attempts reaches a threshold (the suggestion is 5).
+
+We will contact engineers out-of-hours if either [SendEmailWorker][] or
+[DigestEmailGenerationWorker][] have not completed at least one successful run
+within a time period (suggestion is 30 minutes) despite work existing. This
```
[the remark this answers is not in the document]
benthorner: Sure, I can explain. Currently our pattern is to only recover lost work [after an hour](https://github.com/alphagov/email-alert-api/blob/469efe788a290d1a2205d8f870c6aafb4c8c9b00/app/workers/recover_lost_jobs_worker/unprocessed_check.rb#L13). Even if we reduced the that to 30 mins, the worker only runs every half hour, so it could be up to an hour before we recover a lost email. If the total number of emails being sent is small in total (let's say we're processing a "tiny" content change at 11PM), then in the worst case it would be (say) 59 mins between the emails being created and a lost one being recovered and sent.

There's always the possibility of being unlucky, of course. Since we know this happens every so often, I was wondering if we should mention it here as a general caveat: the thresholds may need to be longer (or the recovery more frequent) to account for one or two losses in work.

Does that help?

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,234 @@
...
+### Sidekiq workers
+
+We will configure critical alerts, that contact engineers out-of-hours, if
+individual Sidekiq jobs fail multiple times for the following workers:
+
+- [DailyDigestInitiatorWorker][]
+- [DigestEmailGenerationWorker][]
+- [ProcessContentChangeWorker][]
+- [ProcessMessageWorker][]
+- [SendEmailWorker][]
```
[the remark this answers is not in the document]
benthorner: Sure, that's OK. I was trying to remember it's behaviour regarding rate limiting, which lead me to suggest this. In general, I think we're asserting that none of these jobs should be prone to failure, so we can treat them all equally.

Does that match your understanding?

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Monitoring Email Alert API
+
+### Sidekiq workers
+
+We will configure critical alerts, that contact engineers out-of-hours, if
+individual Sidekiq jobs fail multiple times for the following workers:
+
+- [DailyDigestInitiatorWorker][]
+- [DigestEmailGenerationWorker][]
```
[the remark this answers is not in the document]
benthorner: Thanks, the new content looks good, except for the next paragraph, which starts off with "Similarly" - I get that it's a similar alert, but the reasoning is (now) quite different. Minor, but could we just remove "Similarly"?

docs/adr/adr-007-monitoring-and-alerting.md
benthorner: Thanks for updating 👍.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+- [ProcessContentChangeWorker][]
+- [ProcessMessageWorker][]
+- [SendEmailWorker][]
+- [WeeklyDigestInitiatorWorker][]
+
+These workers will use the database to store the number of times we have
+attempted to perform an action. An alert will be raised if the quantity of
+failed attempts reaches a threshold (the suggestion is 5). To ensure that these
```
1pretz1, line 115: Could you elaborate on why 5 was chosen? 

Does this number also need to be adjusted per worker? 5 failed attempts trying to send a single email seems minor, 5 failed attempts trying to process a daily digest seems major.

benthorner:
Thanks @kevindew. I've left a few replies on some of the comments. Could you take a look at those? I thought I'd postpone re-reading the whole thing while it may still change, if that's OK.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+jobs actually run we will configure the [RecoverLostJobsWorker][] to identify
+any of these that need to run and attempt them, this protects from lost work
+problems such as [Sidekiq crashing][sidekiq-reliability].
```
1pretz1, line 118: Is this sentence referring to the work we've already done in the RecoverLostJobsWorker, or future work?

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+reasonable time frame of the retries being exhausted (the Sidekiq default of 21
+days is very long).
+
+We will modify the [latency checks for email queues][latency-checks]. We
+will be treating [transactional emails][] different to other latency checks
+as these emails have have an [expectation of a prompt
+delivery][transactional-delivery-expectation] and the lack of them blocks a
+user from progressing with their task. These will raise a warning after 5 mins
+and go critical (with engineers contacted out-of-hours) should the latency
+reach 15 mins.
```
1pretz1, line 163: Are we planning to call up people if this only applies to a single transactional email which is not sending?

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+considering whether the load balancer has any healthy instances.
+
+### Removal of Email Alert Monitoring
+
+[Email Alert Monitoring][email-alert-monitoring] is the aforementioned
+end-to-end check that monitors whether travel advice and medical safety content
+changes result in an email received by a Gmail inbox. We intend to retire this
+check as we believe the new approach to monitoring the email alert system will
+cover the aspects of this check that are within scope of the email alert system.
+Broadly the Email Alert Monitoring verifies that:
```
1pretz1, line 223:
```suggestion
Broadly Email Alert Monitoring verifies that:
```

1pretz1 (commented):
Looks great @kevindew! Very clear and easy to read!

Just left a few comments. I think a few of them probably relate to my lack of context around some of the decisions made.

docs/adr/adr-007-monitoring-and-alerting.md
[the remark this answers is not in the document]
kevindew: Yup, rate limiting doesn't result in a failure (the worker is [rescheduled](https://github.com/alphagov/email-alert-api/blob/1711885b4df9c80aed5d8d2333e4e007c778badd/app/workers/send_email_worker.rb#L16-L17)). `temporary-failure` is also something we hear about in a call back rather than at sending time.

Yeah, from the doc our decision was to treat these the same. With the external service there is a greater risk of a problem in SendEmailWorker but I recall us concluding the odds of 5 intermittent fails in a row were still absurdly low.

docs/adr/adr-007-monitoring-and-alerting.md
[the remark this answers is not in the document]
kevindew: Thanks, I understand now and is a valid point. 

The time period in RecoverLostJobsWorker should ideally be less than the period before we alert and really shouldn't be greater than it. I've modified the ADR to include a reference to this but haven't gone the whole way in defining the solution now - hopefully we can that out when working out how to retry SendEmailWorker (and apply the same to the DigestEmailGenerationWorker).

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+check][email-alert-monitoring] which verifies that a Gmail account
+has received an email within a time frame. Both of these approaches are flawed.
+
+Monitoring for high quantities of work, and whether work was completed in a
+time frame, has left Email Alert API vulnerable to producing alerts when the
+application is busy, yet fully operational. These alerts are unactionable for
+support engineers as the alerts resolve themselves once the work is completed
+and there is little an engineer can do to speed this process up. For example,
+if the application has accumulated 1,000,000 emails to send it raises a
+critical alert even if it is sending these at the fastest rate Notify
```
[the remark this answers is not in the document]
kevindew: Thanks, that's a good idea.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+
+Since these alerts can't be relied upon to reliably suggest the
+application is broken, they aren't used for any out-of-hours alerts. Therefore,
+even if they did reflect a genuine problem no engineers would be contacted.
+An engineer instead learns of a problem out-of-hours due to the other type
+of monitoring approach: the end-to-end checks. These alert when content
+published through [Travel Advice Publisher][] does not arrive in
+a particular Gmail inbox within a specified time period.
+
+Monitoring an inbox for a limited set of emails has its own distinct flaws.
```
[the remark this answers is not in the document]
kevindew: I don't think we need to try make this list exhaustive. This seems useful extra arguments if we need to work harder to convince someone of their problems, but that seems beyond the scope of this ADR.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+[if and only if](https://en.wikipedia.org/wiki/If_and_only_if) the email alert
+system is broken. The majority of these alerts should contact engineers
+out-of-hours.
+
+Unpacking this, we consider the email alert system to be comprised of 3
+applications: Email Alert API,
+[Email Alert Frontend](https://github.com/alphagov/email-alert-frontend) and
+[Email Alert Service](https://github.com/alphagov/email-alert-service); this
+excludes the monitoring of whether GOV.UK publishing is working and whether
+Notify is working as they are independently monitored. The use of
```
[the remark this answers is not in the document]
kevindew: Thanks 👍

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+
+We will configure critical alerts, that contact engineers out-of-hours, if
+individual Sidekiq jobs fail multiple times for the following workers:
+
+- [DailyDigestInitiatorWorker][]
+- [DigestEmailGenerationWorker][]
+- [ProcessContentChangeWorker][]
+- [ProcessMessageWorker][]
+- [SendEmailWorker][]
+- [WeeklyDigestInitiatorWorker][]
```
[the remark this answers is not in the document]
kevindew: They're already alphabetical - the casulty of alphabetical was the Digest ones were no longer grouped, however that made the whole list ordering arbitrary.

docs/adr/adr-007-monitoring-and-alerting.md
```diff
@@ -0,0 +1,267 @@
...
+
+These workers will use the database to store the number of times we have
```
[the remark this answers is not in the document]
kevindew: That seems more of an implementation detail.

docs/adr/adr-007-monitoring-and-alerting.md
kevindew: I've added a brief comment to justify 5. The same would be applied to all the workers. We worked out that with intermittent network problems we'd be extremely unlucky to have the same email 5 times. All the other workers don't make external requests to services so should be very unlikely to fail.

There's an aspect from the original document of seeing if we got any SendEmailWorker false positives due to intermittent failures. Ideas on approaching it could be automatically retrying within the worker on SendEmailWorker failures or setting that worker to be more lenient. Hopefully this won't be needed though.

I'm not too sure how to add this into the ADR without it going off on a tangent... 🤔

docs/adr/adr-007-monitoring-and-alerting.md
kevindew: A bit of both - we've done some things and still have more to go.

docs/adr/adr-007-monitoring-and-alerting.md
kevindew: It's based on the queue latency and doesn't consider the number of items. In theory it could be possible to have one item in the queue and a latency of 15 minutes but in practice that should be virtually impossible as you'd only have queue latency when there is a high number of items in the queue.

kevindew:
Thanks @benthorner and @1pretz1 I've just pushed up another round of amends.

docs/adr/adr-007-monitoring-and-alerting.md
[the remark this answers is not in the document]
benthorner: Yes, I agree we can sort this out as an implementation detail.

docs/adr/adr-008-monitoring-and-alerting.md
```diff
@@ -0,0 +1,269 @@
...
+Monitoring for high quantities of work, and whether work was completed in a
+time frame, has left Email Alert API vulnerable to producing alerts when the
+application is busy, yet fully operational. These alerts are unactionable for
+support engineers as the alerts resolve themselves once the work is completed
+and there is little an engineer can do to speed this process up. For example,
+if the application has accumulated 1,000,000 emails to send it raises a
+critical alert even if it is sending these at the [fastest rate Notify
+would allow][rate-limit]. These alerts, therefore, waste the time of support
+engineers and create a risk that a support engineer might miss an alert that
+requires their intervention.
```
benthorner, line 34: Minor: we could acknowledge that it is still necessary to monitor these things for capacity planning, but that this should not be conflated with alerts for when the system is (definitely) broken.

docs/adr/adr-008-monitoring-and-alerting.md
```diff
@@ -0,0 +1,269 @@
...
+of monitoring approach: the end-to-end checks. These alert when content
+published through [Travel Advice Publisher][] does not arrive in
+a particular Gmail inbox within a specified time period.
+
+Monitoring an inbox for a limited set of emails has its own distinct flaws.
+If Email Alert API breaks outside of office hours then engineers would not be
+notified unless travel advice content was published. This alert also monitors
+beyond the boundaries of the email alert system - considering whether
+GOV.UK publishing, Notify and Gmail are all working - leading to alerts
+representing problems outside the control of GOV.UK engineers (such as Gmail
```
benthorner, line 49:
```suggestion
representing problems that are irrelevant to the system, or outside the control of GOV.UK engineers (such as Gmail
```

Minor: previously I didn't clock that we're implying problems with GOV.UK Publishing might be outside the control of GOV.UK engineers.

docs/adr/adr-008-monitoring-and-alerting.md
```diff
@@ -0,0 +1,269 @@
...
+failed attempts reaches a threshold (the suggestion is 5 - based on this being
+evidence of consistent failure). To ensure that these jobs actually run we
+will configure the [RecoverLostJobsWorker][] to identify any of these that
+need to run and attempt them, this protects from lost work problems such
+as [Sidekiq crashing][sidekiq-reliability].
+
+We expect [SendEmailWorker][] and [DigestEmailGenerationWorker][] jobs to run
+quickly (durations of up to a few seconds) and know that there can be high
+volumes of them to run (both can regularly reach volumes of > 100,000).
+Therefore, we are concerned about a scenario where these consistently fail
```
benthorner, line 124:
```suggestion
Therefore, we are concerned about a scenario where these consistently fail,
```

(They're not failing because it could take a long...)

docs/adr/adr-008-monitoring-and-alerting.md
```diff
@@ -0,0 +1,269 @@
...
+To reflect this concern we will contact engineers if either of these workers
+has not completed one successful run within a time period despite work
+existing (suggested time period is 30 minutes). The [Sidekiq retry
+cadence][sidekiq-retry] and [recovery window][] for these jobs will need to be
+updated to ensure the jobs are always attempted within the time period, this
+is to ensure we don't alert while work isn't being attempted.
+
+We will contact engineers out-of-hours if we determine that
+[DigestRuns][DigestRun] are not created within a reasonable time frame
+(suggestion is 2 hours). Creating these is a [trivial][digest-run-creation],
```
benthorner, line 136: Minor: the wording here is slightly at-odds with the definition in the README, which describes the "batch" combo of the DigestRun object and all the DigestRunSubscribers. It may help to explain why we're describing the creation of DigestRun records separately - they are (trivially) created on a schedule, and we're monitoring the schedule.

docs/adr/adr-008-monitoring-and-alerting.md
```diff
@@ -0,0 +1,269 @@
...
+
+- [Content Changes unprocessed after 2 hours][unprocessed-content-changes]
+- [Messages unprocessed after 2 hours][unprocessed-messages]
+- [Incomplete DigestRuns after 2 hours][incomplete-digest-runs]
+
+### Monitoring Email Alert Service
+
+Email Alert Service listens for [Publishing API][] events and communicates
+these to Email Alert API. We currently [only have alerts][rabbitmq-alert] that
+occur in-hours if this system fails, so there is no visibility for any
```
benthorner, line 203:
```suggestion
are monitored in-hours if this system fails, so there is no visibility for any
```

(The alerts still exist out-of-hours.)

benthorner (approved):
Thanks for addressing all my comments @kevindew 💯. I spotted a couple of minor things on reading through again - just suggestions to help someone else understand the content, as they don't change the outcome from my perspective.

docs/adr/adr-007-monitoring-and-alerting.md
1pretz1: My thoughts were around as we have the `send_email_immediate_high` queue which has the same priority as as the transactional queue then could it be feasible that a single transactional email gets stuck behind a load of high priority emails?

docs/adr/adr-007-monitoring-and-alerting.md
1pretz1: Your brief comment does help but could we add something along the lines of your first paragraph as I think it's helpful to highlight the decision up on why 5 was chosen and applied to all workers equally?

docs/adr/adr-007-monitoring-and-alerting.md
kevindew: Oh I had thought the transactional email was a higher priority than immediate high, that is a surprise. It probably should be higher as they are more important. @benthorner do you have thoughts on that? It doesn't really have an impact as we don't expect that queue to grow big.

Even with them being the same priority it can't be blocked by send_email_immediate_high. The priority indicates how frequently the queue is checked so each time Sidekiq needs to find a new job it'll check one of it's queues, and the priority indicates that it'll check send_email_transactional 8 times more frequently than the default queue.  In practice this means that means that even with a huge send_email_immediate_high queue to work through the first items in every other queue will still get worked on rapidly.

The situation that could block this is if all the Sidekiq workers were busy with a slow running task. For example if we had 300 content changes to process, which take 30 mins to process, and they exhausted all workers then we could find all email is blocked. If we were worried about that scenario we'd probably want to run a separate Sidekiq process for email sending.

1pretz1: Ah, thank you for the detailed explanation and for clearing up my concern! 

I agree it feels like transactional emails should be deemed as a higher priority even if the priority isn't going to make much of a difference in practice, due to as you say, there is only going to be a handful of them in the queue at any one time.

benthorner: Thanks for spotting @1pretz1.

I don't mind much either way. An argument for keeping it as-is would be that it emphasises how the intuitive interpretation of "priority" does not match the way Sidekiq works. As long as they are processed more quickly than most other email, I think it's fine to leave it as-is, but I also don't mind if we increase it symbolically.

docs/adr/adr-007-monitoring-and-alerting.md
kevindew: Sure, I've updated it some more.

kevindew:
Thanks @benthorner and @1pretz1 I've applied your suggestions. I think this is ready to go unless I've introduced any new problems.

docs/adr/adr-007-monitoring-and-alerting.md
1pretz1: Much clearer, thanks!

1pretz1 (approved):
Looks good to go from my end! 

Ta for resolving the comments!

ChrisBAshton (approved):
A detailed, sensible proposal 👍  We may want to make it an explicit part of the ADR that each alert scenario has corresponding documentation for the on-call engineer to follow, i.e. we won't raise any new alerts without documenting what should be done about them.

That said, documenting the resolution steps themselves would be out of scope for this ADR.

chao-xian (commented):
This all makes sense to me. Chris has already mentioned this, but I echo pointing to docs on how to address some of the potential issues mentioned.

kevindew:
Thanks @ChrisBAshton and @chao-xian ⭐  Yup, it's considered implicit that there is documentation that accompanies any alert, I'm not sure it needs calling out explicitly as writing documentation isn't necessarily a part of the architectural decision.

deborahchua:
Thanks for putting this together Kevin, really detailed and well explained. They all sound like good improvements!

docs/adr/adr-008-monitoring-and-alerting.md
```diff
@@ -0,0 +1,277 @@
...
+application becoming unavailable - unless travel advice were published. We
+noted there wasn't a particularly consistent approach to this across GOV.UK
+([with some apps raising an out-of-hours alert when a single instance fails
+healthchecks][publishing-api-alert]) and felt it would be good to take a simple
+approach to avoid surprising engineers.
+
+We decided the most appropriate place to alert would be when [the application
+load balancer has no healthy instances of the application][aws-lb-check]. This
+represents that no instances of the web application are healthy and
+[no requests will succeed][aws-alb-health].
```
sengi, line 197: nit: all the load balancer healthchecks failing doesn't _necessarily_ mean that user requests aren't succeeding, because in that situation the load balancer will behave as if all backends are healthy (because the healthchecks might be returning false negatives so it might as well try — and this type of scenario isn't quite as rare as we might be tempted to imagine).

I think it's still a reasonable surrogate measure (if a little unorthodox) and probably a good tradeoff for now in terms of implementation simplicity. It's also no bad thing to alert in the case where the healthchecks are all broken, even if the service might still be up - since that scenario would mean the alerting is blind to any real outage that might happen until the healthchecks are fixed.

sengi (commented):
I think this is a giant leap in the right direction for GOV.UK's alerting and it makes me so happy to see! Thanks for putting this together!

docs/adr/adr-008-monitoring-and-alerting.md
sengi: The reason I mention this is that it's something worth mentioning when someone comes to write the playbook entry for this alert — i.e. there are two possible causes: either the backends are all unavailable or (less likely but by no means impossible) the healthchecks are broken or timing out and therefore the alert is broken but the service may or may not still be up.

kevindew: @sengi Thanks for your comments ⭐ 

The behaviour for no healthy instances varies whether you are using an [application load balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html) or a [classic load balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html). Key phrasing:

Application Load Balancer: "If a target group contains only unhealthy registered targets, the load balancer nodes route requests across its unhealthy targets."
Classic Load Balancer:  "The load balancer routes requests only to the healthy instances. When the load balancer determines that an instance is unhealthy, it stops routing requests to that instance"

As Email Alert API uses a classic load balancer it will serve 503 responses if all instances of the app are failing the healthcheck - we found this out when the whole app went down and we were wondering where all the stats were (we anticipated high 5xx rate for instance).

Email Alert Frontend is behind an application load balancer so will switch to serving to all instances if all are unhealthy.

---

The tricky bit around this is trying to establish if there is a single point that can be monitored for web application health. If just one box is broke (which admittedly doesn't seem to have much precedent) it doesn't feel necessary to call some up for one, nor is it much fun for a support engineer to get different calls for every box if they all break together. This all seems to tie into smells that are alerting system is a bit outdated as I imagine there are better approaches to all this that the rest of the world have adopted.

sengi: Oh gosh that's shocking! You're absolutely right though. I hadn't come across that before but it's definitely the case — classic ELB responds `503 Service Unavailable: Back-end server is at capacity` when all its backends are unhealthy.

If I'd known this I would have replaced these with ALBs during the AWS migration 😞

We might want to increase the healthcheck timeout in light of this. It's only 3 seconds at the moment whereas I'm sure the timeout on the app itself is longer than that (the nginx timeout is 15 s, couldn't easily find where the Rails one is set).

kevindew: Yeah, it was a surprise for me too. To be honest I find it makes quite a lot of sense logically that it's genuinely severe when nothing is healthy - whereas you seem to enter this frustratingly ambiguous world when all healthchecks fail means you're serving all instances (compared to say the oddness of 15 of 16 instances failing healthchecks and you only serving from).

I'm assuming though that "classic" is a synonym for legacy so this behaviour is on it's way out.

There tends to not be a Rails timeout, I think when that is done it is implemented in Rack and I've not seen that we've done that anywhere (we also have plenty of evidence of Rails requests cruising to take > 30s). 15s would sound a better number - I have a feeling that icinga may time out at 10s for healthchecks so could be consistent with that. Mostly the healthchecks should be very fast, but there is a chance they could get stalled if one of their dependent services (a db for instance) has some sort of connection issue.

sengi: Yeah we actually saw that kind of timeout scenario for real in integration while working on the AWS migration (must have been a different app though because we didn't find out about the surprising blackhole behaviour of the load balancer!) Classic ELB is the first load balancer I've come across which doesn't have some sort of last-resort mode (it's a common enough concept that it has a name). Oh well, the more you know!

PR #1478, opened 2020-11-20
Rewrite content for subscription auth email

benthorner:
https://trello.com/c/g3cR5QGg/618-update-confirm-subscription-email-content

This is part of the new design for creating a new subscription.

## Before

Screenshot 2020-11-20 at 15 47 21

## After

Screenshot 2020-11-24 at 10 12 22

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,5 +1,10 @@
 en:
   emails:
+    subscription_auth:
+      frequency:
+        daily: You’ll get one email a day
+        weekly: You’ll get one email a week
+        immediate: You’ll get an email each time we add or update a page
     message:
       subject: "Update from GOV.UK – %{title}"
       opening_line: Update on GOV.UK.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -4,7 +4,7 @@ en:
       frequency:
         daily: You’ll get one email a day
         weekly: You’ll get one email a week
-        immediate: You’ll get an email each time we add or update a page
+        immediately: You’ll get an email each time we add or update a page
     message:
       subject: "Update from GOV.UK – %{title}"
       opening_line: Update on GOV.UK.
--- a/spec/integration/subscriptions_auth_token_spec.rb
+++ b/spec/integration/subscriptions_auth_token_spec.rb
@@ -6,8 +6,12 @@ RSpec.describe "Subscriptions auth token", type: :request do
   describe "creating an auth token" do
     let(:path) { "/subscriptions/auth-token" }
     let(:address) { "test@example.com" }
-    let(:topic_id) { "business-tax-corporation-tax" }
     let(:frequency) { "daily" }
+
+    let(:topic_id) do
+      create(:subscriber_list, slug: "business-tax-corporation-tax").slug
+    end
+
     let(:params) do
       {
         address: address,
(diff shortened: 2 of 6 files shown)


benthorner:
Adding "DNM" since we need to get approval for the URL at the end first.

spec/integration/subscriptions_auth_token_spec.rb
```diff
@@ -6,8 +6,12 @@
   describe "creating an auth token" do
     let(:path) { "/subscriptions/auth-token" }
     let(:address) { "test@example.com" }
-    let(:topic_id) { "business-tax-corporation-tax" }
     let(:frequency) { "daily" }
+
+    let(:topic_id) do
+      (create :subscriber_list, slug: "business-tax-corporation-tax").slug
```
1pretz1, line 12: Can we use `create(:subscriber_list, slug: "business-tax-corporation-tax").slug` here as the current syntax doesn't look very familiar?

config/locales/en.yml
```diff
@@ -4,7 +4,7 @@ en:
       frequency:
         daily: You’ll get one email a day
         weekly: You’ll get one email a week
-        immediate: You’ll get an email each time we add or update a page
+        immediately: You’ll get an email each time we add or update a page
```
1pretz1, line 7: Could we move this to the first commit? It's a bit confusing it's here as the previous key would not actually be found when finding this locale for an immediate email.

1pretz1 (approved):
Checked the content and it all looks good, nice work!

Approving as the comments are minor.

config/locales/en.yml
```diff
@@ -4,7 +4,7 @@ en:
       frequency:
         daily: You’ll get one email a day
```
gavinwye, line 5: In the design, there is a full stop after the word `day` does this need to be here too?

config/locales/en.yml
```diff
@@ -4,7 +4,7 @@ en:
...
         weekly: You’ll get one email a week
```
gavinwye, line 6: In the design, there is a full stop after the word `week` does this need to be here too?

config/locales/en.yml
benthorner: That's covered here: https://github.com/alphagov/email-alert-api/pull/1478/files#diff-05a846651f6792b461f7abfa940657266c0d62655a4dbd40d6ee2f2b33527d59R39.

config/locales/en.yml
benthorner: That's covered here: https://github.com/alphagov/email-alert-api/pull/1478/files#diff-05a846651f6792b461f7abfa940657266c0d62655a4dbd40d6ee2f2b33527d59R39.

config/locales/en.yml
benthorner: I'd like to keep it here, as it's the motivation behind the specs in this commit, which will fail without this change. While ideally each commit should be correct in itself, I think it's OK to have other commits to fix things, as long as they tell a clear story. I've updated the commit message to explain better, though.

spec/integration/subscriptions_auth_token_spec.rb
benthorner: Sure, I see where you're coming from - updated 👍.

benthorner:
> Adding "DNM" since we need to get approval for the URL at the end first.

Updated to use the full link as we want to get this out, and expect the upcoming rewrite of the page will change the URL anyway - we'll make a note to revisit this then.

PR #1479, opened 2020-11-20
Updates our analytics doc

1pretz1:
Updates our analytics doc to match our new data [retention ADR].

Removes the information around how to query archived emails in Athena as
this process is now deprecated.

The rake task section has been left alone in this stream of work as
there is a story coming up in the near future to decipher which
analytics tasks are going to be kept/combined. This doc can be updated
once we've made those decisions.

Dependant on:

- [x] https://github.com/alphagov/email-alert-api/pull/1460

Trello:

[retention ADR]: https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-007-retain-data-for-up-to-one-year.md#decision

### DIFF
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -4,26 +4,27 @@ Email Alert API does not currently have many automated means to generate
 analytics and data insights. This document serves as a guide to understand
 how to produce this information.
 
-## How data is organised
+## Available data
 
-Most of the data in this application is stored in the database for perpetuity,
-thus for most cases querying the database through Rails console or through
-PostgreSQL is appropriate. However as this application creates such a large
-volume of emails these are instead archived with minimal data in
-[Amazon S3](https://aws.amazon.com/s3/) and can be queried
-with [Amazon Athena][athena].
+We retain most data the system doesn't use for one year. This is outlined in
+the [data retention strategy] for Email Alert API. We do have a couple of
+exceptions to this: emails are removed 7 days after they are sent and subscriber
+lists that have never had a subscriber are also removed after 7 days.
 
-Once an email has been sent or is no longer being sent (such as a
-permanent failure) the record will be archived in a reduced form and sent to
-Amazon S3 in hourly batches. After 14 days the email will be removed from the
-database. Thus for recent information the `emails` database table can be used
-otherwise the email archive should be queried.
+In addition we [nullify subscriber's email addresses] that have been
+unsubscribed from all subscriptions for more than 28 days. This is carried out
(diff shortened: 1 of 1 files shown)


docs/analytics.md
```diff
@@ -4,27 +4,31 @@ Email Alert API does not currently have many automated means to generate
...
+are deleted after a week. More data isn't retained as we have no evidence of
+needing it.
 
-Once an email has been sent or is no longer being sent (such as a
-permanent failure) the record will be archived in a reduced form and sent to
-Amazon S3 in hourly batches. After 14 days the email will be removed from the
-database. Thus for recent information the `emails` database table can be used
-otherwise the email archive should be queried.
+See [ADR-7] for the most up to date data retention strategy. An overview of our
+data cleanup mechanisms can also be found [here][cleanup mechanisms].
```
kevindew, line 15: there's a few minor issues across these paragraphs that I think could be worked out with some rephrasing:

- unusued data - seems a bit weird,
- pollutes analytics - seems exotic
- up to date - seems time based
- should try avoid using a link on "here" if you can, seems less natural than summarising something with a link to wider info.

I'd try something like

We have established a [data retention strategy][adr-7] for Email Alert API. Emails are stored in the system for 7 days after they are sent. Subscriber lists that have never had a subscriber are also removed after 7 days. When a subscriber no longer has any subscriptions [their personally identifiable information](https://github.com/alphagov/email-alert-api/blob/master/docs/data-cleanup-mechanisms.md#nullifying-email-addresses) is removed after 28 days. Past subscriptions, content changes, messages and digest run information are stored for 1 year.

1pretz1: I've taken inspiration from your suggestion and restructured the paragraphs. 

I wasn't really keen on adding individual models to this description as there is quite a few more than what you've listed and I think it's sufficient to just name out the ones that are different to the one year retention rule, if they really want to dig into the specifics they can look at the ADR.

1pretz1:
Thanks for the review @kevindew. Should be ready for another peek. 👀

kevindew (approved):
Looks good, nice one Pete.

By the way, if I remember right, the link style of `[test]` rather than `[test][]` doesn't work in govuk-developer-docs so you might want to use the latter approach so this works there too. @ChrisBAshton should be able to confirm whether I'm remembering right or not.

1pretz1:
Just checked and the `[something]` syntax renders a link fine without the need to add an empty set of brackets. Maybe this got fixed somewhere down the line @kevindew 😄

PR #1480, opened 2020-11-23
Remove redundant SubscriberList question methods

benthorner:
https://trello.com/c/PaZNXoMS/320-dev-pain-we-have-special-email-signup-pages-for-foreign-travel-advice

These are only used for testing generic functionality of the system,
and do not need to exist.

Original commit: b8e133f13ce530435dcd643df7383561f0757789


kevindew (approved):
🎖️

PR #1482, opened 2020-11-24
Only send pending emails and do so in a lock

kevindew:
Trello: https://trello.com/c/x0vYijvw/459-make-sending-an-email-idempotent

This changes how the SendEmailWorker runs so that it only attempts to send emails that are in a pending state (to avoid risks of resending the same email) and uses a DB lock to isolate the running (to ensure two workers can't be trying to send the same email at the same time).

This then updates the methods to send failed emails to work with the pending constraint.

More info in the commits.

### DIFF
--- a/app/workers/send_email_worker.rb
+++ b/app/workers/send_email_worker.rb
@@ -18,12 +18,13 @@ class SendEmailWorker < ApplicationWorker
       return
     end
 
-    increment_rate_limiter
+    ActiveRecord::Base.transaction do
+      email = Email.lock.find_by(id: email_id, status: :pending)
+      return unless email
 
-    SendEmailService.call(
-      email: Email.find(email_id),
-      metrics: parsed_metrics(metrics),
-    )
+      increment_rate_limiter
+      SendEmailService.call(email: email, metrics: parsed_metrics(metrics))
+    end
   end
 
   def self.perform_async_in_queue(email_id, metrics = {}, queue:)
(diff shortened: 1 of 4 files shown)


1pretz1 (approved):
Changes look good @kevindew!

I think retrying 'failed' emails still could be necessary in certain cases, an example being if Notify was to experience issues for a few hours we would stop retrying emails and mark them as failed after 2h 39m 42s (according to Sidekiq docs for an exponential retry count of 9). Maybe we could increase the retry count, or adjust the frequency in which retries occur so the time window is wider which should mean we never have to re-send a failed email (unless something horrific occurs)? 

I did ponder on the usefulness of the support tasks for resending emails existing in the first place, as they are / were super simple! They're now a bit more complicated so it probably makes sense to keep them around.

kevindew:
> Changes look good @kevindew!
> 
> I think retrying 'failed' emails still could be necessary in certain cases, an example being if Notify was to experience issues for a few hours we would stop retrying emails and mark them as failed after 2h 39m 42s after the first attempt (according to Sidekiq docs for an exponential retry count of 9). Maybe we could increase the retry count, or adjust the frequency in which retries occur manually so the time window is wider which should mean we never have to re-send a failed email (unless something horrific occurs)?
> 
> I did ponder on the usefulness of the support tasks for resending emails existing in the first place, as they are / were super simple! They're now a bit more complicated so it probably makes sense to keep them around.

Yeah - I agree, I think the task existence is a sign that the retry period is too short. Luckily the work from https://github.com/alphagov/email-alert-api/pull/1476 should change them to essentially retry forever which may render these tasks obsolete.

PR #1483, opened 2020-11-26
Recover sending individual emails

1pretz1:
We now recover and retry all emails that have a `pending` state which are an hour old. Historically we ended up with a cruft of `pending` emails that never got sent or removed from the system. All emails now get removed after 7 days so the cruft is limited but it's still true that some emails never get sent which this PR addresses.

See the commits for more info ->

## Prerequisites

- [x] Test this branch on integration by queuing up a large amount of emails to send and monitor the progress. We should check this change doesn't adversely effect performance / doesn't throw up any surprises.

Trello:
https://trello.com/c/iSL6WUwO/460-recover-sending-individual-emails

### DIFF
--- a/app/workers/recover_lost_jobs_worker.rb
+++ b/app/workers/recover_lost_jobs_worker.rb
@@ -2,5 +2,6 @@ class RecoverLostJobsWorker < ApplicationWorker
   def perform
     RecoverLostJobsWorker::UnprocessedCheck.new.call
     RecoverLostJobsWorker::MissingDigestRunsCheck.new.call
+    RecoverLostJobsWorker::OldPendingEmailsCheck.new.call
   end
 end
--- a/app/workers/send_email_worker.rb
+++ b/app/workers/send_email_worker.rb
@@ -4,11 +4,19 @@ class SendEmailWorker < ApplicationWorker
   RATE_LIMIT_THRESHOLD = 21_600 # max requests in a minute, equates to 350 a second
   RATE_LIMIT_INTERVAL = 60
 
-  sidekiq_options retry: 9
-
   sidekiq_retries_exhausted do |msg|
     Email.find_by(id: msg["args"].first, status: :pending)
          &.update!(status: :failed)
+  end
+
+  sidekiq_options lock: :until_executing,
+                  unique_across_queues: true,
+                  unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
+                  on_conflict: :log,
+                  retry_count: 9
+
+  def self.uniqueness_with(args)
+    [args.first]
   end
 
   def perform(email_id, metrics, queue)
--- a/app/workers/send_email_worker.rb
+++ b/app/workers/send_email_worker.rb
@@ -4,11 +4,6 @@ class SendEmailWorker < ApplicationWorker
   RATE_LIMIT_THRESHOLD = 21_600 # max requests in a minute, equates to 350 a second
   RATE_LIMIT_INTERVAL = 60
 
-  sidekiq_retries_exhausted do |msg|
-    Email.find_by(id: msg["args"].first, status: :pending)
-         &.update!(status: :failed)
-  end
-
   sidekiq_options lock: :until_executing,
                   unique_across_queues: true,
                   unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem "plek"
 gem "ratelimit"
 gem "redcarpet"
 gem "sidekiq-scheduler"
+gem "sidekiq-unique-jobs"
 gem "with_advisory_lock"
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -315,6 +315,10 @@ GEM
     sidekiq-statsd (2.1.0)
       activesupport
       sidekiq (>= 3.3.1)
+    sidekiq-unique-jobs (6.0.25)
+      concurrent-ruby (~> 1.0, >= 1.0.5)
+      sidekiq (>= 4.0, < 7.0)
+      thor (>= 0.20, < 2.0)
     simplecov (0.19.0)
       docile (~> 1.1)
       simplecov-html (~> 0.11)
@@ -384,6 +388,7 @@ DEPENDENCIES
   rspec-rails
   rubocop-govuk
   sidekiq-scheduler
+  sidekiq-unique-jobs
   simplecov
   webmock
   with_advisory_lock
--- a/app/workers/recover_lost_jobs_worker/old_pending_emails_check.rb
+++ b/app/workers/recover_lost_jobs_worker/old_pending_emails_check.rb
@@ -1,0 +1,18 @@
+class RecoverLostJobsWorker::OldPendingEmailsCheck
+  def call
+    old_pending_emails = Email.where(status: :pending)
+                              .where("created_at <= ?", 1.hour.ago)
+
+    recover(old_pending_emails)
+  end
+
+private
+
+  def recover(old_pending_emails)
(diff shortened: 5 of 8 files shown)


app/workers/send_email_worker.rb
```diff
@@ -4,13 +4,21 @@ class SendEmailWorker < ApplicationWorker
...
   RATE_LIMIT_INTERVAL = 60
 
-  sidekiq_options retry: 9
-
   sidekiq_retries_exhausted do |msg|
     Email.find_by(id: msg["args"].first, status: :pending)
          &.update!(status: :failed)
   end
 
+  sidekiq_options lock: :until_executed,
```
benthorner, line 12:
```suggestion
  sidekiq_options lock: :until_executing,
```

This was explained in https://github.com/alphagov/email-alert-api/commit/7d7b1b9ff43e0eb8567b2f63c90684c03433b506.

app/workers/send_email_worker.rb
```diff
@@ -4,13 +4,21 @@ class SendEmailWorker < ApplicationWorker
...
+                  unique_across_queues: true,
```
benthorner, line 13: Note to self: [the doc for this option](https://github.com/mhenrixon/sidekiq-unique-jobs#unique_across_queues) is confusing, but [the actual code makes sense](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/fb45ea748732310ae9299dd60820b36b31d8bc18/lib/sidekiq_unique_jobs/lock_digest.rb#L62), and also clarifies that the class is still considered as part of the uniqueness.

app/workers/recover_lost_jobs_worker.rb
```diff
@@ -2,5 +2,18 @@ class RecoverLostJobsWorker < ApplicationWorker
   def perform
     RecoverLostJobsWorker::UnprocessedCheck.new.call
     RecoverLostJobsWorker::MissingDigestRunsCheck.new.call
+    recover_old_pending_emails
```
benthorner, line 5: Could you explain the decision to put the code in the worker?

benthorner (commented):
Nice explanation about the uniqueness across queues - glad that worked out 😌. I think we need to change the locking behaviour, though, as it's redundant and also makes the code more brittle if a worker is killed.

app/workers/send_email_worker.rb
1pretz1: Ah I thought it was a typo in the card as `executed` made more sense to me, didn't realise it came with it's own issues. Corrected.

app/workers/send_email_worker.rb
1pretz1: Yeah the doc isn't clear at all (apparently their documentation is being overhauled in version 7). 

Hopefully the commit message and specs give a better indication of its use. 😄

app/workers/recover_lost_jobs_worker.rb
1pretz1: Sure thing - I initially popped it in the `UnprocessedCheck` class but felt it didn't really sit right as the email recovery code is quite different (batches, different fields compared and different worker call method), my main reasons were:

* The class is called `Unprocessed*` and all recovery inside the class focuses on the `processed_at` field (makes sense)

* Pending email recovery can't share the methods used in all other recovery attempts so it just felt like a bespoke bolt on

As a benefit of doing this we've added a spec for the worker in the process which tests the super simple class calls which is similar to our approach when [collecting metrics](https://github.com/alphagov/email-alert-api/blob/master/spec/workers/metrics_collection_worker_spec.rb).

benthorner: I agree the `UnprocessedCheck` is doing something a bit different to what we need here, and those two points are strong arguments that it's a separate concern, like missing digest runs, which we've encapsulated in a separate class. What I'm confused about is why the code to recover emails is inconsistent with how we've structured the other recovery code. Why not put it in its own class?

Also, I think adding tests to the existing worker class is unrelated to this change, so we could always get the benefit of that separately.

1pretz1: I didn't think it warranted it's own class because the recovery of pending emails is simpler than the recovery code contained within `UnprocessedCheck` (multiple models) and `MissingDigestRunsCheck` (more complicated logic).

If you feel strongly I can create its own class.

benthorner: I do feel strongly, but I think this is more than subjective: while complexity is indeed an argument for what merits a class, consistency is also a strong argument. If I'm reading this code for the first time, then I'm faced with:

- A new concept called "Check", with code in two classes.
- A method, which says it does recovery.

The "Check" concept and the method actually represent the same thing. Conversely, the purpose of the worker is to recovery "everything", but its private methods are specific to one kind of recovery.

If I'm adding a new check in the future, it's also unclear whether I should put it in worker directly, or in its own class, as there's now a precedent for both.

1pretz1: Hmm I see your point. 

I did think about naming the method in respect to the `*Check` classes which may have been easier to parse, e.g. `old_pending_emails_check` but decided against it as it is not really a 'check' as the method is actively recovering work. 

The more I think about the more I wonder the benefit of having these class suffixed with 'Check' as it's a little confusing as they do more than just check something. However, I've added the class as requested so hopefully this should be ready to go!

benthorner: Thanks. Yeah, I see what. you mean with `Check`. We could probably do just lose the suffix, as it was just a subjective choice of how to describe the classes, which are arguably described by their parent class.

benthorner (approved):
Looks good to me 🏥.

Have we tried testing it by loading up Integration? We could try updating a few hours' worth of sent emails to `pending` and see how it copes.

1pretz1:
Thanks @benthorner.

I've deployed the branch to integration yesterday and tested various things including adding emails to the various queues which worked fine but haven't done any substantial testing. Your suggestion seems sensible and we should do that before releasing to prod! 

Adding DNM until this has been done.

app/workers/send_email_worker.rb
```diff
@@ -4,13 +4,21 @@ class SendEmailWorker < ApplicationWorker
...
   sidekiq_retries_exhausted do |msg|
     Email.find_by(id: msg["args"].first, status: :pending)
          &.update!(status: :failed)
   end
 
+  sidekiq_options lock: :until_executing,
+                  unique_across_queues: true,
+                  unique_args: :uniqueness_with, # in upcoming version 7 of sidekiq-unique-jobs, :unique_args is replaced with :lock_args
+                  on_conflict: :log,
+                  retry_count: 9
```
kevindew, line 16: I was thinking that we'd probably want something to automatically release the lock after a certain amount of time - just to really try cut out the possibility that a lock is held without the job being in Sidekiq which would block the future attempt from recovering.

However looking through version 6 of sidekiq-unique-jobs that [doesn't look to be possible](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/4ce41b33f53662671d34c3ccc44aee7642e170bc/redis/lock.lua#L50-L60). Perhaps it is in [version 7](https://github.com/mhenrixon/sidekiq-unique-jobs/blob/master/lib/sidekiq_unique_jobs/lua/lock.lua#L74-L89) - we may want to document that there is an unlikely possibility that jobs can get perpetually stuck.

All looks good besides that

1pretz1: Thanks for the comment Kev!

Hmm your suggestion does make sense and it's a shame it's not available in the current version - I've added a mention in the commit around the concern.

1pretz1:
Thanks for the reviews @benthorner & @kevindew.

I checked that the application could handle large amounts of email recovery and the scenario where lots of duplicate jobs attempt to get recovered. In both scenarios the `SendEmailWorker` processed jobs promptly and no surprises were thrown.

I've also added a commit removing the exhausted retries logic as specified in the card.

benthorner (approved):
Looks good. Nice one testing it out 🏅.

PR #1488, opened 2020-11-30
Revert "Remove archived_at column"

benthorner:
https://trello.com/c/UMGXUk5R/654-retire-archiving-emails

This reverts commit caec57b6b47fcd4ac8d76c2c0d292ffba3c2d845.

Applying this on Integration has lead to a table lock that's lasted
for over 30 minutes. We should investigate how we can make this change
without causing any downtime.

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -1,6 +1,8 @@
 # Any validations added this to this model won't be applied on record
 # creation as this table is populated by the #insert_all bulk method
 class Email < ApplicationRecord
+  self.ignored_columns = %w[archived_at]
+
   COURTESY_EMAIL = "govuk-email-courtesy-copies@digital.cabinet-office.gov.uk".freeze
 
   enum status: { pending: 0, sent: 1, failed: 2 }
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Hmm interesting that this is slow. I thought dropping columns was normally pretty fast. Wonder if there was a deadlock appearing somewhere 🤔

benthorner:
Yeah, that's what I thought as well, and I can't find anything to the contrary. Redeploying also shows it's still running: `Cannot run migrations because another migration process is currently running.`.

I'm going to merge this now to be on the safe side.

kevindew:
Sure, I reckon it's worth trying again tomorrow morning once the DB has been reset. 

It is listed as one of Braintree's safe operations: https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/#:~:text=The%20good - I'm not sure if the index plays a role. Though what seems more likely to me is that there could have been something still running (maybe an archive job) that's using it.

PR #1493, opened 2020-12-01
Revert "Move rate_limiter out of Services module" 

kevindew:
Trello: https://trello.com/c/sa9Qgdz7/661-look-into-sendemailworker-occasionally-taking-ages-to-run

We seem to have picked up an intermittent performance issue due to this
change that we've not been able to diagnose.

We have seen the occasional SendEmailWorker take crazy amounts of time
(around 4 hours). In trying to diagnose this on integration we found it
relatively easy to have a pool of workers take > 10 mins to
SendEmailWorker. We found that this 10 minutes was spent incrementing
the RateLimiter. Reverting this commit resolved this problem.

I'm not sure where the time gets lost in this, certainly not for hours.
I'm assuming there is some sort of deadlock that occurs which is
resolved by a timeout. This look a total pain to debug and we don't
have much time left as a team so I'm taking the rather ignorant route of
reverting without understanding exactly why the problem occurs.

### DIFF
--- a/app/workers/send_email_worker.rb
+++ b/app/workers/send_email_worker.rb
@@ -35,9 +35,9 @@ class SendEmailWorker < ApplicationWorker
 private
 
   def rate_limit_exceeded?
-    rate_limiter.exceeded?("requests",
-                           threshold: RATE_LIMIT_THRESHOLD,
-                           interval: RATE_LIMIT_INTERVAL)
+    Services.rate_limiter.exceeded?("requests",
+                                    threshold: RATE_LIMIT_THRESHOLD,
+                                    interval: RATE_LIMIT_INTERVAL)
   end
 
   # Sidekiq uses JSON for a workers arguments, so richer objects are not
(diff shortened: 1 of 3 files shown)


app/workers/send_email_worker.rb
```diff
@@ -54,8 +54,6 @@ def increment_rate_limiter
   end
 
   def rate_limiter
-    @rate_limiter ||= Sidekiq.redis do |redis|
-      Ratelimit.new("send_email", redis: redis)
```
benthorner, line 58: The main difference here appears to be that the `Services` version is memoised at the class level. Could we try that and see if it fixes things?

benthorner (approved):
Oh well, at least it was easy to revert. I think we've still got time to do a bit of investigation, but this can also just go straight in 👍.

app/workers/send_email_worker.rb
kevindew: Yeah, I think that would have the same effect. It seems a bit of a smell though as it still has the same thread safety issues as Services and the oddness of using a distinct version of Redis to what Sidekiq uses.

kevindew: Yup, good idea.

PR #1494, opened 2020-12-02
Remove code and tasks for load testing

kevindew:
Trello: https://trello.com/c/G35ncFfw/645-simplify-configuration-for-email-sending

There aren't currently plans for future work on this application and
thus not a foreseeable need for load testing. This code is therefore
being removed from the repo.

This code will be linked to from our documentation so that this can be
resurrected should the need to load test return.


benthorner (approved):
💥

PR #1496, opened 2020-12-02
Add ADR about recovering lost work

benthorner:
https://trello.com/c/pslgihlC/659-write-an-adr-for-automatic-recovery-of-lost-sidekiq-jobs

### DIFF
--- a/docs/adr/adr-009-sidekiq-lost-job-recovery.md
+++ b/docs/adr/adr-009-sidekiq-lost-job-recovery.md
@@ -1,0 +1,82 @@
+# 9. Sidekiq Lost Job Recovery
+
+Date: 2020-12-02
+
+## Context
+
+We use [Sidekiq](https://github.com/mperham/sidekiq) to do background processing in Email Alert API. This includes critical tasks, such as generating and sending email. However, it's possible for these jobs to be lost:
+
+1. If the code to enqueue the job in Redis fails, either for transient reasons, or due to a bug.
+
+2. If a worker process quits ungracefully, while processing a job.
+
+The second issue has occurred frequently due to workers being [forcibly restarted][icinga-memory] when they consume too much memory. A memorable example was the loss of a Travel Advice content change, which was detected by the [alert][travel-advice-alert] for travel advice emails, and lead to an incident. Conversely, we don't think the first issue does not occur frequently.
+
+It's an effortful task to recover lost jobs, due to the complexity of how we generate and send email in the system. The job will no longer exist in Redis, but a record of it will exist in the database, with a flag to indicate its completion. For example, we can check the [`status` field][email-status] of an Email to see if the [job][send-email-worker] to send it has completed. By analysing these and similar flags and timestamps in other tables, we can work out which work has not been completed in good time.
+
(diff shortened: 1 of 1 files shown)


docs/adr/adr-009-sidekiq-lost-job-recovery.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Context
+
+We use [Sidekiq](https://github.com/mperham/sidekiq) to do background processing in Email Alert API. This includes critical tasks, such as generating and sending email. However, it's possible for these jobs to be lost:
+
+1. If the code to enqueue the job in Redis fails, either for transient reasons, or due to a bug.
+
+2. If a worker process quits ungracefully, while processing a job.
+
+The second issue has occurred frequently due to workers being [forcibly restarted][icinga-memory] when they consume too much memory. A memorable example was the loss of a Travel Advice content change, which was detected by the [alert][travel-advice-alert] for travel advice emails, and lead to an incident. Conversely, we don't think the first issue does not occur frequently.
```
kevindew, line 13: You've got a double negative in the last sentence here.

docs/adr/adr-009-sidekiq-lost-job-recovery.md
```diff
@@ -0,0 +1,82 @@
...
+
+It's an effortful task to recover lost jobs, due to the complexity of how we generate and send email in the system. The job will no longer exist in Redis, but a record of it will exist in the database, with a flag to indicate its completion. For example, we can check the [`status` field][email-status] of an Email to see if the [job][send-email-worker] to send it has completed. By analysing these and similar flags and timestamps in other tables, we can work out which work has not been completed in good time.
```
kevindew, line 15: I found this paragraph a bit jarring. I think it's because I read "lost jobs" in the generic whereas it jumps quickly to email specific. I've got a couple of suggestions but it might be better to have a fresh attempt at writing it as I think it veers a bit from it's opening sentence and doesn't really capture why it's an effortful task.

Perhaps if this is aimed at the specific it could start with "It's an effortful task to recover lost SendEmailWorker jobs"

I also got a bit stuck around "The job will no longer exist in Redis, but a record of it will exist in the database, with a flag to indicate its completion" which may benefit from being more verbose, for example: "We can tell, due to flags on database records, whether an item is a job that should be completed but it is unclear whether the job is in Redis".

kevindew (commented):
Sounds good. Nice one putting this together so quickly. There's just a couple of bits I'd suggest modifying.

docs/adr/adr-009-sidekiq-lost-job-recovery.md
benthorner: Thanks for the discussion around this @kevindew, which helped me articulate what a rewrite of this paragraph could look like, and how it does improve the ADR. I've pushed a commit with the new wording: https://github.com/alphagov/email-alert-api/pull/1496/commits/053469603be30dcee91562df4427cac34d393c48.

Could you take a look?

kevindew (approved):
Nice one, looks good to me

PR #1499, opened 2020-12-04
Remove link to feedback survey in emails

benthorner:
https://trello.com/c/7EUikumz/665-close-smart-survey-which-we-link-to-in-emails-we-send

This is being removed since the Notifications team is being paused,
and there is no one to look at the feedback collected. While it may
normally be useful to continuously gather feedback, we also know that
the questions in the current survey are flawed i.e. there's no point
someone filling it in, because the data isn't useful.

Note that, if a user unsubscribes after an individual-type email is
generated, they will see a pointless separator at the bottom of the
email, due to the way the builder for this email is structure. This
doesn't like a big issue, and we're intending to rewrite the template
for them anyway, so I've left it that way for now.

### DIFF
--- a/app/builders/content_change_email_builder.rb
+++ b/app/builders/content_change_email_builder.rb
@@ -64,14 +64,12 @@ private
   end
 
   def footer(subscriptions, address)
-    return feedback_link if subscriptions.empty?
+    return "" if subscriptions.empty?
 
     <<~BODY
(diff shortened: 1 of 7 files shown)


benthorner:
Note about how to close the survey:

- Login to Smart Survey
- Switch to the "User Research" user
- Find the survey in the list for that user

kevindew (approved):

PR #1501, opened 2020-12-07
Stop logging every email (fills up the disk!)

benthorner:
Previously we emulated the behaviour of the former PseudoProvider
    by writing pseudo-sent email contents to a log file. However, this
    never used to happen in Integration/Staging, since we actually used
    the NotifyProvider with an allowed list [1]. This removes some of
    the logging, since the sheer volume of logs has actually filled
    one of the disks in Integration.

    [1]: https://github.com/alphagov/email-alert-api/pull/1489

### DIFF
--- a/app/services/send_email_service/send_pseudo_email.rb
+++ b/app/services/send_email_service/send_pseudo_email.rb
@@ -11,8 +11,6 @@ class SendEmailService::SendPseudoEmail
     Rails.logger.info <<~INFO
       Logging email (#{email.id}) we'd have attempted to send to #{email.address}
       Subject: #{email.subject}
-      Body:
-      #{email.body}
     INFO
 
     Metrics.sent_to_pseudo_successfully


kevindew (commented):
It'd be good to have less logging rather than none, otherwise these things just say they're sent without any clues. I'd suggest just removing the body field. That should massively reduce the size of every log.

benthorner:
I thought about that, and while removing the body would reduce the size a lot, it's still a lot of logging that I don't think anyone knows about - you'd have to know about the notify vs. pseudo quirk in order to go looking for them. But then, if you know about the quirk, you probably don't need to look at the logs.

Can you think of a scenario when someone would go looking for these logs?

kevindew:
> I thought about that, and while removing the body would reduce the size a lot, it's still a lot of logging that I don't think anyone knows about - you'd have to know about the notify vs. pseudo quirk in order to go looking for them. But then, if you know about the quirk, you probably don't need to look at the logs.
> 
> Can you think of a scenario when someone would go looking for these logs?

Yeah, I think removing body would be about a 90% reduction in size (quick at an item had 1741 bytes with body 170 without). I'd see the use cases being that you're wondering what the app is doing, particularly that it's not sending an email it claims is sent, and when you look at the logs it tells you it didn't send it. Otherwise, in the absence of errors, there are no clues that a run of SendEmailWorker didn't send an email. This is from the perspective that I don't consider the log file to be something particularly precious. As for someone going looking, the docs tell them that this behaves this way.

benthorner:
> Otherwise, in the absence of errors, there are no clues that a run of SendEmailWorker didn't send an email.

This is what I'm wondering. Would that actually happen in practice? We agree the docs tell them it behaves this way, and I'm not sure how someone would find these logs amongst all the others, if they didn't already know they were there.

I don't mind keeping some logging. I just want to clarify: do you think someone would really come across these logs?

kevindew:
> > Otherwise, in the absence of errors, there are no clues that a run of SendEmailWorker didn't send an email.
> 
> This is what I'm wondering. Would that actually happen in practice? We agree the docs tell them it behaves this way, and I'm not sure how someone would find these logs amongst all the others, if they didn't already know they were there.
> 
> I don't mind keeping some logging. I just want to clarify: do you think someone would really come across these logs?

I'd have thought the sidekiq log would be the first thing you looked at it if you were investigating why something did or didn't behave in such a way? Maybe I'm missing part of your point?

benthorner:
@kevindew not me, I'm afraid. I find Kibana so unusable and the variety of log files we have (on multiple machines, too) so great that I often try other things before checking logs, so I'd probably see the documentation first. However, as long as we think someone could benefit from it, I don't mind leaving it in. I've updated the PR, if you could take another look.

kevindew:
> @kevindew not me, I'm afraid. I find Kibana so unusable and the variety of log files we have (on multiple machines, too) so great that I often try other things before checking logs, so I'd probably see the documentation first. However, as long as we think someone could benefit from it, I don't mind leaving it in. I've updated the PR, if you could take another look.

Cool thanks - ah that surprises em, but do agree the logging situation is a mess but I find myself in Kibana a lot.

kevindew (approved):

PR #1502, opened 2020-12-08
Remove passthrough for "redirect" in auth token

benthorner:
https://trello.com/c/sgJIkObA/662-authenticate-before-unsubscribing-from-digest-emails

Previously the subscriber auth API accepted a "redirect" parameter,
which Email Alert Frontend could then use to forward the subscriber
to a particular page, after authenticating at the "destination" [1].
In practice this is not used, and makes it harder to understand the
sign in process (just because it's not used). This removes support
for the parameter, noting it's never been sent by the frontend [2].

[1]: https://github.com/alphagov/email-alert-api/pull/560/commits/b609e2464fe4734675931492a220d202fcc736fd
[2]: https://github.com/alphagov/email-alert-frontend/commit/ceac5ac17412f6afaac41642767d9f898e8d4d74#diff-6966e9ac443db47e60dc97c2beb2864ce79ab9a4ce5e477d5280232620d4baedR19-R22

### DIFF
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -15,10 +15,7 @@ class SubscribersAuthTokenController < ApplicationController
 private
 
   def generate_token(subscriber)
-    AuthTokenGeneratorService.call(
-      "subscriber_id" => subscriber.id,
-      "redirect" => expected_params[:redirect],
-    )
+    AuthTokenGeneratorService.call("subscriber_id" => subscriber.id)
   end
 
   def build_email(subscriber, token)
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Looks good.

In the long running chain of a distributed system it looks like we've also got https://github.com/alphagov/gds-api-adapters/blob/8c5224f9c331cf6a22d6065ef7f5de9fcbbbed20/lib/gds_api/email_alert_api.rb#L231-L238 to snipe.

benthorner:
Oh yes, thanks for spotting. I'll try and finish off this tangent with a PR on that 😸.

PR #1503, opened 2020-12-08
Reinstate confirmation email for subscriptions

benthorner:

Previously we removed this because it adds no value for a standard
GOV.UK subscription, except in the case where the list description
was present i.e. a link to Brexit Checker results. This reinstates
the confirmation email, but only when it's useful.

Note that this also removes the previous conditional around whether
an existing subscription was overridden, since this has no bearing
on the expectations of the user, and helps simplify the code.


hannako (approved):
LGTM 👍

PR #1506, opened 2020-12-09
Simplify and enhance POST /subscriptions endpoint

benthorner:

https://trello.com/c/Z8tTO9pr/368-dev-pain-the-subscription-create-endpoint-does-basically-everything-that-the-update-endpoint-does

https://trello.com/c/hdOrNhuC/669-sign-in-after-subscribing

This PR makes a number of changes:

- DRYs up the code between the "create" and "update" actions
for subscriptions.

- DRYs up and simplifies the integration tests for the actions,
which were really confusing.

- Starts returning some extra data for the "create" action, to
support subsequent design changes.

- Does a minor update of the (out-of-date) API doc.

This PR is best reviewed commit-by-commit.


app/services/create_subscription_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      subscription = Subscription.find_by(
+        subscriber_list: subscriber_list,
+        subscriber: subscriber,
+      )
+
+      if subscription
+        return subscription if subscription.frequency == frequency
+
+        subscription.end(reason: :frequency_changed)
```
sihugh, line 23: this seems nicer than before 👍

app/services/create_subscription_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    ApplicationRecord.transaction do
+      subscriber.lock!
+
+      subscription = Subscription.find_by(
+        subscriber_list: subscriber_list,
+        subscriber: subscriber,
+      )
+
+      if subscription
+        return subscription if subscription.frequency == frequency
```
sihugh, line 21: it would be fine to maintain the 200/201 api interface by returning something like `[:found, subscription]`/`[:created, subscription]` from this service (or some other approach). Would we want to use the knowledge of create vs update at some point?

spec/integration/update_subscription_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    context "without an existing subscription" do
+      it "returns a 404" do
+        patch "/subscriptions/xxxx", params: { frequency: "daily" }
+        expect(response.status).to eq(404)
+      end
+    end
+  end
+
+  context "without authentication" do
```
sihugh, line 36: nice additions 👉 👇 . I agree that splitting these out makes things like this easier to spot if they're missing

spec/integration/create_subscription_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+      it "sends a confirmation email" do
+        stub_notify
+        create_subscription
+        expect(a_request(:post, /notifications/)).to have_been_made.at_least_once
+      end
+    end
+
+    context "with a parameter to skip sending the confirmation email" do
+      def create_subscription
```
sihugh, line 62: I wondered whether this should be named differently from the other `create_subscription`? It applies to quite a small block at the moment though, so I'm not too worried.

app/controllers/subscriptions_controller.rb
```diff
@@ -17,7 +17,7 @@ def create
       SendEmailWorker.perform_async_in_queue(email.id, queue: :send_email_transactional)
     end
 
-    render json: { id: subscription.id }
+    render json: { id: subscription.id, subscription: subscription }
```
sihugh, line 20: Are we just going to be using the frequency of the subscription? I wonder whether we're being too leaky by rendering the entire subscription object rather than just picking the bits that we want? It might make it hard for us to refactor at a later date if we need to.

sihugh (commented):
A couple of philosophical questions, but the code looks correct for what you're trying to achieve :-)

app/services/create_subscription_service.rb
benthorner: Yeah, I did wonder about returning some kind of flag - we could also return the previous subscription (if any), so the service doesn't need to know how it's used.

I decided to go the other way and question: why does it matter, as long as we do what was requested? If we do want to use this knowledge in future, then we could always add the differing response codes back in. However, we generally mask these kinds of nuances from the user - in this case, we always say "[you've subscribed](https://github.com/alphagov/email-alert-frontend/blob/master/app/controllers/subscription_authentication_controller.rb#L28)", irrespective of how many times they click the link to confirm their subscription.

spec/integration/create_subscription_spec.rb
benthorner: Oh, how did it read for you? I deliberately named it the same to indicate it should be used in preference to the outer one, but looking again I also see a bit of duplication going on.

I've added another commit with a different approach here - could you take a look?

app/controllers/subscriptions_controller.rb
benthorner: The plan is actually to use the subscriber ID, which is embedded in this object (due to the way `as_json` recurses through ActiveRecord associations). While it is more data than we need, it's:

- Consistent with the way we respond to [an update](https://github.com/alphagov/email-alert-api/blob/master/app/controllers/subscriptions_controller.rb#L78).

- Being selective here in the past is making it harder to change, since we didn't use a nested hash.

A compromise might be something like (after we remove the redundant, top-level ID key):

```
render json: { subscription: { id: <id>, subscriber: { id: <subscriber_id> } } }
```

Since we're already returning the whole thing on the "update" action, though, I reckon this approach is more surprising than just being consistent with the code below it.

What do you think?

app/services/create_subscription_service.rb
sihugh: Cool, I guess if we've taken that as a firm decision, then this seems fine 👍

spec/integration/create_subscription_spec.rb
```diff
@@ -59,18 +65,14 @@ def create_subscription
...
       it "does not send a confirmation email" do
         stub_notify
-        create_subscription
+        create_subscription(extra_params: { skip_confirmation_email: true })
         expect(a_request(:post, /notifications/)).to_not have_been_made
       end
 
       it "returns status code 200" do
-        create_subscription
+        create_subscription(extra_params: { skip_confirmation_email: true })
```
sihugh, line 75: Yep, I prefer this way I think. It's immediately more obvious what the difference is.

spec/integration/create_subscription_spec.rb
sihugh: Yep - I think I prefer the way you went in the new commit

app/controllers/subscriptions_controller.rb
sihugh: Cool, thanks for the explanation. I might feel different if this was a public API, but it isn't - we control all consumers of it. Being consistent is probably more important then, I reckon, and it's a good point about our previous choices making it annoying.

Happy to go with what we've got 👍

sihugh (approved):

app/controllers/subscriptions_controller.rb
benthorner: Oh, yes that's a good point. I guess you've got quite a different perspective on it from e.g. your work on search (API). I'll bear that in mind when I'm working on APIs in the future - thanks!

PR #1509, opened 2020-12-10
Replace courtesy copies with a real subscription

benthorner:

This functionality is redundant, as we can achieve a more
realistic outcome by subscribing the current Google Group to
the "all-government-publishing-updates" list.

Other things to do:

- Actually subscribe the group to the list.
- Update documentation about the group.
- Consider renaming the group to match Integration / Staging.

### DIFF
--- a/app/builders/content_change_email_builder.rb
+++ b/app/builders/content_change_email_builder.rb
@@ -27,7 +27,7 @@ private
         {
           address: address,
           subject: subject(recipient_and_content.fetch(:content_change)),
-          body: body(recipient_and_content.fetch(:content_change), recipient_and_content.fetch(:subscriptions), address),
+          body: body(recipient_and_content.fetch(:content_change), recipient_and_content.fetch(:subscriptions).first, address),
           subscriber_id: recipient_and_content.fetch(:subscriber_id),
           created_at: now,
           updated_at: now,
@@ -40,21 +40,21 @@ private
     I18n.t!("emails.content_change.subject", title: content_change.title)
   end
 
-  def body(content_change, subscriptions, address)
+  def body(content_change, subscription, address)
     <<~BODY
       #{I18n.t!('emails.content_change.opening_line')}
 
       ---
-      #{presented_content_change(content_change, subscriptions)}
+      #{presented_content_change(content_change, subscription)}
       ---
-      #{footer(subscriptions, address).strip}
+      #{footer(subscription, address).strip}
     BODY
   end
 
-  def presented_content_change(content_change, subscriptions)
+  def presented_content_change(content_change, subscription)
     copy = ContentChangePresenter.call(content_change)
 
-    subscriber_list = subscriptions.first.subscriber_list
+    subscriber_list = subscription.subscriber_list
     if subscriber_list.description.present?
       copy += "\n#{subscriber_list.description}\n"
     end
@@ -62,9 +62,9 @@ private
     copy
   end
 
-  def footer(subscriptions, address)
+  def footer(subscription, address)
     <<~BODY
-      #{permission_reminder(subscriptions.first.subscriber_list)}
+      #{permission_reminder(subscription.subscriber_list)}
 
       #{ManageSubscriptionsLinkPresenter.call(address)}
     BODY
--- a/app/builders/content_change_email_builder.rb
+++ b/app/builders/content_change_email_builder.rb
@@ -53,7 +53,6 @@ private
 
   def presented_content_change(content_change, subscriptions)
     copy = ContentChangePresenter.call(content_change)
-    return copy if subscriptions.empty?
 
     subscriber_list = subscriptions.first.subscriber_list
     if subscriber_list.description.present?
@@ -64,8 +63,6 @@ private
   end
 
   def footer(subscriptions, address)
-    return "" if subscriptions.empty?
-
     <<~BODY
       #{permission_reminder(subscriptions.first.subscriber_list)}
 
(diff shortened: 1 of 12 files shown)


app/builders/content_change_email_builder.rb
```diff
@@ -53,7 +53,6 @@ def body(content_change, subscriptions, address)
 
   def presented_content_change(content_change, subscriptions)
     copy = ContentChangePresenter.call(content_change)
-    return copy if subscriptions.empty?
 
     subscriber_list = subscriptions.first.subscriber_list
```
kevindew: Are we confident there isn't any risk that this is lazily loaded and could end up with an empty array? I can't recall the load ordering but I wondered if there's a rare risk that someone has unsubscribed at the point of generating the email and we end loading up no active, immediate subscriptions. Hopefully not.

kevindew (approved):
This looks good. It's nice to see that ❄️  code go.

Just a query about an area where I wondered if this could break. If not looks good to merge. 

Looks great that you're planning to update the documentation, it's probably worth including in there how you set up the subscription. Just incase we ever hit the situation where someone accidentally clicks a fancy one-click unsubscribe link 😆 .

I can see you also mentioned about integration and staging. I'm not sure if you already have a plan there but a word of warning from my thoughts on the matter before. Without having any set of environment variable that does an override it's a real pain to get emails to go to integration or staging consistently due to the data sync. For example, if you wanted every content change to get a courtesy copy you'd need production to be signed up to the same list because of the duplicate database (and then every email to that address would fail in prod). Integration is different too due to the email address anonymising.

benthorner:
Thanks @kevindew. It's surprising this is so easy to remove!

> it's probably worth including in there how you set up the subscription

Sure, I've made a note of that.

> I can see you also mentioned about integration and staging.

I assume you're referring to "Consider renaming the group to match Integration / Staging". To be clear, this was purely about renaming "govuk-email-courtesy-copies@" to "email-alert-api-production@" for consistency. I'm aware that:

- Integration will stop sending courtesy copies, as the new, replacement subscription/address will have been anonymised.
- Staging will stop sending courtesy copies (in any case), as it's not in the allowed list of recipients.

I didn't mention that in the commit/PR because I'm not sure anyone was previously aware we also got courtesy copies from these environments, but at least we have a record of stopping it here.

On reflection, I don't think it's worth renaming the (now production-only) courtesy copies group, since it still functions in the same way, and that function is somewhat different to the "test" groups for integration / staging.

Does that clarify the next steps for you?

benthorner:
Tested and works on Integration ✅.

thomasleese:
> * Integration will stop sending courtesy copies, as the new, replacement subscription/address will have been anonymised.
> * Staging will stop sending courtesy copies (in any case), as it's not in the allowed list of recipients.
> 
> I didn't mention that in the commit/PR because I'm not sure anyone was previously aware we also got courtesy copies from these environments, but at least we have a record of stopping it here.

I don't feel too strongly (because I can work around it), but in the past I've definitely found this useful when testing that emails are sending correctly in integration/staging without needing to sign up to something. An example of this would be when deploying Travel Advice Publisher and testing that emails are still being sent out.

kevindew:
> Thanks @kevindew. It's surprising this is so easy to remove!
> 
> > it's probably worth including in there how you set up the subscription
> 
> Sure, I've made a note of that.
> 
> > I can see you also mentioned about integration and staging.
> 
> I assume you're referring to "Consider renaming the group to match Integration / Staging". To be clear, this was purely about renaming "govuk-email-courtesy-copies@" to "email-alert-api-production@" for consistency. I'm aware that:
> 
> * Integration will stop sending courtesy copies, as the new, replacement subscription/address will have been anonymised.
> * Staging will stop sending courtesy copies (in any case), as it's not in the allowed list of recipients.
> 
> I didn't mention that in the commit/PR because I'm not sure anyone was previously aware we also got courtesy copies from these environments, but at least we have a record of stopping it here.
> 
> On reflection, I don't think it's worth renaming the (now production-only) courtesy copies group, since it still functions in the same way, and that function is somewhat different to the "test" groups for integration / staging.
> 
> Does that clarify the next steps for you?

Ah gotcha, I interpreted your note in a very different way. Yes, I agree to not renaming the courtesy copy group.

> I don't feel too strongly (because I can work around it), but in the past I've definitely found this useful when testing that emails are sending correctly in integration/staging without needing to sign up to something. An example of this would be when deploying Travel Advice Publisher and testing that emails are still being sent out.

Unfortunately this is a feature that is already dead regardless of Ben's changes. It died with https://github.com/alphagov/email-alert-api/pull/1489 - documentation is here: https://docs.publishing.service.gov.uk/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html. I'm still to push this to a wider audience by pinging #govuk-developers.

thomasleese:
> > I don't feel too strongly (because I can work around it), but in the past I've definitely found this useful when testing that emails are sending correctly in integration/staging without needing to sign up to something. An example of this would be when deploying Travel Advice Publisher and testing that emails are still being sent out.
> 
> Unfortunately this is a feature that is already dead regardless of Ben's changes. It died with #1489 - documentation is here: https://docs.publishing.service.gov.uk/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html. I'm still to push this to a wider audience by pinging #govuk-developers.

I'm clearly behind the times! Makes sense, that solution looks a lot neater. 👍

This courtesy copy code was always a bit of a weird special case anyway so it's good that the same codepaths are executed now.

benthorner:
Sounds like this is good to merge, so here we go...

PR #1513, opened 2020-12-14
Update individual-type emails to match design

benthorner:

This includes the introduction of a one-click unsubscribe
link. Note that GOV.UK Notify only support a subset of
Markdown, which means we can't have the bold/strong style
for the subscriber list title in the footer.

## Before (in Google Groups)

Screenshot 2020-12-14 at 11 27 10
Screenshot 2020-12-14 at 11 27 26

## After (personal inbox)

(See https://github.com/alphagov/email-alert-api/pull/1513#issuecomment-745161732 for an explanation of using a personal inbox here.)

Screenshot 2020-12-15 at 12 29 08
Screenshot 2020-12-15 at 12 29 21


benthorner:
Note there were some issues previewing this PR with Google Groups: the Groups UI and its conversation view seems to cause issues with the layout of the emails, as the following screenshots show. Since this appears to be limited to Google Groups, which does not reflect the majority of subscribers, I don't think it's any cause for concern.

## Before (in Google Groups)

Screenshot 2020-12-14 at 15 25 02

## After (in Google Groups)

Screenshot 2020-12-14 at 15 24 52

benthorner:
Code to generate the sample emails:

```
s = Subscriber.find_by_address("email-alert-api-integration@digital.cabinet-office.gov.uk")

SendEmailWorker.new.perform(ImmediateEmailBuilder.call([{ address: s.address, content: ContentChange.last, subscriptions: [s.subscriptions.last], subscriber_id: s.id }]).first, {}, "default")

ENV["GOVUK_NOTIFY_RECIPIENTS"] = "email-alert-api-integration@digital.cabinet-office.gov.uk,ben.thorner@digital.cabinet-office.gov.uk"
```

spec/lib/public_urls_spec.rb
```diff
@@ -7,7 +7,7 @@
   end
 
   describe ".subscription_url" do
-    it "returns the GOV.UK for the new subscription page" do
+    it "returns the GOV.UK url for the new subscription page" do
```
kevindew, line 10: Nice spot 👍

app/builders/content_change_email_builder.rb
```diff
@@ -41,42 +41,34 @@ def subject(content_change)
...
 
-  def permission_reminder(subscriber_list)
-    topic = if subscriber_list.url
-              "[#{subscriber_list.title}](#{Plek.new.website_root}#{subscriber_list.url})"
-            else
-              subscriber_list.title
-            end
-
-    I18n.t!("emails.content_change.permission_reminder", topic: topic)
+  def middle_section(list, content_change)
```
kevindew: Minor, I found this method name a bit surprising as it didn't feel very descriptive. I can see problems with the previous `presented_content_change` name too though and don't have any better suggestions so don't consider this a blocking comment.

app/builders/immediate_email_builder.rb
```diff
@@ -1,4 +1,4 @@
-class ContentChangeEmailBuilder
+class ImmediateEmailBuilder
```
kevindew, line 1: Nice job with this refactoring.

kevindew (approved):
Looks great, nice job. It is super to see Gavin's work become concrete.

app/builders/content_change_email_builder.rb
benthorner: Yeah, it was annoying to have to add it, due to the optional description field. I can't think of a significantly clearer name either, so I'll leave it as-is.

benthorner:
@kevindew while this is merged, I was also contemplating another change to these builders: removing the use of I18n. The use of it seems quite arbitrary, since:

- We only send emails in English.
- We don't use it [uniformly](https://github.com/alphagov/email-alert-api/blob/master/app/presenters/manage_subscriptions_link_presenter.rb#L11).
- We have to invent keys for each line.
- Most of the keys are only used once.
- We don't benefit from it in the [tests](https://github.com/alphagov/email-alert-api/blob/2a6207f2d36df1806881d49aef1c9662f5fbb1bf/spec/builders/immediate_email_builder_spec.rb#L49-L69).

What do you think?

kevindew:
Sure, that sounds reasonable to me. I can't recall an occasion where we've benefited from the abstraction.

PR #1514, opened 2020-12-15
Modify rake task support:view_emails to support users who haven't yet subscribed

bilbof:
This task will help when investigating cases where a user reports that they have not received an invite email / any other emails, since they may not be a subscriber when we investigate.

### DIFF
--- a/lib/tasks/support.rake
+++ b/lib/tasks/support.rake
@@ -23,17 +23,13 @@ namespace :support do
   task :view_emails, %i[email_address limit] => :environment do |_t, args|
     email_address = args[:email_address]
     limit = args[:limit].to_i || 10
-    subscriber = Subscriber.find_by_address(email_address)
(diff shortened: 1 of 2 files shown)


benthorner:
Agree with reporting emails irrespective of having a subscriber. Looking at the existing task, could we adapt it to achieve the same thing? The only [part](https://github.com/alphagov/email-alert-api/blob/1fc05e45d9c18a7690acca96211497a64f10f0e9/lib/tasks/support.rake#L41) that appears to be specific to subscriptions is already agnostic to whether there are any. Could we take your approach to fetching emails and use it in that task instead?

bilbof:
Thanks @benthorner, I've made that change.

benthorner (approved):
Much simpler 🥛.

Minor: this would be a good time to add [a basic test](https://github.com/alphagov/email-alert-api/blob/master/spec/lib/tasks/support_spec.rb) to make sure the task runs OK; we added them here and there to make sure the tasks can at least be run.

PR #1516, opened 2020-12-15
DRY-up interface for builders / presenters

benthorner:

This introduces a new parent class to DRY-up the interface for
Builder and Presenter classes, and make them consistent.

While it's unusual to have Presenter classes with a single method,
this is hopefully still a net improvement to the current code.

### DIFF
--- a/app/builders/application_builder.rb
+++ b/app/builders/application_builder.rb
@@ -1,0 +1,7 @@
+class ApplicationBuilder
+  def self.call(*args, **kwargs)
+    new(*args, **kwargs).call
+  end
+
+  private_class_method :new
+end
--- a/app/builders/bulk_subscriber_list_email_builder.rb
+++ b/app/builders/bulk_subscriber_list_email_builder.rb
@@ -1,19 +1,13 @@
-class BulkSubscriberListEmailBuilder
+class BulkSubscriberListEmailBuilder < ApplicationBuilder
   def initialize(subject:, body:, subscriber_lists:)
     @subject = subject
     @body = body
     @subscriber_lists = subscriber_lists
   end
 
-  def self.call(**args)
-    new(**args).call
-  end
-
   def call
     records.any? ? Email.insert_all!(records).pluck("id") : []
   end
-
-  private_class_method :new
 
 private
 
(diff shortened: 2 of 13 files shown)


app/builders/application_builder.rb
```diff
@@ -0,0 +1,7 @@
+class ApplicationBuilder
+  def self.call(*args, **kwargs)
+    new(*args, **kwargs).call
```
hannako, line 3: I've read some documentation but I'm still not sure I'm super clear on `*args`, `**args` and `**kwargs`

Prior to your refactor, `BulkSubscriberListEmailBuilder` was expecting `**args`, but it is passed keyword arguments, so I would have expected its `.call` method to have been 
` def self.call(**kwargs)` instead. Your new `ApplicationBuilder` does `new(*args, **kwargs).call` so is it fine that it doesn't have a `**args`?

benthorner: Ah, yes, the whole `**` v `*` thing can be confusing. Really we shouldn't need `**` here at all, but I didn't want to start modifying the constructors of all the classes. Here's a summary:

- `*anyvarname` is a way to collect any arguments passed to the method
- `*anyvarname` is a way to collect any keyword arguments (as in `key: value`) passed to the method
- &block is a way to collect any block passed to the method (not used here)

First thing to note is the naming is irrelevant. Second thing to note, by way of explanation, is that using all three of the above is one way to alias one method to another, like this:

```
# convoluted alias technique
def my_alias(*args, **kwargs, &block)
  real_method(*args, **kwargs, block)
end

# better alias technique
alias_method :my_alias, :real_method
```

We're not really doing that here, but hopefully that makes it clearer what's going on?

Might also help to say this is the same technique as we use elsewhere e.g. here: https://github.com/alphagov/content-publisher/blob/master/app/services/application_service.rb.

hannako: > First thing to note is the naming is irrelevant

OK that makes it all much clearer! 

> Really we shouldn't need ** here at all, but I didn't want to start modifying the constructors of all the classes.

Yes makes sense. Thanks.

hannako (approved):
👍 Thanks for the explanation :-)

PR #1519, opened 2020-12-16
Prepare to delete unused group_id field

benthorner:

This was added in expectation of making it easier to group lists for
analytics [1], but this has not been the case in practice [2]. Since
the field is currently only used for Brexit Checker lists, we could
always reproduce its current content in future, because these lists
all share the same "tags" structure.

[1]: https://github.com/alphagov/email-alert-api/pull/973
[2]: https://github.com/alphagov/email-alert-api/pull/1485/commits/443ac32d52c9ec7a2f8eb54110b9956c180ab429


hannako (approved):
👍

PR #1520, opened 2020-12-16
Only use I18n where it adds value

benthorner:

Previously we extracted much of the text of individual-type emails
to I18n keys, which is just unnecessary indirection:

- We only support emails in English.
- It makes it harder to grasp the overall content.
- Each line of the content needs its own (arbitrary) key.
- The tests don't make use of I18n at all.

This replaces uses of I18n where the text is only used once, without
any modifiers. We can still benefit from I18n:

- For some text, as a way to keep it consistent across the app.
- For semi-dynamic text like frequencies, as a simple switch mechanism.

Note that I've only applied this approach to the ImmediateEmailBuilder,
to avoid blocking upcoming design changes to digest emails.

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -33,14 +33,14 @@ private
   end
 
   def subject(content)
-    I18n.t!("emails.immediate.subject", title: content.title)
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Nice one.

PR #1521, opened 2020-12-16
Add "Email is too large" catch to Undeliverable Failures

Tetrino:
Notify has a hard limit on the size of an email in bytes that can be
send. This cannot be bypassed at this step so if any emails come through
that are too large, we want to fail them outright rather than attempting
to resend them.

### DIFF
--- a/app/services/send_email_service/send_notify_email.rb
+++ b/app/services/send_email_service/send_notify_email.rb
@@ -42,6 +42,9 @@ private
   def undeliverable_failure?(error)
     return false unless error.is_a?(Notifications::Client::BadRequestError)
 
-    error.message.end_with?("Not a valid email address")
+    return true if error.message.end_with?("Not a valid email address")
+
+    # We have a hard limit in notify that cannot be bypassed here.
+    error.message.start_with?("BadRequestError: Your message is too long")
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/send_email_service/send_notify_email.rb
```diff
@@ -43,5 +43,8 @@ def undeliverable_failure?(error)
     return false unless error.is_a?(Notifications::Client::BadRequestError)
 
     error.message.end_with?("Not a valid email address")
```
kevindew, line 45: your change would make this line have no consequences. we'd want this to be `return true if error.message.end_with?("Not a valid email address")`

app/services/send_email_service/send_notify_email.rb
```diff
@@ -43,5 +43,8 @@ def undeliverable_failure?(error)
...
+
+    # We have a hard limit in notify that cannot be bypassed here.
+    error.message.include?("Your message is too long.")
```
kevindew, line 48: It'd be good to use `start_with?` here to be consistent with the earlier approach.

app/services/send_email_service/send_notify_email.rb
Tetrino: Aye I noticed my dumb mistake almost immediately (pinged you on slack). Been a long couple days.

kevindew (approved):
Thank you - this looks good.

PR #1522, opened 2020-12-16
Add a rake task to update multiple subscription lists with new tags

benjamineskola:
https://trello.com/c/xnGP5kS0/2236-5-mhra-request-to-rename-specialist-publisher-alert-categories

### DIFF
--- a/lib/tasks/data_migration.rake
+++ b/lib/tasks/data_migration.rake
@@ -10,6 +10,22 @@ namespace :data_migration do
     SubscriberListMover.new(**args).call
   end
 
+  desc "Rename an alert type and all combinations of subscriptions to it"
+  task :rename_alert_subscription_lists, %i[from_slug to_slug] => :environment do |_t, args|
+    SubscriberList.where("tags->'alert_type' IS NOT NULL").find_each do |list|
+      next unless list.tags[:alert_type][:any].include? args[:from_slug]
+
+      new_tags = (list.tags[:alert_type][:any] - [args[:from_slug]] + [args[:to_slug]]).uniq
+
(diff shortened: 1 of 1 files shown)


deborahchua (approved):
LGTM! It might also be worth adding to the [support task](https://github.com/alphagov/email-alert-api/blob/master/docs/support-tasks.md) doc.

lib/tasks/data_migration.rake
```diff
@@ -10,6 +10,22 @@ namespace :data_migration do
     SubscriberListMover.new(**args).call
   end
 
+  desc "Rename an alert type and all combinations of subscriptions to it"
+  task :rename_alert_subscription_lists, %i[from_slug to_slug] => :environment do |_t, args|
+    SubscriberList.where("tags->'alert_type' IS NOT NULL").all.each do |list|
```
thomasleese, line 15:
```suggestion
    SubscriberList.where("tags->'alert_type' IS NOT NULL").find_each do |list|
```

This might be slightly more efficient, although I guess there aren't many of these.

lib/tasks/data_migration.rake
```diff
@@ -10,6 +10,22 @@ namespace :data_migration do
...
+      next unless list.tags[:alert_type][:any].include? args[:from_slug]
+
+      new_tags = (list.tags[:alert_type][:any] - [args[:from_slug]] + [args[:to_slug]]).uniq
+
+      if (new_list = SubscriberList.where("tags->'alert_type' IS NOT NULL").find_all { |l| l.tags[:alert_type][:any].sort == new_tags.sort }.first)
```
thomasleese, line 20: I'm guess there's the potential for the `new_list` and the `old_list` to end up the same? For example if there's a subscriber list for alerts types that aren't related to the old or new slug. Do we need a check for this?

benjamineskola: I'm not entirely sure of the circumstances this would occur in, but it can't hurt to make sure.

PR #1524, opened 2020-12-17
Switch to one email per digest subscription

benthorner:
https://trello.com/c/KEcIcyix/673-update-template-for-digest-emails-to-match-design

This PR makes a couple of changes:

- Switches the digest functionality so that we send one
email per digest subscription, instead of an aggregation
of aggregations.

- Updates the digest email template to match the final
design, now that the emails are split.

### Single Subscription

Screenshot 2020-12-17 at 13 17 07

### Multiple Subscriptions (top part)

Screenshot 2020-12-17 at 13 17 36

### Multiple Subscriptions (bottom part)

(many more news content items above here!)

Screenshot 2020-12-17 at 13 17 57

### First subscription

Screenshot 2020-12-17 at 13 43 16

### Second subscription

Screenshot 2020-12-17 at 13 43 31

### DIFF
--- a/app/builders/digest_email_builder.rb
+++ b/app/builders/digest_email_builder.rb
@@ -1,7 +1,7 @@
 class DigestEmailBuilder < ApplicationBuilder
-  def initialize(address:, digest_items:, digest_run:, subscriber_id:)
+  def initialize(address:, digest_item:, digest_run:, subscriber_id:)
     @address = address
-    @digest_items = digest_items
+    @digest_item = digest_item
     @digest_run = digest_run
     @subscriber_id = subscriber_id
   end
@@ -17,7 +17,7 @@ class DigestEmailBuilder < ApplicationBuilder
 
 private
 
-  attr_reader :address, :digest_items, :digest_run, :subscriber_id
+  attr_reader :address, :digest_item, :digest_run, :subscriber_id
 
   def body
     <<~BODY
@@ -31,8 +31,7 @@ private
   end
 
   def presented_results
-    digest_items.map { |result| presented_segment(result) }
-                .join("\n&nbsp;\n\n")
+    presented_segment(digest_item)
   end
 
   def presented_segment(segment)
--- a/app/builders/digest_email_builder.rb
+++ b/app/builders/digest_email_builder.rb
@@ -9,7 +9,10 @@ class DigestEmailBuilder < ApplicationBuilder
   def call
     Email.create!(
       address: address,
-      subject: I18n.t!("emails.digests.#{digest_run.range}.subject"),
+      subject: I18n.t!(
+        "emails.digests.#{digest_run.range}.subject",
+        title: digest_item.subscriber_list_title,
+      ),
       body: body,
       subscriber_id: subscriber_id,
     )
@@ -21,60 +24,51 @@ private
 
   def body
     <<~BODY
-      #{I18n.t!("emails.digests.#{digest_run.range}.opening_line")}
+      #{I18n.t("emails.digests.#{digest_run.range}.opening_line")}
+
+      #{title_and_optional_description}
+
+      ---
 
       #{presented_results}
-      #{I18n.t!("emails.digests.#{digest_run.range}.permission_reminder")}
 
-      #{ManageSubscriptionsLinkPresenter.call(address)}
+      ---
+
+      # Why am I getting this email?
+
+      #{I18n.t("emails.digests.#{digest_run.range}.footer_explanation")}
+
+      #{digest_item.subscriber_list_title}
+
+      [Unsubscribe](#{unsubscribe_url})
+
+      [#{I18n.t!('emails.digests.footer_manage')}](#{PublicUrls.authenticate_url(address: address)})
     BODY
   end
 
   def presented_results
-    presented_segment(digest_item)
-  end
-
-  def presented_segment(segment)
-    <<~RESULT
-      #{presented_header(segment)}
-
-      #{presented_content(segment.content)}
-      ---
-
-      #{UnsubscribeLinkPresenter.call(segment.subscription_id, segment.subscriber_list_title)}
-    RESULT
-  end
-
-  def presented_header(segment)
-    copy = "# #{presented_title(segment)} &nbsp;"
-
-    if segment.subscriber_list_description.present?
-      copy += "\n\n#{segment.subscriber_list_description}"
+    changes = digest_item.content.map do |item|
+      presenter = "#{item.class.name}Presenter".constantize
+      presenter.call(item, frequency: digest_run.range)
     end
 
-    copy
+    changes.join("\n---\n\n").strip
   end
 
-  def presented_title(segment)
-    if segment.subscriber_list_url
-      "[#{segment.subscriber_list_title}](#{Plek.new.website_root}#{segment.subscriber_list_url})"
-    else
-      segment.subscriber_list_title
+  def title_and_optional_description
+    result = "# " + digest_item.subscriber_list_title
+
+    if digest_item.subscriber_list_description.present?
+      result += "\n\n" + digest_item.subscriber_list_description
     end
+
+    result
   end
 
-  def presented_content(content)
-    changes = content.map do |item|
-      case item
-      when ContentChange
-        ContentChangePresenter.call(item, frequency: digest_run.range)
-      when Message
-        MessagePresenter.call(item, frequency: digest_run.range)
-      else
-        raise "Unexpected content type: #{item.class}"
-      end
-    end
-
-    changes.join("\n---\n\n")
+  def unsubscribe_url
+    PublicUrls.unsubscribe(
+      subscription_id: digest_item.subscription_id,
+      subscriber_id: subscriber_id,
+    )
   end
 end
--- a/app/workers/digest_email_generation_worker.rb
+++ b/app/workers/digest_email_generation_worker.rb
@@ -2,10 +2,10 @@ class DigestEmailGenerationWorker < ApplicationWorker
   sidekiq_options queue: :email_generation_digest
 
   def perform(digest_run_subscriber_id)
-    email = nil
+    email_ids = []
+
     DigestRun.transaction do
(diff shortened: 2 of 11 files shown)


benthorner:
Notes for creating the test emails:

```
# create subscriptions
s = Subscriber.create(address: "ben.thorner@digital.cabinet-office.gov.uk")
d = DigestRun.order(created_at: :desc).first
s.subscriptions.create(subscriber_list: SubscriberList.find_by(slug: "personal-tax"), frequency: "daily")
s.subscriptions.create(subscriber_list: SubscriberList.find_by(slug: "news-and-communications"), frequency: "daily")
dr = DigestRunSubscriber.create(subscriber: s, digest_run: d)

# create emails
s = Subscriber.find_by_address("ben.thorner@digital.cabinet-office.gov.uk")
d = DigestRun.order(created_at: :desc).first
dr = DigestRunSubscriber.find_by(subscriber: s, digest_run: d)
ENV["GOVUK_NOTIFY_RECIPIENTS"] += ",ben.thorner@digital.cabinet-office.gov.uk"	
items = DigestItemsQuery.call(dr.subscriber, dr.digest_run)
items.each { |item| DigestEmailBuilder.call(address: s.address, digest_item: item, subscriber_id: s.id, digest_run: d) }
Email.where("created_at > ?", 5.minutes.ago).each { |e| e.update!(status: "pending"); SendEmailWorker.new.perform(e.id, {}, 
"default") }
```

app/builders/digest_email_builder.rb
```diff
@@ -21,60 +24,53 @@ def call
 
   def body
     <<~BODY
-      #{I18n.t!("emails.digests.#{digest_run.range}.opening_line")}
+      Update from GOV.UK for:
```
sihugh, line 27: Github's diff here is 🤮

app/workers/digest_email_generation_worker.rb
```diff
@@ -2,51 +2,58 @@ class DigestEmailGenerationWorker < ApplicationWorker
...
     end
   end
 
-  def fill_subscription_content(email, digest_items, digest_run_subscriber)
+  def fill_subscription_content(email_ids, digest_items, digest_run_subscriber)
     now = Time.zone.now
-    records = digest_items.flat_map do |result|
-      result.content.map do |content|
+
+    records = digest_items.zip(email_ids).flat_map do |digest_item, email_id|
```
sihugh, line 52: There's quite a bit happening on this line and the next. My brain's struggling a bit to keep digest_items, records, and emails in play, probably because I'm not familiar with the code. Is there a way of pulling it apart a little that doesn't make it worse?

app/builders/digest_email_builder.rb
```diff
@@ -21,60 +24,53 @@ def call
...
-      when Message
-        MessagePresenter.call(item, frequency: digest_run.range)
-      else
-        raise "Unexpected content type: #{item.class}"
-      end
+  def title_and_optional_description
+    result = "# #{digest_item.subscriber_list_title}"
+
+    if digest_item.subscriber_list_description.present?
+      result += "\n\n" + digest_item.subscriber_list_description
```
sihugh, line 64: is it worth keeping the string construction consistent here? We've got `"# #{digest_item.subscriber_list_title}"` and  `"\n\n" + digest_item.subscriber_list_description` a couple of lines apart

config/locales/en.yml
```diff
@@ -13,13 +13,14 @@ en:
       footer_manage: "Manage your email preferences"
     digests:
       daily:
-        subject: "Daily update from GOV.UK"
-        opening_line: Daily update from GOV.UK.
-        permission_reminder: "^You’re getting this email because you subscribed to daily updates on these topics on GOV.UK."
+        subject: "Daily update from GOV.UK for: %{title}"
+        heading: "Yesterday’s changes"
```
sihugh, line 17: Should we use curly quotes here? Designers seem to love them, so I'm wondering if they should be in here :-)

(I think that's a larger conversation, so let's not do that now)

sihugh (approved):
I think this looks OK. Minor comments which might be less important than shipping this and learning from it.

app/builders/digest_email_builder.rb
benthorner: Could you explain a bit more what you mean here? I'm not following, after reading a couple times.

sihugh: I sort of meant that we should either have:

config/locales/en.yml
benthorner: Oh yes, they're intentional here: we've done similarly for confirmation emails ([here](https://github.com/alphagov/email-alert-api/blob/55726c8f73aabda0fd58588a6d3a942d3603ee86/config/locales/en.yml#L5-L7)).

app/builders/digest_email_builder.rb
```diff
@@ -1,81 +1,76 @@
...
     @subscriber_id = subscriber_id
   end
 
   def call
     Email.create!(
       address: address,
-      subject: I18n.t!("emails.digests.#{digest_run.range}.subject"),
+      subject: I18n.t!(
+        "emails.digests.#{digest_run.range}.subject",
+        title: digest_item.subscriber_list_title,
```
kevindew, line 14: Do we know if there are any character limits on this and whether any of our titles may blow it? I guess we run a similar risk with content changes but they don't use the list title and I know we have some crazy long list titles.

app/builders/digest_email_builder.rb
benthorner: Oh, I see what you mean. Sure, I've amended the last commit to be more consistent.

app/workers/digest_email_generation_worker.rb
benthorner: I also stared at this for quite a while. Hopefully it's not much worse than it was before. I couldn't think of any meaningful variables to abstract things with - other than `digest_items_together_with_their_email`.

Fundamentally this is hard to understand because SubscriptionContent is hard to understand ([pain card for that](https://trello.com/c/0ge1rXr5/314-dev-pain-the-subscriptioncontent-model-is-confusing-and-hard-to-work-with)). But also, we push creating these join records away from creating the emails themselves i.e. new emails are initially orphaned. I think that's intentional ([see similarly here](https://github.com/alphagov/email-alert-api/blob/55726c8f73aabda0fd58588a6d3a942d3603ee86/app/services/immediate_email_generation_service/batch.rb#L15)) - i.e. new emails are temporarily orphaned from their content change / message / digest run - but does mean we end up with some heavy code to build the joining records manually, later on. Hopefully that makes some sense. I'm not sure there's much we can do with the code, though, short of looking again at why the SubscriptionContent model exists at all.

kevindew (commented):
Nice job - it is a surprising low amount of changes. 

One thing I was pondering is that, if you are user with multiple subscriptions, whether it is a better or worse user experience to get the emails sent in rapid succession. As a consequence  of doing a digest run subscriber at a time there's a high chance they'll get a rapid burst of emails - which could look weird. But it does seem like it'd require re-thinking the whole digest generation to have different subscriptions distributed across a run. 

One thing to consider for future is that by dealing with individual lists we could calculate a digest once and re-use it for each subscriber. But that seems a big coding jump.

app/workers/digest_email_generation_worker.rb
kevindew: A small amend you could make this easier to understand would be to insert into subscription contents multiple times as part of the [digest_items loop](https://github.com/alphagov/email-alert-api/pull/1524/files#diff-8e57211715641f37fe547d46d2a9665bc9e53503623cb9f7c1c598759cd6867bR16-R18). I can't imagine we benefit much from this being optimised to do only a single insert given we're already doing multiple email inserts.

app/builders/digest_email_builder.rb
benthorner: Oh, interesting idea. Our maximum title is 2163 characters, which is fine. I tried with longer titles - eventually getting an OOM error at 1 billion trillion (or something). Should be fine!

kevindew: Is that with Notify? I wondered if SES or Notify imposed any limits. The StackOverflow question I can find on it seems there's just vague recommendations: https://stackoverflow.com/questions/1592291/what-is-the-email-subject-length-limit

benthorner: Oh sorry, I was being dim (I thought you meant with I18n). I've had a go and 2163 chars is fine. Seems to be fine up to 10K chars. I also chatted with David from GOV.UK Notify. They don't know what the actual limit is, as there is indeed nothing written about it. But for now, I don't think it'll cause any problems - even for our insanely long titles.

benthorner:
@kevindew @sihugh I'm going to merge this in now, so we can get it running in production. Appreciate there a couple of things to follow-up on afterwards, which I can look at tomorrow:

- Update the screenshots to reflect the final PR.
- Look again @sihugh's comment around subscription contents.
- Respond to @kevindew's comment about the future of digest runs.

app/workers/digest_email_generation_worker.rb
benthorner: Oh yeah, that might be worth a try. I'll make a note to look at that in a refactoring PR.

benthorner:
I've updated the screenshots in the description, and made some notes on the card about doing a follow-up refactoring PR, as there were a couple of things I skipped over in order to get this out quickly.

@kevindew to respond to your comment...

> But it does seem like it'd require re-thinking the whole digest generation to have different subscriptions distributed across a run.

That might be quite challenging to achieve! I think whether it's weird is quite subjective, as there's no clear expectation here: if I'm expecting to receive multiple, similar emails, then why would they not come at the same time?

> One thing to consider for future is that by dealing with individual lists we could calculate a digest once and re-use it for each subscriber. But that seems a big coding jump.

It certainly would! It would make more sense, though: we could process by list, instead of by subscriber, where we now undo our own aggregation. We'd end up with fewer jobs, so the downside might be we'd run into the same awkward [batching situation](https://github.com/alphagov/email-alert-api/blob/master/app/services/immediate_email_generation_service/batch.rb) as with individual-type emails. Do you think there would be much advantage in processing digests by list?

PR #1533, opened 2020-12-21
Refactor creating SubscriptionContent for digests

benthorner:
https://trello.com/c/KEcIcyix/673-split-digest-emails-so-we-send-one-email-per-subscription

Relates to: https://github.com/alphagov/email-alert-api/pull/1524#discussion_r545157027

This is to make the association clearer between the digest item and
the email. In future, we should reconsider if there are clearer ways
to articulate the association of emails to digest runs, messages and
content changes, as opposed to SubscriptionContent.


sihugh (approved):
Aha, this directly addresses that scary `zip.reverse.interleave.flip.undo.transmogrify` line!  Thanks, this is much easier to understand for a bear of unfamiliar brain.

PR #1536, opened 2020-12-30
Add task to rename Brexit checker list titles

benthorner:
https://trello.com/c/RHB4pu1A/748-change-get-ready-for-2021-to-brexit-checker-results

### DIFF
--- a/lib/tasks/data_migration.rake
+++ b/lib/tasks/data_migration.rake
@@ -56,4 +56,11 @@ namespace :data_migration do
       puts "Updated #{args[:key]} in #{list.title} to #{new_criteria}"
     end
   end
+
+  desc "Temporary task to update all existing Brexit checker list titles"
+  task temp_update_all_brexit_list_titles: :environment do
+    SubscriberList
+      .where("tags->'brexit_checklist_criteria' IS NOT NULL")
+      .update_all(title: "Brexit checker results")
+  end
 end


benthorner:
Tested and works in Integration ✅.

lib/tasks/data_migration.rake
```diff
@@ -56,4 +56,11 @@ namespace :data_migration do
       puts "Updated #{args[:key]} in #{list.title} to #{new_criteria}"
     end
   end
+
+  desc "Temporary task to update all existing Brexit checker list titles"
+  task temp_update_all_brexit_list_titles: :environment do
+    SubscriberList
+      .where("tags->'brexit_checklist_criteria' IS NOT NULL")
```
huwd, line 63: Yep I think this makes sense to me.
So because this is criteria based it will rename both all those lists _with_ actions and those _without_ actions.

And in the null case of "you didn't answer any questions" there is still no option to subscribe right?

If so the query makes sense!

huwd (approved):

lib/tasks/data_migration.rake
benthorner: Hmm, not quite following you there, but the query isn't trivial so warrants a bit of explanation: only lists for Brexit checker results have a `tags` key of `brexit_checklist_criteria` (see [here](https://github.com/alphagov/finder-frontend/blob/f55dfc4bcb5e29926629e503ebd4c42b720f937e/app/controllers/brexit_checker_controller.rb#L176)). The value of the key is the criteria generated by answering the checker questions.

The `->` query is used to query on the value of the key in the JSON `tags` column, to see if it exists (see [here](https://www.postgresql.org/docs/9.3/functions-json.html)).

Does that help?

huwd: It does! Thanks for confirming

benthorner: Discussed IRL and clarified: 

> there’s a one-to-one correspondence between “brexit checker lists” and “lists with that tag pattern in it”.

PR #1538, opened 2020-12-30
Stop printing Message title in email body

benthorner:
https://trello.com/c/N98texVc/735-update-static-parts-of-email-templates-so-theyre-consistent-with-other-navigation-options-2b

Previously this was used in both the email body and in the subject of
individual-type emails. While this coupling works for content changes,
it doesn't for messages, since they are by definition ad-hoc [1], and
we should avoid making any assumptions about the structure of the body
content for the email. This does the simplest thing to decouple the
subject line from the email body content - the Brexit Checker is still
the only consumer of the /messages API, and we will update it to render
a suitable replacement to the current title.

Arguably the subject should also be completely flexible, and not start
with "Update from GOV.UK for", but that would be a wider change.

Note that the tracking for the title is necessarily removed, and is now
entirely the concern of the API consumer to add tracking on any links,
which the Brexit Checker is already doing [2]. (It is actually quite
confusing that we showed two different pieces of text that link to the
same URL, but with different tracking attributes!)

[1]: https://github.com/alphagov/email-alert-api/blob/master/docs/adr/adr-004-message-concept.md#introduction
[2]: https://github.com/alphagov/finder-frontend/blob/f55dfc4bcb5e29926629e503ebd4c42b720f937e/app/helpers/brexit_checker_helper.rb#L38-L49

### DIFF
--- a/app/presenters/message_presenter.rb
+++ b/app/presenters/message_presenter.rb
@@ -5,53 +5,10 @@ class MessagePresenter < ApplicationPresenter
   end
 
   def call
-    [
-      title_markdown,
-      body,
-    ].compact.join("\n\n") + "\n"
+    message.body + "\n"
   end
 
 private
 
-  attr_reader :message, :frequency
-
-  delegate :title, :body, :url, to: :message
-
-  def absolute_url
-    return unless url
-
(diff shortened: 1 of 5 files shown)


huwd (approved):
Looks great

PR #1539, opened 2020-12-30
Fix not showing link to some checker results

benthorner:
Previously we adopted this hack in favour of manually specifying a
fake "description" for each subscriber list [1]. However, we forgot
that the URL for checker results has changed. This fixes the hack
to cope with both versions of the checker results URL.

[1]: 520be069


huwd (approved):

benthorner:
Tested and works in Staging ✅.

PR #1540, opened 2020-12-31
Prepare to remove URL column for Messages

benthorner:
https://trello.com/c/N98texVc/735-update-static-parts-of-email-templates-so-theyre-consistent-with-other-navigation-options-2b

This is no longer used since 5b9a81af.


sihugh (approved):
Sorry, thought I'd done this one. 🔪

PR #1542, opened 2021-01-04
Enhance bulk email functionality

benthorner:

This makes a few changes to the existing bulk email builder:

- It is now batches generation of large quantities of email.
- The email now ends with a standard unsubscribe / manage footer.
- A list URL can be substituted for a "%LISTURL%" placeholder.

Arguably the builder is now doing too much: it should not be
deciding which emails to generate, but simply generating them.
This has lead to the test setup becoming a bit convoluted. In
future work, we should refactor the builder to separate out the
selection of email, like we do for immediate emails [1].

## Before (Link goes to "%LISTURL%" i.e. doesn't work)

Screenshot 2021-01-04 at 14 54 06

## After (link goes to checker results page for subscriber)

Screenshot 2021-01-04 at 14 44 07

[1]: https://github.com/alphagov/email-alert-api/blob/d248a8d63324cb2255f6db36104cce9f3ed0bf2f/app/services/immediate_email_generation_service.rb

### DIFF
--- a/app/builders/bulk_subscriber_list_email_builder.rb
+++ b/app/builders/bulk_subscriber_list_email_builder.rb
@@ -58,7 +58,7 @@ private
     )
 
     <<~BODY
-      #{body.gsub('%LISTURL%', list.url.to_s)}
+      #{body.gsub('%LISTURL%', PublicUrls.url_for(base_path: list.url.to_s))}
 
       ---
 
--- a/app/builders/bulk_subscriber_list_email_builder.rb
+++ b/app/builders/bulk_subscriber_list_email_builder.rb
@@ -10,7 +10,7 @@ class BulkSubscriberListEmailBuilder < ApplicationBuilder
 
   def call
     batches.flat_map do |batch|
-      records = records_for_batch(batch)
+      records = records_for_batch(batch.to_h)
       Email.insert_all!(records).pluck("id")
     end
   end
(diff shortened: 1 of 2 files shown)


app/builders/bulk_subscriber_list_email_builder.rb
```diff
@@ -1,40 +1,81 @@
...
-          body: body,
-          subscriber_id: subscriber_id,
-          created_at: now,
-          updated_at: now,
-        }
-      end
+  attr_reader :subject, :body, :subscriber_lists, :now
+
+  def records_for_batch(batch)
+    subscribers = Subscriber.find(batch.keys).index_by(&:id)
```
hannako, line 25: Feel free to ignore:
I'm not v familiar with email alert api. It took me a while to figure out that calling `.to_h` on the array produced by the scope `.subscription_ids_by_subscriber` would mean that `batch.key` was a subscriber id. and like wise `batch.values` are subscription ids. Perhaps `batch.keys` and `batch.values` could be their own more explicitly named methods?

hannako (approved):
LGTM

app/builders/bulk_subscriber_list_email_builder.rb
benthorner: Yes, it's pretty obscure isn't it! Apparently it's [a Rails-ism](https://apidock.com/ruby/v2_5_5/Array/to_h), which I copied from [here](https://github.com/alphagov/email-alert-api/blob/master/app/services/immediate_email_generation_service.rb#L36). I'm hoping to make this clearer in the follow-up PR, but for now I've amended [the fourth commit](https://github.com/alphagov/email-alert-api/pull/1542/commits/c872fa49f54eebb2bb67f6054edd254ca2c27387) so it's hopefully a little clearer. 

How does it look now?

hannako:
Have looked at the commits made after approval, and still looks good 👍

PR #1548, opened 2021-01-06
Add tracking to source URLs in all emails

benthorner:

This refactors the existing code to build URLs to make
it easy to create tracked URLs. We then make use of this
to add tracking for source URLs in the three emails in
which they occur. An example of a source URL is for lists
generated from Brexit Checker results.


lib/public_urls.rb
```diff
@@ -1,26 +1,28 @@
 module PublicUrls
   class << self
-    def url_for(base_path:)
-      URI.join(website_root, base_path).to_s
+    def url_for(base_path:, **params)
+      uri = URI.join(website_root, base_path)
+      query = Hash[URI.decode_www_form(uri.query.to_s)]
```
hannako, line 5: what is the reason for having `Hash[URI.decode_www_form(uri.query.to_s)]` rather than just `{params}`?

spec/lib/public_urls_spec.rb
```diff
@@ -1,9 +1,14 @@
 RSpec.describe PublicUrls do
   describe ".url_for" do
-    it "returns the GOV.UK url for the content item" do
+    it "returns a GOV.UK for a base path" do
       result = subject.url_for(base_path: "/foo/bar")
```
hannako, line 4: it returns a GOV.UK url for a base path?

app/builders/digest_email_builder.rb
```diff
@@ -57,7 +57,12 @@ def presented_results
 
   def title_and_optional_url
     result = "# " + digest_item.subscriber_list_title
-    source_url = SourceUrlPresenter.call(digest_item.subscriber_list_url)
+
+    source_url = SourceUrlPresenter.call(
+      digest_item.subscriber_list_url,
+      utm_source: digest_item.subscriber_list_slug,
+      utm_content: digest_run.range,
```
hannako, line 64: Not sure I follow the commit message about the relevance of the utm_content param. Could you explain?

lib/public_urls.rb
benthorner: So there are two sets of params at play here: any existing params in the "base_path" (e.g. Brexit Checker criteria) and the ones we want to append. Because we don't know if the "base_path" has query params, we need to fish them out and merge them with the ones we want to add. `URI.decode_www_form` returns a key/value array, so we need to use `Hash` to get the initial hash to merge `params` with.

Does that help?

hannako: ah that makes sense, got it thank you

spec/lib/public_urls_spec.rb
benthorner: Oops - I've updated the second commit with "URL" 👍.

app/builders/digest_email_builder.rb
benthorner: Sure, the whole [UTM thing](https://en.wikipedia.org/wiki/UTM_parameters) is quite bizarre, as I think it was designed for links on web pages, rather than links in emails. The overall goal is that we want to know which links are being used in the different emails we send. All the info has to be in the link itself, since that's all we get after the user clicks on it.

In this case, we know that the link came from a particular list (UTM Source). The UTM Content param is optional, and another way to distinguish the link. Here, it will tell us whether the link was in a weekly, daily, or individual-type email. We do a similar thing already for one of the other links ([here](https://github.com/alphagov/email-alert-api/blob/f1687ad93e261a1ec4368072407b5221d68fcb7a/app/presenters/content_change_presenter.rb#L30)).

It's pretty speculative, but it would allow us to see if e.g. users interact with digest emails more than individual-type emails.

In case it helps, I wrote a bit about tracking with UTM params [here](https://github.com/alphagov/content-publisher/blob/master/docs/approach-to-analytics.md#utm--inbound-links).

hannako: Thanks for the explanation!

hannako (approved):
Thanks for the explanations Ben. Looks good 👍

benthorner:
Tested and works on Integration ✅.

@hannako thanks for the review! In testing this I found an issue with the unsubscribe links, which had become mangled and had stopped working. I've added another commit for that - could you take a look?

hannako:
@benthorner Last commit looks good, thanks for the links to docs.

PR #1551, opened 2021-01-07
Remove bulk task for Brexit emails

benthorner:

We have no plan to run this again, and can easily re-add it if we
do. The task itself is problematic, as it can't be used with the
standard "Run rake task" Jenkins job, which only supports a single
line input i.e. we can't specify multiline body content.


hannako (approved):
👋 😭

PR #1556, opened 2021-01-07
Update the comment for the "message is too long" fix

kevindew:
This comment has been updated to reflect the non-ideal nature of this
fix. It is presented with the hope it doesn't encourage people to add
similar things in future.

As I reference stop-gap in the commit I have no doubt doomed this code
to live in the code for another decade.


benthorner (approved):
Thanks 🏅.

PR #1560, opened 2021-01-08
Fix tests for opt-in / manage auth tokens

benthorner:
Previously we ignored these in order to quickly fix a bug.
The original fault was due to the removal of URL unescaping
for all tests [1], which allowed the feature and integration
tests to continue passing in spite of the URLs being invalid.

[1]: https://github.com/alphagov/email-alert-api/pull/1548/commits/57f2fca2d83391b88c75aaa3b9862b470715df00#diff-48cb7b27ea61deeffa1a37e42451aaeb4c9fa56ea4b07076c4733cdd858a0a93L11

### DIFF
--- a/spec/builders/subscriber_auth_email_builder_spec.rb
+++ b/spec/builders/subscriber_auth_email_builder_spec.rb
@@ -12,6 +12,12 @@ RSpec.describe SubscriberAuthEmailBuilder do
       )
     end
 
+    before do
+      allow(PublicUrls).to receive(:url_for)
+        .with(base_path: destination, token: token)
+        .and_return("auth_url")
+    end
+
     it { is_expected.to be_instance_of(Email) }
 
     it "creates an email" do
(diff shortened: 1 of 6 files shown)


spec/builders/subscriber_auth_email_builder_spec.rb
```diff
@@ -12,6 +12,13 @@
       )
     end
 
+    before do
+      allow(PublicUrls).to receive(:url_for)
+        .with(base_path: destination,
+              token: token)
```
ChrisBAshton, line 18: Nitpick: I think this would look neater moved onto line 17.

ChrisBAshton (approved):
LGTM 👍

spec/builders/subscriber_auth_email_builder_spec.rb
benthorner: Sure, I've amended that commit 👍.

PR #1561, opened 2021-01-08
Update confirmation email and add tracking

benthorner:

This updates the template for the confirmation received after
a user has subscribed. In doing so, we also adds tracking to
all "unsubscribe" and "manage" links, in all emails in which
they appear. Please see the commits for more details.

## Before

Screenshot 2021-01-08 at 17 15 48

## After

Screenshot 2021-01-08 at 17 17 14


hannako (approved):
LGTM

PR #1562, opened 2021-01-08
Add temporary rake task to send apology email

benthorner:
Screenshot 2021-01-08 at 16 20 57

### DIFF
--- a/spec/lib/tasks/bulk_email_spec.rb
+++ b/spec/lib/tasks/bulk_email_spec.rb
@@ -40,4 +40,35 @@ RSpec.describe "bulk_email" do
       Rake::Task["bulk_email:for_lists"].invoke(subscriber_list.id)
     end
   end
+
+  describe "temp_incident_new_subscribers" do
+    let(:affected_subject) { "Confirm that you want to get emails from GOV.UK" }
+
+    before do
+      Rake::Task["bulk_email:temp_incident_new_subscribers"].reenable
+    end
+
+    it "sends a single apology email to affected subscribers" do
+      # affected emails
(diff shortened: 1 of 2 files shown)


benthorner:
Tested and works as expected on Integration: the number of generated emails is a fraction of the affected number, which is because we only have the affected emails from yesterday in Integration at the moment.

spec/lib/tasks/bulk_email_spec.rb
```diff
@@ -40,4 +40,35 @@
...
+      create(:email, subject: affected_subject, created_at: Time.zone.local(2021, 1, 8, 13, 50))
+      create(:email, subject: "something else", created_at: Time.zone.local(2021, 1, 8, 13, 40))
+
+      expect(SendEmailWorker)
+        .to receive(:perform_async_in_queue)
+        .with(instance_of(String), queue: :send_email_immediate)
+        .once
+
+      expect { Rake::Task["bulk_email:temp_incident_new_subscribers"].invoke }
+        .to change { Email.count }.by(1)
```
ChrisBAshton, line 67: Why 1, not 2? I'd expect 2 based on https://github.com/alphagov/email-alert-api/pull/1562/files#diff-08e6964a2fb00793bbdeb6bfc868a53eaf60ec65d46676f3538a677d2ffa7d85R52-R54

benthorner: It's meant to be 1: we only want to send a single email, irrespective of how many times a given would-be-subscriber tried to subscribe.

ChrisBAshton (approved):
Thanks for jumping on this so quickly.

PR #1564, opened 2021-01-11
Simplify code to generate unique emails

benthorner:

This introduces a new "dedup_by_subscriber" scope as an alternative
to the current "subscription_ids_by_subscriber", and switches bulk
and immediate email generation to use it. In doing this, we simplify
the code to focus on single subscription objects.

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -16,14 +16,14 @@ private
     @records ||= begin
       now = Time.zone.now
       recipients_and_content.map do |recipient_and_content|
-        subscriber = recipient_and_content.fetch(:subscriber)
         content = recipient_and_content.fetch(:content)
-        subscriptions = recipient_and_content.fetch(:subscriptions)
+        subscription = recipient_and_content.fetch(:subscription)
+        subscriber = subscription.subscriber
 
         {
           address: subscriber.address,
           subject: subject(content),
-          body: body(content, subscriptions.first, subscriber),
+          body: body(content, subscription, subscriber),
           subscriber_id: subscriber.id,
           created_at: now,
           updated_at: now,
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -1,6 +1,7 @@
 class ImmediateEmailBuilder < ApplicationBuilder
-  def initialize(recipients_and_content)
-    @recipients_and_content = recipients_and_content
+  def initialize(content, subscriptions)
+    @content = content
+    @subscriptions = subscriptions
   end
 
   def call
@@ -10,20 +11,18 @@ class ImmediateEmailBuilder < ApplicationBuilder
 
 private
 
-  attr_reader :recipients_and_content
+  attr_reader :content, :subscriptions
 
   def records
     @records ||= begin
       now = Time.zone.now
-      recipients_and_content.map do |recipient_and_content|
-        content = recipient_and_content.fetch(:content)
-        subscription = recipient_and_content.fetch(:subscription)
+      subscriptions.map do |subscription|
         subscriber = subscription.subscriber
 
         {
           address: subscriber.address,
-          subject: subject(content),
-          body: body(content, subscription, subscriber),
+          subject: subject,
+          body: body(subscription, subscriber),
           subscriber_id: subscriber.id,
           created_at: now,
           updated_at: now,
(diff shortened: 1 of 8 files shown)


app/models/subscription.rb
```diff
@@ -36,6 +36,13 @@ class Subscription < ApplicationRecord
             .to_h
         }
 
+  scope :dedup_by_subscriber,
+        lambda {
+          select("DISTINCT ON (subscriptions.subscriber_id) subscriber_id, subscriptions.id as id")
+            .order("subscriptions.subscriber_id DESC, subscriptions.created_at desc")
+            .map(&:id)
```
kevindew, line 36: This seems to create actual models only to throw them away with the map. I wonder if this would be better as:

```
order("subscriptions.subscriber_id DESC, subscriptions.created_at desc")
.pluck("DISTINCT ON (subscriptions.subscriber_id) subscriber_id, subscriptions.id as id")
.map(&:last)
```

app/services/immediate_email_generation_service/batch.rb
```diff
@@ -43,34 +43,16 @@ def subscription_content_records(email_ids)
...
-        subscription_ids_by_subscriber_id.values.flatten,
-      )
-
-      subscription_ids_by_subscriber_id
-        .each_with_object({}) do |(subscriber_id, subscription_ids), memo|
-          subscriber = all_subscribers[subscriber_id]
-          subscriptions = all_subscriptions.values_at(*subscription_ids).compact
-          memo[subscriber] = subscriptions if subscriber && subscriptions.any?
-        end
-    end
```
kevindew, line 61: nice 👍

app/builders/immediate_email_builder.rb
```diff
@@ -16,14 +16,14 @@ def records
     @records ||= begin
       now = Time.zone.now
       recipients_and_content.map do |recipient_and_content|
-        subscriber = recipient_and_content.fetch(:subscriber)
         content = recipient_and_content.fetch(:content)
-        subscriptions = recipient_and_content.fetch(:subscriptions)
+        subscription = recipient_and_content.fetch(:subscription)
```
kevindew: You've got a minor typo on the commit message for this.

app/services/immediate_email_generation_service/batch.rb
```diff
@@ -27,19 +27,16 @@ def email_parameters
...
         end
       end
     end
 
     def subscription_content_records(email_ids)
       email_parameters.flat_map.with_index do |params, index|
-        params[:subscriptions].map do |subscription|
-          { subscription_id: subscription.id, email_id: email_ids[index] }
-        end
+        { subscription_id: params[:subscription].id, email_id: email_ids[index] }
```
kevindew: Hmm this looks like a subtle change. It looks like where previously we'd have had every subscription inserted as a subscription content this would just be the first one.

I guess this doesn't have any real world impacts. Could, only be a touch confusing if you were debugging but probably just another warning flag about subscription content's usage/purpose.

kevindew (commented):
Looks good to me. Just a couple of minor things to resolve.

app/models/subscription.rb
benthorner: Oh yes, that's what I was aiming for - totally missed that "pluck" can have arbitrary SQL. I've amended the first commit to do that - it actually required `Arel.sql` (see updated commit message).

app/builders/immediate_email_builder.rb
benthorner: I assume you mean `and` (-> `an`). Fixed.

app/services/immediate_email_generation_service/batch.rb
benthorner: Oh yes, I remember thinking that and realise I didn't write it anywhere. I've amended the third commit message to explain how it has no effect on the processing. I suppose it could lead to an interesting debugging problem if someone was querying _from_ a particular subscription, expecting to find an email that may actually have been associated with another. I agree it's mostly a signal of pain with the model itself, rather than the processing.

I've made a note on [the pain card about it](https://trello.com/c/0ge1rXr5/314-dev-pain-the-subscriptioncontent-model-is-confusing-and-hard-to-work-with).

benthorner:
Thanks for the review @kevindew - hopefully the next people who work on this system will appreciate the simplicity! I've also done a bit of performance testing of this in Integration (bars on the right are with the change). If anything, the time is reduced, which makes sense because there's generally "less stuff" than there was before.

Screenshot 2021-01-13 at 09 45 16

kevindew (approved):
Nice one. LGTM 👍

PR #1569, opened 2021-01-12
Prepare to auto-update list titles / URLs

benthorner:
https://trello.com/c/dgdLhy2H/777-update-transition-to-brexit-in-email-notifications-for-transition-taxon

Currently we have a "find_or_create" API adapter that requires two
complementary actions to find ("index") or "create" a list [1]. For
the purpose of keeping existing list titles and URLs up-to-date with
GOV.UK, this copies the "index" functionality to the "create" action.
The next step will be to change the API adapter to call "create", and
then delete the (now) unused "index" action.

Currently we have to manually update list titles [2][3], and many lists
don't have a URL populated, since we only started sending this kind of
data recently, when creating Brexit Checker lists.

A potential smell here is the naming of the "create" action, which is
really doing a find/create/update. We do have a small precedent with
the "update" endpoint for subscriptions. One possible remedy would be
to rename the endpoint to "find_or_create", which is mostly accurate,
but would be easier to do in a follow-up PR.

[1]: https://github.com/alphagov/gds-api-adapters/blob/3d96f0fe726d5f42d628950ce69016b5cc3b5b1f/lib/gds_api/email_alert_api.rb#L12
[2]: https://github.com/alphagov/govuk-developer-docs/blob/1dfd9ce9a863c649145fc4cd7c359c5a75933402/source/manual/rename-a-country.html.md#5-update-email-alert-api
[3]: https://github.com/alphagov/email-alert-api/pull/1566

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -25,7 +25,7 @@ class SubscriberListsController < ApplicationController
 
   def create
     subscriber_list = CreateSubscriberListService.call(params: params, user: current_user)
-    render json: subscriber_list.to_json, status: :created
+    render json: subscriber_list.to_json
   end
 
 private
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -24,22 +24,11 @@ class SubscriberListsController < ApplicationController
   end
 
   def create
-    subscriber_list = SubscriberList.create!(subscriber_list_params)
+    subscriber_list = CreateSubscriberListService.call(params: params, user: current_user)
     render json: subscriber_list.to_json, status: :created
   end
 
 private
-
-  def subscriber_list_params
-    title = params.fetch(:title)
-
-    find_exact_query_params.merge(
-      title: title,
-      slug: slugify(title),
-      url: params[:url],
-      signon_user_uid: current_user.uid,
-    )
-  end
 
   def convert_legacy_params(link_or_tags)
     link_or_tags.transform_values do |link_or_tag|
@@ -55,16 +44,5 @@ private
       email_document_supertype: params.fetch(:email_document_supertype, ""),
       government_document_supertype: params.fetch(:government_document_supertype, ""),
     }
-  end
-
-  def slugify(title)
-    slug = title.parameterize.truncate(255, omission: "", separator: "-")
-
-    while SubscriberList.where(slug: slug).exists?
-      slug = title.parameterize.truncate(244, omission: "", separator: "-")
-      slug += "-#{SecureRandom.hex(5)}"
-    end
-
-    slug
   end
 end
--- a/app/services/create_subscriber_list_service.rb
+++ b/app/services/create_subscriber_list_service.rb
@@ -5,7 +5,9 @@ class CreateSubscriberListService < ApplicationService
   end
 
   def call
-    SubscriberList.create!(subscriber_list_params)
+    subscriber_list = FindExactQuery.new(**find_exact_query_params).exact_match
+    return SubscriberList.create!(subscriber_list_params) unless subscriber_list
+    subscriber_list
   end
 
 private
--- a/app/services/create_subscriber_list_service.rb
+++ b/app/services/create_subscriber_list_service.rb
@@ -1,0 +1,52 @@
+class CreateSubscriberListService < ApplicationService
+  def initialize(params:, user:)
+    @params = params
+    @user = user
+  end
+
+  def call
+    SubscriberList.create!(subscriber_list_params)
+  end
+
+private
+
+  attr_reader :params, :user
+
+  def subscriber_list_params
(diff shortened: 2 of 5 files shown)


app/controllers/subscriber_lists_controller.rb
```diff
@@ -25,7 +25,7 @@ def show
 
   def create
     subscriber_list = CreateSubscriberListService.call(params: params, user: current_user)
-    render json: subscriber_list.to_json, status: :created
+    render json: subscriber_list.to_json
```
sihugh, line 28: I thought about whether this should return different codes depending on create vs update vs "found". I think you're right that it's not worth differentiating at this point.

app/services/create_subscriber_list_service.rb
```diff
@@ -43,14 +41,14 @@ def find_exact_query_params
     }
   end
 
-  def slugify(title)
-    slug = title.parameterize.truncate(255, omission: "", separator: "-")
+  def slug
+    result = params[:title].parameterize.truncate(255, omission: "", separator: "-")
```
sihugh, line 45: Do you think there's any risk of this being accidentally used multiple times now that it's named a bit more like a variable than a method? I wonder whether it's useful to memoize this to prevent `n` different slugs being created for the same thing in some horrible future?
Admittedly it's quite edgey as edge cases come.

app/services/create_subscriber_list_service.rb
```diff
@@ -0,0 +1,54 @@
...
+    )
+  end
+
+  def convert_legacy_params(link_or_tags)
+    link_or_tags.transform_values do |link_or_tag|
+      link_or_tag.is_a?(Hash) ? link_or_tag : { any: link_or_tag }
+    end
+  end
+
+  def find_exact_query_params
```
sihugh, line 34: do you think we need to do anything (perhaps in future) to address this duplication? I'm not sure these keys change very often, and I don't mind a bit, so long as it doesn't bite me :-)

spec/services/create_subscriber_list_service_spec.rb
```diff
@@ -0,0 +1,132 @@
+RSpec.describe CreateSubscriberListService do
+  describe ".call" do
+    let(:user) { create :user }
+    let(:existing_list_query) { double(FindExactQuery, exact_match: nil) }
+
+    let(:params) do
+      ActionController::Parameters.new(
+        title: "This is a sample title",
```
sihugh, line 8: Should we be testing our permitted params expectations somewhere? Ensuring that we don't accidentally open a hole here

app/services/create_subscriber_list_service.rb
benthorner: That's a fair point, and it makes sense to call out the expectation now. I've amended the last commit to memoise.

app/services/create_subscriber_list_service.rb
benthorner: (I assume you're referring to the duplication in the controller.) Yes, once we switch [the adapter](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/email_alert_api.rb#L12) to just use the "create" endpoint, the "index" action, and its associated params/methods, can be deleted i.e. the duplication should only be temporary.

spec/services/create_subscriber_list_service_spec.rb
benthorner: Hmm, interesting. Could you expand a bit on what you mean by expectations? I initially thought you meant protecting against the code "dumbly" passing the entire params hash into the record. But I don't think that's a likely coding avenue, as it's quite hard to go down that route in practice.

Perhaps you're referring more to the params that _are_ permitted? So can we inject a nasty title, for example? I think we're generally protected against that by ActiveRecord and the API adapters escaping everything, everywhere. I'm still interested, though: is there a particular scenario you're concerned about?

app/services/create_subscriber_list_service.rb
sihugh: Perfect 👍

spec/services/create_subscriber_list_service_spec.rb
sihugh: Yeah, it was more about the former. I think I was looking at the `signon_user_id` field and wondering whether that could get overridden.

benthorner: Thanks for clarifying. I do think it's unlikely, as this is an internal API. There doesn't seem to be a precedent for testing for bad behaviour, except for params that are sourced from user input in frontend apps.

Interestingly, the app doesn't use `signon_user_id` for anything. Perhaps the idea was to support auditing. I could add a test to prove that it's not overridable by params, but the value seems pretty low.

Do you think there's a missing test here?

sihugh: Possibly - looking back it does seem as though it was added everywhere to support something that might have been done in the future. Not in scope for this change though

sihugh (approved):

PR #1571, opened 2021-01-14
Add new report about potentially dead lists

benthorner:
https://trello.com/c/PjRE1A0G/200-email-alert-api-has-dead-lists-that-will-never-send-any-email

This means we can start to quantify the scale of the tech debt, in
which the email system gets out-of-sync with GOV.UK over time.


1pretz1 (approved):
💯

PR #1579, opened 2021-01-19
Refactor finding existing subscriber lists

benthorner:
https://trello.com/c/dgdLhy2H/777-update-transition-to-brexit-in-email-notifications-for-transition-taxon

Previously we hoped to move all the code for this into the new
CreateSubscriberListService [1], but it turns out that Content
Tagger still requires separate access to the "index" endpoint.
This reverts some of the changes in [1], and then ensures the
resulting code is fully and just sufficiently tested.

[1]: https://github.com/alphagov/email-alert-api/pull/1569


spec/integration/create_subscriber_list_spec.rb
```diff
@@ -18,23 +18,9 @@
...
-          tags
-          links
-          email_document_supertype
-          government_document_supertype
-          active_subscriptions_count
-          tags_digest
-          links_digest
-        ].to_set.sort,
+      expect(subscriber_list.keys).to include(
+        "id", "title", "slug", "document_type",
```
sihugh, line 22: it doesn't _quite_ match the list you linked (url is expected here, but not there). If we want it to match exactly, would it be worth defining as a constant somewhere?

spec/integration/create_subscriber_list_spec.rb
```diff
@@ -62,6 +61,28 @@
       expect(subscriber_list["tags_digest"]).to eq(digested(subscriber_list["tags"]))
     end
 
+    context "with legacy links / tags" do
```
sihugh, line 50: 👍

spec/integration/create_subscriber_list_spec.rb
```diff
@@ -5,26 +5,17 @@
...
-
+      create_subscriber_list
       expect(SubscriberList.count).to eq(1)
     end
 
     it "returns a 200" do
-      create_subscriber_list(tags: { topics: { any: ["oil-and-gas/licensing"] },
-                                     location: { all: %w[france germany] } })
-
+      create_subscriber_list
```
sihugh, line 13: I wonder if it's worth some of these tests being combined? We're doing the same thing multiple times, and then asserting on different aspects of the result. Seems like we could shave off some seconds in a test run and save the planet?

spec/integration/browsing_subscriber_lists_spec.rb
```diff
@@ -94,6 +94,14 @@
         get_subscriber_list({})
         expect(response.status).to eq(404)
       end
+
+      it "copes if the (legacy) links / tags are not in a hash" do
```
sihugh, line 98: 👍

spec/integration/browsing_subscriber_lists_spec.rb
```diff
@@ -126,5 +116,9 @@
     def get_subscriber_list(query_payload)
       get "/subscriber-lists", params: query_payload, headers: json_headers
     end
+
+    def response_subscriber_list
```
sihugh, line 128: 👍

spec/integration/browsing_subscriber_lists_spec.rb
```diff
@@ -111,52 +100,6 @@
         expect(subscriber_list.fetch("id")).to eq(subscriber_list_links_and_document_type.id)
       end
 
-      it "finds subscriber lists that match tags and document type" do
```
sihugh, line 114: 👍

spec/integration/browsing_subscriber_lists_spec.rb
```diff
@@ -161,80 +161,6 @@
         get_subscriber_list({})
         expect(response.status).to eq(404)
       end
-
```
sihugh, line 164: 👍

app/controllers/subscriber_lists_controller.rb
```diff
@@ -24,7 +24,13 @@ def show
   end
 
   def create
-    subscriber_list = CreateSubscriberListService.call(params: params, user: current_user)
+    subscriber_list = CreateSubscriberListService.call(
```
sihugh, line 27: 👍

sihugh (commented):
Mostly great, just had a question about the subscriber_list keys not being quite as similar as the commit message promised ;-)

spec/integration/create_subscriber_list_spec.rb
benthorner: Oh yes, thanks for checking. I've removed `url`, as it's also not used by any of the callers. The APIs are free to diverge, so I don't think it's worth introducing coupling between the tests at this stage. Rather, the existing test was just a helpful precedent to reduce the verbosity of this one.

spec/integration/create_subscriber_list_spec.rb
benthorner: Sure, I've done that. I had the same thought myself, but it seemed relatively minor at the time.

sihugh (approved):
🚀

PR #1585, opened 2021-01-25
Update login email to match design

benthorner:
https://trello.com/c/l2cC0fY4/678-update-sign-in-journey-email-screens-to-match-design

This makes a couple of user-facing changing:

- Fixes a longterm typo in the common email footer
- Updates the login email design to match other emails

And then there are a few minor refactorings. Please see
the commits for more details.

## Before

Screenshot 2021-01-25 at 11 56 54

## After

Screenshot 2021-01-25 at 12 03 55

### DIFF
--- a/app/builders/subscriber_auth_email_builder.rb
+++ b/app/builders/subscriber_auth_email_builder.rb
@@ -19,24 +19,21 @@ private
   attr_reader :subscriber, :destination, :token
 
   def subject
-    "Manage your GOV.UK email subscriptions"
+    "Change your GOV.UK email preferences"
   end
 
   def body
     <<~BODY
-      # Manage your GOV.UK email subscriptions
+      # Click the link to confirm your email address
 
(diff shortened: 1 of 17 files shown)


kevindew (approved):
LGTM 👍 

There's a part of me that finds the copy a little jarring, but I consider that outside the scope of my review.  There seems a contrast between "Click the link to confirm your email address" and "Yes, I want to change my GOV.UK email preferences" - I think it's because I find the "yes..." to be an answer on a different topic to the title of "Click..."

benthorner:
> There's a part of me that finds the copy a little jarring, but I consider that outside the scope of my review.

"Yes, I also wondered about that" 😸. It didn't come up in the design review, so I'm going to focus on it being a net improvement over the current one. We're recommending a future Notifications team does a usability study on the whole journey, which should be more decisive in mopping up potential issues like this.

PR #1592, opened 2021-02-04
Add ability to specify report headers

oscarwyatt:
Only some headers are required for FCDO report
Also need the ability to get the total number
of subscribers

Trello: https://trello.com/c/ipkKc0gg/751-to-automate-the-rake-tasks-associated-with-updating-fcdo-living-in-guides-eu-countries-and-travel-advice

### DIFF
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -8,7 +8,7 @@ Email Alert API does not currently have an automated means to generate analytics
 
 ## Subscription data for each subscriber list
 
-This generates a CSV report about each subscriber list, for a specific date. In order to help disambiguate between lists with similar titles, each row of the report includes all of the [matching criteria](matching-content-to-subscriber-lists.md) for a list.
+This generates a CSV report about each subscriber list, for a specific date. In order to help disambiguate between lists with similar titles, by default each row of the report includes all of the [matching criteria](matching-content-to-subscriber-lists.md) for a list.
 
 ```bash
 report:csv_subscriber_lists["2020-11-05"]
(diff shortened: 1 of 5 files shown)


docs/analytics.md
```diff
@@ -23,6 +23,12 @@ TAGS_PATTERN=brexit_checklist_criteria report:csv_subscriber_lists["2020-11-05"]
 LINKS_PATTERN=countries report:csv_subscriber_lists["2020-11-05"]
 ```
 
+If you want just some specific headers, you can choose which headers you want as follows:
+```bash
+HEADERS=title,individual_subscribers,daily_subscribers,weekly_subscribers,total_subscribers report:csv_subscriber_lists["2020-11-05"]
+```
+All available headers are used by default, simply pass in the name of the headers as they are in the first row of the default output
```
chao-xian, line 30: Nitpick but please remove `simply`.

chao-xian: Can you also add an example below on how to run the rake task equivalent please?

lib/reports/subscriber_lists_report_row.rb
```diff
@@ -0,0 +1,81 @@
+class Reports::SubscriberListsReportRow
+  def initialize(date, headers, list)
+    @date = date
+    @headers = headers
+    @list = list
+  end
+
+  def call
+    headers.map { |header| send(header) }
```
chao-xian, line 9: What's the advantage of doing this versus:

```ruby

	headers.map { |header| &:header }
```

(...apart from being confusingly named)?

oscarwyatt: I'm not sure I understand your question...it's calling send on each member of the array `headers` - ie so headers might be `[:title, :slug]` - line 9 simply iterates over the array and calls the relevant function by passing the member of the array to the `send` function, which calls the relevant function (ie for the element `:title` the `title` method of the class will get called. AFAIK calling `&:header` will just try to call the header function for every member of the array - or have I misunderstood something?

docs/analytics.md
oscarwyatt: 👍

lib/reports/subscriber_lists_report_row.rb
chao-xian: Ahhhh sorry I didn't grok this properly. Thanks for the explainer. And validate headers stops the report entirely?

oscarwyatt: Yup. I figured if someone was entered the wrong name by accident it's better to tell them than not include that header (which might lead to a lot of head scratching)

chao-xian (approved):

PR #1593, opened 2021-02-04
Remove link from "double opt-in" email

benthorner:
This was originally intended to be part of a standard transactional
footer, but we since decided against this in all other such emails
[1]. We've just changed the URL for the link, which is prompting a
removal of it here, as opposed to updating it to avoid a redirect.

[1]: https://github.com/alphagov/email-alert-api/commit/ad414433e77353097c055584514d7e1c80aab7d4#diff-2a2c736444a20e59b27c8585e61759e5c8a54d7a8eeb336c6636ab980d972750

### DIFF
--- a/app/builders/subscription_auth_email_builder.rb
+++ b/app/builders/subscription_auth_email_builder.rb
@@ -36,7 +36,6 @@ private
 
       Thanks 
       GOV.UK emails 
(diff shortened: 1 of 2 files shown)


1pretz1 (approved):
Looks good Ben.

Tiny nitpick being that there is a link missing from your commit / PR.

benthorner:
Oh yes, thanks for spotting. I've amended with the missing link.

PR #1599, opened 2021-03-09
Update Brexit subscriber list criteria

sihugh:
[We removed an option from the checker](https://github.com/alphagov/finder-frontend/pull/2447)
which means that a criterion will never get a match for notifications.

Fortunately, this (`personal-eu-org`) was an umbrella criterion for:
  - personal-eu-org-process
  - personal-eu-org-use
  - personal-eu-org-provide

This migration updates the subscriber lists that have the deprecated criterion and replaces that specific one with the other three.

The CSV was generated by running a [report](https://github.com/alphagov/email-alert-api/blob/master/lib/tasks/report.rake#L9) against the `personal-eu-org` criterion, then creating a new CSV using the `slug`, `matching_criteria` columns, and adding a new column which is a copy of the `matching_criteria` column but with the new criteria swapped in.

https://trello.com/c/NSgSLzOS/1341-review-wording-of-data-sharing-question-in-the-checker


hannako (approved):
How did you run the report for the criterion? Tags pattern?

PR #1609, opened 2021-03-29
Convert to date in SubscriberListsReport

barrucadu:
A time-at-midnight-with-a-timezone compares as less than a date:

    irb(main):001:0> with_ts = Time.zone.parse(Time.zone.today.to_s)
    => Mon, 29 Mar 2021 00:00:00.000000000 BST +01:00

    irb(main):002:0> without_ts = Time.zone.today
    => Mon, 29 Mar 2021

    irb(main):003:0> with_ts < without_ts
    => true

So the check that the date must be in the past isn't working right: a
test that `Time.zone.today` gets rejected is failing, because the time
is being converted through `to_s`/`Time.zone.parse`, and ends up as
comparing less than its original value.

This report uses the date, not the time, so convert the input back
into a timezone-less date object.

Using `Time.zone.today.in_time_zone` for the comparison instead also
works.  I don't think there's a difference in practice, because the
`@date` gets used in database queries in the context of
`beginning_of_day` and `end_of_day`, which are timezone aware:

    irb(main):001:0> Time.zone.today.beginning_of_day
    => Mon, 29 Mar 2021 00:00:00.000000000 BST +01:00
    irb(main):002:0> Time.zone.today.in_time_zone.beginning_of_day
    => Mon, 29 Mar 2021 00:00:00.000000000 BST +01:00

    irb(main):003:0> Time.zone.today.end_of_day
    => Mon, 29 Mar 2021 23:59:59.999999999 BST +01:00
    irb(main):004:0> Time.zone.today.in_time_zone.end_of_day
    => Mon, 29 Mar 2021 23:59:59.999999999 BST +01:00

### DIFF
--- a/lib/reports/subscriber_lists_report.rb
+++ b/lib/reports/subscriber_lists_report.rb
@@ -2,7 +2,7 @@ class Reports::SubscriberListsReport
   attr_reader :date, :slugs, :tags_pattern, :links_pattern, :headers
 
   def initialize(date, slugs: "", tags_pattern: nil, links_pattern: nil, headers: nil)
-    @date = Time.zone.parse(date)
+    @date = Date.parse(date)
     @slugs = slugs.split(",")
     @tags_pattern = tags_pattern
     @links_pattern = links_pattern
@@ -31,7 +31,6 @@ private
   end
 
   def validate_date
-    raise "Invalid date" if date.blank?
     raise "Date must be in the past" if date >= Time.zone.today
     raise "Date must be within a year old" if date <= 1.year.ago
   end
--- a/spec/lib/reports/subscriber_lists_report_spec.rb
+++ b/spec/lib/reports/subscriber_lists_report_spec.rb
@@ -90,11 +90,6 @@ RSpec.describe Reports::SubscriberListsReport do
       .to raise_error("Lists not found for slugs: other-list,list")
   end
 
-  it "raises an error if the date is invalid" do
-    expect { described_class.new("blahhh").call }
-      .to raise_error("Invalid date")
-  end
-
   it "raises an error if the date isn't in the past" do
     expect { described_class.new(Time.zone.today.to_s).call }
       .to raise_error("Date must be in the past")


edwardkerry (approved):

lib/reports/subscriber_lists_report.rb
```diff
@@ -2,7 +2,7 @@ class Reports::SubscriberListsReport
   attr_reader :date, :slugs, :tags_pattern, :links_pattern, :headers
 
   def initialize(date, slugs: "", tags_pattern: nil, links_pattern: nil, headers: nil)
-    @date = Time.zone.parse(date)
+    @date = Time.zone.parse(date)&.to_date
```
kevindew, line 5: Could this be `Date.parse(date)` ?

barrucadu: There's a little extra exception handling, to make `@date` `nil` rather than raise a `Date::Error`, I did that to be consistent with all the exception-raising validation logic being in `call`.

lib/reports/subscriber_lists_report.rb
```diff
@@ -2,7 +2,12 @@ class Reports::SubscriberListsReport
   attr_reader :date, :slugs, :tags_pattern, :links_pattern, :headers
 
   def initialize(date, slugs: "", tags_pattern: nil, links_pattern: nil, headers: nil)
-    @date = Time.zone.parse(date)
+    begin
+      @date = Date.parse(date)
+    rescue Date::Error
+      @date = nil
+    end
```
kevindew, line 9: Ah sorry I hadn't understood the delicate balance this has. 

It seems a touch convoluted to me to deal with a nil here that only raises an exception later. So this could be:

```
    begin
      @date = Date.parse(date)
    rescue Date::Error
      raise "invalid date" 
    end
```

but I don't really think that's necessary as it doesn't provide any more or less information than the `Date::Error` exception of `Date::Error (invalid date)`. 

So I'd suggest doing just `Date.parse(date)` here and removing https://github.com/alphagov/email-alert-api/blob/913a48b2816248182b2ae357a2644ea58b0cc487/lib/reports/subscriber_lists_report.rb#L39 and the associated test.

kevindew (approved):

PR #1610, opened 2021-04-09
Decouple Raven from Email-alert-api

callumknights:
Removed direct references to Raven object and remaned appropriately

Trello: https://trello.com/c/S1QvqyCE/2459-decouple-govuk-apps-from-raven-3


thomasleese (commented):
I think the `config/initializers/raven.rb` file will need to be removed. I also think the convention would be to name this file `config/initializers/govuk_error.rb`. but I'm not sure it matters too much.

thomasleese (approved):

PR #1611, opened 2021-04-09
Bugfix: Fetch content changes for digest uniquely by subscription

bilbof:
This fixes a bug that would prevent a content change appearing in
a digest email.

Explanation:

Say you have two subscriptions `A` and `B`, and a piece of content `C`
that is changed twice in a week, generating content changes `Ca` and
`Cb`.

The query currently attempts to ensure that content change `Ca` is
not included in the digest, since we just want the most recent
content change, to keep emails short and sweet (and another reason
related to batching digests into a single email which I will get back
to).

The query uses `uniq(&:content_id)` to do ensure a piece of content
appears once in an email, even if there have been many changes to it.

However, the effect of this is that while change `Ca` is omitted from both
weekly digest emails (for subscriptions `A` and `B`), change `Cb` is
also omitted from the digest email for subscription `B`.

Fix:

To fix this I've changed the query so that content changes are distinct
based on content id and subscription id. Now change `Cb` will appear
in both digest emails for subscription `A` and `B`.

Addendum:

We had a user write in who experienced exactly this situation I described
above. It was pretty confusing to debug, and confusing for our subscriber
too!

I think folks will only have started noticing this in December, when
we stopped batching weekly digests into single emails [1]. Since,
even if one subscription didn't include the content change, the other
one would.

Corresponding ticket: https://govuk.zendesk.com/agent/tickets/4494504

[1] https://github.com/alphagov/email-alert-api/pull/1524

### DIFF
--- a/app/queries/digest_items_query.rb
+++ b/app/queries/digest_items_query.rb
@@ -35,7 +35,8 @@ private
       .where(subscriptions: { id: subscriptions })
       .where("content_changes.created_at >= ?", digest_run.starts_at)
       .where("content_changes.created_at < ?", digest_run.ends_at)
-      .uniq(&:content_id)
+      .order(created_at: :desc)
+      .uniq { |content| [content.content_id, content.subscription_id] }
       .group_by(&:subscription_id)
   end
 
--- a/spec/queries/digest_items_query_spec.rb
+++ b/spec/queries/digest_items_query_spec.rb
@@ -144,6 +144,52 @@ RSpec.describe DigestItemsQuery do
           )
       end
 
+      it "includes a content item uniquely per list" do
+        content_id = SecureRandom.uuid
+
+        content_change1 = create(
+          :content_change, :matched,
+          content_id: content_id,
+          subscriber_list: subscriber_list1,
+          created_at: digest_run.starts_at + 1.hour
+        )
+
+        create(
+          :content_change, :matched,
+          content_id: content_id,
+          subscriber_list: subscriber_list1,
+          created_at: digest_run.starts_at
+        )
+
+        content_change3 = create(
+          :content_change, :matched,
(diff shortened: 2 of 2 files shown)


kevindew (approved):
Great catch 👍 , this seems good to me.

I did struggle a bit to comprehend the situation described, but I think I do, I just wanted to clarify: Given I am signed up to get a daily email of all of [HMRC's publishings](https://www.gov.uk/email-signup?link=/government/organisations/hm-revenue-customs) and the [Paying HMRC](https://www.gov.uk/email-signup?link=/topic/dealing-with-hmrc/paying-hmrc) topic taxon. If HMRC publish a news article like [this](https://www.gov.uk/government/news/tax-cut-for-employers-of-veterans-brought-in) which can generate an email for both, then the current behaviour would result in only one of those subscriptions receiving the message, whereas we expect them both to?

bilbof:
@kevindew yes exactly.

In the support ticket the example given was essentially the same. A user signed up for a weekly email digest for the [Living in Austria](https://www.gov.uk/guidance/living-in-austria) page and [Brexit](https://www.gov.uk/transition) page, both of which are tagged to the Brexit taxon.

The user was confused that content changes to the Living in Austria page did not appear in the Brexit email digest, whereas changes for content such as 'Living in Germany' would.

The reason that was happening is that the digest query removes duplicate values using only the content_id:

```ruby
content_changes_in_period = [{content_id: 1, change: :a, sub: 1},{content_id: 1, change: :a, sub: 2}]
old_digest = content_changes_in_period.uniq { _1[:content_id] } # count = 1, loses second sub's digest content changes
new_digest = content_changes_in_period.uniq { [_1[:content_id], _1[:sub]] } # count = 2, all good
```

PR #1638, opened 2021-06-23
Add an endpoint to authenticate a subscriber using a GOV.UK Account

barrucadu:
This commit adds a new endpoint

    POST /subscribers/govuk-account

Which takes a `govuk_account_session` parameter, fetches the `email`
and `email_verified` attributes from account-api, and returns:

- 401 if the session identifier is invalid

- 403 if `email_verified` is false

- 404 if there is no subscriber with the same `email`

- 200 and the subscriber details otherwise

The 403, 404, and 200 responses may also include a new
`govuk_account_session` in the JSON response body.

---

[Trello card](https://trello.com/c/qPIGG5rp/872-backend-work-to-allow-logging-in-to-email-alert-api-with-an-account)

### DIFF
--- a/app/controllers/subscribers_govuk_account_controller.rb
+++ b/app/controllers/subscribers_govuk_account_controller.rb
@@ -1,0 +1,41 @@
+class SubscribersGovukAccountController < ApplicationController
+  before_action :validate_params
+
+  def authenticate
+    account_response = GdsApi.account_api.get_attributes(
+      attributes: %i[email email_verified],
+      govuk_account_session: expected_params[:govuk_account_session],
+    )
+
+    email_verified = account_response.dig("values", "email_verified")
+    email = account_response.dig("values", "email")
+
+    api_response = { govuk_account_session: account_response["govuk_account_session"] }.compact
+
+    render status: :forbidden, json: api_response and return unless email_verified
+    render status: :not_found, json: api_response and return unless email
+
+    subscriber = Subscriber.find_by_address(email)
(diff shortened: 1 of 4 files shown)


app/controllers/subscribers_govuk_account_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      govuk_account_session: expected_params[:govuk_account_session],
+    )
+
+    email_verified = account_response.dig("values", "email_verified")
+    email = account_response.dig("values", "email")
+
+    api_response = { govuk_account_session: account_response["govuk_account_session"] }.compact
+
+    render status: :forbidden, json: api_response and return unless email_verified
+    render status: :not_found, json: api_response and return unless email
```
huwd, line 16: Interesting, I like this being here as a defensive measure but is there a scenario where we could actually see this triggered so far?

- we require an email to sign up (currently?)
- so... unless someone has tampered with a payload (in which case `GdsApi::HTTPUnauthorized` anyway?)?

huwd (approved):

app/controllers/subscribers_govuk_account_controller.rb
barrucadu: I think at the moment this could only trigger if there's a bug in the account manager and so it hasn't sent the attribute to the attribute service or account-api.

huwd: great to have it there anyway incase we screw something up later.

PR #1640, opened 2021-06-24
Single content notifications

huwd:
[Trello](https://trello.com/c/3oXvMEPj/875-backend-work-to-allow-signing-up-to-notifications-for-a-single-content-item)

We are experimenting with allowing users to montior single pages for changes.
This will require a new kind of subscription list which can be created, queried and trigger content change notifications based on `content_id`

That's what this PR adds.

### DIFF
--- a/app/queries/find_exact_query.rb
+++ b/app/queries/find_exact_query.rb
@@ -2,6 +2,7 @@ class FindExactQuery
   class InvalidFindCriteria < StandardError; end
 
   def initialize(
+    content_id:,
     tags:,
     links:,
     document_type:,
@@ -9,6 +10,7 @@ class FindExactQuery
     government_document_supertype:,
     slug: nil
   )
+    @content_id = content_id.presence
     @tags = tags.deep_symbolize_keys
     @links = links.deep_symbolize_keys
     @document_type = document_type
@@ -29,6 +31,7 @@ private
   def base_scope
     @base_scope ||= begin
       scope = SubscriberList
+        .where(content_id: @content_id)
         .where(document_type: @document_type)
         .where(email_document_supertype: @email_document_supertype)
         .where(government_document_supertype: @government_document_supertype)
--- a/db/migrate/20210608161920_add_content_id_to_subscriber_lists.rb
+++ b/db/migrate/20210608161920_add_content_id_to_subscriber_lists.rb
@@ -1,0 +1,6 @@
+class AddContentIdToSubscriberLists < ActiveRecord::Migration[6.1]
+  def change
+    add_column :subscriber_lists, :content_id, :uuid
+    add_index :subscriber_lists, :content_id
+  end
+end
--- a/spec/queries/find_exact_query_spec.rb
+++ b/spec/queries/find_exact_query_spec.rb
@@ -107,6 +107,12 @@ RSpec.describe FindExactQuery do
       expect(query.exact_match).to eq(subscriber_list)
     end
 
+    it "matched when subscriber list has the same tags and matching content_id" do
+      query = build_query(links: { policies: { any: %w[beer] } }, content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+      subscriber_list = create_subscriber_list(links: { policies: { any: %w[beer] } }, content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+      expect(query.exact_match).to eq(subscriber_list)
+    end
+
     it "not matched when subscriber list has different links" do
       query = build_query(links: { policies: { any: %w[aa-11] } })
       _subscriber_list = create_subscriber_list(links: { policies: { any: %w[11-aa] } })
@@ -128,6 +134,12 @@ RSpec.describe FindExactQuery do
     it "not matched on document type - even if they match" do
       query = build_query(links: { policies: { any: %w[aa-11] } }, document_type: "travel_advice")
       _subscriber_list = create_subscriber_list(document_type: "travel_advice")
+      expect(query.exact_match).to be_nil
+    end
+
+    it "not matched on content_id - even if they match" do
+      query = build_query(links: { policies: { any: %w[beer] } }, content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+      _subscriber_list = create_subscriber_list(content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
       expect(query.exact_match).to be_nil
     end
   end
@@ -241,6 +253,12 @@ RSpec.describe FindExactQuery do
       expect(query.exact_match).to eq(subscriber_list)
     end
 
+    it "matched when subscriber list has the same tags and matching content_id" do
+      query = build_query(tags: { policies: { any: %w[beer] } }, content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+      subscriber_list = create_subscriber_list(tags: { policies: { any: %w[beer] } }, content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+      expect(query.exact_match).to eq(subscriber_list)
+    end
+
     it "not matched when subscriber list has different tags" do
       query = build_query(tags: { policies: { any: %w[beer] } })
       _subscriber_list = create_subscriber_list(tags: { policies: { any: %w[cider] } })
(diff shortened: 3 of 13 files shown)


huwd:
Much of this should look very familliar from your spike @barrucadu

app/queries/find_exact_query.rb
```diff
@@ -2,13 +2,15 @@ class FindExactQuery
   class InvalidFindCriteria < StandardError; end
 
   def initialize(
+    content_id:,
```
barrucadu, line 5: In commit 69becf60384cec34f3e25d8b04ea02afafdfc276, could you add `content_id: nil` to the `find_exact_query_params` in the controller, so that the controller specs pass?  It's a small change, but it would make this commit bisectable..

spec/queries/find_exact_query_spec.rb
```diff
@@ -130,6 +136,12 @@
       _subscriber_list = create_subscriber_list(document_type: "travel_advice")
       expect(query.exact_match).to be_nil
     end
+
+    it "not matached on content_id - even if they match" do
```
barrucadu, line 140: *matched

spec/queries/find_exact_query_spec.rb
```diff
@@ -258,6 +276,12 @@
...
```
barrucadu, line 280: *matched

spec/queries/find_exact_query_spec.rb
```diff
@@ -290,8 +314,21 @@
...
   end
 
+  it "matched on content_id only" do
+    query = build_query(content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+    subscriber_list = create_subscriber_list(content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+    expect(query.exact_match).to eq(subscriber_list)
+  end
+
+  it "not matched on different content_id" do
+    query = build_query(tags: { policies: { any: %w[beer] } }, content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
```
barrucadu, line 324: Remove the `tags`, as currently I think this is testing the same thing as the tags + content_id matching above - but really it should just be testing the content_id.

spec/queries/subscriber_list_query_spec.rb
```diff
@@ -1,6 +1,7 @@
 RSpec.describe SubscriberListQuery do
```
barrucadu, line 1: Could you add a test for `content_id: ""`?  That's a valid thing which can be sent to the API, based on the docs, so we need to make sure that behaves the same as `content_id: nil`.

huwd: [How's this?](https://github.com/alphagov/email-alert-api/pull/1640/commits
/3f918b552120f3bd473f7e7e6dd0aaa04cb43d67#diff-fae1d811d2f95c3d665906fe8e504e659a3026697b594b5e3aeb5b5607f2da69R16)

`it { is_included_in_links tags_or_links, content_id: nil}` returns the same so they should be the same.
however as nil is in the defaults i don't think we just stick with this new L16

barrucadu: Yeah, that's good

barrucadu: Aha, I see, `""` is turned into `nil` when matching a UUID field:

```
irb(main):003:0> puts SubscriberList.where(content_id: [nil, ""]).to_sql
SELECT "subscriber_lists".* FROM "subscriber_lists" WHERE ("subscriber_lists"."content_id" = NULL OR "subscriber_lists"."content_id" IS NULL)
```

Looks like the same happens for any blank value:

```
irb(main):006:0> puts SubscriberList.where(content_id: [false, "     "]).to_sql
SELECT "subscriber_lists".* FROM "subscriber_lists" WHERE "subscriber_lists"."content_id" IN (NULL)
```

barrucadu: Hmm, that doesn't actually look like correct behaviour from ActiveRecord though, because `NULL = NULL` is false in SQL (SQL uses a 3-valued logic):

```ruby
irb(main):010:0> SubscriberList.create!(title: "test", slug: "slug")

irb(main):015:0> pp SubscriberList.where(content_id: "").to_a ; nil
  SubscriberList Load (0.4ms)  SELECT "subscriber_lists".* FROM "subscriber_lists" WHERE "subscriber_lists"."content_id" = $1  [["content_id", nil]]
[]

irb(main):018:0> pp SubscriberList.where(content_id: nil).to_a ; nil
  SubscriberList Load (0.4ms)  SELECT "subscriber_lists".* FROM "subscriber_lists" WHERE "subscriber_lists"."content_id" IS NULL
[#<SubscriberList:0x00005617a3f12398
  id: 2,
  title: "test",
  created_at: Fri, 25 Jun 2021 13:18:18.062446000 BST +01:00,
  updated_at: Fri, 25 Jun 2021 13:18:18.062446000 BST +01:00,
  document_type: "",
  tags: {},
  links: {},
  email_document_supertype: "",
  government_document_supertype: "",
  signon_user_uid: nil,
  slug: "slug",
  url: nil,
  tags_digest: nil,
  links_digest: nil,
  content_id: nil>]

irb(main):019:0> pp SubscriberList.where(content_id: ["", "     "]).to_a ; nil
  SubscriberList Load (0.4ms)  SELECT "subscriber_lists".* FROM "subscriber_lists" WHERE "subscriber_lists"."content_id" IN (NULL)
[]
```

db/migrate/20210608161920_add_content_id_to_subscriber_lists.rb
```diff
@@ -0,0 +1,5 @@
+class AddContentIdToSubscriberLists < ActiveRecord::Migration[6.1]
+  def change
+    add_column :subscriber_lists, :content_id, :uuid
```
barrucadu, line 3: I think we want an index on this field too, or matching content to these subscriber lists will be slow

spec/queries/find_exact_query_spec.rb
```diff
@@ -290,8 +314,21 @@
...
+    subscriber_list = create_subscriber_list(content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+    expect(query.exact_match).to eq(subscriber_list)
+  end
+
+  it "not matched on different content_id" do
+    query = build_query(content_id: "11a51f5e-3f17-4516-87c7-02afba87ef40")
+    _subscriber_list = create_subscriber_list(content_id: "ef8a3fd2-2b15-4c0e-a2d4-2a34187f337a")
+    expect(query.exact_match).to be_nil
+  end
+
```
barrucadu, line 328: Can you add a test to this file that a find_exact query with `content_id: ""` matches a subscriber_list with `content_id: nil`?

```ruby
  it "matched on blank content_id" do
    query = build_query(content_id: "")
    subscriber_list = create_subscriber_list(content_id: nil)
    expect(query.exact_match).to eq(subscriber_list)
  end
```

This currently fails.  I think in the initialiser we'll want

```ruby
@content_id = content_id.blank? ? nil : content_id
```

spec/queries/subscriber_list_query_spec.rb
barrucadu: I think there's another change we need to make for `FindExactQuery` (see other comment)

barrucadu: I'll open an issue on Rails to ask about why it's trying to compare `NULL`s for equality

spec/queries/find_exact_query_spec.rb
barrucadu: Ah, `presence` works here:

```ruby
@content_id = content_id.presence
```

spec/queries/subscriber_list_query_spec.rb
huwd: nb: https://github.com/rails/rails/issues/42600

db/migrate/20210608161920_add_content_id_to_subscriber_lists.rb
```diff
@@ -0,0 +1,6 @@
...
+    add_index :subscriber_lists, :content_id, unique: true
```
barrucadu, line 4: Don't make this a `unique` index.  I can't think why we'd want two subscriber lists matching the same content ID, but I'm not confident enough in how email-alert-api works to say we can rule that out.

barrucadu (approved):
Looks good to me

PR #1644, opened 2021-06-29
Link subscribers to a GOV.UK Account & add endpoints to manage them

barrucadu:
This PR adds a `govuk_account_id` field to the `Subscriber` model, which gets returned from all the endpoints which return a `Subscriber`  (by virtue of not excluding it in `as_json`), and adds two endpoints to manage such users:

- `POST /subscribers/govuk-account/link` (with a `govuk_account_session`) looks up the GOV.UK Account and subscriber (in the same way as the auth endpoint), and then "links" them by saving the `govuk_account_id` against the `Subscriber` record.
- `GET /subscribers/govuk-account/:govuk_account_id` (without a `govuk_account_session`) looks up a subscriber by `govuk_account_id` and returns it if there is one.

The former endpoint will be called when we migrate a user over.  The latter endpoint will be used by account-api to know whether to update a user's email address:

1. User changes email address in Identity Provider
2. Identity Provider notifies Account API
3. Account API queries Email Alert API for the subscriber
4. If there is one, Account API calls Email Alert API to update the
address of the subscriber

---

### DIFF
--- a/app/controllers/subscribers_govuk_account_controller.rb
+++ b/app/controllers/subscribers_govuk_account_controller.rb
@@ -1,28 +1,40 @@
 class SubscribersGovukAccountController < ApplicationController
   before_action :validate_params
 
-  def authenticate
-    account_response = GdsApi.account_api.get_attributes(
-      attributes: %i[email email_verified],
+  before_action do
+    account_response = GdsApi.account_api.get_user(
       govuk_account_session: expected_params[:govuk_account_session],
     )
 
-    email_verified = account_response.dig("values", "email_verified")
-    email = account_response.dig("values", "email")
+    @govuk_account_id = account_response["id"]
+    email = account_response["email"]
+    email_verified = account_response["email_verified"]
 
-    api_response = { govuk_account_session: account_response["govuk_account_session"] }.compact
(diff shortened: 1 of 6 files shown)


app/controllers/subscribers_govuk_account_controller.rb
```diff
@@ -1,29 +1,41 @@
...
+    @api_response = { govuk_account_session: account_response["govuk_account_session"] }.compact
 
-    render status: :forbidden, json: api_response and return unless email_verified
-    render status: :forbidden, json: api_response and return unless email
+    if !email
+      render status: :forbidden, json: @api_response
+    elsif !email_verified
+      render status: :forbidden, json: @api_response
+    else
+      @subscriber = Subscriber.resilient_find_or_create(email, signon_user_uid: current_user.uid)
```
huwd: Where are thse consumed as instance ariables?
I can see L27 and L32, but they seem to be taking regular variables?

And i can't see an attr_reader nywhere?

barrucadu: There's an `attr_reader` line after the `private`

barrucadu:
Tests now passing with the updated gds-api-adapters

huwd (approved):

PR #1646, opened 2021-07-01
Forbid logging in with an auth token for GOV.UK Account users

barrucadu:
We've decided that, if you have linked your Subscriber record to your
GOV.UK Account, then the account should be the only way to
authenticate yourself to get into email-alert-frontend.

---

### DIFF
--- a/app/controllers/subscribers_auth_token_controller.rb
+++ b/app/controllers/subscribers_auth_token_controller.rb
@@ -3,6 +3,8 @@ class SubscribersAuthTokenController < ApplicationController
 
   def auth_token
     subscriber = Subscriber.find_by_address!(expected_params[:address])
+    head :forbidden and return if subscriber.linked_to_govuk_account?
+
     token = generate_token(subscriber)
     email = build_email(subscriber, token)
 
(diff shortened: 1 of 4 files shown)


app/controllers/subscribers_auth_token_controller.rb
```diff
@@ -3,6 +3,8 @@ class SubscribersAuthTokenController < ApplicationController
 
   def auth_token
     subscriber = Subscriber.find_by_address!(expected_params[:address])
+    head :forbidden and return if subscriber.linked_to_govuk_account?
```
huwd, line 6: We've used this patten everywhere and I generally like it.
It's pretty human readable.

But it just occured to me, i'm not sure if that's just a style preference or is there actually a difference when compared to:


```
return head :forbidden if subscriber.linked_to_govuk_account?
```

Remind me how we ended up using this so much? As I recall it was somthing neat you or Bruce found?

huwd (approved):

app/controllers/subscribers_auth_token_controller.rb
barrucadu: I think Bruce came up with it.  I personally find it more readable than `return head :forbidden`, because that to me reads like it should be `return(head) :forbidden` which would be a syntax error... but Ruby actually seems to handle it just fine.

huwd: ok great, so both work but we prefer this for readability.

I agree, it's nice! just wanted t remind myself why we'd standarised around it!

PR #1647, opened 2021-07-01
Allow merging subscribers on address clash

barrucadu:
As part of bringing notifications and accounts together, we need to be
able to automatically handle the case where a user:

1. Has notifications set up for `foo` and `bar`
2. Registers a GOV.UK account to manage their notifications for `foo`
3. Changes the email address of their account to `bar`

Currently, email-alert-api will reject this change: it doesn't let you
change to an email address that's already in use.

But for the change to become active, the user has proven to the
account that they own both `foo` (we're not letting them do any
notifications things without confirming their email address) *and*
`bar` (we won't update the address until it's been confirmed).

We've decided in this case to merge the two subscribers in
email-alert-api.  This is pretty straightforward, other than the case
where both subscribers have an active subscription to the same topic,
with different frequencies.  We've decided to keep the most frequent
if that happens.  The other is marked as ended with a new
`subscriber_merged` reason, so we can see how much this happens.

This new behaviour is only used if `on_conflict=merge` is passed along
as a parameter.  We'll make account-api do that, but
email-alert-frontend will retain the old behaviour.

---

### DIFF
--- a/app/controllers/subscribers_controller.rb
+++ b/app/controllers/subscribers_controller.rb
@@ -11,7 +11,18 @@ class SubscribersController < ApplicationController
   end
 
   def change_address
-    subscriber.update!(address: new_address)
+    Subscriber.transaction do
+      if subscriber_params[:on_conflict] == "merge"
+        MergeSubscribersService.call(
+          subscriber_to_keep: subscriber,
+          subscriber_to_absorb: Subscriber.find_by_address(new_address),
+          current_user: current_user,
+        )
+      end
+
+      subscriber.update!(address: new_address)
+    end
+
     render json: { subscriber: subscriber }
   end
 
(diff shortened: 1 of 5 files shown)


huwd (approved):
Nice, thanks for the really clear commit message.

The Service module and tests are also good, I like them - they are clear and leave the actual change to the controller nice and simple.

Looks great!

app/controllers/subscribers_controller.rb
```diff
@@ -11,16 +11,18 @@ def subscriptions
   end
 
   def change_address
-    if subscriber_params[:on_conflict] == "merge"
-      Subscriber.transaction do
+    Subscriber.transaction do
```
huwd, line 14: Interesting what is the thinking behind moving the if statement into the transaction?
I'm guessing it still does nothing if the `if` doesn't fire, but will rollback if the call doesn't work?

barrucadu: It's so that the `subscriber.update` on line 23 happens in the same transaction as the merging

huwd (approved):

PR #1670, opened 2021-10-18
Send an email on linking to a GOV.UK account

barrucadu:
After a subscriber gets linked to a GOV.UK account, they will have to
log in via the account to manage their subscriptions.  So, if they
already had active subscriptions, send them an email telling them that
what they are used to has changed.

---

[Trello card](https://trello.com/c/wUYJLuJ1/1064-implement-confirm-you-want-to-get-emails-page-account-linking-logic)


huwd (approved):

app/builders/linked_account_email_builder.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+    items.join("\n")
+  end
+
+  def sign_in_url
+    GovukPersonalisation::Urls.sign_in
+  end
+
+  def contact_url
+    GovukPersonalisation::Urls.feedback
```
barrucadu, line 59: Before merging this I'll check with Ale which of the feedback forms this should be - the one we link to from the footer, or the one we link to from the beta banner.

PR #1676, opened 2021-11-09
Brexit checker retirement email

sihugh:
We need to be able to send separate emails to subscribers who have an account and those who don't, as the behaviour of the checker is different.

In the first commit we introduce the ability to filter out any emails that are in a particular file. We could do this by querying the account API, but that would be extremely slow.

In the second commit we create another task that is identical to the first one, but allows us to only send an email if the same email address file _includes_ their address. In this way, we will split the emails, and only send one email per address.

This will be reverted by the end of the month when the Brexit checker gets retired.


hannako (approved):
👍

PR #1681, opened 2021-12-01
Add task unsubscribing all from Brexit checker

sihugh:
The Brexit checker has been retired and will not be sending out any further updates. Because of these, we're closing all the subscriptions.

Once we're totally happy, then we'll delete them altogether, but we're not quite there yet.

This task is intended to be short-lived!

### DIFF
--- a/lib/tasks/support.rake
+++ b/lib/tasks/support.rake
@@ -91,6 +91,16 @@ namespace :support do
     end
   end
 
+  desc "Unsubscribe all Brexit checker subscriptions"
+  task unsubscribe_all_brexit_checker_subscriptions: :environment do
+    unsubscribe_time = Time.zone.now
+
+    brexit_subscriptions = Subscription.joins(:subscriber_list).where("ended_at IS NULL").where("subscriber_lists.tags ->> 'brexit_checklist_criteria' IS NOT NULL")
+
+    puts "Unsubscribing #{brexit_subscriptions.count} subscriptions"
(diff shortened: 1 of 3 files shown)


lib/tasks/support.rake
```diff
@@ -91,6 +91,16 @@ namespace :support do
     end
   end
 
+  desc "Unsubscribe all Brexit checker subscriptions"
+  task unsubscribe_all_brexit_checker_subscriptions: :environment do
+    unsubscribe_time = Time.zone.now
+
+    brexit_subscriptions = Subscription.joins(:subscriber_list).where("ended_at IS NULL").where("subscriber_lists.tags::text like '%brexit_checklist_criteria%'")
```
huwd, line 98: What's the format of a subscriber_list tag name?

I thought it was: `brexit_checker_criteria-abc1231231` or something?
if so could we drop the first % or is there something before it?

I'm always unnessiarily nevous of fuzzy matching strings that could be in the middle of others.
Matching from a known-start to unknown end... feels a tad safer, if we can get away with it?

huwd (approved):
One thought, but it's non-blocking

lib/tasks/support.rake
sihugh: Ah, that's a slug you're thinking of.  The tags look like:

`{ brexit_checklist_criteria: { any: ["travelling", "nationality-eu"] } }`

It's a good shout though and I'll think on it

sihugh: Have done a better query on the tags now I think

huwd (approved):

PR #1687, opened 2021-12-09
Change the sign in link in linked emails to the production home

barrucadu:
[Trello](https://trello.com/c/1tMOYsbk/1218-change-sign-in-link-to-use-di-domain)

This is a nicer URL than `/sign-in/redirect`, and it doesn't really
matter if emails in non-production environments have the wrong link.

---

[Trello card](https://trello.com/c/1tMOYsbk/1218-change-sign-in-link-to-use-di-domain)


app/builders/linked_account_email_builder.rb
```diff
@@ -52,7 +50,7 @@ def bulleted_active_subscriptions
   end
 
   def sign_in_url
-    GovukPersonalisation::Urls.sign_in
+    "https://www.gov.uk/account"
```
huwd, line 53: Humm, we have had cause before to see the output of the email - I think in a log or /tmp text format.
This would make testing things harder on local, as well as not possible to follow links through from integration and staging, wouldn't it?

barrucadu:
We're going to make the normal sign-in route `/account` instead.

PR #1690, opened 2021-12-15
Modify `utm_campaign` to show notification type

huwd:
[trello](https://trello.com/c/36JCWgL3/1138-annotate-notification-links-with-notification-type)

This is so we can compare the click-through rates for the existing
topic-based subscriptions and the new single-page subscriptions.

The default remains `govuk-notifications`.

### DIFF
--- a/spec/presenters/content_change_presenter_spec.rb
+++ b/spec/presenters/content_change_presenter_spec.rb
@@ -1,5 +1,6 @@
 RSpec.describe ContentChangePresenter do
-  let(:subscription) { create(:subscription, frequency: "immediately") }
+  let(:subscriber_list) { create(:subscriber_list) }
+  let(:subscription) { create(:subscription, subscriber_list: subscriber_list) }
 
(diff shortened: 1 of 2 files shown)


spec/presenters/content_change_presenter_spec.rb
```diff
@@ -44,6 +55,27 @@
       expect(result).to eq(expected.strip)
     end
 
+    context "when subscriber list is for a content id" do
+      let(:subscriber_list) { create(:subscriber_list, content_id: SecureRandom.uuid) }
+
+      it "generates a url with utm_list_type=single-page" do
```
barrucadu, line 61: This test name is wrong, should be `utm_campaign=govuk-notifications-single-page`

huwd: ah, great spot - missed that in the title.
Done!

barrucadu (approved):

PR #1693, opened 2022-01-05
Use text type for long varchar database fields

kevindew:
Trello: https://trello.com/c/GRGnnikf/63-plan-production-upgrade-for-email-alert-api-postgresql-database

This change has been made to resolve a problem experienced when using
this application with AWS Database Migration Service. We have that when
the schema has these has VARCHAR type the service will only migrate the
first 510 characters, whereas if they are TEXT there are no problems.

Aside from the above this change has very little impact on the
application, behind the scenes PostgreSQL represents VARCHAR and TEXT
very similarly [1]. Arguably TEXT is a more semantic choice for these long
values.

[1]: https://stackoverflow.com/questions/4848964/difference-between-text-and-varchar-character-varying

### DIFF
--- a/db/migrate/20220105173124_use_text_for_long_subscriber_lists_columns.rb
+++ b/db/migrate/20220105173124_use_text_for_long_subscriber_lists_columns.rb
@@ -1,0 +1,15 @@
+class UseTextForLongSubscriberListsColumns < ActiveRecord::Migration[6.1]
+  def up
+    change_table :subscriber_lists, bulk: true do |t|
+      t.change :title, :text
+      t.change :url, :text
+    end
+  end
+
+  def down
+    change_table :subscriber_lists, bulk: true do |t|
+      t.change :title, :string
+      t.change :url, :string
+    end
+  end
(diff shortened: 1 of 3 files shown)


db/migrate/20220105173124_use_text_for_long_subscriber_lists_columns.rb
```diff
@@ -0,0 +1,15 @@
...
+    change_table :subscriber_lists, bulk: true do |t|
+      t.change :title, :text
+      t.change :url, :text
+    end
+  end
+
+  def down
+    change_table :subscriber_lists, bulk: true do |t|
+      t.change :title, :text
+      t.change :url, :text
```
barrucadu, line 12: Should these `:text`s be `:string` in this `down` migration?

kevindew: Oops, yes. Thanks for catching.

barrucadu (approved):

PR #1694, opened 2022-01-06
Add endpoint to unsubscribe all users from a list

barrucadu:
We're going to use this with the single-page notifications feature to
alert a user when a page is removed or redirected, and remove their
now-dead current subscription.

I went back and forth on the naming for this a bit.  I originally used
"destroy", but actually we're not destroying the list here, we're just
unsubscribing everyone and (optionally) sending an email.  We could
destroy the list, but there isn't really any need to, as it'll be
cleared out by the HistoricDataDeletionWorker - and in fact I'm wary
of adding a new way of deleting data when there is already an
established pattern people may expect to hold.

So I went for "bulk-unsubscribe" instead.

If a message body is given, a sender_message_id is also required.
This is to ensure we don't send multiple unsubscribe emails if the
same publishing event is processed twice (this matches the behaviour
of content changes).  But the sender_message_id has to be a UUID.

To generate this UUID, generate a v5 UUID with:

- namespace: content_id
- name: public_updated_at

ActiveSupport has a helpful Digest::UUID.uuid_v5 method to do this.

---

[Trello card](https://trello.com/c/mQNmwCo2/1140-send-users-an-email-when-a-single-page-is-unpublished)

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -31,6 +31,27 @@ class SubscriberListsController < ApplicationController
     render json: subscriber_list.to_json
   end
 
+  def bulk_unsubscribe
+    subscriber_list = SubscriberList.find_by(slug: params[:slug])
+
+    if subscriber_list.nil?
+      render json: { error: "Could not find the subscriber list" }, status: :not_found
+    elsif bulk_unsubscribe_params[:body] && !bulk_unsubscribe_params[:sender_message_id]
+      render json: { error: "Unprocessable entity" }, status: :unprocessable_entity
+    elsif bulk_unsubscribe_already_requested?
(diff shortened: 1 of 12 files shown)


huwd (approved):

huwd:
Looks good, pending: https://github.com/alphagov/email-alert-api/pull/1696

huwd (changes requested):
Oh just realised are we missing this in /docs/api?

barrucadu:
Pushed two fixup commits where I've:

- Made it return a 202 (rather than a 204), to match the messages endpoint
- Rearranged the tests slightly so that message creation is tested regardless of whether or not there's a user
- Added the missing `GET /subscriber-lists/xxx` endpoint to the docs
- Added our new endpoint to the docs

huwd (approved):

PR #1696, opened 2022-01-13
Add ADR 010 - reinstate unpublish emails for single page subscriptions

huwd:
[trello](https://trello.com/c/mQNmwCo2/1140-send-users-an-email-when-a-single-page-is-unpublished)
Moved repo from [email-alert-service](1) for previous discssion.

[1]: https://github.com/alphagov/email-alert-service/pull/480

### DIFF
--- a/docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
+++ b/docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
@@ -1,0 +1,79 @@
+# Decision record: reinstate unpublish emails for single page subscriptions
+
+Date: 2021-12-03
+
+## Context
+
+We have just launched a new email notifications feature that allows users to subscribe to updates on a single page.
+Currently this feature is only enabled on a few selected pages (eg. [Open standards for government][open-standards-for-government]),
+but it is expected to roll out to the majority of pages and content types by early 2022.
+
+Unpublished content will become unavailable and receive no further updates. Users who have demonstrated a desire to stay up to date
+with an item of content may find it unexpected if content they were subscribed to disappears without notification, or becomes unavailable through their account. This change is part of meeting a user need to stay up to date with changing content, including when that content is changed through unpublication.
+
+The email alert API used to send emails when a page in a taxon subscription was unpublished.
+This feature was [removed in Jan 2021][remove-previous-unpublish-emails] for good reasons.
+
+[open-standards-for-government]: https://www.gov.uk/government/publications/open-standards-for-government
+[remove-previous-unpublish-emails]: https://github.com/alphagov/email-alert-api/issues/1572
+
+## Decision
+
+We will add an endpoint to email-alert-api to unsubscribe everyone from a given subscriber list, and optionally send them an email with a reason why (using the existing [message](https://github.com/alphagov/email-alert-api/blob/main/docs/adr/adr-004-message-concept.md) machinery).  We will change email-alert-service to call this new endpoint when a page gets unpublished, to notify any users subscribed to that specific page.
+
+We will not delete the now-empty list, as it will be cleaned up by the [HistoricalDataDeletionWorker](https://github.com/alphagov/email-alert-api/blob/main/app/workers/historical_data_deletion_worker.rb), and introducing another mechanism that destroys lists could be a point of confusion.
+
+### Addressing the concerns from the previous version
+
+The previous version of this feature was removed for several good reasons. We believe readding the unpublish emails only for single page subscriptions addresses these concerns:
+
+#### It only worked for topic taxons
+
+Single page notification work generates a clearer user need to inform about subscriptions in the more frequent scenario that individual pages are unpublished.
+
+Given the specificity of the need that drives this, we don't think limiting it to just those subscriptions is a problem.
+
+#### It only worked if the unpublished page was redirected
+
+The new unpublished emails will handle pages that are redirected or published in error with or without an alternative URL. This covers the publisher-facing [unpublishing types][unpublishing-types].
+
+#### It wasn't monitored so we didn't know if it was working
+
+We don't have a specific plan for monitoring this. Ideally monitoring for this would be taken on by a team that owns this portion of the system in the log run, and incorporated into their usual service monitoring.
+
+#### It rendered email using ERB, unlike the rest of the system
+
+The new unpublished emails will use the same method to render emails as the messages endpoint.
+
+#### It removed any subscriber lists that included the page, even those for a different taxon
+
+We are taking a very different approach from the previous unpubication of taxons. The previous route had to infer which lists should be removed as they represented a collection of materials.
+
+Here we are unpublishing a single page, and matching it to a subscription list by the same content ID. The email alert api doesn't need to do anything clever to match the two, we just tell it which subscription list should be cleared, and it does so.
+
+#### There was no evidence it met user needs
+
+We have user research demonstrating a need for users to stay up to date with specific items of government guidance. Examples might be recent coronavirus guidance that changes over time and can have a large impact on lives and what is possible to do.
+
+Unpublication is a form of change to that content that is currently unmonitored. This could have an impact in scenarios such as when a specific item of guidance is consolodated into a more general page when rules are made the same. A user may also be confused if an item they have subscribed to disappears from their account without notification, and may also wish to know if an item they subscribed to was removed as published in error.
+
+We have focused in on these as derived needs that emerge from the general user need to be kept up to date with individual pages.
+
+### Other related issues
+
+The email alert API has subscriber lists that will never send an email. This is documented as [GOV.UK Tech Debt][email-alert-api-dead-lists].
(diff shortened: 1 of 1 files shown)


huwd:
Moved this discussion from: https://github.com/alphagov/email-alert-service/pull/480

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,74 @@
...
+The email alert API used to send emails when a page in a taxon subscription was unpublished.
+This feature was [removed in Jan 2021][remove-previous-unpublish-emails] for good reasons.
+
+[open-standards-for-government]: https://www.gov.uk/government/publications/open-standards-for-government
+[remove-previous-unpublish-emails]: https://github.com/alphagov/email-alert-api/issues/1572
+
+## Decision
+
+We will reinstate emails when a page is unpublished, but only to users with a single page subscription to that page.
+We will also remove the subscriber list for a single page when it is unpublished to reduce the build up of empty lists in Email Alert API.
```
barrucadu, line 23: Given the behaviour of #1694, I think a better phrasing would be

> We will add an endpoint to email-alert-api to unsubscribe everyone from a given subscriber list, and optionally send them an email with a reason why (using the existing [message](https://github.com/alphagov/email-alert-api/blob/main/docs/adr/adr-004-message-concept.md) machinery).  We will change email-alert-service to call this new endpoint when a page gets unpublished, to notify any users subscribed to that specific page.
>
> We will not delete the now-empty list, as it will be cleaned up by the [HistoricalDataDeletionWorker](https://github.com/alphagov/email-alert-api/blob/main/app/workers/historical_data_deletion_worker.rb), and introducing another mechanism that destroys lists could be a point of confusion.

As we're intentionally not adding a concept of "unpublishing" to email-alert-api.  We're adding a more general thing, which email-alert-service (which *is* aware of the publishing workflow) uses to implement the behaviour we want.

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,74 @@
...
+
+The previous version of this feature was removed for several good reasons. We believe readding the unpublish emails only for single page subscriptions addresses these concerns:
+
+#### It only worked for topic taxons
+
+The new unpublished emails will only apply to single page subscriptions. This is a deliberate decision to keep the scope small and allow us to address the other concerns about handling an unpublished page in a taxon.
+
+#### It only worked if the unpublished page was redirected
+
+The new unpublished emails will handle pages that are redirected published in error with or without an alternative URL. This covers the publisher-facing [unpublishing types][unpublishing-types].
```
barrucadu, line 35: Missing an "or" between "redirected" and "published in error"

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### It wasn't monitored so we didn't know if it was working
+
+Because the new emails will only be for single page subscriptions, it will be easier to test the whole system. We will also use existing methods for templating and sending emails so we can benefit from the monitoring that's already in place.
```
barrucadu, line 39: I don't think we've actually addressed this concern until we implement monitoring.  So better to not make it sound like there's some existing monitoring which will automatically check this works.

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,74 @@
...
+We will reinstate emails when a page is unpublished, but only to users with a single page subscription to that page.
+We will also remove the subscriber list for a single page when it is unpublished to reduce the build up of empty lists in Email Alert API.
+
+### Addressing the concerns from the previous version
+
+The previous version of this feature was removed for several good reasons. We believe readding the unpublish emails only for single page subscriptions addresses these concerns:
+
+#### It only worked for topic taxons
+
+The new unpublished emails will only apply to single page subscriptions. This is a deliberate decision to keep the scope small and allow us to address the other concerns about handling an unpublished page in a taxon.
```
barrucadu, line 31: "We'll make it only work for Y instead" isn't a great response to "it only worked for X!"

Better to say something like

> We have a more clear user need that this feature is useful for subscriptions to individual pages, so the feature being limited to just those subscriptions isn't a problem.

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,74 @@
...
+
+Because the new emails will only be for single page subscriptions, it will be easier to test the whole system. We will also use existing methods for templating and sending emails so we can benefit from the monitoring that's already in place.
+
+#### It rendered email using ERB, unlike the rest of the system
+
+The new unpublished emails will use the same method to render emails as the content changes.
+
+#### It removed any subscriber lists that included the page, even those for a different taxon
+
+There's only one single page subscriber list for each page, so it's much easier to know which list to remove when a page is unpublished.
```
barrucadu, line 47: We're adopting a fundamentally different approach here: the old method took a taxon and tried to infer which lists should be removed.  But in #1694, we're adding an endpoint to remove a specific, identified, subscriber list.  There's no cleverness in figuring out which lists apply: it just removes the one list it's told to remove.

Explaining that is a much stronger argument than "there's only one list".

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### There was no evidence it met user needs
+
+Single page notifications are an answer to the user need to stay up to date with specific pieces of government guidance. Especially when the area might be subject to frequent updates or changes.
+
+Our hypothesis is that some users will subscribe to taxons that are broader than their actual area of interest. So they will receive updates about areas they are not interested in, in order to get the few relevant updates to them.
+
+Our hypothesis is that users who subscribe to be notified for major updates, will also want to know if the guidance is unpublished as part of staying up-to-date, and will find it unexpected if guidance they have been monitoring disappears from their account subscriptions without warning.
```
barrucadu, line 55: We've done user research on this.  Do we have something more concrete than a hypothesis?  eg, can we say "we've seen in user research sessions that being able to subscribe to specific pages of guidance is useful to many users, and meets a different need than subscribing to a group of pages (such as taxon), and that users with these subscriptions expect to be notified if the page is unpublished"?

huwd: Yeah that last phrase is key, do we actually have UR to that point?
I think it's a fairly solid assumption, but I wasn't sure if it's an expectation we generated or if it came our of UR?

barrucadu: I'm sure one of the UCD people would know

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Other related issues
+
+The email alert API has subscriber lists that will never send an email. This is documented as [GOV.UK Tech Debt][email-alert-api-dead-lists].
+This proposal does not directly address this Tech Debt, but by removing the single page subscriber lists as a page is unpublished we can try to prevent the impact of the tech debt from increasing.
```
alex9smith, line 65: I think we need to update this line now the email alert API endpoint no longer deletes the subscriber list.

Perhaps:
```suggestion
This proposal does not directly address this Tech Debt, but by removing subscriptions from the subscriber list once the emails have been sent the [HistoricalDataDeletionWorker](https://github.com/alphagov/email-alert-api/blob/main/app/workers/historical_data_deletion_worker.rb) can periodically remove the empty lists.
```

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
huwd: cool have reworded

docs/adr/adr-010-send-unpublish-emails-for-single-pages.md
```diff
@@ -0,0 +1,79 @@
...
+
+The new unpublished emails will handle pages that are redirected or published in error with or without an alternative URL. This covers the publisher-facing [unpublishing types][unpublishing-types].
+
+#### It wasn't monitored so we didn't know if it was working
+
+We don't have a specific plan for monitoring this. Ideally monitoring for this would be taken on by a team that owns this portion of the system in the log run, and incorporated into their usual service monitoring.
+
+#### It rendered email using ERB, unlike the rest of the system
+
+The new unpublished emails will use the same method to render emails as the content changes.
```
barrucadu, line 46: * as the messages endpoint.

huwd: ace amended

barrucadu (approved):

PR #1698, opened 2022-01-19
Fix sending unsubscribe notifications immediately

barrucadu:
We use the message machinery to send a notification to single-page
subscribers when the page gets unpublished and we unsubscribe them
from it.

We need these emails to go out immediately so we know when to
unsubscribe the user.  So in this commit, I add an
`override_subscription_frequency_to_immediate` field to the `messages`
table, and use that to skip the frequency scope check in the
`ImmediateEmailGenerationService`.

Alternative approaches would be:

- We could change the frequency of the `Subscription` when we process
the bulk-unsubscription.  This doesn't feel in keeping with the spirit
of email-alert-api, where normally data is not mutated in cases like
that.

- We could include unsubscription notifications in batches.  But then
we don't know when we can end the subscription (worst case we have to
wait a week), and if the page is gone anyway the other updates in the
batch are likely useless.

I think this is a least-bad approach.  But it is a feature we should
be careful not to abuse for other purposes.  We offer users frequency
controls for a reason.

---

[Trello card](https://trello.com/c/mQNmwCo2/1140-send-users-an-email-when-a-single-page-is-unpublished)

### DIFF
--- a/app/queries/digest_items_query.rb
+++ b/app/queries/digest_items_query.rb
@@ -45,6 +45,7 @@ private
       .select("messages.*, subscriptions.id AS subscription_id")
       .joins(matched_messages: { subscriber_list: :subscriptions })
       .where(subscriptions: { id: subscriptions })
+      .where(messages: { override_subscription_frequency_to_immediate: false })
       .where("messages.created_at >= ?", digest_run.starts_at)
       .where("messages.created_at < ?", digest_run.ends_at)
       .uniq(&:id)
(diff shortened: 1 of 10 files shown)


huwd (approved):
Looks good, nice one

barrucadu:
This also needs the immediate messages excluded from digest runs: https://github.com/alphagov/email-alert-api/blob/94580a7c9b5c5bcfd1227641f94b0fc283837361/app/queries/digest_items_query.rb#L43

PR #1704, opened 2022-01-21
Patch subscriber lists

huwd:
[Trello](https://trello.com/c/IkYegsOt/1225-update-subscriberlist-titles-on-major-minor-publishing-change)

A part of work to update subscriber list titles upon a major / minor content change update.

See commit messages for more details

### DIFF
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -31,6 +31,27 @@ class SubscriberListsController < ApplicationController
     render json: subscriber_list.to_json
   end
 
+  def update
+    permitted_params = params.permit(updatable_parameters)
+
+    if permitted_params.empty?
+      render json: {
+        error: "Must include at least one of: #{updatable_parameters.join(', ')}",
+      }, status: :unprocessable_entity
+      return
+    end
+
+    subscriber_list = SubscriberList.find_by(slug: params[:slug])
+
+    unless subscriber_list
+      render json: { error: "Could not find the subscriber list" }, status: :not_found
+      return
+    end
+
+    subscriber_list.update!(permitted_params)
+    render json: subscriber_list.to_json
+  end
+
   def bulk_unsubscribe
     subscriber_list = SubscriberList.find_by(slug: params[:slug])
 
@@ -53,6 +74,10 @@ class SubscriberListsController < ApplicationController
   end
 
 private
+
+  def updatable_parameters
+    [:title]
+  end
 
   def convert_legacy_params(link_or_tags)
     link_or_tags.transform_values do |link_or_tag|
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,6 +4,7 @@ Rails.application.routes.draw do
     resources :subscriber_lists, path: "subscriber-lists", only: %i[create]
     get "/subscriber-lists", to: "subscriber_lists#index"
     get "/subscriber-lists/:slug", to: "subscriber_lists#show"
+    patch "/subscriber-lists/:slug", to: "subscriber_lists#update"
     post "/subscriber-lists/:slug/bulk-unsubscribe", to: "subscriber_lists#bulk_unsubscribe"
 
     resources :content_changes, only: %i[create], path: "content-changes"
--- a/docs/api.md
+++ b/docs/api.md
@@ -73,6 +73,30 @@ It will respond with the JSON response for the `GET` call above.
 
 Returns a `404 Not Found` if there is no such list.
 
+### `PATCH /subscriber-lists/xxx`
+
+Update data that helps describe the subscriber list, such as the title.
+It requires at least one parameter to update.
+
+```json
+{
+  "title": "A new Subscriber list title",
+}
+```
+
+The following fields are accepted:
+- title: The title of this particular list, which will be shown to the user;
+  email sent to a user;
+
+Any additional parameters will be ignored.
+
+Returns a `200 OK` with the details of the subscriber list including the
+new parameters, for an acceptable request.
+
+Returns a `404 Not Found` if there is no such list.
+
+Returns a `422 Bad Request` if no allowed parameters are provided to update.
+
 ### `POST /subscriber-lists/xxx/bulk-unsubscribe`
 
 Unsubscribes all subscribers from that list, and optionally sends an email to them.
--- a/spec/integration/update_subscription_list_spec.rb
+++ b/spec/integration/update_subscription_list_spec.rb
@@ -1,0 +1,64 @@
+RSpec.describe "Updating a subscriber list", type: :request do
+  let!(:subscriber_list) { create(:subscriber_list) }
+  let(:update_subscriber_lists_path) { "/subscriber-lists/#{slug}" }
+  let(:slug) { subscriber_list.slug }
+  let(:update_params) do
+    {
+      "title" => "A new title",
+    }
+  end
+
+  context "with authentication and authorisation" do
+    before do
+      login_with_internal_app
+    end
+
+    it "returns a 200" do
+      patch update_subscriber_lists_path, params: update_params.to_json, headers: json_headers
+      expect(response.status).to eq(200)
+    end
+
+    it "returns the subscriber list with updated title" do
+      patch update_subscriber_lists_path, params: update_params.to_json, headers: json_headers
+
+      expect(JSON.parse(response.body).dig("subscriber_list", "title")).to eq(update_params["title"])
+    end
+
+    it "returns 422 if no allowed permitted parameters are provided" do
+      patch update_subscriber_lists_path, params: {}, headers: json_headers
+      expect(response.status).to eq(422)
+    end
+
+    it "returns 422 if a title with value of nil is provided" do
+      patch update_subscriber_lists_path, params: { title: nil }.to_json, headers: json_headers
+      expect(response.status).to eq(422)
+    end
+
+    context "for an unknown subscriber list" do
+      let(:slug) { "not-a-real-slug" }
+
+      it "returns 404" do
+        patch update_subscriber_lists_path, params: update_params.to_json, headers: json_headers
+        expect(response.status).to eq(404)
+      end
+    end
+  end
+
+  context "without authentication" do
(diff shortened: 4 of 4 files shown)


app/controllers/subscriber_lists_controller.rb
```diff
@@ -54,6 +72,14 @@ def bulk_unsubscribe
 
 private
 
+  def updatable_parameters
+    [:title]
+  end
+
+  def has_updatable_parameters?
+    updatable_parameters.map { |field| params.include?(field) }.any?
```
barrucadu, line 80: `any?` takes a block, so this would be more efficient as

```ruby
updatable_parameters.any? { |field| params.include?(field) }
```

config/routes.rb
```diff
@@ -5,6 +5,7 @@
     get "/subscriber-lists", to: "subscriber_lists#index"
     get "/subscriber-lists/:slug", to: "subscriber_lists#show"
     post "/subscriber-lists/:slug/bulk-unsubscribe", to: "subscriber_lists#bulk_unsubscribe"
+    patch "/subscriber-lists/:slug", to: "subscriber_lists#update"
```
barrucadu, line 8: Nitpick: I think this should be above the `post`, so it's next to the other `/subscriber-lists/:slug` route

docs/api.md
```diff
@@ -73,6 +73,30 @@ It will respond with the JSON response for the `GET` call above.
...
+  email sent to a user;
+
+Any additional parameters will be ignored.
+
+Returns a `200 OK` with the details of the subscriber list including the
+new parameters, for an acceptable request.
+
+Returns a `404 Not Found` if there is no such list.
+
+Returns a `400 Bad Request` if no allowed parameters are provided to update.
```
barrucadu, line 98: Should this be a 422 instead?  eg, that's what Rails uses for missing params

app/controllers/subscriber_lists_controller.rb
```diff
@@ -31,6 +31,24 @@ def create
...
+      render json: {
+        error: "Must include at least one of: #{updatable_parameters.join(', ')}",
+      }, status: :bad_request
+      return
+    end
+
+    subscriber_list = SubscriberList.find_by(slug: params[:slug])
+
+    unless subscriber_list
+      render json: { error: "Could not find the subscriber list" }, status: :not_found and return
```
barrucadu, line 45: The `... and return` pattern is useful in the construct `do_thing and return if condition`, but if you're not using an inline condition I think it's better to split it across two lines.

app/controllers/subscriber_lists_controller.rb
```diff
@@ -54,6 +72,14 @@ def bulk_unsubscribe
 
 private
 
+  def updatable_parameters
+    [:title]
```
barrucadu, line 76: Since #1705 has been merged, this should be `%i[title description]` now

app/controllers/subscriber_lists_controller.rb
huwd: ah great shout i'd forgotten any? can take a block

docs/api.md
huwd: Ah interesting, if that's the case will change

app/controllers/subscriber_lists_controller.rb
huwd: Yeah, I'd like to keep that spearate for the moment, I think it will help if these are packaged together?

app/controllers/subscriber_lists_controller.rb
```diff
@@ -31,6 +31,25 @@ def create
     render json: subscriber_list.to_json
   end
 
+  def update
```
barrucadu, line 34: You could avoid needing to specify the `:title` in three places (twice on line 49 and once on line 77) by taking a hash of updates, like so:

```ruby
def update
  updates = params.permit(updateable_parameters).to_h # the `to_h` might not be necessary - check that

  if updates.empty?
    # do the unprocessable entity here
  end

  # etc

  subscriber_list.update!(updates)

  # etc
end
```

spec/integration/update_subscription_list_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    before do
+      login_with_internal_app
+    end
+
+    it "returns a 200" do
+      patch update_subscriber_lists_path, params: update_params.to_json, headers: json_headers
+      expect(response.status).to eq(200)
+    end
+
+    it "returns a the subscription with updated title" do
```
barrucadu, line 21: `returns the subscriber list with updated title`

app/controllers/subscriber_lists_controller.rb
barrucadu: Also, do you want to protect against nulls?  eg, PATCHing `{"title":null}` feels like it shouldn't blank out the title to me.  In which case your current code doesn't handle that.  In my suggestion you can add `.compact` onto the `updates` line.

spec/integration/update_subscription_list_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      patch update_subscriber_lists_path, params: update_params.to_json, headers: json_headers
+
+      expect(JSON.parse(response.body).dig("subscriber_list", "title")).to eq(update_params["title"])
+    end
+
+    it "returns 422 if no allowed updatable parameters are provided" do
```
barrucadu, line 27: If you do want to forbid nulls, you should have a test for `params: { title: nil }.to_json`

app/controllers/subscriber_lists_controller.rb
huwd: yeah you are right, permit is the right tool here, I'm doing something non-standard

huwd:
cool those amendments are made

spec/integration/update_subscription_list_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    it "returns the subscriber list with updated title" do
+      patch update_subscriber_lists_path, params: update_params.to_json, headers: json_headers
+
+      expect(JSON.parse(response.body).dig("subscriber_list", "title")).to eq(update_params["title"])
+    end
+
+    it "returns 422 if no allowed permitted parameters are provided" do
+      patch update_subscriber_lists_path, params: { title: nil }.to_json, headers: json_headers
+      expect(response.status).to eq(422)
+    end
```
barrucadu, line 30: How is this test passing?  In the controller you're not doing `.compact` in the params, and this is false:

```ruby
irb(main):001:0> ActionController::Parameters.new({ title: nil }).permit(:title).empty?
=> false
```

So why does this case return a 422?

huwd: oh neat, so turns out Rails is helping.
If you try to update with unpermitted params, it is going to 422.

So the response body I'm getting is:
```
JSON.parse(response.body)
=> {"error"=>"Unprocessable Entity", "details"=>{"title"=>["can't be blank"]}}
```

And I can [drop all of this](https://github.com/alphagov/email-alert-api/pull/1704/files#diff-8eea030d9562e2c20b45db190d8639669d34b0f7017ea6e8c39fd0060de1e3efR37-R42) and get the same result

huwd: Interestingly without that block:
```
patch update_subscriber_lists_path, params: { }.to_json, headers: json_headers
```
returns 200, which is not what we want.

So this must be ActiveRecord valdiations bubbling up from the schema requirements:
https://github.com/alphagov/email-alert-api/blob/main/db/schema.rb#L119

barrucadu: Hmm, that'll work for the title, but the description is nullable, and presumably we don't want `{ "description": null }` to blank out the description.

huwd: I think I want to test for both these scenarios, so will leave that block and add a second test.

huwd: Amended

huwd: I think the point is we should get two different types of 422 here, no?

1. 422 from ActiveRecord if title (or description / any future one) violates DB or ActiveRecord validation
2. 422 from our controller if we don't pass at least one permitted parameter

So I reckon if descriptions should not be nullable the thing to do would be to add validation to the subscriber list model?
Otherwise if you can have a subscriber list without a description, you should be able to patch it to null via the API, no?

huwd: I guess we could do something like `permitted_params.empty? || permitted_params.values.all?(nil)`

huwd: or perhaps `permitted_params.values.all?(&:blank?)` that'd get `""` too I think

huwd: though tbh that starts to feel a bit eww, what if we want a nullable thing we can patch in future?
We're getting the controller to do validation that should be on the model?

If the problem is - what do we do with existing subscriptions that have no descriptions then I think we should scope validation to single page notificaitons only and put a migration in to add descriptions (we likely need to do that anyway?)

huwd: Probably another question we should be asking... can publishers remove page descriptions from content?
or are they required over there?

Perhaps this one is best addressed as a part of that work?

alex9smith: Description is required yes, it's the summary field in Whitehall. So every page **should** have something in the description

barrucadu: If we're going to allow blanking out descriptions that's fine, but in that case we should remove the `.compact` from https://github.com/alphagov/gds-api-adapters/pull/1131/files

huwd: Ah yeah sorry when I said migraition, i guess it could be a rake task too or something. Was thinking something that would iterate over all db values for single page notifications, likely query the content store / publishing api with the content ID and backfill all subscriber lists for pre-existing subscriber lists?

I've seen us do that with temp rake tasks and migrations on GOV.UK... I think rake tasks are better, they can best tested?

barrucadu (approved):

PR #1708, opened 2022-01-27
Add rake task to add descriptions to single page subscriber lists

alex9smith:
We added an optional description field to subscriber lists in [[1]]. We
intend to use the page summary as this description for single page
subscriber lists, and will include it in some emails.

New single page lists get the description passed through from email
alert frontend when the list is created, but it's null for any that
existed before this change.

Add a rake task that iterates over all the single page subscriber lists
and fetches the description from content store so every single page 
list has a description.

[Trello](https://trello.com/c/dm5ZxFVB/1226-pass-through-page-descriptions-to-subscriberlists-on-major-minor-publishing-change)

[1]: https://github.com/alphagov/email-alert-api/pull/1705

### DIFF
--- a/lib/tasks/data_migration.rake
+++ b/lib/tasks/data_migration.rake
@@ -76,4 +76,14 @@ namespace :data_migration do
       puts "Updated #{args[:key]} in #{list.title} to #{new_criteria}"
     end
   end
+
+  desc "Fetch descriptions for single page subscription lists which don't have descriptions"
+  task fetch_subscriber_list_descriptions: :environment do |_t, _args|
+    SubscriberList.where.not(content_id: nil).where(description: nil).find_each do |list|
+      new_description = Services.content_store.content_item(list.url)["description"]
+      list.update!(description: new_description)
+
+      puts "Subscriber list for #{list.url} updated with new description"
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/data_migration.rake
```diff
@@ -76,4 +76,15 @@ namespace :data_migration do
       puts "Updated #{args[:key]} in #{list.title} to #{new_criteria}"
     end
   end
+
+  desc "Fetch descriptions for single page subscription lists which don't have descriptions"
+  task fetch_subscriber_list_descriptions: :environment do |_t, _args|
+    SubscriberList.where.not(content_id: nil).where(description: nil).find_each do |list|
+      new_description = Services.content_store.content_item(list.url).to_h["description"]
```
barrucadu, line 83: `.to_h` isn't needed here, the [`GdsApi::Response` delegates `:[]`](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/response.rb#L95).

lib/tasks/data_migration.rake
```diff
@@ -76,4 +76,15 @@ namespace :data_migration do
...
+      list.description = new_description
+      list.save!
```
barrucadu, line 85: This could just be `list.update!(description: new_description)`

alex9smith: Thanks, I'm still not used to `?` and `!` in method names

barrucadu (approved):

PR #1709, opened 2022-01-27
Allow subscriber list description to be updated

alex9smith:
We will be extending the queue processors added in [[1]] to also check
for updates to a page's description and pass these through to the API
to be stored for use in some emails.

Before we can do that, we need to add `description` to the list of
updateable parameters in the API.

[Trello](https://trello.com/c/dm5ZxFVB/1226-pass-through-page-descriptions-to-subscriberlists-on-major-minor-publishing-change)

[1]: https://github.com/alphagov/email-alert-service/pull/493

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -81,12 +81,14 @@ It requires at least one parameter to update.
 ```json
 {
   "title": "A new Subscriber list title",
+  "description": "A new subscriber list description",
 }
 ```
 
 The following fields are accepted:
 - title: The title of this particular list, which will be shown to the user;
   email sent to a user;
+- description: A description of the content this list represents, used by [email-alert-service](https://github.com/alphagov/email-alert-service) to construct emails when a page is unpublished. 
 
 Any additional parameters will be ignored.
 
--- a/app/controllers/subscriber_lists_controller.rb
+++ b/app/controllers/subscriber_lists_controller.rb
@@ -77,7 +77,7 @@ class SubscriberListsController < ApplicationController
 private
 
   def updatable_parameters
-    [:title]
+    %i[title description]
   end
 
   def convert_legacy_params(link_or_tags)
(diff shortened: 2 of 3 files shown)


docs/api.md
```diff
@@ -81,12 +81,14 @@ It requires at least one parameter to update.
...
 {
   "title": "A new Subscriber list title",
+  "description": "A new subscriber list description",
 }
 ```
 
 The following fields are accepted:
 - title: The title of this particular list, which will be shown to the user;
   email sent to a user;
+- description: A description of the content this list represents, shown to the user in some emails.
```
alex9smith, line 91: I don't love this explanation. Suggestions for a better one are welcome!

barrucadu: It's actually *not* used in any of the email bodies that email-alert-api generates (email-alert-service constructs the unpublishing email bodies).  Which does make this field a bit weird, as it's essentially arbitrary extra text data, passed over when the list is created, with no specific use...

alex9smith: Yeah true. It will eventually end up in some of the emails (we'll pass it back to the service, the service will construct the email body, and then send the email through the API), but it's more of a case of the API database being the best place to store something we'll need later.

 I looked back at the description the [last time it existed ](https://github.com/alphagov/email-alert-api/pull/1527/files#diff-9eddf4dc51bf6a0125ca7fb094468ad284112270385c41cebce4f8f0a29620abL46) which explains the rogue `email sent to a user;` on line 90, but otherwise isn't very helpful.
 
 I think I'm going to go with: 
 "A description of the content this list represents, used by [email-alert-service](https://github.com/alphagov/email-alert-service) to construct emails when a page is unpublished."
 
 That'll at least give people somewhere to look for more information

huwd (approved):

PR #1717, opened 2022-02-18
Add rake task to clean up subscriptions for archived Specialist Topics

chao-xian:


KludgeKML (approved):
Looks okay to me.

AgaDufrat (approved):
I think it's good to merge after that tiny change. I double checked the topics list and it looks good to me. I agree that adding those outputs when we run it with rake `--dry-run` will do the job. 🎉

lib/tasks/email_and_unsubscribe_archived_topics.rake
```diff
@@ -0,0 +1,223 @@
...
+            subject: subject,
+            body: body,
+            subscriber_lists: [sub_list],
+          )
+
+          email_ids.each do |id|
+            SendEmailWorker.perform_async_in_queue(id, queue: :send_email_immediate)
+          end
+
+          puts "Destroying subscription list #{sub_list.url} (ID: #{sub_list.id})"
```
hannako, line 52: Because I'm a bit scared of deleting things (and email alert api in general I think!) I would probably have done the destruction of the subscription lists as a clean up task after everyone has been emailed. So this task would be to send the email, and unsubscribe users to the old lists ([this](https://github.com/alphagov/email-alert-api/pull/1681/files) sort of thing). And then a second task would be to delete them.

lib/tasks/email_and_unsubscribe_archived_topics.rake
```diff
@@ -1,12 +1,16 @@
 namespace :archived_topics do
   desc "Send a bulk email to subscribers to these archived Specialist Topics."
-  task :send_emails, [] => :environment do |_t, _args|
-    topic_urls.each do |topic|
+  task :send_emails, [:check_output] => :environment do |_t, args|
+    args.with_defaults(check_output: "false")
```
hannako, line 4: Is it safer to have `check_output: "true"` by default. And then you have to pass in an extra argument to make the task run for real?

Instead of lines 32 to 41 you could just wrap the `BulkSubscriberListEmailBuilder` in an unless `check_output: false`. So you see exactly what's going to happen?

lib/tasks/email_and_unsubscribe_archived_topics.rake
chao-xian: That was originally the plan but felt that it was being too delayed and also once process would eventually how we want to do it ultimately in https://trello.com/c/CQlLYnrF/296-automate-unsubscribing-from-archived-specialist-topics

lib/tasks/email_and_unsubscribe_archived_topics.rake
chao-xian: Good shout - thanks

chao-xian: Addressed in https://github.com/alphagov/email-alert-api/commit/3585b4808f5fce5e4b321d7b2633490aea90a5ee

chao-xian:
I've tested this on integration and pretty happy with it.

PR #1722, opened 2022-03-03
Report on subs by url

huwd:
Add Reporting class for counting active subscribers to a list by URL

This should work across:
- Organisation pages ([one with some subs](https://deploy.blue.staging.govuk.digital/job/run-rake-task/218128/console)!)
- Single page notifications (one [with no subs](https://deploy.blue.staging.govuk.digital/job/run-rake-task/218127/console), one [with some](https://deploy.blue.staging.govuk.digital/job/run-rake-task/218129/console))
- Taxon pages (tested on `/education` [see here](https://deploy.blue.staging.govuk.digital/job/run-rake-task/218131/console), nb [/coronavirus](https://deploy.blue.staging.govuk.digital/job/run-rake-task/218130/console) does not work, i suspect there's special magic behind that taxon)

All these subscriber lists seem to have URLs, most often requests arrive
for subscriber stats from stakeholders who know the URL of a page.

Put it all in a rake task so we can run it on jenkins and update two things in the docs:

1. The existing stuff seems to export a much grander CSV, with lots of
data and that uses the existing matching functions. Clarify that a bit.

2. Document the new simple count by URL report.

### DIFF
--- a/docs/support-tasks.md
+++ b/docs/support-tasks.md
@@ -110,12 +110,31 @@ Depending on the number of emails to send, the Rake task can take a few minutes 
 bundle exec rake 'support:resend_failed_emails:by_id[<email_one_id>,<email_two_id>]'
 ```
 
-## Get subscriber numbers for a list
+## Get subscriber list csv report
 
-To see the number of subscribers for a given list:
+To see a csv report for the number of subscribers for a given list:
 
 ```bash
 bundle exec rake 'report:csv_subscriber_lists[<on_date>] SLUGS=<list_slug>'
 ```
 
 The date should be in ISO8601 format, for example 2020-01-01.
+
+## Get a simple count of subscribers to a list by its URL
+
+To see a simple count of the number of subscribers for a given list:
+
+```bash
+bundle exec rake 'report:csv_subscriber_lists[<url>]'
+```
+This will always report on the number of active subscriptions for a given url, as of the end of today.
+
+You can also pass it a :active_on_datetime which will count how many active subscriptions there were at the end of the day on a particular date.
+(active is defined as created before that datetime and not with an "ended_on" datetime by the given date)
+
+```bash
+bundle exec rake 'report:csv_subscriber_lists[<url>, <active_on_date>]'
+```
+
+The active_on_date should be in ISO8601 format, for example 2022-03-03T12:12:16+00:00.
+The time will always be rounded to the end of the day, even if that is in the future.
--- a/lib/reports/subscriber_list_subscriber_count_report.rb
+++ b/lib/reports/subscriber_list_subscriber_count_report.rb
@@ -1,0 +1,33 @@
+class Reports::SubscriberListSubscriberCountReport
+  class BadDateError < StandardError; end
+
+  attr_reader :url, :active_on_date
+
+  def initialize(url, active_on_date = nil)
+    @url = url
+    @active_on_date = active_on_date || Time.zone.now.end_of_day
+  end
+
+  def call
+    list = SubscriberList.find_by_url(url)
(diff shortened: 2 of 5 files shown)


lib/reports/subscriber_list_subscriber_count_report.rb
```diff
@@ -0,0 +1,31 @@
...
+    @url = url
+    @active_on_date = active_on_date || Time.zone.now
+  end
+
+  def call
+    list = SubscriberList.find_by_url(url)
+    if list
+      list.subscriptions
+          .active_on(formatted_date)
+          .count
```
alex9smith, line 16: Could you format the count into a string? I think just having a number on a line is easy to miss or not understand.

I'm thinking something like
`"Subscriber list for #{url} had #{count} subscribers on #{formatted_date}."`

docs/support-tasks.md
```diff
@@ -110,12 +110,30 @@ Depending on the number of emails to send, the Rake task can take a few minutes
...
+This will always report on the number of active subscriptions as of now.
+
+You can also pass it a :active_on_datetime which will count how many active subscriptions there were on a particuar date.
+(active is defined as created after that datetime and not with an "ended_one" datetime by the given date)
+
+```bash
+bundle exec rake 'report:csv_subscriber_lists[<url>, <active_on_date>]'
+```
+
+The active_on_date should be in ISO8601 format, for example 2022-03-03T12:12:16+00:00.
```
alex9smith, line 139: Not a blocker, but it'd be nice if this could be just a date (and then the task reports on the number of subscribers by the end of that day)

docs/support-tasks.md
```diff
@@ -110,12 +110,30 @@ Depending on the number of emails to send, the Rake task can take a few minutes
...
+
+To see a simple count of the number of subscribers for a given list:
+
+```bash
+bundle exec rake 'report:csv_subscriber_lists[<url>]'
+```
+This will always report on the number of active subscriptions as of now.
+
+You can also pass it a :active_on_datetime which will count how many active subscriptions there were on a particuar date.
+(active is defined as created after that datetime and not with an "ended_one" datetime by the given date)
```
alex9smith, line 133: Should this be `before`? (also you have a typo)

```suggestion
(active is defined as created before that datetime and not with an "ended_on" datetime by the given date)
```

docs/support-tasks.md
huwd: Yeah the end of the day bit is a good shout, I could complicate this a bit.
As is, this will default to the start of the day 🤔 perhaps end_of_day if there's no time...

I did try a more open `Time.zone.parse` that might have given more leeway on input here but freaked out at how permissive that is. Check this out:
https://gds.slack.com/archives/CADGP8PB4/p1646314848714819

Tldr...

```
Time.zone.parse("20th of Furberry")
```
gets made into a valid date.

docs/support-tasks.md
huwd: yep! thanks!

I changed the implimentation of this and missed these docs

docs/support-tasks.md
huwd: I'm gonna part this one for the moment, let's look at extended the task with this depending on what the demand looks like?

alex9smith: Yeah I saw that descent into madness 🙈 

I hadn't realised that it'd default to the start of the day but that does make sense. Maybe just call that out in the docs and it'll do for my original thought of "I don't want to type out a timestamp when I'm on support"

alex9smith (approved):

alex9smith (approved):
Nice!

PR #1730, opened 2022-03-14
Report on single page notificaiton subscription figures

huwd:
[Trello](https://trello.com/c/Lz5wJCHC/1301-put-single-page-notification-subscription-reports-into-a-rake-task)

We ran some reports last week to better understand traffic spikes.
I've moved the reports I ran into tested report classes and have provided a rake task for easy future reuse.


alex9smith (approved):
Looks good

PR #1759, opened 2022-06-08
Remove archived_topics:email_and_unsubscribe rake task

AgaDufrat:
This functionality was build into the Collections Publisher app in https://github.com/alphagov/collections-publisher/pull/1623

https://trello.com/c/CQlLYnrF/1038-automate-unsubscribing-from-archived-specialist-topics-m


Tetrino (approved):

KludgeKML:
Ah, sorry - I was doing other stuff that needs to be deployed and I thought since this one was approved and passing it might as well be merged.

Since email-alert-api doesn't (yet) have CD, this hasn't actually been deployed, but I was going to deploy it manually as soon as I could. Should I hold off? How often do users need to be unsubscribed manually?

AgaDufrat:
Let's just bin this. Every 1-2 month or so. Even if we have to do it manually again we would need to raise a PR to update the yaml so reverting it is not going to save us any button pressing ;)

KludgeKML:
Cool.

PR #1826, opened 2022-10-31
deleted messages_controller and all pertaining tests

kashifatcha:

https://trello.com/c/RYRwEqia/1539-check-whether-anything-is-still-using-email-alert-api-as-substitute-database


1pretz1:
I can see that we’ve removed the controller and routing for the messages endpoint (and associated tests), which is a good start. However, there is still a fair amount of code to remove and some database migrations to run.

I’d suggest stepping through the code (from the messages controller) and searching where the associated classes are called and removing them and their references. `Message` is part of the [nomenclature](https://github.com/alphagov/email-alert-api#nomenclature) in Email Alert API, meaning references to ‘Message’, message’, ‘messages’ etc all should all be related. As we’re removing the Messages endpoint, we should also remove all code relating to a ‘message’.

An example being, [MessageHandlerService](https://github.com/alphagov/email-alert-api/blob/main/app/services/message_handler_service.rb), which is called from the [MessagesController](https://github.com/alphagov/email-alert-api/blob/main/app/controllers/messages_controller.rb#L5). From within `MessageHandlerService`, we also see that `Message`, `ProcessMessageWorker` and `message_created` are all used. All of this ‘messages’ code (classes + references) should be removed.

There are a lot more instances of messages code than the above, you’ll either need to step through each controller and check for any message relating code, or (which will be a lot quicker): search in the repo for references to ‘Message’, ‘messages’ etc and remove accordingly. There are a lot of hits for `Message` so that should give you a good starting point. Searching in the repo by filename will also work well here.

We’ll also need to run some database migrations to remove the [messages and matched_message tables](https://github.com/alphagov/email-alert-api/blob/main/db/schema.rb#L92-L116) and remove a `message_id` column from an existing table (https://github.com/alphagov/email-alert-api/blob/main/db/schema.rb#L160). This should be done after the application code has been removed (could still be within the same PR). 

For the deleting of tables - we’ll want to have one migration per table. Some guidance on this can be found at: https://gist.github.com/chand/3c646d7ef8f32599ea17ae37c6ebde86. 

For the removal of the `message_id` column, some guidance can be found at: https://riptutorial.com/ruby-on-rails/example/3491/remove-an-existing-column-from-a-table.

1pretz1 (approved):
Great work Kash - changes look good to me! (just need to resolve some merge conflicts) 🎉

PR #1869, opened 2023-01-11
Improve documentation of simple subscriber list count rake task.

KludgeKML:
Improve documentation of simple subscriber list rake task, including warning that it does not tell the whole picture.

(See rendered version here: https://github.com/alphagov/email-alert-api/blob/update-docs/docs/support-tasks.md)

### DIFF
--- a/docs/support-tasks.md
+++ b/docs/support-tasks.md
@@ -125,12 +125,17 @@ The date should be in ISO8601 format, for example 2020-01-01.
 To see a simple count of the number of subscribers for a given list:
 
 ```bash
-bundle exec rake 'report:csv_subscriber_lists[<url>]'
+bundle exec rake 'report:csv_subscriber_lists[<path>,<active_on_date (optional)>]'
 ```
-This will always report on the number of active subscriptions for a given url, as of the end of today.
+This will report on the number of active subscriptions for a given path.
 
-You can also pass it a :active_on_datetime which will count how many active subscriptions there were at the end of the day on a particular date.
-(active is defined as created before that datetime and not with an "ended_on" datetime by the given date)
+> **Warning**
+>
+> This is called simple because it only gets subscriber lists for a direct path. There may still be subscriptions by topic or link, so finding an empty or missing subcription list does not mean that no-one will receive email updates about that page! See below for a more detailed description of how to get the subscriber list.
+
+The path should be the full path on gov.uk (for instance /government/statistics/examples if the page you're interested in is https://www.gov.uk/government/statistics/examples)
+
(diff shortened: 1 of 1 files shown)


docs/support-tasks.md
```diff
@@ -125,26 +125,24 @@ The date should be in ISO8601 format, for example 2020-01-01.
 To see a simple count of the number of subscribers for a given list:
 
 ```bash
-bundle exec rake 'report:csv_subscriber_lists[<url>]'
+bundle exec rake 'report:csv_subscriber_lists[<path>,<active_on_date (optional)>]'
 ```
-This will always report on the number of active subscriptions for a given url, as of the end of today.
+This will always report on the number of active subscriptions for a given path, as of the end of today.
```
maxgds, line 130: I think this line needs a tweak in light of adding the data param in the snippet.

docs/support-tasks.md
```diff
@@ -125,26 +125,24 @@ The date should be in ISO8601 format, for example 2020-01-01.
...
 
-You can also pass it a :active_on_datetime which will count how many active subscriptions there were at the end of the day on a particular date.
-(active is defined as created before that datetime and not with an "ended_on" datetime by the given date)
+> **Warning**
+>
+> This is called simple because it only gets subscriber lists for a direct path. There may still be subscriptions by topic or link, so finding an empty or missing subcription list does not mean that no-one will receive email updates about that page!
```
maxgds, line 134: Can we document, or link to documentation on how to find those things? I can imagine being on 2nd line and knowing there are no direct subscribers but not having a clue what to do next in response to a ticket as it stands.

maxgds (changes requested):

docs/support-tasks.md
KludgeKML: I was sort of hoping, because it's horrific, that as a follow-up we might work out some sort of technical fix to that so we didn't have to document it!

Essentially you need to do a subscriber list query with all of these parameters:

SubscriberListQuery.new(content_id:, tags:, links:, document_type:, email_document_supertype:  government_document_supertype:)

...which if they come from an existing content change is not too bad (ie if an email has been triggered, you can do this all on email-alert-api), but if it's for hypothetical changes to a page you need to manually look up a content item in another app that can access the content store and then copy and paste them across.

If email-alert-api could read from content store that would be relatively simple, but currently it doesn't have access and I don't know if there's a security concern with allowing it. Probably not? 

I'll write this up anyway, but there's definitely another card or two to be made about this.

maxgds (approved):

PR #1905, opened 2023-04-05
Add temporary rake task to send bulk email following incident

brucebolt:
During an incident, a number of users were subscribed to the incorrect list.
    
This rake task will select the affected users and send them an email asking them to subscribe again.

Most of the code has been taken from https://github.com/alphagov/email-alert-api/pull/1562, but updated to send to only the affected users in this incident.  It also retrieves the affected users in batches, as we were finding the query would cause the console to quit unexpectedly in EKS.

### DIFF
--- a/lib/tasks/bulk_email.rake
+++ b/lib/tasks/bulk_email.rake
@@ -26,4 +26,47 @@ namespace :bulk_email do
     end
     puts "Sending #{email_ids.count} emails to subscribers on the following lists: #{subscriber_lists.pluck(:slug).join(', ')}"
   end
+
+  desc "Send an apology email following an incident"
+  task :temp_incident_new_subscribers, [] => :environment do
+    affected_subscribers = []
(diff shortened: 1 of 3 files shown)


leenagupte (commented):
It looks good to me, and the test is great.

This is difficult to test locally, but I see from internal chat that you have tested getting the affected Subscriptions in a console?

Where you able to test this rake task too? Can you link to the results?

brucebolt:
> Where you able to test this rake task too? Can you link to the results?

I ran this in integration but can't link to the output as we no longer have Jenkins.

Here's a screenshot instead (the task has no output):
Screenshot showing rake task run on a console with no output

Before doing this, I updated one of the subscriptions (which had been anonymised by the data sync) to use `email-alert-api-integration@digital.cabinet-office.gov.uk`.  I could see the email arrive in [that inbox](https://groups.google.com/a/digital.cabinet-office.gov.uk/g/email-alert-api-integration/c/jycqB07gfZE) when I ran the task (albeit a number of times whilst I tried to perfect how the line breaks and links were handled by Notify).

leenagupte:
> I ran this in integration but can't link to the output as we no longer have Jenkins.

Thanks for confirming!

leenagupte (approved):

PR #1910, opened 2023-04-17
Cleanup unicorn conf

KludgeKML:
Removes obsolete references to unicorn in config and procfile, and removes an obsolete Jenkinsfile.


Procfile
```diff
@@ -1,2 +0,0 @@
-web: bundle exec unicorn -c ./config/unicorn.rb -p ${PORT:-3088}
-worker: bundle exec sidekiq -C ./config/sidekiq.yml
```
leenagupte, line 2: Why is sidekiq being removed as well?

KludgeKML: It's more that this entire file is being removed, I think nothing uses it any more (but I'm willing to be corrected). Deployments use the Dockerfile, Kubernetes deployments use something else. I'll confirm this, though.

KludgeKML: Thread here: https://gds.slack.com/archives/CAB4Q3QBW/p1681806083832859

leenagupte (approved):

PR #1943, opened 2023-06-07
BadListRemover class

hannako:
https://trello.com/c/nd9miwbj/1846-move-users-subscribed-to-bad-specialist-topic-lists-to-good-specialist-topic-lists-m-l


lib/bad_list_subscriptions_migrator.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def subscriber_lists
+    @subscriber_lists ||= SubscriberList.where("url LIKE ?", "%/#{prefix}/%")
+  end
+
+  def move_users_from_bad_lists_to_good_list(lists)
+    valid_lists = lists.select { |list| valid_links_based_list?(list) }
+    bad_lists = lists - valid_lists
+
+    if valid_lists.count == 1
```
BeckaL: Just a check - do we have any examples where there is no valid list, but there is a bad list? I imagine it's unlikely, but could theoretically have happened if all the signups were when the bug was in place, and none before / after?

hannako: Yep good shout, thanks 👍

BeckaL: And I guess also are there any with a valid lists count of 2?

lib/subscriber_list_mover.rb
```diff
@@ -42,8 +42,6 @@ def call
         )
 
         if subscribed_to_destination_subscriber_list.nil?
-          puts "Moving #{subscriber.address} with ID #{subscriber.id} to #{destination_subscriber_list.title} list"
```
BeckaL, line 45: Could we leave it in but just take out the subscriber address? I e.g. quite like having some output that I can put in a trello ticket for an audit trail in case we need it

lib/bad_list_subscriptions_migrator.rb
```diff
@@ -41,4 +41,10 @@ def valid_links_based_list?(list)
   def has_links_or_tags(list)
     list.links.present? || list.tags.present?
   end
+
+  def clear_target_for_migration?(lists)
+    return unless lists.count == 1
+
+    lists.first.active_subscriptions_count.positive?
```
BeckaL, line 60: I might be misunderstanding, but here, why does it matter if the valid list has any active subscribers? In my head, it only matters if the bad list has no active subscribers - if not, it doesn't matter about migrating. But if there are no active subscribers on the good list, we'd still want to migrate

spec/lib/bad_list_subscriptions_migrator_spec.rb
```diff
@@ -1,75 +1,100 @@
...
+  describe "#process_all_lists" do
+    let(:prefix) { "topic" }
 
-  let(:child_benefit_base_params) do
-    {
-      title: "Child Benefit",
-      url: "/topic/benefits-credits/child-benefit",
-    }
-  end
+    let(:child_benefit_base_params) do
```
BeckaL, line 5: minor: it'd be nice if this didn't show up in the diff but not bothered if we want to prioritise getting this out

BeckaL (commented):
Looking great! I think that's it for my comments on the implementation, tests look good but haven't gone through them in massive detail since I know you were possibly thinking about adding in extra

lib/bad_list_subscriptions_migrator.rb
hannako: I've used the `active_subscriptions_count` as a blunt tool to determine if a list is a valid candidate for migration. I've followed the pattern of the [historic_data_deletion_worker](https://github.com/alphagov/email-alert-api/blob/main/app/workers/historical_data_deletion_worker.rb) which removes lists (created before the max retention period) with no subscribers. So doing the same here seems sensible.

These bad lists were created some months ago, so it also seems fairly unlikely that the equivalent good list will have no accrued subscribers by now.

BeckaL: I guess my question is, does it matter at all if there are no active subscribers to a good list? I'm not sure how that impacts on its validity or not - maybe we could chat through?

lib/subscriber_list_mover.rb
hannako: Ah, we actually get a loooot of output with this task without this line, from [here](https://github.com/alphagov/email-alert-api/blob/df756cf9d320e80752b26b33f01835000cf69067/lib/subscriber_list_mover.rb#L55). Example output:

```
Bad subscriptions count for prefix 'topic':2334
Running migration...
185 active subscribers moving from safeguarding-children to schools-colleges-and-children-s-services-safeguarding-children
185 active subscribers moved from safeguarding-children to schools-colleges-and-children-s-services-safeguarding-children.
32 active subscribers moving from mot-manuals-and-special-notices to mot-manuals
Migration complete
There are 159 remaining bad subscriptions for 'topic' lists.
These pages still have bad lists: ["/topic/immigration-operational-guidance/immigration-rules", "/topic/immigration-operational-guidance/asylum-policy", "/topic/immigration-operational-guidance/fees-forms", "/topic/immigration-operational-guidance/nationality-guidance", "/topic/immigration-operational-guidance/entry-clearance-guidance", "/topic/immigration-operational-guidance/modernised-guidance", "/topic/immigration-operational-guidance/enforcement", "/topic/immigration-operational-guidance/windrush-scheme"].

So unless you feel strongly, I would prefer to remove this additional output.

```

spec/lib/bad_list_subscriptions_migrator_spec.rb
hannako: If you click to the settings cog at the top (to the right of Conversations), there's a checkbox for hide whitespace!

lib/bad_list_subscriptions_migrator.rb
hannako: Ah OK gotcha. Yes it does matter. Because if a topic page has been archived since the bad lists were created, all users belonging to the good subscriber lists have been [bulk unsubscribed](https://github.com/alphagov/collections-publisher/blob/24a7108e588f419892b52c4a8fa4584ea3c27496/app/services/tag_archiver.rb#L61). If the list is within the max retention period for the historic data deletion worker it won't yet have been deleted from the database. Does that make sense? ie We don't want to migrate users to these lists because the page the list is linked to has been archived. 

The task outputs the url's of the pages that still have bad lists associated with them (in the case of specialist topics it's only 8 pages). So we can handle those manually.

BeckaL: Ah OK got you, so 0 active subscribers is being used as a proxy for "it might be the case that this content has been archived". That makes sense

spec/lib/bad_list_subscriptions_migrator_spec.rb
BeckaL: Yeah I used that, more thinking as a cleaning up of git blame

hannako: Once we've run this task I'll be removing the whole file, so if it's OK with you I might save myself the rebase wrangling.

lib/bad_list_subscriptions_migrator.rb
hannako: Joining up the comments... Keith looked into this on staging (comment [here](https://docs.google.com/document/d/1Ci61Neb8sJqGn31DUNuhVS-pT1-RJnM6gtPpYNe8wi8/edit?disco=AAAAyi_ScYQ))

hannako: Keith's comment:
```
I checked on Staging, and out of 37156 subscriber lists, there are 3 urls which have 3 subscriber lists: 

- /government/organisations/hm-passport-office
- /government/organisations/groceries-code-adjudicator
- /government/organisations/building-regulations-advisory-committee.

They all follow the same pattern:
1st list is for the slug "all-types-of-document-about-<whatever>", and has a links digest, 
2nd is for the slug "<whatever>" and has a links digest, 
3rd is for the slug "<whatever>-<8-digit-hex-number>" and has a content_id)
```
So let's look at the 3 subscriber lists with the url "/government/organisations/hm-passport-office"

The two good lists both have the same links (and therefore the same links digest). They both have a similar number of active subscriptions. And fetching the most recently created subscription for these lists reveals both are actively being signed up for. 

So this is what I'm looking into...

How has this situation been created?
Does it matter?
If it matters --> leave this check in, and manually handle the exceptions after running the task.
If it doesn't matter --> remove the check

hannako: - How has this situation been created?

When Email Alert API creates a new subscriber list, the [slug is formed](https://github.com/alphagov/email-alert-api/blob/c445e17ee1598c40d6fca7a985ddda19eae5dec5/app/services/create_subscriber_list_service.rb#L34) from the title parameter [sent from email alert frontend](https://github.com/alphagov/email-alert-frontend/blob/main/app/services/subscriber_list_params/generate_links_based_list_params_service.rb#L10). So it's possible that at some point the code creating slugs in email alert api, or the code populating the title parameter in email alert frontend, has changed? I can't find evidence of this.

- Does it matter?

When a user clicks to sign up to a links based email subscription, email alert frontend asks email alert api to [find or create a subscriber list with the provided params](https://github.com/alphagov/email-alert-frontend/blob/757c60919a9fc11452d642876e6c54b5f96d73fb/app/controllers/content_item_signups_controller.rb#LL36C25-L36C25).

In email alert api, the [create_subscriber_list_service](https://github.com/alphagov/email-alert-api/blob/c445e17ee1598c40d6fca7a985ddda19eae5dec5/app/services/create_subscriber_list_service.rb#L13) uses the [find_exact_query](https://github.com/alphagov/email-alert-api/blob/main/app/queries/find_exact_query.rb#L23) class to match the provided params to existing subscriber lists in its database. Because links are present, it will [match on links_digest](https://github.com/alphagov/email-alert-api/blob/c445e17ee1598c40d6fca7a985ddda19eae5dec5/app/queries/find_exact_query.rb#L57) first. I think [this](https://stackoverflow.com/questions/42855244/how-it-works-find-by-in-rails) is what rails does under the hood (looking for a better ref) ie it returns the first record that matches. So in our case it will sometimes return the "all-types-of-document-about-<whatever>" list, and sometimes it will return "<whatever>" list.  So users are being subscribed to either of these. 

When a content change is sent to email alert api, it uses the SubscriberListQuery - will it match against both of these lists, or only one? To be continued...

hannako: I believe a matched content change will match on both of these lists according to the [matched_for_notification](https://github.com/alphagov/email-alert-api/blob/c445e17ee1598c40d6fca7a985ddda19eae5dec5/app/queries/matched_for_notification.rb#L20) query. It does a `select` for all SubscriberLists that match on links. Rather than a `find_by` that returns the first match it finds.

So where does that leave us!

I can remove this check for if there is only one good list, and just use the first list. 

So TLDR:

No, it doesn't matter which list the user is subscribed to

BeckaL (approved):
🥳 looks good to me, tests look sensible too

KludgeKML (approved):
Approved with caveat that possibly some of the commits could be squashed.

lib/bad_list_subscriptions_migrator.rb
hannako: Further probing into this confirms that both lists are valid and either can be used as the target for migration:

```
hm_lists = SubscriberList.where(url: "/government/organisations/hm-passport-office")
hm_lists.count
=> 3

ids = hm_lists.pluck(:id)
249080 is a bad list (content id no links)
4256 and 4655 are both good lists that differ only by slug and creation/updated_at timestamps

4256 slug: "all-types-of-document-about-all-topics-by-hm-passport-office",
4655 slug: slug: "hm-passport-office",

content_change_ids_for_4655 =  MatchedContentChange.where(subscriber_list_id: 4655).pluck(:content_change_id).sort

content_change_ids_for_4256 =  MatchedContentChange.where(subscriber_list_id: 4256).pluck(:content_change_id).sort

content_change_id_id_4655 == content_change_id_id_4256
=> true
```

PR #1995, opened 2023-09-06
Allow HistoricDataDeletionWorker to delete Subscribers/SubscriberLists that have older inactive subscriptions.

KludgeKML:
- add new scope for subscriptions that are either active or ended less than a week ago
- add this scope to the selections in the worker so it can delete Subscribers or SubscriberLists if they only contain older ended subscriptions, so they more closely match what the original specs actually asked for.
- replace .delete calls with .destroy calls in worker to allow rails to maintain db relation consistency (slower but correct, and worker does not need to be super performant)

https://trello.com/c/HFy9tpyE/2215-investigate-whether-historic-email-subscription-data-can-be-more-aggressively-deleted

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -30,6 +30,7 @@ class Subscription < ApplicationRecord
 
   scope :active, -> { where(ended_at: nil) }
   scope :ended, -> { where.not(ended_at: nil) }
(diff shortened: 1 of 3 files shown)


KludgeKML:
Closing as I think this will be unnecessary if we simply clear the existing confusing bad lists.

PR #2024, opened 2023-11-02
Add additional information to Email model

KludgeKML:
Adds content_id and notify_status columns to the Email model, and the support code to populate them - the content_id from the immediate email builder and the notify_status from the callback endpoint. Also adds an index on the existing email id field so that we have a fast lookup during the callback endpoint rather than a field scan.

https://trello.com/c/xdMu72UJ/2245-add-new-information-to-email-alert-api-email-model

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -28,6 +28,7 @@ private
           subscriber_id: subscriber.id,
           created_at: now,
           updated_at: now,
+          content_id: content.try(:content_id),
         }
       end
     end
--- a/app/controllers/status_updates_controller.rb
+++ b/app/controllers/status_updates_controller.rb
@@ -23,6 +23,13 @@ class StatusUpdatesController < ApplicationController
       UnsubscribeAllService.call(subscriber, :non_existent_email) if subscriber
     end
 
+    begin
+      Email.find(reference).update!(notify_status: status)
+    rescue StandardError
+      # At the moment we don't want to do anything if
+      # the reference can't be found.
+    end
+
     head :no_content
   end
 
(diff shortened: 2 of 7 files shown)


app/builders/immediate_email_builder.rb
```diff
@@ -28,6 +28,7 @@ def records
           subscriber_id: subscriber.id,
           created_at: now,
           updated_at: now,
+          content_id: content.instance_of?(ContentChange) ? content.content_id : nil,
```
1pretz1, line 31: We could just do `content.try(:content_id)` here

app/controllers/status_updates_controller.rb
```diff
@@ -23,6 +23,13 @@ def create
       UnsubscribeAllService.call(subscriber, :non_existent_email) if subscriber
     end
 
+    begin
+      Email.where(id: reference).update!(notify_status: status)
```
1pretz1, line 27: `Email.where(id: reference).update!(notify_status: status)` ->  `Email.find(reference).update!(notify_status: status)`

1pretz1 (approved):
Looks good to me - let a couple minor comments. I think you're probably right about updating the notify status synchronously but agreed we should test it 👍

app/builders/immediate_email_builder.rb
KludgeKML: Oh yeah, that's a bit more future proof as well (maybe messages will have content ids in them at some point)

app/controllers/status_updates_controller.rb
KludgeKML: Yeah... I have one very mild reservation about this, which is that find returns an array and I can't call explain on it to check how it's doing the find. But I'm going to assume it does the right thing behind the scenes, so I'll make the change.

1pretz1: I think `find` will only return an array if an array is passed to it, passing just a string should just return the email object (if it exists), but yeah you can't find the query information from the return object. However, I'd be very surprised if `find` didn't use the newly created index (assume that's what you're verifying). Could also check this via Postgres logs.

PR #2026, opened 2023-11-09
Scheduled job checking alerts

KludgeKML:
Adds scheduled jobs and workers for Medical Alerts (and eventually Travel Advice Alerts), and prometheus metrics arising from them (these can be compared, if not equal, an alert can be raised)

https://trello.com/c/ONGpFW0k/2258-add-scheduled-job-to-email-alert-api-to-check-delivery-status-of-alert-emails

- [x] Add tests for Medical Alerts
- [x] Add tests for Travel Advice Alerts
- [x] Implement Medical Alert check
- [x] Implement Travel Advice Alert check

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -39,3 +39,11 @@
     recover_lost_jobs:
       every: '30m'
       class: RecoverLostJobsWorker
+    check_medical_safety_alerts:
+      every: '15m'
+      class: AlertCheckWorker
+      args:  ["medical_safety_alert"]
+    check_travel_advice_alerts:
+      every: '15m'
+      class: AlertCheckWorker
+      args:  ["travel_advice_alert"]
--- a/docs/alert_check_scheduled_jobs.md
+++ b/docs/alert_check_scheduled_jobs.md
@@ -1,0 +1,17 @@
+# Alert Check Scheduled Jobs
+
+Two sets of alerts (Medical Safety Alerts and Travel Advice Alerts) are critical, so it's important to check that they have gone out correctly. To do this we have a pair of scheduled sidekiq jobs. These jobs use the public search API to find a list of relevant Content IDs, then iterate that list checking for the existence of at least one Email record which has that Content ID and a notify_status of "delivered", and which was created after the public update time . Via the prometheus collector, we return metrics of the number of alerts and the number of alerts that have at least one delivery record. Ideally those number should be the same (0 & 0 is okay, 2 & 2 is okay, but 2 & 1 is an alert state).
+
+The jobs run every 15 minutes, and checks for alerts that meet the following criteria from the Search API results:
+- document type is `medical_safety_alert` or `travel_advice`
+- is in the 50 most recently updated documents of that type
+- created at least one hour ago (very new alerts are excluded)
+- created no more than 2 days ago (older alerts are excluded)
+
+## Multiple changes to a content item
+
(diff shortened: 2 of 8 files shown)


app/workers/medical_safety_alert_check_worker.rb
```diff
@@ -0,0 +1,24 @@
+class MedicalSafetyAlertCheckWorker < ApplicationWorker
```
1pretz1, line 1: Now both works are pretty much the same, we could either combine both into one worker, or keep distinct workers but use a parent class to DRY up the code (+ specs) and just pass through the alert type.

config/sidekiq.yml
```diff
@@ -39,3 +39,9 @@
     recover_lost_jobs:
       every: '30m'
       class: RecoverLostJobsWorker
+    check_medical_safety_alerts:
+      every: '30m'
```
1pretz1, line 43: 30m -> 15m

docs/alert_check_scheduled_jobs.md
```diff
@@ -0,0 +1,19 @@
...
+
+The jobs run every 15 minutes, and checks for alerts that meet the following criteria from the Search API results:
+- document type is `medical_safety_alert` or `travel_advice`
+- is in the 50 most recently updated documents of that type
+- created at least one hour ago (very new alerts are excluded)
+- created no more than 2 days ago (older alerts are excluded)
+
+## Multiple changes to a content item
+
+It is possible that a given travel advice item might be updated more than once in two days, but there's no way to know about that through the Search API (we can only find out the last update to a given page). In this case we can only check that emails were delivered after the last public update. The 1 hour lag before an alert becomes checkable plus the 15 minute update frequency means that if a change happens between 60 and 75 minutes after the previous change the system can only confirms that emails were sent out for the most recent update, and if further proof of the previous update is needed a developer will have to check that manually. If the change happens later than 75 minutes, both changes will have been checked by separate runs of the job.
```
1pretz1, line 13: `It is possible that a given travel advice item` -> `It is possible that a given travel advice item or medical alert`

1pretz1 (commented):
Nice work Keith! I've left a few comments, some are consequences of our offline chat (making alert thresholds the same).

lib/search_alert_list.rb
```diff
@@ -0,0 +1,12 @@
+module SearchAlertList
+  MAX_RESULTS = 50
+
+  def get_alert_content_items(alert_type:, min_age:, max_age:)
+    results = GdsApi.search.search(count: MAX_RESULTS, fields: "content_id,link,public_timestamp", filter_format: alert_type, order: "-public_timestamp")
+
+    results.to_hash["results"]
+      .select { |result| Time.zone.parse(result["public_timestamp"]) < Time.zone.now - min_age }
+      .select { |result| Time.zone.parse(result["public_timestamp"]) > Time.zone.now - max_age }
```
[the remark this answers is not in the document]
KludgeKML: Done (sort of - compact_map isn't available for this array, so to keep rubocop happy I did map to a var then compact on that. Also worth noting that between?'s params are start/end, so we want max/min in this case. But tiny changes, basic idea nice).

app/workers/medical_safety_alert_check_worker.rb
KludgeKML: Yeah, I kept thinking this, it made sense when they were checking different feeds, but now not so much. Since we can add params in the sidekiq schedule, single class makes sense.

1pretz1 (approved):
Nice - looks good to me! Left two minor comments but non blocking

PR #2047, opened 2023-12-04
Get data hygiene SQL from helm-charts

KludgeKML:
- removes local copy, ensures tests are always run against the hygiene code that's actually being used in the actual db backup.


sengi (approved):
Thanks for addressing this — clears an item from my todo list!

This adds a network dependency (and a third-party one at that) to your testsuite — if you're really ok with that then not gonna stand in your way, but you might wanna consider fetching the file ahead of time (outside the testsuite itself).

KludgeKML:
Yeah, it's not 100% ideal, but I think reasonably the dependency here is on a github site, so since it'll be running from dev machines or github actions, we'll probably have larger problems if it fails. I might just double check that it's not making multiple calls, though (I think a let shouldn't, but I might be making a bad assumption)

sengi:
No probs 👍  it's arguably less bad than not having the test automated 🤷😅

KludgeKML:
It was making multiple calls. Sorted that out, which makes it a bit easier to add a non-network fallback later on if this ends up causing problems.

PR #2080, opened 2024-01-11
Honour RAILS_LOG_LEVEL in Sidekiq workers.

sengi:
For https://github.com/alphagov/govuk-helm-charts/pull/1715.

See also https://github.com/alphagov/locations-api/pull/337.


KludgeKML (approved):
Thanks!

PR #2084, opened 2024-01-16
Add PollingAlertCheckWorker to copy with 429s blocking Notify callbacks

KludgeKML:
The problem of blocks of 429s sometimes being returned when notify attempts to make callbacks for delivered emails means that we're at risk of getting false alarms for emails not being sent when they actually have been, but the callback has failed. As a fallback, we add in an active polling step.

- If a single positive callback has been recorded, we act as normal (no alert).
- If no positive callbacks have been records, we start actively polling Notify with the IDs (reference ids in Notify parlance) of emails that we've had no callback for, querying whether they have been received. We expect in most cases this will result in one or two calls to Notify before a positive delivery is found, although the worst case (no emails delivered at all) would result in us asking Notify about every single email.
 
https://trello.com/c/KR2THVrN/2323-add-active-polling-to-mitigate-false-alarms-from-429-responses


spec/workers/polling_alert_check_worker_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  include SearchAlertListHelpers
+  include NotifyRequestHelpers
+
+  describe "#perform", caching: true do
+    def perform
+      described_class.new.perform("medical_safety_alert")
+    end
+
+    context "there are no alerts older than an hour" do
+      before { stub_medical_safety_alert_feed(content_id: SecureRandom.uuid, age: 30.minutes) }
```
1pretz1, line 11: Could rename `stub_medical_safety_alert_feed` to `stub_medical_safety_alert_query` as feed reminds me of atom 😄

spec/workers/polling_alert_check_worker_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      it "should put 1/0 in the cache" do
+        expect(Rails.cache).to receive(:write).with("current_medical_safety_alerts", 1, expires_in: 15.minutes)
+        expect(Rails.cache).to receive(:write).with("delivered_medical_safety_alerts", 0, expires_in: 15.minutes)
+        perform
+      end
+    end
+
+    context "there is a valid alert with apparently undelivered emails but they actually were delivered" do
```
1pretz1, line 59: -> "there are no alerts as polling confirmed they were actually delivered"

spec/workers/polling_alert_check_worker_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      before { stub_medical_safety_alert_feed(content_id: SecureRandom.uuid, age: 3.days) }
+
+      it "should put 0/0 in the cache" do
+        expect(Rails.cache).to receive(:write).with("current_medical_safety_alerts", 0, expires_in: 15.minutes)
+        expect(Rails.cache).to receive(:write).with("delivered_medical_safety_alerts", 0, expires_in: 15.minutes)
+        perform
+      end
+    end
+
+    context "there is a valid alert with delivered emails" do
```
1pretz1, line 30: `"there is a valid alert with delivered emails"` -> `"there are no alerts with delivered emails"`

app/services/check_notify_email_service.rb
```diff
@@ -0,0 +1,35 @@
...
+    )
+
+    results.collection.count.positive?
+  rescue Notifications::Client::RequestError, Net::OpenTimeout, Net::ReadTimeout, SocketError => e
+    Rails.logger.debug("Unable to contact Notify to determine status of email reference #{reference}: #{e}")
+  end
+
+private
+
+  attr_reader :email, :client
```
1pretz1, line 21: Swap `:email` with `:status`

1pretz1 (commented):
Excellent work on this Keith! Overall everything looks great, just left a few comments.

We should also remove [AlertCheckWorker](https://github.com/alphagov/email-alert-api/blob/9c7d2bd0f1e6486652e57d9a0a1f2141df0a7ce3/app/workers/alert_check_worker.rb#L1), the [spec](https://github.com/alphagov/email-alert-api/blob/9c7d2bd0f1e6486652e57d9a0a1f2141df0a7ce3/spec/workers/alert_check_worker_spec.rb#L1).

Also need to update the [testing documentation](https://github.com/alphagov/email-alert-api/blob/9c7d2bd0f1e6486652e57d9a0a1f2141df0a7ce3/docs/alert_check_scheduled_jobs.md?plain=1#L50), I don't think the current method would suffice anymore as the canary account will receive the email and Notify should respond with the email delivery. I think we could probably achieve the same test by just deleting all the emails vs marking the Notify status as `nil`.

app/services/check_notify_email_service.rb
```diff
@@ -0,0 +1,35 @@
...
+    @status = status
+    @client = Notifications::Client.new(Rails.application.credentials.notify_api_key)
+  end
+
+  def present?(reference)
+    results = @client.get_notifications(
+      template_type: "email",
+      status: @status,
+      reference:,
+    )
```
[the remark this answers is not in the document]
KludgeKML: Actually, the readers probably don't add anything (we don't actually need to read externally, and we're only using the var in a couple of places), so I've removed them entirely along with the method below.

app/services/check_notify_email_service.rb
```diff
@@ -0,0 +1,35 @@
...
+    results.collection.count.positive?
+  rescue Notifications::Client::RequestError, Net::OpenTimeout, Net::ReadTimeout, SocketError => e
+    Rails.logger.debug("Unable to contact Notify to determine status of email reference #{reference}: #{e}")
+  end
+
+private
+
+  attr_reader :email, :client
+
+  def undeliverable_failure?(error)
```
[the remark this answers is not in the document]
KludgeKML: Copy-pasto! I've removed now

spec/workers/polling_alert_check_worker_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    context "there is a valid alert with apparently undelivered emails but they actually were delivered" do
+      before do
+        content_id = SecureRandom.uuid
+        stub_medical_safety_alert_feed(content_id:, age: 2.hours)
+        email = create(:email, content_id:, notify_status: nil, status: "sent")
+        stub_notify_email_delivery_status_delivered(email.id)
+      end
+
+      it "should put 1/0 in the cache" do
```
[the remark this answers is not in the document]
KludgeKML: Done

app/services/check_notify_email_service.rb
KludgeKML: Removed as above.

spec/workers/polling_alert_check_worker_spec.rb
KludgeKML: I've renamed all of these to make the wording clear that it's "email alerts" we're concerned about in the context (ie the thing we're watching), not a generated alert (ie what Prometheus does when the two metrics don't match)

app/workers/polling_alert_check_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+    return true if Email.where("notify_status = 'delivered' AND content_id = ? AND created_at > ?", content_id, valid_from).exists?
+
+    Rails.logger.info("First pass couldn't find any delivered emails for #{document_type.titleize} records with content id #{content_id}, reverting to polling")
+
+    service = CheckNotifyEmailService.new("delivered")
+    Email.sent.where("content_id = ? AND created_at > ? AND notify_status IS NULL", content_id, valid_from).first(100).each do |email|
+      return true if service.present?(email.id)
+    end
+
+    false
```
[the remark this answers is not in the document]
KludgeKML: Done, as per conversation in slack.

spec/workers/polling_alert_check_worker_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+      it "should put 1/1 in the cache" do
+        expect(Rails.cache).to receive(:write).with("current_medical_safety_alerts", 1, expires_in: 15.minutes)
+        expect(Rails.cache).to receive(:write).with("delivered_medical_safety_alerts", 1, expires_in: 15.minutes)
+        perform
+      end
+
+      it "doesn't poll Notify once it's found a delivered email" do
+        perform
+        expect(a_request(:get, "https://api.notifications.service.gov.uk/v2/notifications?reference=#{@no_poll_email.id}&status=delivered&template_type=email")).not_to have_been_made
```
1pretz1, line 81: We could probably clean up the code slightly here:
```
        email3 = create(:email, content_id:, notify_status: nil, status: "sent")
        request = stub_notify_email_delivery_status_delivered(email3.id)
        perform
        expect(request).not_to have_been_made
```

1pretz1 (approved):
Changes look good, thanks for working through the comments! Left one minor suggestion but non blocking so approving 🍏

PR #2121, opened 2024-03-20
Subscriber list stats for finders

KludgeKML:
Adds a rake task that will get subscriber lists for a finder (since these don't work well with existing rake tasks)

https://trello.com/c/ZM53KHEM/2406-consider-how-email-alert-api-reports-mailing-list-sizes-for-finders

### DIFF
--- a/app/queries/subscriber_lists_for_finder_query.rb
+++ b/app/queries/subscriber_lists_for_finder_query.rb
@@ -1,0 +1,31 @@
+class SubscriberListsForFinderQuery
+  attr_reader :govuk_path
+
+  class NotAFinderError < StandardError; end
+
+  def initialize(govuk_path:)
+    @govuk_path = govuk_path
+  end
+
+  SELECT_FOR_TAGS_TEMPLATE = <<-SQL.freeze
+    SELECT EXISTS (
+      SELECT formats FROM (
+        SELECT json_array_elements_text(tags -> ? -> 'any') AS formats
+      ) AS allowed_formats WHERE formats = ?
+    )
+  SQL
+
+  def call
+    content_item = GdsApi.content_store.content_item(govuk_path).to_hash
+    raise NotAFinderError unless content_item["document_type"] == "finder"
+
+    lists = []
+    content_item["details"]["filter"].each_key do |filter_name|
+      Array(content_item["details"]["filter"][filter_name]).each do |filter_value|
+        lists += SubscriberList.where(SELECT_FOR_TAGS_TEMPLATE, filter_name, filter_value)
+      end
+    end
+
+    lists.uniq
+  end
+end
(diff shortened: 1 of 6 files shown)


app/queries/subscriber_lists_for_finder_query.rb
```diff
@@ -0,0 +1,38 @@
...
+        FROM
+            (
+                SELECT
+                    json_array_elements_text(
+                        tags -> ? -> 'any'
+                    ) AS formats
+            ) AS allowed_formats
+        WHERE
+            formats = ?
+    )".freeze
```
1pretz1, line 23: I've had a stab at making this more readable, what do you think?
```
SELECT_FOR_TAGS_TEMPLATE = <<-SQL 
SELECT EXISTS (
  SELECT formats FROM (
    SELECT json_array_elements_text(tags -> ? -> 'any') AS formats
  ) AS allowed_formats
  WHERE formats = ?
)
SQL
```

app/queries/subscriber_lists_for_finder_query.rb
```diff
@@ -0,0 +1,38 @@
...
+  def call
+    content_item = GdsApi.content_store.content_item(govuk_path).to_hash
+    raise NotAFinderError unless content_item["document_type"] == "finder"
+
+    lists = []
+    content_item["details"]["filter"].each_key do |filter_name|
+      Array(content_item["details"]["filter"][filter_name]).each do |filter_value|
+        lists += SubscriberList.where(SELECT_FOR_TAGS_TEMPLATE, filter_name, filter_value)
+      end
+    end
```
1pretz1, line 34: Do we expect there to be anything other than `format` in the `filter` JSON? If we don't then this could be a lot simpler, e.g:
```
SubscriberList.where(SELECT_FOR_TAGS_TEMPLATE, "format",  content_item["details"]["filter"]["format"])
```

docs/support-tasks.md
```diff
@@ -113,11 +113,27 @@ This will report on the number of active subscriptions for a given path.
...
+To see a report on the number of subscribers to a finder, :
+
+```bash
+kubectl -n apps exec -it deploy/email-alert-api -- bundle exec rake 'report:finder_statistics[<path>]'
+```
+This will report on the number of subscriber lists to the finder on a given path.
+
+The path should be the full path on gov.uk (for instance /cma-cases if the page you're interested in is https://www.gov.uk/cma-cases)
+
+Because finder subscriptions can include the filters active on the finder at the point the subscription was created, there may be more than one subscriber list. Finders require an email_alert_signup item in the links section of their content item to allow alerts, so if there are no subscriptions to an finder
```
1pretz1, line 133: "to an finder" -> "to a finder"

1pretz1 (commented):
Excellent work Keith! Just left a few comments

app/queries/subscriber_lists_for_finder_query.rb
KludgeKML: No, we can definitely have other things than `format` (for instance search/news-and-communications has this:

```
filter: {
  content_purpose_subgroup: [
    "news",
    "speeches_and_statements"
  ]
},
```

KludgeKML: I think there's also some with `content_purpose_supergroup`.

1pretz1 (approved):
Sweet!

PR #2136, opened 2024-04-12
Remove specialist topic code from email alert API

unoduetre:

## What

When all specialist topic pages have been archived and redirected, we need to remove the specialist topic related code from Email alert API

## Why

Remove tech debt.

[Trello ticket](https://trello.com/c/RObMmjXG/2422-remove-specialist-topic-code-from-email-alert-api-l)

## Dependencies

The [following PR](https://github.com/alphagov/email-alert-api/pull/2147) has been extracted from the current one and needs to be merged first.
The [following PR](https://github.com/alphagov/gds-api-adapters/pull/1250) fixes Pact tests.

### DIFF
--- a/lib/email_alert_criteria.rb
+++ b/lib/email_alert_criteria.rb
@@ -65,7 +65,6 @@ class EmailAlertCriteria
 
   def contains_supported_attribute?
     supported_attributes = %w[
-      topics
       policies
       service_manual_topics
       taxons
--- a/lib/valid_tags.rb
+++ b/lib/valid_tags.rb
@@ -65,7 +65,6 @@ class ValidTags
     subject
     therapeutic_area
     tiers_or_standalone_items
-    topics
     tribunal_decision_categories
     tribunal_decision_category
     tribunal_decision_country


Tetrino (approved):

spec/factories.rb
```diff
@@ -100,7 +100,7 @@
   factory :subscriber_list do
     sequence(:title) { |n| "title #{n}" }
     sequence(:slug) { |n| "title-#{n}" }
-    tags { { topics: { any: ["motoring/road_rage"] } } }
```
hannako, line 103: Policies are a really old type of tag. Policies were retired, and you can no longer tag content to a `policy` anymore from Whitehall. In fact digging into documentation from previous teams, I think all policy subscriber lists were meant to have been bulk-unsubscribed. Similar to what we're doing here! So it might be safer to swap out `topics` with `taxons` instead. As taxons are the least likely to be retired AFAIK!

spec/queries/find_exact_query_spec.rb
```diff
@@ -169,79 +173,79 @@
...
+      bad_query = build_query(tags: { browse_pages: { all: %w[uuid-888] } })
+      good_query = build_query(tags: { browse_pages: { all: %w[uuid-888], any: %w[uuid-777] } })
       expect(bad_query.exact_match).to be_nil
       expect(good_query.exact_match).to eq(subscriber_list)
     end
 
     it "matched when subscriber tags has the same tags" do
       query = build_query(tags: { policies: { any: %w[beer] },
-                                  topics: { all: %w[taxon] } })
+                                  browse_pages: { all: %w[taxon] } })
```
hannako, line 234: `browse_pages` aren't a valid tag - let's update this to be `taxons` as well.

spec/queries/matched_for_notification_spec.rb
```diff
@@ -26,122 +26,124 @@ def execute_query(query_hash, field: :tags)
...
       end
     end
 
-    context "matches on all topics and any policies" do
+    context "matches on all browse pages and any policies" do
       before do
-        @all_topics_tax_vat_any_policies_economy_industry = create_subscriber_list_with_tags_facets(topics: { all: %w[vat tax] }, policies: { any: %w[economy industry] })
-        @all_topics_vat_any_policies_economy_industry = create_subscriber_list_with_tags_facets(topics: { all: %w[paye schools] }, policies: { any: %w[economy industry] })
+        stub_const("ValidTags::ALLOWED_TAGS", ValidTags::ALLOWED_TAGS + %i[browse_pages])
+        @all_browse_pages_tax_vat_any_policies_economy_industry = create_subscriber_list_with_tags_facets(browse_pages: { all: %w[vat tax] }, policies: { any: %w[economy industry] })
```
hannako, line 89: As above, browse_pages aren't a valid tag, so we should replace with taxons

hannako (changes requested):
Thanks Mat,

I've requested some changes, hopefully the comments makes sense?

I'd like to split this PR to make it easier for us to get the code reviewed, and make the intention of the work a bit easier to understand. I can see you extracted the pact tests into their own PR, but presumably that's not going to pass against a branch of gds-api-adapters without the SubsciberList factory being updated (which is still in this PR)? 

Can I suggest we have one PR that removes `topic` as a valid tag for email alert api. This would contain your changes to
- lib/email_alert_criteria.rb
- lib/valid_tags.rb
- spec/validators/tags_validator_spec.rb
- spec/lib/email_alert_criteria_spec.rb

And then have a lower risk PR, which includes the remaining changes to the pact tests and factories and all specs, swapping out references to `topics` to `taxons`. That PR could go first, and would be about removing legacy references to specialist_topics. Ie not application code.

spec/queries/matched_for_notification_spec.rb
hannako: There are more incidences below, but I won't comment on each one

spec/factories.rb
unoduetre: The only possibile tags than can be used are policies and browse_pages, if I understand things correctly. Please, check this line: https://github.com/alphagov/publishing-api/blob/b08fbe0a3f58d827f38403c714920b69826f608f/content_schemas/dist/formats/email_alert_signup/notification/schema.json#L782

I think you are thinking about links, here: https://github.com/alphagov/publishing-api/blob/b08fbe0a3f58d827f38403c714920b69826f608f/content_schemas/dist/formats/email_alert_signup/notification/schema.json#L182

But this part of the tests checks tags, not links.

spec/queries/find_exact_query_spec.rb
unoduetre: Please, check the comment above.

spec/queries/matched_for_notification_spec.rb
unoduetre: Please, see the comment above.

This was done this way, because the only valid tags now are browse_pages and policies, but browse_pages are not in ValidTags::ALLOWED_TAGS. Some tests require two different tags to work, but after removing topics we are left with only one tag which is also in ValidTags::ALLOWED_TAGS, namely policies. So by adding browse_pages it allowed for two different tags, so the tests can work again.

unoduetre:
> Thanks Mat,
> 
> I've requested some changes, hopefully the comments makes sense?
> 
> I'd like to split this PR to make it easier for us to get the code reviewed, and make the intention of the work a bit easier to understand. I can see you extracted the pact tests into their own PR, but presumably that's not going to pass against a branch of gds-api-adapters without the SubsciberList factory being updated (which is still in this PR)?
> 
> Can I suggest we have one PR that removes `topic` as a valid tag for email alert api. This would contain your changes to
> 
>     * lib/email_alert_criteria.rb
> 
>     * lib/valid_tags.rb
> 
>     * spec/validators/tags_validator_spec.rb
> 
>     * spec/lib/email_alert_criteria_spec.rb
> 
> 
> And then have a lower risk PR, which includes the remaining changes to the pact tests and factories and all specs, swapping out references to `topics` to `taxons`. That PR could go first, and would be about removing legacy references to specialist_topics. Ie not application code.

I think we can safely move all the tests to a separate PR. Can I move them to [this PR](https://github.com/alphagov/email-alert-api/pull/2147) (I will change the name and the description of the PR accordingly), so we would have 1 PR dealing with tests and 1 PR applying the actual change?

spec/factories.rb
hannako: Ah yep, taxons was not the right suggestion...

This is a SubscriberList factory. SubscriberLists are valid with any of these tags right? https://github.com/alphagov/email-alert-api/blob/main/lib/valid_tags.rb so you can pick any of these.

unoduetre: OK, I think I get it now. Specialist-publisher bypasses publishing-api so I can use tags from it. I'll modify the code to use some of those.

spec/queries/matched_for_notification_spec.rb
unoduetre: I'll use some of the specialist-publisher tags instead.

spec/queries/find_exact_query_spec.rb
unoduetre: Yes, that makes sense. Specialist publisher bypasses publishing-api, so its tags are not there. I'll use some of those instead of policies.

hannako (commented):
This looks good now Mat - thank you for splitting the PR's!

Minor thing: Please could you add the changes to spec/validators/tags_validator_spec.rb and spec/lib/email_alert_criteria_spec.rb from https://github.com/alphagov/email-alert-api/pull/2147 to this PR

hannako (approved):
Amazing work Mat, thanks for all your efforts

PR #2162, opened 2024-05-23
Add one-click-unsubscribe params when calling Notify

KludgeKML:
- Save subscription in Email
- Calculate unsubscribe token with new path and add that to header when sending emails via Notify.

This allows Notify to create one-click unsubscribe headers in our outgoing emails, to improve the unsubscribe journey in supporting email agents and reduce the chances of our emails being marked as spam instead of being unsubscribed from. Note that our default is to cancel a specific subscription, not to unsubscribe the user from all subscriptions. We may revisit that if it proves to be a problem.

https://trello.com/c/ONisAn0b/52-add-one-click-unsubscribe-support-to-email-alert-api-frontend

Related: https://github.com/alphagov/email-alert-frontend/pull/1773

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -26,6 +26,7 @@ private
           subject:,
           body: body(subscription, subscriber),
           subscriber_id: subscriber.id,
+          subscription_id: subscription.id,
           created_at: now,
           updated_at: now,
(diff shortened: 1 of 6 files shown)


leohemsted (dismissed):
Definitely not an endorsement that this is mergeable, but the notify side looks fine to me 👍

PR #2171, opened 2024-06-06
Revert "Add one-click-unsubscribe params when calling Notify"

kevindew:
Reverts alphagov/email-alert-api#2162

We believe this change is the source of [an incident](https://docs.google.com/document/d/1ty12B5eBWB9YSfnD9xY1mr5rtTQxdNxRdmEGgibilN0/edit) so expect reverting it will help resolve the issue.

### DIFF
--- a/app/builders/immediate_email_builder.rb
+++ b/app/builders/immediate_email_builder.rb
@@ -26,7 +26,6 @@ private
           subject:,
           body: body(subscription, subscriber),
           subscriber_id: subscriber.id,
-          subscription_id: subscription.id,
           created_at: now,
           updated_at: now,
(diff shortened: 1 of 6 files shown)


MuriloDalRi (approved):

kevindew:
We understand the problem is caused when passing Notify a value of nil for one_click_unsubscribe_url:

This causes Notify to return an error:

```
(irb):29:in `<main>': ValidationError: one_click_unsubscribe_url is not a valid https url (Notifications::Client::BadRequestError)
```

This occurs anytime an email is generated without a subscription_id - which I guess is any emails that were in the queue before this was deployed or any emails that don't use ImmediateEmailBuilder

PR #2172, opened 2024-06-10
Re-Add one-click-unsubscribe params when calling Notify

georges1996:
Reverts alphagov/email-alert-api#2171

Ticket: https://trello.com/c/d3c5dOsE/171-email-alert-api-incident-in-production, [Jira issue PNP-7976](https://gov-uk.atlassian.net/browse/PNP-7976)

Going off the comment from the reverted PR https://github.com/alphagov/email-alert-api/pull/2171#issuecomment-2153351035, **(tldr: Problem is caused when passing Notify a value of nil for one_click_unsubscribe_url)** we now only include the `one_click_unsubscribe_url` field in the `email_params` if `subscription_id` and `subscription` are both present

### DIFF
--- a/app/builders/subscription_auth_email_builder.rb
+++ b/app/builders/subscription_auth_email_builder.rb
@@ -36,8 +36,8 @@ private
 
       If you did not request this email, you can ignore it.
 
-      Thanks 
-      GOV.UK emails 
+      Thanks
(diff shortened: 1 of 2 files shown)


KludgeKML (approved):
LGTM

PR #2200, opened 2024-08-01
Adjust alert check flag lifetime

KludgeKML:
- The flag expiry time of 15 minutes sometimes caused little blip alerts where at times of high load the elements expired before being renewed. Although this is usually safe (the collector defaults to 0 for all metrics, which is a no alarm state), occasionally the read has happened between the metrics expiring, causing false positive alarms which resolve themselves a few minutes later. We double the expiry time to make sure that the flags don't rely on the job being perfectly on time.
- Note that this doesn't affect the happy path - under normal circumstances, the flags are still getting regenerated every 15 minutes, so we can still detect a problem within 15 minutes.
- We also move to generating the expiry time once and using expire_at: so the two caches have the exact same expiry time.

### DIFF
--- a/spec/workers/polling_alert_check_worker_spec.rb
+++ b/spec/workers/polling_alert_check_worker_spec.rb
@@ -2,6 +2,20 @@ RSpec.describe PollingAlertCheckWorker do
   include SearchAlertListHelpers
   include NotifyRequestHelpers
 
+  def expect_cache_flags_to_contain(current, delivered)
+    expect(Rails.cache).to receive(:write) do |name, count, hash|
+      expect(name).to eq("current_medical_safety_alerts")
+      expect(count).to eq(current)
+      expect(hash[:expires_at]).to be_within(5.seconds).of(Time.zone.now + 30.minutes)
+    end
+
+    expect(Rails.cache).to receive(:write) do |name, count, hash|
+      expect(name).to eq("delivered_medical_safety_alerts")
+      expect(count).to eq(delivered)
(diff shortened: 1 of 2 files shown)


spec/workers/polling_alert_check_worker_spec.rb
```diff
@@ -2,6 +2,20 @@
   include SearchAlertListHelpers
   include NotifyRequestHelpers
 
+  def expect_cache_flags_to_contain(current, delivered)
+    expect(Rails.cache).to receive(:write) do |name, count, hash|
+      expect(name).to eq("current_medical_safety_alerts")
+      expect(count).to eq(current)
+      expect(hash[:expires_at]).to be_within(5.seconds).of(Time.zone.now + 30.minutes)
```
leenagupte, line 9: Nice use of `be_within` 🥇

leenagupte (approved):
You gave a great explanation of the problem this change fixes in slack, so I'll add it here for posterity:

> So the current situation is that the task runs every 15 minutes, and it sets two flags - the number of emails that should have been sent out, and the number that actually were (and if there’s a discrepancy between those two numbers, the alert goes off). The problem is that I had originally set these flags to expire after 15 minutes, and if the job got delayed slightly, the alarm would go off. Not common, but it has happened. It confused 2L, because it solved itself on the next run 15 minutes later.
>
> What this does is sets the flags to expire only after 30 minutes, so that if we get delayed a little bit the old flags still exist and the alarm doesn’t go off. Normally it won’t matter, because the flags will get recreated after 15 minutes anyway, it just makes the thing a bit less brittle if the server is under heavy load.

👍

